PR #1, opened 2012-03-07
feature/suricata merge

obilodeau:
starting suricata review / merge

see both [#1141: Support for surricata IDS](http://packetfence.org/bugs/view.php?id=1141) and [#956: confusion between trapping.range, pf.conf's interfaces and networks.conf](http://packetfence.org/bugs/view.php?id=956).


obilodeau:
Minor changes done. This is ready for testing. A few questions for you @fgaudreault:
1. /var/log/snort-compat/ vs /var/log/snort/ in our remote probe's initrd, what should we do with that? I vaguely recall something about that... Found it: http://packetfence.org/bugs/view.php?id=956. So, is suricata putting it's alert pipe in snort-compat? should we tell users to do so in our guides (and for snort) or should we do something else?
2. I noticed suricata's default-log-dir is %%install_dir%%/logs. Does this means we need to add new log files to our log rotation?
3. I noticed changes between internal_range usage and trapping_range usage. Is this consistent with were we want to go (last time I recall we were talking about getting rid of trapping_range)? We also discuss this here http://packetfence.org/bugs/view.php?id=956 and I noticed we've started to use internal-nets + routed-nets in apache's configuration...

Also, before merge:
1. documentation: at least be on-par with snort

fgaudreault:
Answers :
1. We decided to move everything to snort-compat so SNORT and Suricata can drop their alert file in there.  We thought it would be less complicated to do so.
2. Probably
3. trapping_range is used for the HOME_NET parameter of the SNORT config.  The internal + routed-nets in the apache config is for something else.

I will add some documentation about Suricata, especially the install part.

obilodeau:
just rebased the branch on top of current stable. Now mergeable. Will go into integration now.

obilodeau:
note to myself: this interacts badly with the upcoming feature/debian branch changes because of tricky conflicts in pf::services.

obilodeau:
note to myself: the cleanup done to get rid of interface.gateway might create a problem. At merge time, make sure that:

``` perl
$tags{'gateways'}      = join( ",", get_gateways() );
$tags{'internal-ips'}  = join( ",", get_internal_ips() );
```

are not present in the lib/pf/services/snort.pm.

fgaudreault:
Rebased from Devel.  I will review it one last time before we merge that into devel.

dwlfrth:
Successfully passed the integration branch.
Things that need to be done before the merge:
- NEWS/UPGRADE documents (i'll manage the conflict... it is better if these documents are updated in the branch)
- according to a recent change, line 49 of lib/pf/services/snort.pm (internal_ips) should be removed

dwlfrth:
Merged into devel with commit 1eddefafffbe3ef6625e2e02ad7139efe0fc92a3

PR #2, opened 2012-03-08
feature/managed-radius

obilodeau:
We are not in a rush to merge this yet but I wanted to add that we should fix this issue in this branch: http://packetfence.org/bugs/view.php?id=1394

Also webservice user / password should be automatically populated by the template parsers.


obilodeau:
oups.. I started a new one... closing this. See pull #4.

PR #4, opened 2012-03-21
managed radius feature integration

obilodeau:
Few questions / comments
- Are the latest radius changes in this branch? (no longer log to syslog, outer-identity-privacy support, avoid double rlm_perl running, etc.)
- webservice's password should be populated
- files in var/ are meant to be somewhat temporary, I think we should move all files to conf/radiusd (or raddb, whatever you prefer)

Not of high priority right now but I think we should get the discussion going by starting the pull request.


obilodeau:
I'll need to rebase this feature branch because of the pf/ -> ./ move done after the git migration.

obilodeau:
I finished the rebase on stable branch. Now I can start the review. I'll also merge into integration for packaging tests tonight.

fgaudreault:
Just did a fresh rebase.  I will look at the remaining stuff so we can bring this in.

dwlfrth:
shouldn't we have an entry in the NEWS document ?

fgaudreault:
We should add the NEWS entry post-merge to avoid fancy conflicts in the NEWS file.

dwlfrth:
Successfully passed the integration branch.
Things that need to be done before the merge:
- NEWS document (i'll manage the conflict... it is better if the NEWS is updated in the branch)
- check for files listed twice warnings
- should we add something accordingly in the admin guide?

PR #7, opened 2012-04-10
code review for feature/roles-experiment

obilodeau:
ready for review.


obilodeau:
@fgaudreault told me some issues to look at:
- ~~locationlog isn't kept in sync so on access re-evaluation if the VLAN didn't change but the role should've (ex: recategorization) the user is not disconnected~~
  - Out of scope for next release. Filed issue [#1421: roles support doesn't handle re-assignment properly yet](http://www.packetfence.org/bugs/view.php?id=1421) to track it for +1
- ~~simple validation should be performed on the format of the roles parameter in switches.conf~~
  - implemented in 3ba16a4

lib/pf/SNMP.pm
```diff
@@ -660,6 +679,58 @@ sub _setVlanByOnlyModifyingPvid {
...
+        # otherwise undef
+        $logger->warn("Roles are configured but no role found in cache for $roleName");
+        return;
+    }
+
+    # not cached, let's perform the lookup then prepare cache
+    # split on ; to get 'category = controller_role'
+    my $roles_assignment_ref;
+    my @category_split = split( /\;/, $this->{'_roles'} );
+    foreach my $r_assign (@category_split) {
```
dwlfrth: r_assign? not really self-documenting ;)

lib/pf/roles.pm
```diff
@@ -0,0 +1,186 @@
...
+
+Constructor. Usually you don't want to call this constructor but use the 
+pf::roles::custom subclass instead.
+
+=cut
+
+sub new {
+    my $logger = Log::Log4perl::get_logger(__PACKAGE__);
+    $logger->debug("instantiating new " . __PACKAGE__ . " object");
+    my ( $class, %argv ) = @_;
```
dwlfrth, line 68: maybe reorder theses one for consistency purposes

lib/pf/SNMP/AeroHIVE.pm
```diff
@@ -55,6 +60,7 @@ use pf::util;
 # access technology supported
 sub supportsWirelessDot1x { return $TRUE; }
 sub supportsWirelessMacAuth { return $TRUE; }
+sub supportsRoleBasedEnforcement { return $TRUE; }
```
dwlfrth: i'd go with an alphabetic order

lib/pf/SNMP.pm
obilodeau: fixed in 1fade0bb09c823beac78c5bebbbec6f5e64a6583

lib/pf/roles.pm
obilodeau, line 68: I think putting instance() on top makes it clearer that we are dealing with a singleton. pf::SwitchFactory is structured that way too.

lib/pf/SNMP/AeroHIVE.pm
obilodeau: fixed in 785d0a1

obilodeau:
thank you dear for your review

dwlfrth:
my pleasure sir!

PR #8, opened 2012-04-10
Introducing DNS rewrite (DNS DNATting) for inline mode

obilodeau:
owner: @dereksky 
assigned: @obilodeau

### DIFF
--- a/lib/pf/iptables.pm
+++ b/lib/pf/iptables.pm
@@ -166,10 +166,10 @@ sub generate_inline_rules {
         next if ( !pf::config::is_network_type_inline($network) );
 
         my $rule = "--protocol udp --destination-port 53";
-        $$nat_prerouting_ref .= "-A $FW_PREROUTING_INT_INLINE $rule --match mark --mark 0x$IPTABLES_MARK_UNREG 
-                --jump DNAT --to $ConfigNetworks{$network}{'gateway'}\n";
-        $$nat_prerouting_ref .= "-A $FW_PREROUTING_INT_INLINE $rule --match mark --mark 0x$IPTABLES_MARK_ISOLATION
-                --jump DNAT --to $ConfigNetworks{$network}{'gateway'}\n";
+        $$nat_prerouting_ref .= "-A $FW_PREROUTING_INT_INLINE $rule --match mark --mark 0x$IPTABLES_MARK_UNREG "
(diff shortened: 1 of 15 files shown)


obilodeau:
~~Not sure I like the fact to assign the DNAT to networks.conf's gateway parameter. This is not self-documenting.~~
We'll leave it that way for the moment. It is documented in PacketFence Administration Guide.

obilodeau:
~~This feature would need a notice in UPGRADE.~~
Commit 1b99d16

obilodeau:
~~Also, I think we talk in the documentation about the limitations of the inline mode (one VLAN only) and portal_redirect stuff. This would need to be removed / improved.~~
See Commits 6a40435 and dd39ae6

fgaudreault:
~~Having a bug with this part I think :
Use of uninitialized value in pattern match (m//) at /usr/local/pf/lib/pf/util.pm line 119.
Use of uninitialized value in concatenation (.) or string at /usr/local/pf/lib/pf/services/named.pm line 104.
Use of uninitialized value in substitution iterator at /usr/local/pf/lib/pf/util.pm line 678, <$template_fh> line 13.~~
- Was trying to generate a template for a non-existing interface
- named was set to disabled
  See commit 5e2a86c and 623e4da

lib/pf/iptables.pm
```diff
@@ -160,19 +160,18 @@ sub generate_inline_rules {
     my ($filter_rules_ref, $nat_prerouting_ref, $nat_postrouting_ref) = @_;
     my $logger = Log::Log4perl::get_logger('pf::iptables');
 
-    $logger->info("Allowing DNS through on inline interfaces to configured DNS servers");
-    foreach my $network (keys %ConfigNetworks) {
+    $logger->info("Adding DNS DNAT rules for unregistered and isolated inline clients.");
+    foreach my $network ( keys %ConfigNetworks ) {
         # skip non-inline interfaces
-        next if (!pf::config::is_network_type_inline($network));
+        next if ( !pf::config::is_network_type_inline($network) );
```
obilodeau, line 167: lets try to not change whitespace only.. makes reviewing and merging more complex. (next if line and the foreach line)

obilodeau:
Made minor changes. Now testing.

PR #9, opened 2012-04-12
Cisco RADIUS Disconnect-Message (RFC3576) support

obilodeau:
unfortunately moving the files around makes the diff a lot messier than it should be :(

we'll probably only test and go ahead with this one ;)


obilodeau:
got the go from @dereksky

PR #10, opened 2012-04-19
feature/debian

obilodeau:
c'est prêt!


debian/packetfence.postinst
```diff
@@ -0,0 +1,93 @@
...
+	export PACKETFENCE=/usr/local/pf
+	chown -R pf.pf /usr/local/pf
+	export PERL5LIB=/usr/local/pf/lib
+	perl -w -e "use strict; use warnings; use Parse::RecDescent; use lib \"./lib\"; use pf::pfcmd::pfcmd; Parse::RecDescent->Precompile(\$grammar, \"pfcmd_pregrammar\");"
+	mv pfcmd_pregrammar.pm $PACKETFENCE/lib/pf/pfcmd/
+	if [ ! -e "$PACKETFENCE/db/pf-schema.sql" ]; then
+		cd $PACKETFENCE/db
+		VERSIONSQL=$(ls pf-schema-* |sort -r | head -1)
+		ln -s $VERSIONSQL ./pf-schema.sql
+	fi
```
obilodeau, line 31: ~~There are no SQL schema generated for every packetfence version, it's only if the schema was updated so this, while clever, won't work.~~

Ignore my comment. It's actually very clever and will work. I'm integrating that for the RPM version. :)

debian/packetfence.postinst
```diff
@@ -0,0 +1,93 @@
...
+	if [ ! -e "$PACKETFENCE/conf/httpd.conf" ]; then
+		cd $PACKETFENCE/conf
+		ln -s httpd.conf.apache22 ./httpd.conf
+	fi
+	if [ ! -e "$PACKETFENCE/conf/named.conf" ]; then
+		cd $PACKETFENCE/conf
+		ln -s named.conf.bind97 ./named.conf
+	fi
```
obilodeau: what's the bind version shipped with debian 6?

debian/packetfence.postinst
```diff
@@ -0,0 +1,93 @@
...
+	cp -r $PACKETFENCE/addons/freeradius-integration/sql/mysql/packetfence.conf /etc/freeradius/sql/mysql
+	cp -r $PACKETFENCE/addons/soh/packetfence-soh.pm /etc/freeradius
+	cp -r $PACKETFENCE/addons/802.1X/packetfence.pm /etc/freeradius
+	cp -r $PACKETFENCE/addons/freeradius-integration/sites-available/packetfence /etc/freeradius/sites-available
+	cp -r $PACKETFENCE/addons/freeradius-integration/sites-available/packetfence-soh /etc/freeradius/sites-available
+	cp -r $PACKETFENCE/addons/freeradius-integration/sites-available/packetfence-tunnel /etc/freeradius/sites-available
+	mv $PACKETFENCE/addons/pfdetect_remote/initrd/pfdetectd /etc/init.d/
+	mv $PACKETFENCE/addons/pfdetect_remote/sbin/pfdetect_remote /usr/local/pf/sbin
+	mv $PACKETFENCE/addons/pfdetect_remote/conf/pfdetect_remote.conf /usr/local/pf/conf
+	if [ ! -e "/etc/freeradius/sites-enabled/control-socket" ]; then
```
obilodeau: ~~(pfdetect_remote is in another package (packetfence-remote-snort-sensor) which conflicts with PacketFence. We should probably get rid of the files (and later do the remote-snort-sensor subpackage))~~ addressed

debian/packetfence.postinst
```diff
@@ -0,0 +1,93 @@
...
+	fi
+	if [ ! -e "/etc/freeradius/sites-enabled/packetfence-soh" ]; then
+		ln -s /etc/freeradius/sites-available/packetfence-soh /etc/freeradius/sites-enabled/packetfence-soh
+	fi
+	if [ ! -e "/etc/freeradius/sites-enabled/packetfence-tunnel" ]; then
+		ln -s /etc/freeradius/sites-available/packetfence-tunnel /etc/freeradius/sites-enabled/packetfence-tunnel
+	fi
+	sed -i "1i\suhosin.session.encrypt=Off\n" /etc/php5/apache2/php.ini
+	chmod 755 /usr/local/pf/sbin/pfdetect /usr/local/pf/sbin/pfdhcplistener /usr/local/pf/sbin/pfmon /usr/local/pf/sbin/pfredirect /usr/local/pf/sbin/pfsetvlan
+	chmod 744 $PACKETFENCE/installer.pl $PACKETFENCE/configurator.pl $PACKETFENCE/html/admin/soh.cgi
```
obilodeau: is bin/pfcmd properly setuid to root?

debian/packetfence.postrm
```diff
@@ -0,0 +1,41 @@
...
+
+
+case "$1" in
+    upgrade|failed-upgrade|abort-install|abort-upgrade|disappear)
+    ;;
+
+    purge|remove)
+        update-rc.d packetfence remove >/dev/null || exit 0
+	rm -fr /usr/local/pf
+    ;;
```
obilodeau: ~~(I think we should rm only on purge.)~~ addressed

debian/packetfence.postinst
fdurand: Bind 9.7.3

debian/packetfence.postinst
fdurand: installer.pl do it

obilodeau: installer.pl is not run on upgrades and I wonder if the permissions won't be overwritten by dpkg on upgrades?

obilodeau:
Other tasks
- ~~(configuration files will need to be handled as we do with the RPM)~~
- ~~(init script with a condrestart for high-availability setups (like the RPM version))~~

debian/rules
```diff
@@ -0,0 +1,108 @@
...
+	install -d -m0700 $(CURDIR)/debian/packetfence$(PREFIX)/$(NAME)/var/rrd
+	install -d -m0700 $(CURDIR)/debian/packetfence$(PREFIX)/$(NAME)/var/session
+	install -d -m0700 $(CURDIR)/debian/packetfence$(PREFIX)/$(NAME)/var/webadmin_cache
+#PacketFence remote Snort sensor
+	install -d -m0700 $(CURDIR)/debian/packetfence-remote-snort-sensor$(PREFIX)/$(NAME)/sbin
+	install -d -m0700 $(CURDIR)/debian/pfdetect_remote.conf $(CURDIR)/debian/packetfence-remote-snort-sensor$(PREFIX)/$(NAME)/conf
+	$(INSTALL) addons/pfdetect_remote/sbin/pfdetect_remote $(CURDIR)/debian/packetfence-remote-snort-sensor$(PREFIX)/$(NAME)/sbin
+	$(INSTALL) addons/pfdetect_remote/conf/pfdetect_remote.conf $(CURDIR)/debian/packetfence-remote-snort-sensor$(PREFIX)/$(NAME)/conf	
+# Build architecture-independent files here.
+binary-indep: build install
```
obilodeau: How does the rules file knows if it's doing pfdetect_remote or not?

fdurand: It knows from the directory $(CURDIR)/debian/packetfence-remote-snort-sensor

obilodeau: Great!

obilodeau:
~~(permission handling changes done. Can ownership changes be made the same way or not?)~~ let's go with what we currently do for now

debian/packetfence.postinst
fdurand: change done

fdurand:
Rebased

obilodeau:
good to go!

PR #11, opened 2012-04-20
fix for #1405: trap limit behavior

obilodeau:
urgent review required


obilodeau:
tested in lab several scenarios (disabled, no action, action shut)
minor issue filed: [#1431: trap limiting threading issue](http://www.packetfence.org/bugs/view.php?id=1431)

PR #12, opened 2012-04-25
Feature/h3c macauth

dwlfrth:
Supports for H3C S5120 switches series


lib/pf/SNMP/H3C.pm
```diff
@@ -0,0 +1,211 @@
...
+    my $result = $this->{_sessionRead}->get_request( -varbindlist => [ "$OID_hh3cLswSysVersion.$slotNumber" ] );
+    $result = $result->{"$OID_hh3cLswSysVersion.$slotNumber"};
+
+    if ( (exists($result->{"$OID_hh3cLswSysVersion.$slotNumber"}))
+             && ($result->{"$OID_hh3cLswSysVersion.$slotNumber"} ne 'noSuchInstance') ) {
+        return $result->{"$OID_hh3cLswSysVersion.$slotNumber"};
+    } else {
+        return;
+    }
+}
```
obilodeau, line 135: add error-handling to log the error in $logger->warn().

Something like (this is the error handling in pf::SNMP::Nortel):

``` perl
    # Error handling
    if (!defined($result)) {
        $logger->warn("Asking for port information failed with " . $this->{_sessionRead}->error());
        return;
    }

    if (!defined($result->{"$oid_rcVlanPortType.$ifIndex"})) {
        $logger->error("Returned value doesn't exist!");
        return;
    }

    if ($result->{"$oid_rcVlanPortType.$ifIndex"} eq 'noSuchInstance') {
        $logger->warn("Asking for port information failed with noSuchInstance");
        return;
    }
```

Instead of the above, what you could do is handle the success case first (return result if defined and not 'noSuchInstance') and then do the error handling stuff.

lib/pf/SNMP/H3C.pm
```diff
@@ -0,0 +1,211 @@
...
+    return ( $self->{_VoIPEnabled} == 1 );
+}
+
+=cut
+
+=item NasPortToIfIndex
+
+Same as pf::SNMP::ThreeCom::Switch_4200G
+
+=cut
```
obilodeau, line 167: add something like:

TODO: consider subclassing ThreeCom to avoid code duplication

but we won't do it now, later is fine.

lib/pf/SNMP/H3C/S5120.pm
```diff
@@ -0,0 +1,46 @@
...
+
+=head1 NAME
+
+pf::SNMP::H3C::S5120 - Object oriented module to access and configure enabled H3C S5120 switches.
+
+=head1 STATUS
+
+This module is currently only a placeholder, see pf::SNMP::H3C.
+
+=cut
```
obilodeau, line 11:
```
change 'see pf::SNMP::H3C' to 'see L<pf::SNMP::H3C>'
```

with this you get a nice hyperlink that works fine on our published documentation on the web site

obilodeau:
review completed. once the very minor things will be in, this will be good to go.

Have you run the test suite on it? ;)

fixed in b07a6d2f

obilodeau:
I will integrate it into integration for tonight's build.

lib/pf/SNMP/H3C.pm
dwlfrth, line 135: fixed in 00d28a28

lib/pf/SNMP/H3C.pm
dwlfrth, line 167: fixed in 00d28a28

lib/pf/SNMP/H3C/S5120.pm
dwlfrth, line 11: fixed in 00d28a28

obilodeau:
made small changes, running tests

obilodeau:
tests passed, merging

PR #13, opened 2012-05-04
task list for feature/catalyst-configurator

obilodeau:
This is too early for real merge. It is only meant to track the work still to do on that branch.
### Before next image:
- b3b3316: ~~if step5 can't list the services, it should display an error why.~~ OB
- ~~apply `init.d/gdm3` changes to the usb image~~
- ~~have the configurator server log somewhere in `/usr/local/pf/logs/` OB~~
- ~~add `ping` and `nslookup` to the usb image~~ FD
- ~~type 'other' management (set to None if no config), going back to step1 must not give 'other'~~
- ~~image: disable dhcp on interfaces~~ FD
### Next image:
- 8b31ef7 ~~bug: `completed->{stepX}` not properly setup when we reach step5.~~ FL
  - ~~problem seems to come from step 2 : completed    => { step1 => 1, step3 => 1, step4 => 1 }~~
- ~~image: SSL self-signed cert to be pre-generated DW~~
  - ~~image: cert not in proper location~~ FD
- d32ce47 ~~step1 inject proper default gateway~~ DW
- d5b245c ~~persistent interface configuration for debian~~ DW
- ~~image: ip_forward should be enabled by default~~ FD

Then, next test iteration, see remaining issues.
### Found issues for next test iteration:
- ~~missing installer steps:~~ OB
  - Creating empty log files
  - chown -R pf:pf on /usr/local/pf/
  - chown root:root bin/pfcmd
  - chmod 6755 bin/pfcmd
- ~~image: user `pf` does not exist~~ FD
- aabc74e ~~image: do not re-open the configurator if it was already run (fix: if currently_at is set, redirect to admin)~~ DW
- ~~image: some things are not persistent. ex: `/etc/network/interfaces`~~
- e2540f26ff ~~interface type 'other' keeps re-appearing (see DW to reproduce)~~
### Before merge:
- making sure that the builds passes
- documentation updates
- new init.d files for configurator (should rename liveusb's configurator to packetfence-{admin,app,appserver,backend} or pf-appserver or something else more meaningful)
### After merge:
- ~~pressing Enter should work in more places (tested not working on set IP dialog and DNS on step1)~~
- ~~interface type 'other' becomes enabled when messing with enforcement types, etc.~~
- 547d567c2a ~~cursor is not located at a friendly location for keyboard navigation between the screens (at least on iceweasel)~~ 
- b89eb925ba ~~add default mask values to 255.255.255.0~~
- add validation for fields, 8361d839a5 ~~especially VLAN tags~~
- e1901f15e6 ~~provide with current gateway~~
- multi-process issues when running with: `CATALYST_ENGINE='HTTP::Prefork' script/yourapp_server.pl` as we would like to do in production
- 5912b5f802 ~~restore enforcement mechanisms checkboxes (inline and/or vlan) in step 1~~
- ~~test persistant gateway / interfaces in debian~~
- ~~add error handling in inject_default_gateway with pf_run...~~
- 40d783a73e ~~step 2 (network) should check for return codes~~
### Ludo bugs list:
- ~~editing a disabled and unconfigured interface keeps the modal opened but the change works~~
- ~~vlan creation reporting that vlan already exists but works~~
- ~~if encounter an error and a modal is opened, the error will also be visible from within the modal~~
- ~~after vlan creation, try enabling it says interface is already enabled... shouldn't we refresh the list on each change cause the vlan creation automatically enable the interface~~ host interface needs to be up for the status to show
- 92ce692 ~~root username by default on database configuration step~~
- multiprocess bug on setting a new password for the mysql root account
- ~~error message when creating the database but works~~
- e3798a0 ~~add a message to say that most of the time, simply enters root without any pass (for mysql unsecure installations)~~
- ~~choose inline enforcement but don't select any inline interface won't show an error but will stays on the page (will give a json error)~~
- f621b7a ~~should have persistence on interfaces type selection between the modal windows~~

Meta-tracker for issues to be resolved by branch: http://packetfence.org/bugs/view.php?id=1444


dwlfrth:
CLOSED since this has been merged in devel for release 3.5.0

PR #15, opened 2012-05-08
fix/perl-suid into devel

obilodeau:
going into devel

See issue [#1438: perl 5.12+ and/or modern distro and setuid script](http://packetfence.org/bugs/view.php?id=1438).

Not ready yet, build system needs to be modified so that it'll handle new subpackages more cleanly. Mimic what has been done under debian.

TODO
- ~~(rpm: bin/pfcmd.pl is still setuid)~~
- ~~(security: we should probably enforce taint-mode in the bin/pfcmd.pl #! line (perl -T I think). taint mode was forced to On when running a setuid binary before but since our little wrapper, I'm pretty it'll need to be explicit (maybe not?))~~


fdurand:
get_all_perl_binaries() has been updated to remove /usr/local/pf/bin/pfcmd
Removed setuid on pfcmd.pl
Added -T to pfcmd.pl

obilodeau:
Another thing I just thought.. the perl-suid subpackage should probably depend on packetfence >= 3.4.0

This way you cannot remove it by accident. After all, it is useless completely useless without packetfence (as opposed to docs or tests).

obilodeau:
This branch should be re-based on stable as it currently can't merge in integration easily. Due to all the packages changes in our catalyst work..

obilodeau:
re-based on devel. will go in integration and will be reviewed shortly.

addons/packages/packetfence.spec
```diff
@@ -36,6 +36,7 @@ URL: http://www.packetfence.org
 AutoReqProv: 0
 BuildArch: noarch
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{rev}-root
+%define debug_package %{nil}
```
obilodeau: what is this for?

fdurand: It´s for disabling the debug package

obilodeau:
sent in integration

obilodeau:
alright, tested on my Fedora box with Perl 5.14.2 and good to go. I'm merging this in devel right now, we will fix the nightly build scripts next week.

obilodeau:
another issue: generated pfcmd-suid sub-package on my fedora was noarch instead of x86-64: `packetfence-pfcmd-suid-3.6.0-0.20120903.fc17.noarch.rpm`

obilodeau:
Looks like we are having a real problem on RHEL regarding arch-specific subpackage:

```
> Unfortuantely noarch package cannot have arch subpackage, so nimbus-theme-gnome
> must be arch as well.
```

from: https://bugzilla.redhat.com/show_bug.cgi?id=488174#c6

obilodeau:
Ok, so it seems the recommended way to do it is a virtual top-level package with one noarch and one arch-specific subpackage. Here's [an example](https://bugzilla.redhat.com/attachment.cgi?id=344264).

obilodeau:
With the latest changes, here's a result of the build:

```

Wrote: /home/olivier/rpmbuild/SRPMS/packetfence-3.6.0-0.20120904fc17.src.rpm
Wrote: /home/olivier/rpmbuild/RPMS/noarch/packetfence-core-3.6.0-0.20120904fc17.noarch.rpm
Wrote: /home/olivier/rpmbuild/RPMS/noarch/packetfence-remote-snort-sensor-3.6.0-0.20120904fc17.noarch.rpm
Wrote: /home/olivier/rpmbuild/RPMS/x86_64/packetfence-pfcmd-suid-3.6.0-0.20120904fc17.x86_64.rpm
```

Looking good. However I'm scared of hard-to-predict side effects of such a change.. @fdurand once you approve, let's have that in devel as soon as we can so we can adapt the build sytem and ask the community to test it out.

obilodeau:
Jean had a good point, I wonder if someone upgrading packetfence will pull in packetfence-core...

Relevant link: http://fedoraproject.org/wiki/Packaging:Guidelines#Renaming.2FReplacing_Existing_Packages

obilodeau:
After some tests in VMs, I decided to try renaming the virtual package (packetfence to packetfence-virtual) and rename packetfence-core to packetfence instead. This was done here: d7f95fec2ae0a9e56d2552fb9e9f1aae27878984

obilodeau:
I just successfully upgraded a CentOS 6 VM from a packetfence stable (only packetfence package) to a splitted packetfence + packetfence-pfcmd-suid package. Setuid C wrapper is properly installed and worked.

Ready for review and should be merged in devel pretty soon. Build scripts will be adapted after.

PR #16, opened 2012-05-08
getting rid of [interface xxx].gateway

obilodeau:
Large change. Reviewing by commit is probably easiest.

Will go into integration.

Meant for devel branch.


obilodeau:
We experienced too many regressions regarding rogue DHCP, so we added detection of locally running dhcp servers based on `conf/networks.conf` information.

dwlfrth:
Seems ok from my point of view.
That is a lot of changes!

PR #17, opened 2012-05-15
Support for Aironet in Wireless Directory Services (WDS) mode

obilodeau:
ready for review

working in a lab

doc TODO
- network device guide documentation:
  - the fact that the radius dynauth server runs on a non-standard port by default (not 3799)
  - MAC in format aaaa.bbbb.cccc
- module POD bugs and limitations:
  - no Service-Type otherwise the WDS crashes (confirmed on IOS 12.3.8JEC3)

We should maybe do this also while we are at it: [#1452: Aironet should perform deauth with RFC3576 Packet of Disconnect (PoD)](http://packetfence.org/bugs/view.php?id=1452)

### DIFF
--- a/lib/pf/SNMP/Cisco/Aironet.pm
+++ b/lib/pf/SNMP/Cisco/Aironet.pm
@@ -55,6 +55,7 @@ use Net::SNMP;
 use base ('pf::SNMP::Cisco');
 
 use pf::config;
+use pf::util qw(format_mac_as_cisco);
 
 =head1 SUBROUTINES
 
@@ -83,17 +84,10 @@ sub deauthenticateMac {
     my ( $this, $mac ) = @_;
     my $logger = Log::Log4perl::get_logger( ref($this) );
 
-    #format MAC
-    if ( length($mac) == 17 ) {
-        $mac =~ s/://g;
-        $mac
-            = substr( $mac, 0, 4 ) . "."
-            . substr( $mac, 4, 4 ) . "."
-            . substr( $mac, 8, 4 );
-    } else {
-        $logger->error(
-            "ERROR: MAC format is incorrect ($mac). Should be xx:xx:xx:xx:xx:xx"
-        );
+    $mac = format_mac_as_cisco($mac);
+    if ( !defined($mac) ) {
+        $logger->error("ERROR: MAC format is incorrect. Aborting deauth...");
+        # TODO return 1, really?
         return 1;
     }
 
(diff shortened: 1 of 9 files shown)


lib/pf/SNMP/Cisco/Aironet.pm
```diff
@@ -83,17 +84,10 @@ sub deauthenticateMac {
...
-            . substr( $mac, 4, 4 ) . "."
-            . substr( $mac, 8, 4 );
-    } else {
-        $logger->error(
-            "ERROR: MAC format is incorrect ($mac). Should be xx:xx:xx:xx:xx:xx"
-        );
+    $mac = format_mac_as_cisco($mac);
+    if ( !defined($mac) ) {
+        $logger->error("ERROR: MAC format is incorrect. Aborting deauth...");
+        # TODO return 1, really?
```
dwlfrth, line 90: return 1 ? i don't get it neither

lib/pf/util.pm
```diff
@@ -181,6 +181,25 @@ sub format_mac_for_acct {
...
+Put the mac address in the cisco format, accepting xx:xx:xx:xx:xx:xx
+
+Returning format aabb.ccdd.eeff
+
+=cut
+sub format_mac_as_cisco {
+    my ($mac) = @_;
+
+    if (defined($mac) && 
+        $mac =~ /^([0-9a-f]{2}):([0-9a-f]{2}):([0-9a-f]{2}):([0-9a-f]{2}):([0-9a-f]{2}):([0-9a-f]{2})$/
```
dwlfrth, line 195: capitals will not pass... but as far as i remember, every mac in PF is lowered... or is it?

lib/pf/SNMP/Cisco/WLC_2100.pm
```diff
@@ -78,17 +79,10 @@ sub _deauthenticateMacSnmp {
...
-            . substr( $mac, 4, 4 ) . "."
-            . substr( $mac, 8, 4 );
-    } else {
-        $logger->error(
-            "ERROR: MAC format is incorrect ($mac). Should be xx:xx:xx:xx:xx:xx"
-        );
+    $mac = format_mac_as_cisco($mac);
+    if ( !defined($mac) ) {
+        $logger->error("ERROR: MAC format is incorrect. Aborting deauth...");
+        # TODO return 1, really?
```
dwlfrth, line 85: you again! seems strange to return 1 on an error...

dwlfrth:
Merged in devel. Thanks for being with us!

PR #18, opened 2012-05-16
dynamic expansion of customVlans

obilodeau:
Ready to roll into stable (or devel) as we wish.

Someone please ACK.

TODO
- once merged, have a 'How can I get more customVlan?' FAQ entry ready that explains that ui.conf must be modified in order to display the new VLANs on the CLI and Admin UI.


dwlfrth:
seems good!

PR #19, opened 2012-05-25
pfdhcplistener hang fixes

obilodeau:
http://packetfence.org/bugs/view.php?id=1453

Please someone review. It's not too complex so it should be fairly quick.


cgx:
Looks good to me!

PR #21, opened 2012-05-31
Fixed guest import w/ Windows line terminators files

obilodeau:
Someone please ACK for integration into stable.


obilodeau:
This should go in 3.3.3, can anyone review? It was tested 'in the field' and worked so a quick glance and a comment like "I haven't puked while looking at this code" should be enough.

dwlfrth:
Seems good. Altough, don't like the multiple PID regex

obilodeau:
The regex in pf::pfcmd would have needed splitting and a rather extensive rework.. Plus it should have been in pf::person in the first place.. Then there's the regex also in pf::pfcmd::pfcmd...

It's a mess. :(

PR #22, opened 2012-05-31
bulk import now supports default node properties in pf.conf

obilodeau:
So in `pf.conf` with:

```
[node_import]
category=staff
```

It will import and register the nodes with a staff category.

pid, category and voip (yes/no) is supported.

Someone please ACK.


cgx:
Looks good to me!

PR #24, opened 2012-06-04
Feature/acct reporting

fgaudreault:
Ready to review.


obilodeau:
I think using the term bandwidth instead of accounting would have been clearer since accounting could also be used for other things (time connected and such).

For violations we went with accounting and we are receiving requests for violation based on accounting session duration so I think bandwidth would have been a better choice there..

I'm not sure if I'm going to change it or not. I'll try it first.

obilodeau:
top 25 should still be a pie chart

obilodeau:
quick task list reminder for myself
- ~~s/accounting/bandwidth/~~ (including odd bug to find and fix. see e8c1bec)
- ~~update help (supported parameters in front of report types)~~
- ~~remove top25 from grammar / help~~
- ~~test: node-accounting pie chart~~
- ~~make clearer that the delay is a window~~
- ~~update grammar so we can't call stuff like this:~~
  
  $ bin/pfcmd report osclass daily
  No such sub report_osclass_daily at line 2022.
  description|percent|count
- ~~php, avoid large copy and pasted if, pack in a foreach with array lookups instead~~
- ~~eliminate hardcoded constants~~

obilodeau:
ready for review by another dev

lib/pf/pfcmd/report.pm
```diff
@@ -580,6 +641,114 @@ sub report_ssid_active {
...
+    return (@return_data);
+}
+
+=item * report_osclassbandwidth_all
+
+Reporting - OS Class bandwitdh usage - All time
+
+=cut
+sub report_osclassbandwidth_all {
+    return report_osclassbandwidth(999999999);
```
dwlfrth: is this 9999999999 completely arbitrary ?
usually when we talk about 'all' or everything, we use 0

obilodeau: 100% arbitrary unfortunately. My guess is that Francois was lazy and didn't want to do separate query without time constraints...

dwlfrth:
seems good!

dwlfrth:
seems good!

obilodeau:
thanks dawg! going in.

PR #26, opened 2012-06-07
pid (username) of the form 'domain\\user' are now allowed

obilodeau:
http://packetfence.org/bugs/view.php?id=1253

Tested in 'in the field' and worked. Pull request opened mostly so reviewers get a hold of our complex grammar setup.


dwlfrth:
if it works "in the field" !! code seems pretty straight forward!

PR #27, opened 2012-06-07
config parameter for HTTP load-balacing support

obilodeau:
It will trigger proper IP lookup in HTTP Headers if source IP is from
the load balancers.

Tested in lab and works.


dwlfrth:
seems fine! ;)

PR #28, opened 2012-06-11
more flexibility in services launcher (required for debian fix #1463)

obilodeau:
Not only flags are defined in pf::services now but because of a sprintf
syntax we can also use it to pass environment variables too.

This change is required for the upcoming Debian fix #1463
http://packetfence.org/bugs/view.php?id=1463


obilodeau:
reviewed under pressure by @dereksky

PR #29, opened 2012-06-12
a fix for #1464: process management issues (pid misuse in listing and killing daemons)

obilodeau:
http://packetfence.org/bugs/view.php?id=1464


dwlfrth:
seems ok! as long as tested "in lab field" ;)

PR #34, opened 2012-07-19
Feature/acct violations

fgaudreault:
For code review.  Still need some finish, but it is working.


lib/pf/accounting.pm
```diff
@@ -62,7 +65,7 @@ Readonly our $ACCOUNTING_TRIGGER_RE => qr/
     ($BANDWIDTH_DIRECTION_RE)     # bandwidth direction
     (\d+)                         # nb of bandwidth units
     ($BANDWIDTH_UNITS_RE)         # bandwidth units
-    (\d+)?($TIME_MODIFIER_RE)?    # optional time window (number + time modifier)
+    ($TIME_MODIFIER_RE)?          # optional time window (number + time modifier)
```
cgx: No more number before the time unit?

fgaudreault: Correct.  We now have a fixed window length (present day, present month, etc) instead of sliding window.

fgaudreault:
You might also check the 2 new icons added (green/red circles).  Maybe you want to have some more 'stylish' icons for those :P

lib/pf/accounting.pm
dwlfrth: should adjust the comment accordingly ;)

lib/pf/accounting.pm
```diff
@@ -234,34 +361,72 @@ sub acct_maintenance {
...
+       	       	    $interval =	"monthly";
+                } elsif	($4 eq 'Y') {
+       	       	    $interval =	"yearly";
+                }
             } 
             # no interval given so we assume from beginning of time
             else {
-                # mysql specific
-                #This should bring us in 1695, and PF was not yet existing :)
-                #The funny thing is that it's the birth year of Jean De La Fontaine.. alright no one cares :)
```
dwlfrth, line 250: i loled :)

lib/pf/accounting.pm
fgaudreault: Fixed.

lib/pf/action.pm
```diff
@@ -32,6 +32,7 @@ Readonly::Scalar our $TRAP => 'trap';
 Readonly::Scalar our $LOG => 'log';
 Readonly::Scalar our $EXTERNAL => 'external';
 Readonly::Scalar our $WINPOPUP => 'winpopup';
+Readonly::Scalar our $CLOSE => 'close';
```
dwlfrth, line 35: usage?

lib/pf/violation.pm
```diff
@@ -463,9 +474,42 @@ sub violation_trigger {
...
-        $logger->info("calling '$bin_dir/pfcmd violation add vid=$vid,mac=$mac' (trigger ${type}::${tid})");
+        # Compute the release date
+        my $date = 0;
+        
+        # Check if we have a window defined for the violation, and act properly 
+        # TODO: Handle the "dynamic" keyword
+        my $class = class_view($vid);
+
+        if (defined($class->{'window'})) {
+          if ($class->{'window'} ne 'dynamic' && $class->{'window'} ne '0' ) {
```
dwlfrth, line 485: indentation seems problematic

lib/pf/action.pm
fgaudreault, line 35: The same as others, nothing :)  I guess it's there to avoid hard-coding the eq 'close' part.

dwlfrth, line 35: ahah ok ! just wondering if i were crazy of seeing a new constant without using it :)

fgaudreault, line 35: I am not using it, just like we are not using the other ones.  It's there in case of.

PR #41, opened 2012-08-06
Feature/ipset

fdurand:
for review

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -15,7 +15,8 @@ For a list of compatibility related changes see the UPGRADE file.
 --------------------------------------------------------------------------------
 Version <stableRelease> released on <releaseDate>
 
-Enhancements
+New Features
+ * Adding ipset support for inline mode
 
 Bug Fixes
 
--- a/NEWS
+++ b/NEWS
@@ -17,6 +17,8 @@ Version <stableRelease> released on <releaseDate>
 
 New Features
  * Adding ipset support for inline mode
+ * Adding support for multivlan in inline mode
+ * Adding support of portal profil by vlan
 
 Bug Fixes
 
--- a/NEWS
+++ b/NEWS
@@ -65,6 +65,7 @@ New Features
  * Adding a new web based configurator that helps the installation and configuration process
    of a new PacketFence installation
  * Adding support for Suricata IDS (#1141)
+ * Using ipset to replace IPTables::ChainMgr (if ipset is not found we use IPTables::ChainMgr)
 
 Enhancements
  * Improved handling of an empty conf/pf.conf configuration file
--- a/NEWS
+++ b/NEWS
@@ -65,7 +65,7 @@ New Features
  * Adding a new web based configurator that helps the installation and configuration process
    of a new PacketFence installation
  * Adding support for Suricata IDS (#1141)
- * Using ipset to replace IPTables::ChainMgr (if ipset is not found we use IPTables::ChainMgr)
+ * Using ipset to replace IPTables::ChainMgr for RHEL6/Centos6/Debian (if ipset is not found we use IPTables::ChainMgr)
 
 Enhancements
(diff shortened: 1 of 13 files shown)


NEWS
```diff
@@ -13,6 +13,13 @@ For more details and developer visible changes see the ChangeLog file.
 For a list of compatibility related changes see the UPGRADE file.
 
 --------------------------------------------------------------------------------
+Version <devel> released on <releaseDate>
+
+New Features
+ * Adding ipset support for inline mode
+ * Adding support for multivlan in inline mode
+ * Adding support of portal profil by vlan
```
dwlfrth: s/profil/profiles

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -304,6 +304,11 @@ server, DHCP server, RADIUS server) using:
   sudo apt-get update
   sudo apt-get install packetfence
 
+In order to use ipset in inline mode, you must install two news dependencies and compile kernel modules:
```
dwlfrth, line 310: That is the Debian based doc... is there anything for RHEL?

fdurand, line 310: For RHEL5 no way to use ipset , for RHEL6 , ipset is in the spec

obilodeau, line 310: It should be documented as such and the NEWS entry should mention it too. I would probably do it in a ipset dedicated section. Not really sure where to put it though..

obilodeau:
Does this fixes #1424 for all supported platforms? http://packetfence.org/bugs/view.php?id=1424

debian/control
```diff
@@ -16,6 +16,8 @@ Depends: ${misc:Depends}, apache2, apache2.2-common, apache2-utils,
  openssl, openssl-blacklist, openssl-blacklist-extra, php-log, snort, 
  mysql-server, libapache2-mod-proxy-html, libapache2-mod-php5, php-pear,
  php5-mysql, php5-gd, perl-suid, make,
+# ipset (ipset must be installed by hand  apt-get install xtables-addons-common  xtables-addons-source)
```
obilodeau, line 19: installed by hand? reason?

fdurand:
Debian , centos 6 and RHEL6, not centos 5

debian/control
fdurand, line 19: You must have to compil the kernel module by module-assistant auto-install xtables-addons.
So when you install packetfence with apt-get you can´t build the module in the init script (it use dpkg that is already locked).
So i prefer to document that and install by hand.

obilodeau, line 19: and how will that be handled in the liveusb creation process?

obilodeau:
re: not on centos 5

Then one of the alternatives we discussed back in june will have to be considered. Unless we are willing to drop support for it in our next feature release. Also, I'm not sure if I buy the sudoers idea. I opposed to it earlier but you still took that direction...

debian/control
fdurand, line 19: The kernel module must be build the first time the live usb is running.
And i have a problem with installing xtables-addons-common  xtables-addons-source during the build, it remove packetfence.
So by hand.

fdurand:
It´s also be possible to implement ipset on centos 5, but we have to make kernel module for at least some of the last kernel version.
For sudoers ,i was inspired by perl Captive::Portal that i think is the better solution

fdurand:
Warning on restart:
http://www.packetfence.org/bugs/view.php?id=1539

dwlfrth:
Do we need to keep this one opened since the whole thing have been rewritten object style?

fdurand:
no, but keep this branch on github yes, it has been deployed in some clients network

dwlfrth:
Closed since this has been reworked on an object oriented base
#59

PR #44, opened 2012-08-09
sponsorship_cc config item for guests notification

obilodeau:
ready for review / integration


dwlfrth:
such an intrusive change... we should discuss it ;)
ahah seems good to be integrated

obilodeau:
passing integration, will be merged today

PR #46, opened 2012-08-15
Sanitizing WS_USER and WS_PASS before passing them to the SOAP call

dwlfrth:


cgx:
Can't we avoid the duplication of the sanitize subroutine?

obilodeau:
@cgx has a good point. Now that we're managed radius, we could share a lot of code in a pf-lib and import it from the modules. Insuring thread safety would be paramount.

I guess, merging the duplicated code in stable is ok to get the fix out as quick as possible and in devel we could experiment with the shared radius perl lib idea.

obilodeau:
one critic: quoted eval construct (eval "STRING") is not recommended and perlcritic doesn't like it

otherwise this is sexyyy

obilodeau:
passing integration

dwlfrth:
as @obilodeau said, we should get this one out as fast as possible but continue working on it in devel.
Anyway we had plan to possibly remove authentication + https when RADIUS is on localhost.

We should then rework the user/pass thing for remote radius but anyway, in the case of remote radius we no longer enter the case of "managed radius"...

obilodeau:
This will be integrated into stable tomorrow.

obilodeau:
related ticket: [#1516: illegal characters in webservices user / pass not properly escaped](http://packetfence.org/bugs/view.php?id=1516)

Needs testing in lab so not merging until the end of this release crunch.

obilodeau:
@cgx's comment is addressed in #58

PR #47, opened 2012-08-16
quick fixes for the guest management login and auth issues (#1518)

obilodeau:
Avoided API changes by splitting subs into admin specific subs at the
cost of code duplication.

http://packetfence.org/bugs/view.php?id=1518


obilodeau:
passing integration

cgx:
Looks good.

PR #52, opened 2012-08-22
person pid parsing regression #1523

obilodeau:
The pid parser allowed spaces even in the unquoted format which means that
it greadily expanded to grab the next token (pid firstname=...) instead of
just sticking to it's pid.

Regression introduced in `manage register` in 3.0 and further exposed to
person commands in 3.4.0.

Regression test included.

http://packetfence.org/bugs/view.php?id=1523


obilodeau:
passing integration

dwlfrth:
Seems good !

PR #53, opened 2012-08-28
override-able deauth

fdurand:
For review


bin/pfcmd_vlan
```diff
@@ -479,7 +479,15 @@ if ($reevaluateAccess) {
         print "Can not instantiate switch $switchDescRegExp ! See log files for details\n";
     } else {
         $logger->debug("start handling 'deauthenticate' command");
-        $switch->deauthenticateMac($mac);
+        my $deAuthMethod = $switch->{_deAuthMethod};
+        $logger->warn($deAuthMethod);
```
cgx: Is this for debugging? If so, call "debug" instead of "warn".

lib/pf/SNMP.pm
```diff
@@ -2748,6 +2751,34 @@ sub returnRadiusAccessAccept {
...
 }
 
+=item supportedDeauthTechniques
+
+return Default Deauthentication Method
+
+=cut
+sub supportedDeauthTechniques {
+    my ( $this ) = @_;
+    my $logger = Log::Log4perl::get_logger( ref($this) );
```
cgx: It doesn't look necessary to instantiate a logger here.

bin/pfcmd_vlan
```diff
@@ -479,7 +479,14 @@ if ($reevaluateAccess) {
...
-        $switch->deauthenticateMac($mac);
+        my $deAuthMethod = $switch->{_deAuthMethod};
+        my %supportedDeauthTechniques = $switch->supportedDeauthTechniques;
+        if ((defined($deAuthMethod)) && (defined($supportedDeauthTechniques{$deAuthMethod}))) {
+            $supportedDeauthTechniques{$deAuthMethod}($switch,$mac);
+        }
+        else {
+            $switch->deauthenticateMacDefault($mac);
+        }
         $logger->debug("finished handling 'deauthenticate' command");
```
obilodeau: The complexity should be in `pf::SNMP` and not the burden of the callers. Looking in _deAuthMethod here is a violation of the abstraction.

Suggestion: The call should still be ->deauthenticateMac(...) and in pf::SNMP's deauthenticateMac then the lookup and calling the appropriate technique should be performed. This also applies to `sbin/pfsetvlan`.

obilodeau:
General notes

I'm not convinced about the hash approach. Every time you ask the technique, you need to call defined() on it and then call the default in the else _even_ if default should already be provided when you ask for the technique.

With an array you would be able to just get the first one ignore the rest and call it. The configuration would just change the array order. But, it _is_ convenient to have the deauth technique style and be able to fix our thread issues with `sbin/pfsetvlan` like I saw you did...

I guess, instead of supportedDeauthTechniques, it could be preferredDeauthTechnique and returns `'name' => CODEREF` (ie: `SSH => { ... }`). This would get rid of the `if {} else {}` (return the default if nothing else is implemented) and handling overloaded configuration would also be possible (with the ugly `if {} else {}` only in one location: the module itself).

Hopefully, this was clear..

Quick items
- globally rename deAuthMethod to deauthMethod
- SSH, Telnet, RADIUS constants should be fully qualified (not even sure why it works right now): `$SNMP::SSH` instead of `$SSH` also make sure to `use pf::SNMP::constants` in all the consumers of the constants.

I always use the hash approch but i change the function supportedDeauthMethod by DeauthMethod that return the method and the ref of the deauth function and if the method doen´t exist i return the default method.

html/admin/configuration/switches_add.php
```diff
@@ -208,6 +208,11 @@
         printSelect( array('' => 'please choose', 'yes' => 'Yes', 'no' => 'No'), 'hash', $val, "name='$key'");
         break;
 
+      case 'deAuthMethod':
+        print "<tr><td></td><td>$pretty_key:</td><td>";
+        printSelect( array('' => 'please choose', 'Telnet' => 'Telnet', 'SSH' => 'SSH', 'SNMP' => 'SNMP', 'Radius' => 'Radius', 'Http' => 'Http', 'Https' => 'Https'), 'hash', $val, "name='$key'");
+        break;
+
       default:
```
obilodeau: Radius -> RADIUS as it is an acronym. same goes for HTTP, HTTPS

html/admin/configuration/switches_edit.php
```diff
@@ -205,6 +205,11 @@
         printSelect( array('' => 'please choose', 'yes' => 'Yes', 'no' => 'No'), 'hash', $val, "name='$key'");
         break;
 
+      case 'deAuthMethod':
+        print "<tr><td></td><td>$pretty_key:</td><td>";
+        printSelect( array('' => 'please choose', 'Telnet' => 'Telnet', 'SSH' => 'SSH', 'SNMP' => 'SNMP', 'Radius' => 'Radius', 'Http' => 'Http', 'Https' => 'Https'), 'hash', $val, "name='$key'");
+        break;
+
       default:
```
obilodeau: Radius -> RADIUS as it is an acronym. same goes for HTTP, HTTPS

lib/pf/SNMP.pm
```diff
@@ -2748,6 +2751,33 @@ sub returnRadiusAccessAccept {
...
+=cut
+sub deauthenticateMacDefault {
+    my ( $this ) = @_;
+    my $logger = Log::Log4perl::get_logger( ref($this) );
+
+    $logger->error("Deauthentication is not supported on switch type " . ref($this));
+    return $FALSE;
+}
+
 =back
```
obilodeau: method vs technique. ideally one name should be used

lib/pf/SNMP.pm
```diff
@@ -2748,6 +2751,33 @@ sub returnRadiusAccessAccept {
...
+sub supportedDeauthTechniques {
+    my ( $this ) = @_;
+
+    my %tech = (
+        'Default' => \&$this->deauthenticateMacDefault,
+    );
+    return %tech;
+}
+
+=item supportedDeauthTechniques
```
obilodeau: wrong POD

lib/pf/SNMP/constants.pm
```diff
@@ -126,6 +126,17 @@ Check http://www.iana.org/assignments/ianaiftype-mib for the full list.
...
+
+Deauth type method constant
+
+=cut
+Readonly my $TELNET => 'Telnet';
+Readonly my $SSH => 'SSH';
+Readonly my $SNMP => 'SNMP';
+Readonly my $RADIUS => 'Radius';
+Readonly my $HTTP => 'Http';
+Readonly my $HTTPS => 'Https';
```
obilodeau: radius, http, https capitalization

lib/pf/pfcmd/pfcmd.pm
```diff
@@ -161,7 +161,7 @@ $grammar = q {
...
+   switchconfig_view_field : 'type' | 'mode' | 'uplink' | 'SNMPVersionTrap' | 'SNMPCommunityRead' | 'SNMPCommunityWrite' | 'SNMPVersion' | 'SNMPCommunityTrap' | 'cliTransport' | 'cliUser' | 'cliPwd' | 'cliEnablePwd' | 'wsTransport' | 'wsUser' | 'wsPwd' | 'vlans' | 'normalVlan' | 'registrationVlan' | 'isolationVlan' | 'macDetectionVlan' | 'guestVlan' | /customVlan\d\d?/ | 'macSearchesMaxNb' | 'macSearchesSleepInterval' | 'VoIPEnabled' | 'voiceVlan' | 'SNMPEngineID' | 'SNMPUserNameRead' | 'SNMPAuthProtocolRead' | 'SNMPAuthPasswordRead' | 'SNMPPrivProtocolRead' | 'SNMPPrivPasswordRead' | 'SNMPUserNameWrite' | 'SNMPAuthProtocolWrite' | 'SNMPAuthPasswordWrite' | 'SNMPPrivProtocolWrite' | 'SNMPPrivPasswordWrite' | 'SNMPUserNameTrap' | 'SNMPAuthProtocolTrap' | 'SNMPAuthPasswordTrap' | 'SNMPPrivProtocolTrap' | 'SNMPPrivPasswordTrap' | 'radiusSecret' | 'controllerIp' | 'roles' | 'deAuthMethod'
```
obilodeau: this has changed, we should rebase on latest release to avoid conflicts in integration and when merging the feature

bin/pfcmd_vlan
fdurand: deauthenticateMac is already used in Switch modules, maybe i could implement deauthenticateMacMethod in pf::SNMP that does the code above.

obilodeau: It would be an API breaking change. External callers were calling deauthenticateMac and they should still be calling deauthenticateMac.

All switches implementation should be updated. Their existing deauthenticateMac converted to deauthenticateMacDefault (I think that's the way you called it).

fgaudreault:
Couple cosmetics to fix :
- deAuthMethods -> deauthMethods
- DeauthTechniques -> deauthTechniques

lib/pf/SNMP/HP/Controller_MSM710.pm
```diff
@@ -169,6 +170,65 @@ sub extractSsid {
...
+
+=cut
+
+sub _deauthenticateMacWithSSH {
+    my ( $this, $mac ) = @_;
+    my $logger = Log::Log4perl::get_logger( ref($this) );
+    $logger->warn(Dumper $this);
+    my $session;
+    eval {
+        $session = new Net::Appliance::Session->new(
```
fgaudreault: Remove the new in front of the Net::Appliance::Session call.  It will not work since it will create a reference.

fgaudreault:
Looks good.  What do you think? Anyone wants to dive into it prior merging?

fgaudreault:
Merged in Devel.

PR #54, opened 2012-08-28
control NAT out and routing support

fdurand:
For review


lib/pf/iptables.pm
```diff
@@ -621,6 +652,21 @@ sub generate_passthrough {
...
 }
 
+=item get_snat_interface
+
+Return the list of network interface to enable SNAT.
+
+=cut
+sub get_snat_interface {
+    my ($self) = @_;
+    my $logger = Log::Log4perl::get_logger('pf::iptables');
```
dwlfrth: We don't log anything in this sub... can get rid of the logger.
If you REALLY want to keep use the PACKAGE rather than the package name

NEWS
```diff
@@ -15,6 +15,10 @@ For a list of compatibility related changes see the UPGRADE file.
 --------------------------------------------------------------------------------
 Version <stableRelease> released on <releaseDate>
 
+New Features
+ * Adding routing support for the inline mode
+ * Adding support to choose multiples DNAT interfaces for inline mode
```
dwlfrth: DNAT or SNAT?

conf/documentation.conf
```diff
@@ -857,6 +857,12 @@ telling users to open their browsers. Enable them via the ports.listeners
 parameter and add the IMAP (143/tcp) and POP3 (110/tcp) here.
 EOT
 
+[inline.interfaceSNAT]
+type=text
+description=<<EOT
+Interface that is use to SNAT the traffic
```
dwlfrth: Interface used to SNAT inline traffic

conf/iptables.conf
```diff
@@ -113,6 +113,11 @@ COMMIT
 # Make sure to adjust the FORWARD rules also to allow traffic back-in.
 :postrouting-int-inline-if - [0:0]
 %%nat_postrouting_inline%%
+#
+# Chain to enable routing instead of NAT
+#
+:postrouting-int-inline-nonat - [0:0]
```
dwlfrth: s/nonat/routed ?

lib/pf/iptables.pm
```diff
@@ -339,7 +370,7 @@ sub iptables_mark_node {
 
     $logger->debug("marking node $mac with mark 0x$mark");
     my $success = $iptables->iptables_do_command(
-        "-A $FW_PREROUTING_INT_INLINE",  "--match mac --mac-source $mac", "--jump MARK --set-mark 0x$mark"
+         "-A $FW_PREROUTING_INT_INLINE",  "--match mac --mac-source $mac", "--jump MARK --set-mark 0x$mark"
```
dwlfrth: why 5 spaces ? :P

lib/pf/iptables.pm
```diff
@@ -234,10 +249,26 @@ sub generate_inline_if_src_to_chain {
...
-            $rules .= "-A POSTROUTING --out-interface $mgmt_int ";
-            $rules .= "--match mark --mark 0x$_ ";
-            $rules .= "--jump $FW_POSTROUTING_INT_INLINE";
-            $rules .= "\n";
+            my @values = split(',', get_snat_interface());
+            foreach my $val (@values) {
+                foreach my $network ( keys %ConfigNetworks ) {
+                    next if ( !pf::config::is_network_type_inline($network) );
+                    my $inline_obj = new Net::Netmask( $network, $ConfigNetworks{$network}{'netmask'} );
+                    my $NAT = $ConfigNetworks{$network}{'nat'};
```
dwlfrth: CAPS ? by convention, caps are used for constants if i'm not mistaken...

obilodeau: you are not mistaken

lib/pf/iptables.pm
```diff
@@ -234,10 +249,26 @@ sub generate_inline_if_src_to_chain {
...
+                    if (defined ($NAT) && ($NAT eq $NO)) {
```
dwlfrth: $NO ? where does it come from?
Should we use the already commonly used isEnabled or isDisabled ?

lib/pf/iptables.pm
```diff
@@ -234,10 +249,26 @@ sub generate_inline_if_src_to_chain {
...
+                        $rules .= "-A POSTROUTING -s $network/$inline_obj->{BITS} --out-interface $val ";
+                        $rules .= "--match mark --mark 0x$_ ";
+                        $rules .= "--jump $FW_POSTROUTING_INT_INLINE_NONAT";
+                        $rules .= "\n";
+                    }
+
+                }
+
```
dwlfrth, line 369: Not sure if i'm following correctly, but for each defined SNAT interface, if you have a defined network that will use routed rather than nat, both rules will be created ?

lib/pf/iptables.pm
fdurand: from config.pm

lib/pf/iptables.pm
fdurand, line 369: It´s in the case your interface use SNAT, but for one network you want to route.
Exemple, you use multiple inline vlan, but for one of these vlan you want to exclude it from the SNAT and use routed network.

fdurand:
Has been merged in the last release.

PR #55, opened 2012-08-31
performance improvement when querying node info w/ dhcp

obilodeau:
Field tested. Ready for review.

commit log:

```
Introduced a new call node_attributes_with_fingerprint that carries less
information than node_view_with_fingerprint but that is a lot less intensive
on CPU. All core code callers didn't require the extended informations
anyway.

node_view_with_fingerprint is explicitly deprecated. To the point where
warnings will be logged when it's used.
```


dwlfrth:
fairly simple but efficient :) seems good!

PR #56, opened 2012-08-31
fix for #848: password validation

obilodeau:
battle tested and ready for integration / review


dwlfrth:
Even if I'm not that familiar with the grammar stuff, that seems fine ;)

PR #58, opened 2012-09-06
refactoring related to #1516: illegal characters in webservices user / pass not properly escaped

obilodeau:
As discussed in pull request #46 the code duplication in the fix for [issue # 1516](http://packetfence.org/bugs/view.php?id=1516) is bad.

Here's a refactoring of the duplicated code into a shared lib.

Targeted to devel because I only tested with a single RADIUS request in the lab and not if there were unintended high-load / multi-threaded consequences.

NEWS entry:

```
Enhancements:
 * Minor refactoring
```

### DIFF
--- a/lib/pf/util/freeradius.pm
+++ b/lib/pf/util/freeradius.pm
@@ -1,0 +1,100 @@
+package pf::util::freeradius;
+
+=head1 NAME
+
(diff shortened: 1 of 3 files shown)


dwlfrth:
Good!

PR #60, opened 2012-09-12
refactoring: web guest vs admin separation

obilodeau:
$portalSession's stuff is only relevant on the captive portal facing scenario. This refactoring splits out client facing and admin facing code so mistakes are harder to make. It also provides render_template(), it's own template directory, regrouped i18n and a stash-like template variables approach just like on the portalSession side.

A commit per commit review is probably the easiest way to review since there's a lot of code movement in there.

Fixes [#1485: pf::web::guest / admin guest creation need some kind of separation](http://packetfence.org/bugs/view.php?id=1485)

Ready for integration and code review.

Aimed at devel branch.


cgx:
Looks fine. That's a very appreciated aggregation. Let's merge!

dwlfrth:
Lots of nice modifications.
That will enhance the web part a lot.
Seems ready to merge (integration!)

PR #61, opened 2012-09-12
getting rid of domain from our gettext po resources (#1433)

obilodeau:
fix for [#1433: get rid of our domain in packetfence's gettext (PO) files](http://packetfence.org/bugs/view.php?id=1433)

simplifies greatly transifex sync

debian package not updated since it's not even creating translations right now. Will fix that in a separate branch.

```
NEWS

Bug fixes:
 * No more domain in our gettext po resources files (#1433)

```


cgx:
Go go go.

PR #63, opened 2012-09-13
Use of radius with inline mode

fdurand:
If the switch is in inline enforcement we return radius ACCEPT and when the node hit the portal on the inline interface iptables set the rules.
Work with mac-auth, 802.1x (with and without AutoRegister).


dwlfrth:
Shouldn't we add the enforcement flag in the "default" switch of switches.conf?

conf/switches.conf
```diff
@@ -19,6 +19,7 @@ customVlan4 =
 customVlan5 =
 VoIPEnabled = no
 voiceVlan =
+Enforcement = VLAN
```
obilodeau: enforcement not Enforcement (aligns with other single word naming convention)

obilodeau:
pretty much globally s/Enforcement/enforcement/g

lib/pf/SNMP.pm
```diff
@@ -1002,6 +1005,9 @@ sub getManagedIfIndexes {
 sub isManagedVlan {
     my ($this, $vlan) = @_;
 
+    if ($this->isSwitchInlineMode) {
+        return 1;
+    }
```
obilodeau: - 1 -> $TRUE
- add a comment explaining why

lib/pf/SNMP.pm
```diff
@@ -2750,6 +2760,23 @@ sub returnRadiusAccessAccept {
...
+Default implementation.
+
+=cut
+sub isSwitchInlineMode {
+    my ($self) = @_;
+    if ($self->{'_Enforcement'} eq "Inline") {
+        return $SNMP::TRUE;
+    } else {
+        return $SNMP::FALSE;
+    }
```
obilodeau: $SNMP's TRUE and FALSE are actually both true values in perl (see below)! These constants are meant to be used when interacting over SNMP only. Use $TRUE, $FALSE instead.

from pf::SNMP::constants:

``` perl

=item TruthValue - defined by RFC 1903 (SNMP Textual Conventions) aka SNMPv2-TC

 true (1)
 false (2)

=cut
Readonly::Scalar our $TRUE => 1;
Readonly::Scalar our $FALSE => 2;
```

obilodeau: this means all the switches would have been considered inline and implies that you haven't even tested the feature thoroughly ;)

lib/pf/SNMP.pm
```diff
@@ -2750,6 +2760,23 @@ sub returnRadiusAccessAccept {
...
+=item isSwitchInlineMode
+
+Check if the switch is configured in inline mode.
+
+Default implementation.
+
+=cut
+sub isSwitchInlineMode {
+    my ($self) = @_;
+    if ($self->{'_Enforcement'} eq "Inline") {
```
obilodeau: hard-coded string alert!

obilodeau:
Feature incomplete.
- no documentation (and it definitely needs explanations / use case)
- complexifies `pf::vlan` for no reason, it could have been dealt with in `pf::radius` since it affects only this code path

Final comment: this will confuse users, setting inline on a non-RADIUS wired switch will have no effect whatsoever.. I think it should be implemented as a new switch type instead of a switch parameter. It will be simpler to maintain and explain that way. Feel free to disagree but provide good arguments for the introduction of this feature as a param instead.

fdurand:
I understand what you mean, but if we implement a new switch type we will lost all the method from the specified switch/AP.
I think we can implement this feature differently:
-It will be great to have a defined Inline vlan (0 mean that packetfence return just access-accept),and we defined that the inline vlan by the ssid or anything else. (in switches.conf enforcementInline = SSID ,MAC ...)
-Add a flag on the switches modules like sub supportsRadiusInline { return $TRUE; }

So it´ll be possible to have enforcement and inline defined on the same switch/AP.

lib/pf/SNMP.pm
```diff
@@ -2750,6 +2761,23 @@ sub returnRadiusAccessAccept {
...
+=item isSwitchInlineMode
+
+Check if the switch is configured in inline mode.
+
+Default implementation.
+
+=cut
+sub isSwitchInlineMode {
+    my ($self) = @_;
+    if ($self->{'_Enforcement'} eq $IF_ENFORCEMENT_INLINE) {
```
dwlfrth: s/Enforcement/enforcement

lib/pf/pfcmd/pfcmd.pm
```diff
@@ -168,7 +168,7 @@ $grammar = q {
...
+   switchconfig_view_field : 'type' | 'mode' | 'uplink' | 'SNMPVersionTrap' | 'SNMPVersion' | 'cliTransport' | 'cliUser' | 'wsTransport' | 'wsUser' | 'vlans' | 'normalVlan' | 'registrationVlan' | 'isolationVlan' | 'macDetectionVlan' | 'guestVlan' | /customVlan\d\d?/ | 'macSearchesMaxNb' | 'macSearchesSleepInterval' | 'VoIPEnabled' | 'voiceVlan' | 'SNMPEngineID' | 'SNMPUserNameRead' | 'SNMPAuthProtocolRead' | 'SNMPPrivProtocolRead' | 'SNMPUserNameWrite' | 'SNMPAuthProtocolWrite' | 'SNMPPrivProtocolWrite' | 'SNMPUserNameTrap' | 'SNMPAuthProtocolTrap' | 'SNMPPrivProtocolTrap' | 'controllerIp' | 'roles' | 'Enforcement'
```
dwlfrth: s/Enforcement/enforcement

dwlfrth:
- Using the same switch module to do both VLAN and Inline is a bit useless from my point of view. I don't understand why someone with hardware that supports VLAN will use both enforcement mode on the same hardware at the same time.
- The new "inline" switch type is also the way to go, again, from my point of view. We won't lose any methods from specific switch modules since anyway, inline mode won't support these features(methods).

candlerb:
> I don't understand why someone with hardware that supports VLAN will use both enforcement mode on the same hardware at the same time.

I can think of one reason: you could have some dumb switches downstream of certain ports, and want to do inline enforcement for those.

But in that case, surely you would statically put those ports into the inline enforcement VLAN? I don't see why enforcement type needs to become a property of the switch, rather than a property of the network as it currently is. I think it would be highly confusing if it had to be set on both.

Or is the problem that you cannot turn off 802.1x/MAC RADIUS authentication for individual ports? If so, what model of switch has this problem?

fgaudreault:
I can give you a use case.  

One of our client is using Aironets (which supports VLAN mode very well), and they have an SSID to give free public internet access using Inline, while they want to keep dynamic VLAN assignment for their other SSIDs. 

I am pretty sure you also know that if you put the Inline vlan on a switchport/SSID, you cannot receive accounting data? You need AAA to get that info, or do SNMP recurrent calls which is ugly and deprecated.  This is another driver for an Hybrid mode for using RADIUS.

dwlfrth:
Finally decided to go that way.
Will issue a final review and should be ready to go after.

fdurand:
Ready for review

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -54,6 +54,9 @@ PacketFence's operation is completely out of band when using VLAN enforcement wh
 In Band (Inline Enforcement)::
 PacketFence can also be configured to be in-band, especially when you have non-manageable network switches or access points. PacketFence can also work with both VLAN and Inline enforcement activated for maximum scalability and security while allowing older hardware to still be secured using Inline enforcement.
 
+Hybrid support (Inline Enforcement with radius support)::
```
dwlfrth: s/radius/RADIUS (only for this line, not the whole paragraph)

lib/pf/SNMP.pm
```diff
@@ -2750,6 +2765,23 @@ sub returnRadiusAccessAccept {
     return [$RADIUS::RLM_MODULE_OK, %$radius_reply_ref];
 }
 
+=item isSwitchInlineMode
```
dwlfrth: i think you for forgot to remove theses...
inline enforcement no longer seems to exists...

fdurand: Yes you are right

lib/pf/vlan.pm
```diff
@@ -62,6 +62,12 @@ sub fetchVlanForNode {
     my ( $this, $mac, $switch, $ifIndex, $connection_type, $user_name, $ssid ) = @_;
     my $logger = Log::Log4perl::get_logger('pf::vlan');
 
+    # check if the switch is in inline mode
```
dwlfrth: i think you for forgot to remove theses...
inline enforcement no longer seems to exists...

lib/pf/radius/constants.pm
```diff
@@ -92,6 +92,17 @@ Readonly::Scalar our $IP => 1;
 Readonly::Scalar our $IPV4 => 1;
 Readonly::Scalar our $ETHERNET => 6;
 
+=item Trigger Inline
```
dwlfrth: I'm not sure I'd put these in the RADIUS package. These are not really some RADIUS specific constants...
Shouldn't we put them in SNMP ?

fdurand: It´s not realy SNMP constant too ...
I realy don´t know where to put them.

dwlfrth: The thing is, if I look at the other RADIUS constants, they are all very specific for RADIUS usage.
In SNMP, we have some more "global" constants...

fgaudreault: pf::config ?

lib/pf/radius.pm
```diff
@@ -472,6 +482,57 @@ sub _rewriteAccessAccept {
...
+=cut
+sub isTriggerInline {
+    my ($self, $switch, $port, $mac, $ssid) = @_;
+    my $logger = Log::Log4perl::get_logger(ref($self));
+    if (defined($switch->{_triggerInline})) {
+        foreach my $trigger (@{$switch->{_triggerInline}})  {
+
+            # TODO we should refactor this into objects where trigger types provide their own matchers
+            # at first, we are liberal in what we accept
+            $logger->warn("Invalid trigger id ($rigger)") if ($trigger !~ /^\w+::(.*)$/);
```
dwlfrth: That should return $FALSE ?
There is no return in case of errors?

lib/pf/radius.pm
```diff
@@ -472,6 +482,57 @@ sub _rewriteAccessAccept {
...
+            $type = lc($type);
+            $tid =~ s/\s+$//; # trim trailing whitespace
+
+            if ($type eq "always") {
+                return $TRUE;
+            }
+            # make sure trigger is a valid trigger type
+            # TODO refactor into an ListUtil test or an hash lookup (see Perl Best Practices)
+            if ( !grep( { lc($_) eq $type } $switch->inlineCapabilities ) ) {
+                $logger->warn("Invalid trigger type ($type), this is not supported by this switch");
```
dwlfrth, line 511: That should return $FALSE ?
There is no return in case of errors?

dwlfrth:
Seems fine.
If possible, retry to make sure everything is fine.
Would be great to add some tests...

dwlfrth:
Please dont forget to add an entry in both NEWS and UPGRADE

fgaudreault:
Merged in Devel.

PR #64, opened 2012-09-14
Feature/remove default switch

fdurand:
Remove the switch 127.0.0.1 in the switches.conf and instantiate it on the code
## 

NEWS entry:
- Removed static configuration of 127.0.0.1 switch in switches.conf to avoid borked configuration by people who deleted this one switch needed for communication between PacketFence and pfsetvlan.

UPGRADE entry:
- We removed the static configuration of 127.0.0.1 switch in switches.conf as described in the NEWS document. You can safely remove this one switch from your current configuration. Failure to do so won't affect the proper functioning of the solution but we encourage you to proceed.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -19,6 +19,7 @@ New Hardware
  * Support for Cisco Catalyst 2960G in 802.1X / MAC-Authentication
 
 Enhancements
+ * Removed 127.0.0.1 switch from the config and added in the code
 
 Bug Fixes
  * Adding radius.log into the logrotate script (#1526)
--- a/UPGRADE
+++ b/UPGRADE
@@ -29,6 +29,8 @@ Upgrading from a version prior to 3.5.0
 
 ----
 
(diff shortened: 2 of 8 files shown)


dwlfrth:
Wondering if we can get rid of the 127.0.0.1 in bin/pfcmd line 995 since this switch is no longer part of the config? (That's a question)

dwlfrth:
Added some NEWS/UPGRADE documents entries (in the pull-request comment). Review of these would be appreciated.

dwlfrth:
- Shouldn't we add a check in the switches.conf section of checkup.pm to make sure 127.0.0.1 is off the config and if not, display a status message at start time?
- What is the behavior if 127.0.0.1 is still present in switches.conf?
- Maybe we should add either an "addon script" that will remove the 127.0.0.1 section and document it in UPGRADE, or manage the instantiation in a different way if the section is present.

fgaudreault:
Merged in Devel.

PR #65, opened 2012-09-19
pfdhcplistener process management fixes - the sequel

obilodeau:
Ready for review.

NEWS:

```

Bug Fixes
 * Additional fixes to pfdhcplistener process management

```


dwlfrth:
Seems ok from my point of view.
Will test in the "field" but should be OK to go.

fgaudreault:
Is this ready to go?

PR #66, opened 2012-09-24
wispr xml inside captive-portal redirection

obilodeau:
Ready for review. 

NEWS

```
Enhancements
 * Better WISPR support
 * New parameter to control captive portal interception: secure_redirect
```


dwlfrth:
commit 06df64f is kind of heavy to review.
been through it attentively and everything seems fine but i think someone else should take a look at it or we should put it under tests.

obilodeau:
I re-tested the whole workflow (including passthroughs) on the iPad in the lab and it worked. I think it would be reasonable to merge this in devel today.

lib/pf/web/dispatcher.pm
```diff
@@ -0,0 +1,149 @@
...
+
+    # passthrough
+    # if the regex is not defined, we skip, this allow us to skip an expensive Config test
+    if (defined $CAPTIVE_PORTAL{'PASSTHROUGH_HOSTS_RE'} ) {
+        return Apache2::Const::DECLINED if (_matches_passthrough($r));
+    }
+
+    # fallback to a redirection
+    $r->uri('/perl/portal-redirect');
+    return Apache2::Const::DECLINED;
```
fdurand: you can go directly to handler without going through a redirection.
            $r->set_handlers( PerlResponseHandler => &handler );
            return Apache2::Const::OK;

obilodeau: I tried with both `return Apache2::Const::OK;` and `return Apache2::Const::DECLINED;` but I still get an undef `$r` once inside the handler.

obilodeau: Ok finally nailed it, I was missing a `$r->handler('modperl');` statement.

PR #70, opened 2012-09-29
Allow GUI to stop daemon when status is "should be stopped" #1562

candlerb:
If a daemon is running but the status is "should be stopped", then we should
still render the stop button in the GUI

### DIFF
--- a/html/admin/administration/services.php
+++ b/html/admin/administration/services.php
@@ -118,14 +118,18 @@ function print_status_table(){
           } else {
             print "<td>&nbsp;</td>\n";
           }
-          if (($services[1] == 1) && ($services[2] != 0)) {
+          if ($services[2] != 0) {
             print "<td align=center>";
             image_button(
                 "$current_top/$current_sub.php", array( 'service' => $services[0], 'action' => 'stop' ), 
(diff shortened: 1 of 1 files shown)


fgaudreault:
If the service is started when it should in theory be offline, the problem is elsewhere.

Exemple, if you change the config, say named -> disabled (from enabled), and you restart PF, the service will not be stopped at first place.  So named will continue running even if "it should be stopped".  Is this patch fixing this?

candlerb:
You are correct: if you configure services.named -> disabled followed by /etc/init.d/packetfence restart, then it remains running. Ditto for "pfcmd service pf restart". However, "pfcmd service named stop" _does_ stop it.

I'd say this is a related but different problem. With this patch you can at least stop the service through the GUI (because the GUI lets you do "pfcmd service named stop")

Possible approaches for the second problem:
- 'pf stop' should apply to `@pf::services::ALL_SERVICES`
- 'pf stop' should apply to `@alreadyRunningServices` (borrowing logic from 'start')
- 'pf stop' should apply to the union of `@services` and `@alreadyRunningServices`

Which do you prefer?

PR #71, opened 2012-09-30
dhcpd startup if all isolation/registration networks are remote #1565

candlerb:
If dhcpd is answering remote requests only, i.e. there is no directly-
connected isolation/registration network, then add an empty subnet { }
declaration so that dhcpd will still listen on the interface.

### DIFF
--- a/lib/pf/services/dhcpd.pm
+++ b/lib/pf/services/dhcpd.pm
@@ -22,6 +22,7 @@ use warnings;
 use Log::Log4perl;
 use POSIX;
 use Readonly;
+use Net::Netmask;
 
 use pf::config;
 use pf::util;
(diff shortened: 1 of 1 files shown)


fgaudreault:
That patch appears to also listen on the management interface, which can lead to many problems on some deployment.  Exclude the management interface from the listen.

Again, this patch appears to fix your case where you try to run PF on 1 interface.  That will cause many regressions to people running PF on multiple interfaces.   This is a NO GO for me.

candlerb:
The patch makes no change to what interfaces dhcpd listens on, nor does it do anything with management interfaces.
- In lib/pf/services.pm, $service_launchers{'dhcpd'} appends @listen_ints as the list of interfaces for dhcpd to listen on.
- In lib/pf/config.pm, @listen_ints is set to every "internal" interface (except those which match /:\d+$/, i.e. IP aliases). Management interfaces are included in @dhcplistener_ints but not in @listen_ints.
- The patch just ensures there is at least one subnet {} declaration for every interface in @listen_ints

Hence I can see no possibility for regression. Management interfaces remain excluded from dhcpd.conf, and they remain excluded from the dhcpd command line.

You are correct that I _am_ suggesting that with this patch, you can configure a single interface to be of type "management" and "internal". The GUI already allows this - it has a multi-selection box.

If you object to this, can you explain what the problem is?

fgaudreault:
Fair enough. Does someone have time to test this in lab?  If it works, we will merge.

fdurand:
I have tested and seems to be ok

PR #72, opened 2012-10-02
Clarify the configuration requirements for routed enforcement

candlerb:
Recent mailing list postings have said that even with remote-only enforcement, you are still required to have local registration and isolation VLANs.

This is a patch to the admin guide to make the requirements unambiguous, and also give a sample ACL for the remote networks.


docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -807,6 +825,30 @@ named=enabled
...
+interface vlan 20
+ ip address 192.168.20.254 255.255.255.0
+ ip helper-address 192.168.2.1
+ ip access-group PF_REGISTRATION in
+----
+
+If your edge switches support 'vlan-isolation' you can also apply the ACL
+there.  This has the advantage of preventing machines in isolation from
+attempting to attack each other.
+
```
fgaudreault, line 851: Maybe we can also refer to the FAQ: http://www.packetfence.org/support/faqs/article/restrict-communications-in-registrationisolation-vlans.html

It explains briefly VLAN mapping in reg/isol.

PR #76, opened 2012-12-19
added Dlink DGS-3100 and Aruba Instant OS

xinity:
added documentation on how to configure D-Link DGS-3100 and Aruba Instant OS with packetfence.


xinity:
seems to be a little bug with github for windows, i'll rollback and make another pull request

PR #83, opened 2013-01-17
Update lib/pf/SNMP/Cisco.pm

afont:
I've added two new subroutines (getIpAddr and getVoIpAddr), to retrieve the IP Address (via SNMP) of the devices connected to the specified interface. 
I think it's useful when using fixed IP addressing, not DHCP.
Tested successfully on Cisco 3560.


dwlfrth:
Closing since we don't see the "need" to add this kind of stuff in the PacketFence core source code. At a certain point, we try to avoid statically assigned IP address since the whole DHCP related stuff is kind of important in the PacketFence workflow.
Also, it seems to be something that fix an issue you ran into and does really need part of the code.

PR #85, opened 2013-01-17
Update bin/pfcmd_vlan

afont:
Added two new options (-getIpAddr and -getVoIpAddr) to retrieve IP address information from devices connected to Cisco Switches with fixed IP addresses.


dwlfrth:
See #83

PR #86, opened 2013-01-28
Packetfence ARP remote scanner

fdurand:
For review.
This module is a mix from pfmon, pfdhcplistener and pfdetect_remote and it is use to sniff arp traffic to send to PacketFence the mac and ip information.
It´s use in the case you can´t redirect dhcp traffic to packetfence.

Memcached is use to prevent sending an http request each time it see a arp request.


dwlfrth:
This sould have been worked from devel rather than from stable...

PR #94, opened 2013-07-10
A series of changes to address shortcomings in our MAB EAP support.

louismunro:
modified:   lib/pf/config.pm

```
Changed so that an EAP WIRED_MAC_AUTH shows as such in the logs rather
than as ETHERNET-noEAP
```

modified:   lib/pf/radius.pm

```
Changed so that MAC_AUTH detection will work even if MAC separators are
other than ':'.
Comparison between username and mac is now case insensitive.
```

modified:   lib/pf/vlan.pm

```
Changed so that EAP WIRED_MAC_AUTH will get the role from node-info.
This case failed before as it matched the EAP connection type and so we
tried every authentication source with a MAC for username.
```

modified:   lib/pf/vlan/custom.pm

```
Changes reflect new lib/pf/vlan.pm.
```

### DIFF
--- a/lib/pf/config.pm
+++ b/lib/pf/config.pm
@@ -202,6 +202,7 @@ Readonly our $WEB_ADMIN_ALL => 4294967295;
     'Ethernet-NoEAP'        => $WIRED_MAC_AUTH,
     'SNMP-Traps'            => $WIRED_SNMP_TRAPS,
     'Inline'                => $INLINE,
+    'WIRED_MAC_AUTH'        => $WIRED_MAC_AUTH,
 );
 %connection_group = (
     'Wireless'              => $WIRELESS,
@@ -214,7 +215,7 @@ Readonly our $WEB_ADMIN_ALL => 4294967295;
     $WIRELESS_802_1X => 'Wireless-802.11-EAP',
     $WIRELESS_MAC_AUTH => 'Wireless-802.11-NoEAP',
     $WIRED_802_1X => 'Ethernet-EAP',
-    $WIRED_MAC_AUTH => 'Ethernet-NoEAP',
+    $WIRED_MAC_AUTH => 'WIRED_MAC_AUTH',
     $WIRED_SNMP_TRAPS => 'SNMP-Traps',
     $INLINE => 'Inline',
     $UNKNOWN => '',
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -315,7 +315,13 @@ sub getNormalVlan {
     my $role = "";
 
     $logger->debug("Trying to determine VLAN from role.");
-    if (defined $user_name && (($connection_type & $EAP) == $EAP)) {
+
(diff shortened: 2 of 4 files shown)


lib/pf/config.pm
```diff
@@ -202,6 +202,7 @@ Readonly our $WEB_ADMIN_ALL => 4294967295;
     'Ethernet-NoEAP'        => $WIRED_MAC_AUTH,
```
louismunro, line 202: Ethernet-NoEAP is NOT the only kind of WIRED_MAC_AUTH

lib/pf/config.pm
```diff
@@ -202,6 +202,7 @@ Readonly our $WEB_ADMIN_ALL => 4294967295;
...
     'SNMP-Traps'            => $WIRED_SNMP_TRAPS,
     'Inline'                => $INLINE,
+    'WIRED_MAC_AUTH'        => $WIRED_MAC_AUTH,
```
louismunro, line 205: You can haz WIRED_MAC_AUTH over EAP.
It's added here so the logs show it as WIRED_MAC_AUTH and not Ethernet-NoEAP when it is actually WIRED_MAC_AUTH over EAP.

lib/pf/config.pm
```diff
@@ -214,7 +215,7 @@ Readonly our $WEB_ADMIN_ALL => 4294967295;
     $WIRELESS_802_1X => 'Wireless-802.11-EAP',
     $WIRELESS_MAC_AUTH => 'Wireless-802.11-NoEAP',
     $WIRED_802_1X => 'Ethernet-EAP',
-    $WIRED_MAC_AUTH => 'Ethernet-NoEAP',
+    $WIRED_MAC_AUTH => 'WIRED_MAC_AUTH',
```
louismunro, line 218: Again, WIRED_MAC_AUTH does not imply NoEAP.

lib/pf/vlan.pm
```diff
@@ -315,7 +315,13 @@ sub getNormalVlan {
     my $role = "";
 
     $logger->debug("Trying to determine VLAN from role.");
-    if (defined $user_name && (($connection_type & $EAP) == $EAP)) {
```
louismunro, line 318: The old line used to match when it was WIRED_MAC_AUTH over EAP.
Thus it tried to call &pf::authentication::match with a MAC for username, resulting in no role ever being assigned because there is not user named bc:c6:db:1c:2b:a4 (for example).

lib/pf/vlan.pm
```diff
@@ -315,7 +315,13 @@ sub getNormalVlan {
...
+
+    # Try MAC_AUTH, then other EAP methods and finally anything else.
+    if ( ( $connection_type & $WIRED_MAC_AUTH ) == $WIRED_MAC_AUTH ) {
```
louismunro, line 320: The new line tries WIRED_MAC_AUTH first (regardless of EAP or not).
So if it is MAC auth we get the role from the DB.

PR #101, opened 2013-08-28
Fix/xml encode

louismunro:
Fixes a bug seen with a client who had RADIUS attributes such as "Business & Accounting" (i.e. with an unencoded XML reserved character ).


cgx:
Looks good to me.

PR #104, opened 2013-09-30
Feature/service refactor

jrouzierinverse:
Integration of pf::services::manager framework into pfcmd and the webgui.

Team I created a new branch feature/service-refactor for the new and improved pfcmd service command.

Notable changes 
- each service has it's own service manager perl module
- pfcmd pf watch - Will now only restart service that are down
- pfcmd <service> stop|start will show if the service was already stopped or started
- pfcmd service is now in color!

Questions 
- Should we have a configuration setting that force an entire restart of pf if any service is down when the watch is done.
  servicewatch.restart_type=soft|hard
- Is there anything else we should add since we are in the code at the moment?


lib/pf/services/manager/memcached.pm
```diff
@@ -0,0 +1,55 @@
...
+use warnings;
+use Moo;
+use pf::file_paths;
+
+extends 'pf::services::manager';
+with 'pf::services::manager::roles::pf_conf_service_managed';
+
+has '+name' => (default => sub { 'memcached' } );
+
+has '+launcher' => (default => sub { "%1\$s -d -p 11211 -u pf -m 64 -c 1024 -P $install_dir/var/run/memcached.pid"});
```
dwlfrth, line 24: Comments...
"Random" values are useless even if needed when undocumented

dwlfrth:
Can you document (pf::services::manager will probably be the best place) all possibles method (mainly for overloading purposes but also to have a good idea of what which method is doing) and the startup/stop flow.

dwlfrth:
What is the purpose of pf::services::manager::roles::pf_conf_service_managed ?

jrouzierinverse:
pf::services:: manager is pretty much documented.

PR #105, opened 2013-10-03
Fix/mac class

louismunro:
A class to implement MAC addresses with some utility methods.
Meant to replace MAC related functions in pf::util and make possible comparisons between MAC addresses in different notations.
Also makes conversions easier ( see as_Cisco, as_Microsoft, as_oid etc.)

### DIFF
--- a/lib/pf/MAC.pm
+++ b/lib/pf/MAC.pm
@@ -6,6 +6,7 @@ pf::MAC
 
 =head1 DESCRIPTION
 pf::MAC implements a class that instantiates MAC addresses objects.
+
 At the moment it is rather minimalist, inheriting from Net::MAC which already does 
 90% of what PacketFence needs.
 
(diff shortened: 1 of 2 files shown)


lib/pf/MAC.pm
```diff
@@ -0,0 +1,220 @@
...
+
+=item macoui2nb
+
+Provided for backwards compatibility with pf::util::macoui2nb.
+
+Equivalent to get_dec_oui().
+
+=cut 
+
+sub macoui2nb { return $_[0]->get_dec_oui(@_); }
```
cgx: You could probably simplify this with `*macoui2nb = \&get_dec_oui;`.

lib/pf/MAC.pm
```diff
@@ -0,0 +1,220 @@
...
+
+=item mac2nb
+
+Provided for backwards compatibility with pf::util::mac2nb.
+
+Equivalent to get_dec_stripped().
+
+=cut
+
+sub mac2nb { return $_[0]->get_dec_stripped(@_); }
```
cgx: You could probably simplify this with `*mac2nb = \&get_dec_stripped;`.

lib/pf/MAC.pm
```diff
@@ -0,0 +1,220 @@
...
+
+=item format_for_acct
+
+Intended for backward compatibility with pf::util::format_mac_for_acct.
+
+Equivalent to as_acct();
+
+=cut 
+
+sub format_for_acct { return $_[0]->as_acct(@_); }
```
cgx: You could probably simplify this with `*format_for_acct = \&as_acct;`.

PR #106, opened 2013-10-04
Feature/net gear fsm7328 s

louismunro:
Fry's requested new Module.


extrafu:
Should we also merge that one?

louismunro:
The client asked for it.

PR #107, opened 2013-10-15
Feature/network-detection

jrouzierinverse:
Added additional event handler for dealing with the onload event not firing for some browsers


cgx:
With what browsers on which platforms has this code been tested so far?

PR #113, opened 2013-11-15
Fix/cache switch

jrouzierinverse:
Add compression to chi for memcached
Perform Santiy check on SwitchConfig coming from the cache


jrouzierinverse:
This is no longer relevent

PR #115, opened 2013-11-25
Fix unresolved user defined role

julsemaan:
Fixes a bug where a user defined role wouldn't be resolved to a vlan category if passed directly to getVlanByName

Happens with module Catalyst_2950 using port-security.

Not sure it's the correct way to fix it but it works.

### DIFF
--- a/lib/pf/SNMP.pm
+++ b/lib/pf/SNMP.pm
@@ -742,6 +742,14 @@ sub getVlanByName {
     my ($this, $vlanName) = @_;
     my $logger = Log::Log4perl::get_logger(ref($this));
 
+    if (defined($this->{'_roles'}) && defined($this->{'_roles'}->{$vlanName})){
+        $logger->info("Resolved user defined role $vlanName as $this->{'_roles'}->{$vlanName}");
+        $vlanName = $this->{'_roles'}->{$vlanName};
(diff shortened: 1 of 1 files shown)


lib/pf/SNMP.pm
```diff
@@ -741,7 +741,19 @@ Input: VLAN name (as in switches.conf)
 sub getVlanByName {
     my ($this, $vlanName) = @_;
     my $logger = Log::Log4perl::get_logger(ref($this));
-
+    
+    if (defined($vlanName) && $vlanName eq ''){
+        $logger->warn("No vlan name or role was provided. Using default");
+        $vlanName = 'default';
```
julsemaan, line 747: The 'default' should be changed to category with id 1 or a network defined in a config file

extrafu:
Still relevant?

julsemaan:
In my opinion yes, when a user is registered without a role (using the admin interface for example) then the VLAN is never resolved for that user and he's assigned mac detection.

It would make more sense if the default vlan was sent back since by definition this is what it means.

Other solution could be to force the user to set a role to the node when the status is reg in the admin interface.

julsemaan:
Will need to be merged will devel though

PR #117, opened 2013-11-28
Reevaluate access after a role change in web ui

julsemaan:
Makes packetfence reevaluate the access of a node when it's role is changed in the web ui like it did in PF3


extrafu:
Still relevant?

julsemaan:
was still broken but fixed in a cleaner way in 10abac5d2b5c193ecf3effa8da08d80ad879c2a4

Closing

PR #119, opened 2014-01-09
Feature/fingerbank

ghost:
this feature adds the version number of the fingerprints database during the submitting by email.
To be integrated in devel


html/pfappserver/lib/pfappserver/Model/Admin.pm
```diff
@@ -43,6 +43,22 @@ sub pf_release {
...
+=head2 fingerbank_version
+
+Returns the version of Fingerbank from conf/dhcp_fingerprins.conf
+
+=cut
+
+sub fingerbank_version {
+    my ($filehandler, $line, $version);
+    open( $filehandler, '<', "$conf_dir/dhcp_fingerprints.conf" )
+        || print "Unable to open $conf_dir/dhcp_fingerprints.conf: $!";
```
cgx, line 55: Log the error instead of simply printing it.

PR #121, opened 2014-01-17
Fix pfdetectd not starting because of pid file

julsemaan:
Fixes the case where either pfdetect crashes or the server crashes and
the pid file is left in the file system. On reboot or on service restart
the service won't start because it thinks it's running.

Now if the PID file is there it will also check that the pid it contains
is running


cgx:
Please, update the NEWS file!

PR #125, opened 2014-03-03
Remove extra entry in logrotate

mrobbert:
This extra entry was causing logrotate to not work.


cgx:
Was fixed back in december in the devel branch (8255430c).

PR #130, opened 2014-03-14
Made Freeradius config self-documenting.

louismunro:
The eap-types in sites-enabled/packetfence are now self documenting.


dwlfrth:
Simple but good one.
I'd also like to see fully commented FreeRADIUS files rather than the one we are actually using from which we stripped comments.

PR #133, opened 2014-03-26
Added the way to move the custom code from vlan/custom.pm to configuration files.

fdurand:
For review please !
Added the way to move the custom code from vlan/custom.pm to configuration files.


fdurand:
Up !

conf/vlan_filters.conf.example
```diff
@@ -0,0 +1,101 @@
...
+#     node_info
+#     switch
+#     ifIndex
+#     mac
+#     connection_type
+#     username
+#     ssid
+#     time
+#
+# attribute can be:
```
cgx, line 63: Also list the possible `operator` values (`is` and `is_not`?)

lib/pf/vlan/filter.pm
```diff
@@ -0,0 +1,363 @@
...
+
+Parse the node_info attribute and compare to the rule. If it matches then perform the action.
+
+=cut
+
+sub node_info_parser {
+    my ($self, $rule, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid) = @_;
+
+    if ($rule->{'operator'} eq 'is') {
+        if ($node_info->{$rule->{'attribute'}} =~ /$rule->{'regexp'}/) {
```
cgx: Should the regexp be more precise (`/^..$/`)? Or should we use `eq`?

fdurand: Yes and no because i would like to match something like 2014-05-21 for the regdate.
We can imagine doing a rule like is the device has been reg yesterday then deny.
But on the other side it can match more than expected.

addons/dev-helpers/dump.pl
```diff
@@ -11,7 +11,7 @@ =head1 NAME
 
 =head1 SYNOPSIS
 
-dump.pl <config|doc_config|floatingdevices|profiles_filters|profiles|sources|switch <id>|switches|admin_roles|chiconfig>
+dump.pl <config|doc_config|floatingdevices|profiles_filters|profiles|sources|switch <id>|switches|admin_roles|chiconfig|vlanfilters>
```
extrafu: Since we have "profile_filters", I suggest we have "vlan_filters" for consistency.

conf/vlan_filters.conf.example
```diff
@@ -0,0 +1,101 @@
...
+# attribute = category
+# regexp = default
+#
+# [1:1]
+# scope = NormalVlan
+# action = nointernet
+#
+# It mean that when PacketFence try to get the normal Vlan for the node and if the category of the node is
+# default then we return the role nointernet (nointernet you configured in the switch config)
+#
```
extrafu, line 19: It means that when PacketFence tries to get the normal VLAN for the node and if the category of the node is the default one, we return the role "nointernet" (which you will have configured in the switch configuration).

conf/vlan_filters.conf.example
```diff
@@ -0,0 +1,101 @@
...
+# regexp = wd {Mon Tue Wed Thu Fri} hr {11am-2pm}
+#
+# [2:2&3&4]
+# scope = NormalVlan
+# action = nointernet
+#
+# It mean when PacketFence try to get the normal Vlan for the node and if the category of the node is
+# default and come from the ssid OpenWrt-SECURE and the time is between 11am and 2pm from Monday to Friday
+# then we return the role nointernet.
+#
```
extrafu, line 45: Same rewording.

conf/vlan_filters.conf.example
```diff
@@ -0,0 +1,101 @@
...
+#    Wireless-802.11-NoEAP
+#    Ethernet-EAP
+#    Ethernet-NoEAP
+#    SNMP-Traps
+#    Inline
+#    WIRED_MAC_AUTH
+#
+# The regexp of a time is specific and follow this logic:
+# http://search.cpan.org/~pryan/Period-1.20/Period.pm
+# per example if you want to allow the access from Monday to Friday from 9am to 11am and from 2pm to 4pm:
```
extrafu: For example, if you want to allow access from Monday to Friday, 9am to 11am and from 2pm to 4pm:

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2376,6 +2376,55 @@ realm ntdomain {
 }
 ----
 
+Vlan Filter Definition
+~~~~~~~~~~~~~~~~~~~~~~
+
+We added the ability to specify filters directly in the portion of code that revaluate the vlan.
```
extrafu: that re-evaluates the VLAN.

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2376,6 +2376,55 @@ realm ntdomain {
 }
 ----
 
+Vlan Filter Definition
+~~~~~~~~~~~~~~~~~~~~~~
+
+We added the ability to specify filters directly in the portion of code that revaluate the vlan. 
+
+These rules are available in different scope:
```
extrafu: scopes

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2376,6 +2376,55 @@ realm ntdomain {
...
+
+These rules are available in different scope: 
+
+  ViolationVlan
+  RegistrationVlan
+  NormalVlan
+  InlineVlan
+  shouldAutoRegister
+
+And can be defined based on differents criteria like:
```
extrafu: using different criteria like:

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2376,6 +2376,55 @@ realm ntdomain {
...
+  switch
+  ifIndex
+  mac
+  connection_type
+  username
+  ssid
+  time
+
+Per example let's define a rule that doesn't allow the connection to a device when his category is default and the ssid is SECURE and the current date is between 11am and 2pm
+from Monday to Friday when it try to connect as a registered device :
```
extrafu, line 2404: For example, lets define a rule that prevents a device from connecting when its category is the "default", when the SSID is "SECURE" and when the current time is between 11am and 2pm:

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2376,6 +2376,55 @@ realm ntdomain {
...
+  [3]
+  filter = time
+  operator = is
+  regexp = wd {Mon Tue Wed Thu Fri} hr {11am-2pm}
+
+  [1:1&2&3]
+  scope = NormalVlan
+  action = nointernet
+
+You can have a look in the file vlan_filters.conf, there are some example how to use and define filters.
```
extrafu: some examples on how to use

lib/pf/vlan/filter.pm
```diff
@@ -0,0 +1,363 @@
...
+   return $self;
+}
+
+=item test
+
+Test all the rules
+
+=cut
+
+sub test {
```
extrafu, line 48: Is it possible to add a unit test for that?

lib/pf/vlan/filter.pm
```diff
@@ -0,0 +1,363 @@
...
+    foreach my $rule  ( sort keys %ConfigVlanFilters ) {
+        if ( defined($ConfigVlanFilters{$rule}->{'scope'}) && $ConfigVlanFilters{$rule}->{'scope'} eq $scope) {
+            if ($rule =~ /^\w+:(.*)$/) {
+                my $test = $1;
+                $test =~ s/(\w+)/$self->dispatch_rule($ConfigVlanFilters{$1},$switch,$ifIndex,$mac,$node_info,$connection_type,$user_name,$ssid,$1)/gee;
+                $test =~ s/\|/ \|\| /g;
+                $test =~ s/\&/ \&\& /g;
+                if (eval $test) {
+                    $logger->info("Match Vlan rule: ".$rule);
+                    my $role = $ConfigVlanFilters{$rule}->{'action'};
```
fdurand: action is not correct i have to use role instead
action(optional) will be reserved for WebAPI call like unregnode

conf/vlan_filters.conf.example
```diff
@@ -0,0 +1,101 @@
...
+# 
+# you can trigger rule on specific scope (NormalVlan, RegistrationVlan, ViolationVlan, AutoRegistration)
+#
+# Make a simple rule like this:
+#
+# [1]
+# filter = node_info
+# operator = is
+# attribute = category
+# regexp = default
```
cgx: Give a proper regexp example. For example: `^default$`.

However, I suggest we replace `regexp` by `value` and accept the `operator` to be either _is_ (`eq`), _is_not_ (`ne`), _match_ (`=~ m//`) , or _match_not_ (`!~ m//`).

conf/vlan_filters.conf.example
```diff
@@ -0,0 +1,101 @@
+# Vlan filter configuration
+# 
+# you can trigger rule on specific scope (NormalVlan, RegistrationVlan, ViolationVlan, AutoRegistration)
+#
+# Make a simple rule like this:
+#
+# [1]
```
cgx: Can you use a significant name for the rule?

fdurand: Yes

cgx: Provide an example using a significant name. Thanks!

lib/pf/vlan/filter.pm
fdurand, line 48: Done

PR #134, opened 2014-04-04
Correcting few bugs in inline mode

Elektordi:
For review:
- Updating iptables rules for captive portal to be reachable even for registered users (needed for "progress-bar redirect", for email validation, etc...)
- Applying patch (#1655 in BTS) for updating ipset in inline mode

Thanks.


janmeier:
The changes to iptables was already applied in stable, but the change to enforcement fixed inline mode for me. Thanks! :)

jrouzierinverse:
This was fixed in stable already closing pull request.

PR #138, opened 2014-04-25
Added release function to the Catalyst Portal

julsemaan:
Adds the release violation feature to the Catalyst Portal


jrouzierinverse:
Will use the apache handler for now

PR #142, opened 2014-04-29
Create pfdetect_SO

GodfatherOfBandwidth:
Adding this as a new file because it will break the current way PF parses SNORT logs if it replaced the stock pfdetect.

With a separate file it is possible to run this version and the stock version at the same time should you be crazy and have both system deployed.

It is a copy of the current pfdetect with the parsing logic re-written for use with Security Onion.

It also adds a new dependency; File::Tail, which AFAIK does not introduce any performance penalty or memory issues.

### DIFF
--- a/sbin/pfdetect_SO
+++ b/sbin/pfdetect_SO
@@ -1,0 +1,212 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+pfdetect_SO - listen to snort alerts and create PF violations
+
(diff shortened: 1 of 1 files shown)


louismunro:
This has been integrated in pfdetect now. We appreciate your contribution but we can't merge this PR as is.

GodfatherOfBandwidth:
I understand, and I appreciate Inverse adding Security Onion support. You guys rock.

PR #143, opened 2014-05-01
Added I18n support to catalyst portal

julsemaan:
For review please


html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
```diff
@@ -95,6 +97,121 @@ sub setupCommonStash : Private {
...
+    my ($locales) = $c->forward('getLanguages');
+
+    my $locale = shift @$locales;
+    $logger->debug("Setting locale to ".$locale);
+    setlocale(POSIX::LC_MESSAGES, "$locale.utf8"); 
+}
+
+=head2 getLanguages
+
+Retrieve the user prefered languages from the following ordered sources:
```
extrafu: preferred

PR #144, opened 2014-05-02
Removed action validation for start and end date of temporary password

julsemaan:
The actions 'valid_from' and 'expiration' don't seem to exist in the authentication module. That prevents the temporary password from being valid since there is no registration window when doing a preregistration.

For review and merge please


cgx:
AFAIK, they are still used when creating and modifying users from the Web admin interface.

cgx:
Not need anymore as the issue was fixed by 62eb807b.

PR #146, opened 2014-05-09
Add 802.1x support to Juniper EX2200 and EX4200

julsemaan:
For review and merge.

Thanks

### DIFF
--- a/docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
+++ b/docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
@@ -2187,11 +2187,6 @@ protocols {
                 access-ports {
                     supplicant multiple;
                     mac-radius;
-                    transmit-period 5;
-                    reauthentication 3600;
-                    supplicant-timeout 5;
-                    server-timeout 5;
-                    maximum-requests 1;
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -1928,7 +1928,7 @@ Exact command-line configuration to be contributed...
 Juniper
 ~~~~~~~
 
-PacketFence supports Juniper switches in MAC Authentication (Juniper's MAC RADIUS) mode.
+PacketFence supports Juniper switches in MAC Authentication (Juniper's MAC RADIUS) mode and 802.1x.
```
extrafu: 802.1x -> 802.1X   - same thing in other files.

lib/pf/Switch/Juniper/EX2200.pm
```diff
@@ -265,7 +277,7 @@ sub wiredeauthTechniques {
         return $method,$tech{$method};
     }
     else{
-        $logger->error("Only wired mac authentication is supported on EX2200");
+        $logger->error("Only wired authentication is supported on EX2200");
```
dwlfrth: I'd go with "This mode is not supported" rather than "Only wired authentication is supported on EX2200"
- Wired authentication is a bit ambiguous
- Don't need to specify the module name in the error (IMO)

dwlfrth:
Need an entry in the NEWS file ;)

cgx:
@dereksky42 :+1:

julsemaan:
Any other corrections or we merge ?

PR #147, opened 2014-05-16
Feature/mab floating device

julsemaan:
RADIUS floating device support for PacketFence.

See : https://wiki.inverse.ca/focus/PacketFence/Devel/RnD/2014/FloatingDevicesImprovement

Works on Juniper EX4200

Detection :
- On an Access-Request of a floating device it disconnects then deletes the macs present in the location log in case another device triggered a request before it.
- It configures the switch to accept an unlimited number of devices on the floating device port and leaves mac auth enabled.
- Then PF will know there is a floating device on this port from the location log entry.
- When other devices connect to this port it returns the configured PVID or the normalVlan depending on the mode of the floating device (see below).
- When the switch sends the accounting stop for the floating device, the port is unconfigured from the floating device mode and it's locationlog entry closed.

Additionnal devices connecting to the port (access floating):
- When the floating device is not configured in trunk in PacketFence, this flow applies.
- PF will receive a RADIUS request and if a floating device is in the location log for that interface it will accept it with the PVID of the floating device.
- In this mode, all devices connecting to this port are accepted with the floating device VLAN.

Additionnal devices connecting to the port (trunk floating):
- PF will receive a RADIUS request and if a floating device is in they location log for that interface it will lookup the VLAN tag to return as it does normally (registration, isolation or normal)
- In this mode access control is done directly by PacketFence. The port becomes a 'sub-switch'.

Detection and flow for switches that don't support multiple untagged vlan per port:
- They will be detected using the RADIUS access request.
- That will trigger the 'classic' flow of the floating devices.
- MAB will be removed on the port and put in either the access vlan it should or in trunk mode
- up traps will be activated
- Once the up trap is received, pf will reactivate MAB on the port

Current limitations :
- In order to not open multiple location log entries per port PF doesn't keep the location of the devices connected behind the floating device. This means in trunk mode there can't be any deauth in case a MAC needs VLAN reevaluation. Access mode works fine since it never needs deauth.
- Devices connected before the floating need to be disconnected to have them on the correct VLAN and this is not done asynchronously. This shouldn't be a big issue since the mac limit on the switch will make us deauth one client in the worst case.
- If a switch is a floating device in trunk mode then the devices in different VLANs behind it could have communications problems between them. Since they are on two different networks outside of the floating but on the same in the floating, a switch could forward the packets between both devices that are on different IP subnets. Tested with a Cisco 2950 as a floating and didn't get this behavior (SSH + ping working fine between devices connected in the floating).
- (Juniper) The flap-on-disconnect option has to be deactivated programaticaly by PF. Because of the way the Juniper switch has to be configured, the flap-on-disconnect cannot be configured through inheritance on Juniper. Doesn't change the behavior but makes ugly switch configs

### DIFF
--- a/lib/pf/radius.pm
+++ b/lib/pf/radius.pm
@@ -32,6 +32,7 @@ use pf::util;
 use pf::trigger;
 use pf::violation;
 use pf::vlan::custom $VLAN_API_LEVEL;
+use pf::floatingdevice::custom;
 # constants used by this module are provided by
 use pf::radius::constants;
 use List::Util qw(first);
@@ -174,6 +175,9 @@ sub authorize {
         return [ $RADIUS::RLM_MODULE_OK, ('Reply-Message' => "Switch is not in production, so we allow this request") ];
     }
 
+    # Check if a floating just plugged in
+    $this->_handleAccessFloatingDevices($switch, $mac, $port);
+
     # Fetch VLAN depending on node status
     my ($vlan, $wasInline, $user_role) = $vlan_obj->fetchVlanForNode($mac, $switch, $port, $connection_type, $user_name, $ssid, $radius_request);
 
@@ -242,6 +246,25 @@ sub accounting {
     my $isUpdate = $radius_request->{'Acct-Status-Type'} eq 'Interim-Update';
 
     if ($isStop || $isUpdate) {
+        my($switch_mac, $switch_ip,$source_ip) = $this->_parseRequest($radius_request);
+
+        $logger->debug("instantiating switch");
+        my $switch = pf::SwitchFactory->getInstance()->instantiate({ switch_mac => $switch_mac, switch_ip => $switch_ip, controllerIp => $source_ip});
+
+        # is switch object correct?
+        if (!$switch) {
+            $logger->warn(
+               "Can't instantiate switch $switch_ip. This request will be failed. "
+                ."Are you sure your switches.conf is correct?"
+            );
+            return [ $RADIUS::RLM_MODULE_FAIL, ('Reply-Message' => "Switch is not managed by PacketFence") ];
+        }
+
+        if($isStop){
+            #handle radius floating devices
+            $this->_handleAccountingFloatingDevices($switch, $mac, $port);
+        }
+
         # On accounting stop/update, check the usage duration of the node
         my ($nas_port_type, $eap_type, $mac, $port, $user_name, $nas_port_id, $session_id) = $switch->parseRequest($radius_request);
         if ($mac && $user_name) {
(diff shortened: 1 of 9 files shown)


extrafu:
Please also add the code for the Cisco 2960.

julsemaan:
The Cisco 2960 doesn't support the new way of dealing with trunk floating devices which is by having multiple radius authenticated devices on different vlans on the same port.

It could support the access way of doing it but not the trunk.

extrafu:
So it's impossible to use the 2960 with MAB and floating devices?

julsemaan:
Not for trunks but for single vlan floating devices, yes

julsemaan:
Support for the floating devices on the 2960 has been added and a way to have the classic type of floating devices but triggered by a radius access request

extrafu:
Nice work!

julsemaan:
Merge me ?

jrouzierinverse:
This needs to be rebased

julsemaan:
I rebased it 20 minutes ago and it changed again, so review and I'll rebase once we're ready to merge

lib/pf/radius.pm
```diff
@@ -239,8 +243,29 @@ sub accounting {
     my $isUpdate = $radius_request->{'Acct-Status-Type'} eq 'Interim-Update';
 
     if ($isStop || $isUpdate) {
-        # On accounting stop/update, check the usage duration of the node
+        my($switch_mac, $switch_ip,$source_ip) = $this->_parseRequest($radius_request);
```
jrouzierinverse: Why are you parsing the request twice?

lib/pf/radius.pm
```diff
@@ -239,8 +243,29 @@ sub accounting {
...
         my ($nas_port_type, $eap_type, $mac, $port, $user_name, $nas_port_id, $session_id) = $switch->parseRequest($radius_request);
+        my $connection_type = $switch->_identifyConnectionType($nas_port_type, $eap_type, $mac, $user_name);
+        $port = $switch->getIfIndexByNasPortId($nas_port_id) || $this->_translateNasPortToIfIndex($connection_type, $switch, $port); 
+
+        $logger->debug("instantiating switch");
+        my $switch = pf::SwitchFactory->getInstance()->instantiate({ switch_mac => $switch_mac, switch_ip => $switch_ip, controllerIp => $source_ip});
```
jrouzierinverse: No need to instantiate the switch another time

lib/pf/radius.pm
```diff
@@ -239,8 +243,29 @@ sub accounting {
...
+
+        # is switch object correct?
+        if (!$switch) {
```
jrouzierinverse: This check was already done

julsemaan:
Fixed issues and rebased for the 3rd time today.

If there's an ideal time to merge, I guess that would be now...

extrafu:
Not yet! Missing NEWS file entry :-)

PR #154, opened 2014-05-26
Removed the old way to do inline with iptables (we use ipset now)

fdurand:
Review please


NEWS.asciidoc
```diff
@@ -20,6 +20,8 @@ New Features
 Enhancements
 ++++++++++++
 
+* Removed old iptables code (ipset is always use for inline)
```
extrafu: ipset is now always used for inline enforcement

extrafu:
Also review the PacketFence Administration guide. There are some references to the inline enforcement using "iptables".

PR #158, opened 2014-05-27
Fix/radius accounting

louismunro:
Backports accounting fix now in devel.


NEWS.asciidoc
```diff
@@ -25,6 +25,7 @@ Bug Fixes
 
 * Compute unregdate and save the role for autoreg 802.1x
 * Fixed portal profile URI filter in new Catalyst-based captive-portal
+* Radius accounting fixed to call the correct method to parse the radius request.
```
extrafu, line 28: Nit here.. but radius -> RADIUS

PR #161, opened 2014-06-04
Added support for Netgear M series

julsemaan:
For review and merge

Thanks

### DIFF
--- a/lib/pf/Switch/Netgear/MSeries.pm
+++ b/lib/pf/Switch/Netgear/MSeries.pm
@@ -21,6 +21,7 @@ use pf::Switch::constants;
 use base ('pf::Switch::Netgear');
 
 sub supportsWiredMacAuth { return $TRUE }
+sub supportsRadiusDynamicVlanAssignment { return $TRUE; }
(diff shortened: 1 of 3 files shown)


lib/pf/Switch/Netgear/MSeries.pm
```diff
@@ -0,0 +1,54 @@
...
+use warnings;
+
+use Log::Log4perl;
+use Net::SNMP;
+use pf::config;
+use pf::Switch::constants;
+
+use base ('pf::Switch::Netgear');
+
+sub supportsWiredMacAuth { return $TRUE }
```
fdurand, line 23: sub supportsWiredDot1x { return $TRUE; } ?

fdurand, line 23: sub supportsRadiusDynamicVlanAssignment { return $TRUE; } ?

julsemaan, line 23: 802.1X is untested and there is no deauth so that's why I didn't add it. I'll add the dynamic assignment.

PR #163, opened 2014-06-06
Added the way to select role or vlan id answer in switch config

fdurand:


UPGRADE.asciidoc
```diff
@@ -17,6 +17,12 @@ Make sure you run the following to update your schema:
 
   mysql -u root -p pf -v < db/upgrade-4.2.0-4.3.0.sql
 
+Configuration changes
+^^^^^^^^^^^^^^^^^^^^^
+
+The parameters `VlanMap` and `RoleMap` have been added in `switches.conf`; be sure to add them in the [default] switch section.
```
extrafu: You could add a note "See conf/switches.conf.example for and example."

PR #165, opened 2014-06-09
Added method to compute unreg date dynamicaly

ghost:
please comment this pull-request.
It allows to set an unreg date with null year or no year.

### DIFF
--- a/lib/pf/config.pm
+++ b/lib/pf/config.pm
@@ -39,6 +39,7 @@ use File::Which;
 use Socket;
 use List::MoreUtils qw(any);
 use Time::Local;
+use Time::Piece;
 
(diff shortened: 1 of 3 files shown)


lib/pf/config.pm
```diff
@@ -39,6 +39,7 @@ use File::Which;
 use Socket;
 use List::MoreUtils qw(any);
 use Time::Local;
+use Time::Piece;
```
fdurand, line 42: Cool a new perl module but there is nothing in the spec file and debian/control file. Does it available for centos/debian/ubuntu ?

fdurand:
NEWS file ?

julsemaan:
Closing to rebase. See #194

PR #166, opened 2014-06-13
Fix the Net::Appliance::Session SSH bug for wired deauth

julsemaan:
- Adds a pfcmd_vlan command to do the wired deauth outside of Packetfence.
- Plugs pf_run in the flow of pfsetvlan when doing wired deauth with SSH or Telnet
- Adds a method in Switch to manage the deauthentication like in wireless deauth 

For review and merge (target 4.3 ?)


julsemaan:
Closing since we're moving to http api deauth

PR #168, opened 2014-06-17
Fixed VLAN format to comply with RFC 2868.

louismunro:
The RFC states that "There is no restriction on the format of group IDs".
We now only make sure that the VLAN has a length of at least one byte.


NEWS.asciidoc
```diff
@@ -53,6 +53,7 @@ Bug Fixes
 * Fixed SQL join with iplog in advanced search of nodes
 * Fixed unreg date calculation in Catalyst captive portal
 * Fixed allowed_device_types array in device registration page (bugid:1809)
+* Fixed VLAN format to comply with RFC 2868.
```
extrafu: Remove trailing dot

PR #171, opened 2014-06-23
Feature/redis

mullagain:
Added pf::ipc module to extend Redis::Fast. It is also using JSON::XS for data serialization of things that need to be human readable and Storable for Objects/cached things.

I have implemented the following methods.
   enqueue - add a string/hash/array to a redis list (perl push). Auto serialization
   dequeue - pull something off of a redis list (perl shift). Auto serialization.
   queueCount - get a count of items on a redis list.
   pfpublish - redis publish with automatic serialization
   pfSubscribeHander - de-serialization of a published method (used with subscribe)
   cache - Similar to CHI's compute method only in Redis and very simple. (probably needs work).
   lcache - Same as cache accept process specific. The "l" stands for local. It prepends the pid to the key.

I have also included sbin/pftrapp that uses the native perl extension features of snmptrapd. This script breaks a trap into a hash and publishes it. This could work similar to tailing a log file, but a lot less complex and very fast as it all happens in memory.

Redis should be configured to be persistent in this case, and if there is a standby server redis should be configured with a slave.

This should greatly improve any ipc in terms of complexity and speed. It should also allow us to move away from the very slow and inefficient threaded perl.

It should also help with the active/active initiative as we could move more of the config into redis (SQL is overkill and inefficient for config imho)

I probably need more documentation and logging.


mullagain:
anybody?

louismunro:
I like it actually. 
But I know that James here has been working on something similar but more generalized that would support additional serialization formats (e.g. Sereal) . 

The short term goal though seems to be to send the traps (properly serialized) to the http API.
The code to replace internally generated traps is already in place in 4.3. It is just not called yet. 

The external SNMP traps are another matter though. Not because we can't serialize them, but because the current interface for Switches relies on a specific format for those traps in parseTrap() . It is too tightly coupled to pfsetvlan to just rip it out without careful planning. There is no doubt in my mind that the snmptrapd.log must die. I am also quite willing to let pfsetvlan die ASAP. 
It's just that we will need to redesign the Switch.pm interface and rethink the relationship with the rest of the daemons. 

pftrapp or something using the embedded perl interpreter in snmptrapd is definitely in the cards. Writing to a log file is slow and not flexible enough.
Things are a bit slow here today and last week. Some people are on vacation to take advantage of the fact that we have two holidays back to back.

mullagain:
I have used this exact method to parse traps into redis and it is blazing fast and stable. Since I added a subscription wrapper multiple processes would be able to see the traps come in as well. I have another pull request for pf::snmp, but it is not complete. I have questions and I wanted to do a pr before I went too far.

I was thinking of a http api into pf::ipc with access controls, etc. I have not seen the other serialization code so I cannot compare it to what I am doing, but I am willing to collaborate with James on this.

mullagain:
I actually like Sereal, I think I will move this over to that. There is no real reason to support more than one since we have control over what is being used. I thought JSON was good for things we may need to read, but we could just set a bit in redis to enable decoding and Dumpering of the ref.

mullagain:
Converted to Sereal and tested.

rouzier:
Thank your for contribution 

As Louie mentioned before we are working on refactoring pfsetvlan.
Some preliminary work is on the following branch.

https://github.com/inverse-inc/packetfence/tree/feature/pfsetvlan-refactor-redis

Our goals are the following.
- Reduce the complexity of pfsetvlan
- Make it easier to add new trap handling in the future.

The basic concept is the following.

For each supported trap OID have a pf::trap module that handles it.

pf::factory::trap is responsible for finding the switch and pf::trap module based off the trap data.

This initial work has code for the trap factory and a few pf::trap modules.

Is there what is currently missing is a working queuing model for the handling traps.

I do like your idea of using redis as a queue.

Could do the following redo the pull request for https://github.com/inverse-inc/packetfence/tree/feature/pfsetvlan-refactor-redis

Later this week I will set aside some time to review your request.

mullagain:
Sure thing!

jrouzierinverse:
This was done a different branch

PR #172, opened 2014-06-23
The upgrade script duplicates changes to the locationlog that are

louismunro:
already in the 4.0-4.1 script.

This causes an error where the last statements are never run and thus
the wrix is never added.


louismunro:
No longer required. Already fixed in devel.

PR #175, opened 2014-06-27
Feature/reuse dot1x credentials

jrouzierinverse:
This feature add the ability to reuse dot1x credentials for a specific profile.

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -220,11 +220,11 @@ sub setRole : Private {
     my $params = $c->stash->{matchParams};
     my $info = $c->stash->{info};
     my $pid = $info->{pid};
-    my $source_id = $session->{source_id};
+    my $source_match = $session->{source_match} || $session->{source_id};
     # obtain node information provided by authentication module. We need to get the role (category here)
     # as web_node_register() might not work if we've reached the limit
     my $value =
-      &pf::authentication::match( $source_id, $params, $Actions::SET_ROLE );
+      &pf::authentication::match( $source_match, $params, $Actions::SET_ROLE );
 
     # This appends the hashes to one another. values returned by authenticator wins on key collision
     if ( defined $value ) {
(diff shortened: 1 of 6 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
```diff
@@ -308,21 +306,32 @@ sub authenticationLogin : Private {
...
-      pf::authentication::authenticate( $username, $password, @sources );
-    if ( defined($return) && $return == 1 ) {
-        # save login into session
-        $c->session->{"username"} = $request->param("username");
-        $c->session->{source_id} = $source_id;
+    if($profile->noPasswordNeeded) {
+        my $mac       = $portalSession->clientMac;
+        my $node_info = node_view($mac);
+        my $username = $node_info->{'last_dot1x_username'};
+        if ($username =~ /^(.*)@/ || $username =~ /^[^\/]+\/(.*)$/ ) {
```
fdurand, line 337: Username is not always something like user@domain or domain/user, it depend of what we configured in site-available/packetfence-tunnel

html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
```diff
@@ -308,21 +306,32 @@ sub authenticationLogin : Private {
...
+        }
+        $c->session(
+            "username"  => $username,
+            "source_id" => $sources[0]->id
+        );
     } else {
-        $c->error($message);
+        # validate login and password
+        ( $return, $message, $source_id ) =
+          pf::authentication::authenticate( $username, $password, @sources );
```
fdurand, line 348: I suppose that @sources are equal to ($profile->getInternalSources, $profile->getExclusiveSources) so the default source list, can it be internal sources associated to the portal profile ?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
rouzier, line 337: We should figure out the best place to normalize the username from dot1x.
Based our conversation it could be normalize either in radius, the webservice call or in the portal (like it is now).

html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
rouzier, line 348: When the profile has reuse dot1x credentials enabled the first source defined is used.

html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
fdurand, line 337: So based on feature/realm, you will be able to check if stripped_user_name has been enabled in the source  and use node_info->{'stripped_user_name'} if defined or $node_info->{'last_dot1x_username'} if stripped_user_name has not been enabled.

jrouzierinverse, line 337: It be simpler to just require realm to be enabled when using reuse_dot1x_credentials.
Would there be any cases that someone would not be able to have realm turned on?

fdurand, line 337: Yes but no because a user athentication require stripped user name (sAMAccount name must match with toto, not DOMAIN\toto) but for a machine auth userprincipalname must match with HOST\my_laptop not my_laptop. So it depend if you are using machine auth or user auth.

PR #176, opened 2014-06-27
Feature/nat mode

dwlfrth:
Adding the support of nat / routed for inlinel2 interfaces from the GUI.
Also adding some informational messages

To be reviewed and merged.

### DIFF
--- a/html/pfappserver/root/interface/create.tt
+++ b/html/pfappserver/root/interface/create.tt
@@ -15,6 +15,8 @@
     [% form.field('fake_mac_enabled').render %]
     [% form.field('dhcpd_enabled').render %]
     [% form.field('nat_enabled').render %]
+    <p class="info_inline">[% l('Remember to enable ip_forward on your operating system for the inline mode to work.') %]</p>
+    <p class="info_routed">[% l('Since NATting mode is disabled, PacketFence will adjust iptables to rules to route traffic rather than natting it. Make sure to add the routes on the system.') %]</p>
   </div>
   <div class="modal-footer">
     <a href="#" class="btn" data-dismiss="modal">[% l('Cancel') %]</a>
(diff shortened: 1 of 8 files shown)


html/pfappserver/root/interface/create.tt
```diff
@@ -14,6 +14,9 @@
     [% form.field('vip').render %]
     [% form.field('fake_mac_enabled').render %]
     [% form.field('dhcpd_enabled').render %]
+    [% form.field('nat_enabled').render %]
+    <p class="info_inline">[% l('Remember to enable ip_forward on your operating system for the inline mode to work.') %]</p>
```
extrafu, line 18: Shouldn't these also be set in .po files for translation?

Same for all l('')s.

lib/pf/iptables.pm
```diff
@@ -214,8 +214,8 @@ sub generate_filter_if_src_to_chain {
             foreach my $network ( keys %ConfigNetworks ) {
                 next if ( !pf::config::is_network_type_inline($network) );
                 my $inline_obj = new Net::Netmask( $network, $ConfigNetworks{$network}{'netmask'} );
-                my $NAT = $ConfigNetworks{$network}{'nat'};
-                if (defined ($NAT) && ($NAT eq $NO)) {
+                my $nat = $ConfigNetworks{$network}{'nat'};
+                if (defined ($nat) && (!isenabled($nat))) {
```
extrafu: Minor nit here but IMO, we should define "isdisabled()".

It's a lot easier to read "if nat is defined and is disabled" then "If nat is defined and is not enabled".

html/pfappserver/root/interface/create.tt
cgx, line 18: We have a script that extracts the localizable strings (`addons/extract_i18n_strings.pl`).

PR #180, opened 2014-07-07
Do not cache profile

jrouzierinverse:


fdurand:
What happen in the case of a second pages for a profile based on a uri filter ? Does we stay on the portal based on uri or does we use another ?

PR #182, opened 2014-07-10
Inline Layer 3 is a remote network of an Inline Layer 2 - New filter based on the network (cidr representation)

fdurand:
For review and merge.

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -25,6 +25,7 @@ Enhancements
 
 * Added support to configure either NATting or routed mode for inline layer 2 interfaces from the GUI
 * Added informational messages in the GUI for inline interfaces
+* Improvement of Inline Layer 3 (Inline L3 can only be defined behind Inline Layer 2 network)
 
 Bug Fixes
 +++++++++
--- a/html/pfappserver/lib/pfappserver/Form/Field/ProfileFilter.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Field/ProfileFilter.pm
@@ -49,6 +49,7 @@ has_field 'type' =>
                { value => 'vlan', label => 'VLAN' },
                { value => 'switch', label => 'SWITCH' },
                { value => 'uri', label => 'URI' },
+               { value => 'ip', label => 'NETWORK' },
               ],
   );
 
--- a/html/pfappserver/lib/pfappserver/Form/Field/ProfileFilter.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Field/ProfileFilter.pm
@@ -49,7 +49,6 @@ has_field 'type' =>
                { value => 'vlan', label => 'VLAN' },
                { value => 'switch', label => 'SWITCH' },
                { value => 'uri', label => 'URI' },
-               { value => 'network', label => 'NETWORK' },
               ],
   );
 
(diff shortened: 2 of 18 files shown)


html/pfappserver/lib/pfappserver/Model/Interface.pm
```diff
@@ -715,9 +713,38 @@ sub ACCEPT_CONTEXT {
     return $object;
 }
 
+=head2 getEnforcement
+
+=cut
+
+sub getEnforcement {
```
cgx, line 721: Can't we reuse the `getType` function? Or call `getEnforcement` from `getType`? Please avoid duplicated code.

fdurand, line 721: Type is not the same thing than Enforcement and what i need is the enforcement value in pf.conf, not from networks.conf

cgx, line 721: I misread the code. Keep it like this. Thanks.

NEWS.asciidoc
```diff
@@ -18,11 +18,12 @@ New Features
 ++++++++++++
 
 * Added the possibility to search by computer name on the nodes page
+* New portal profile filter based on the ip address of the device
```
extrafu: New portal profile filter based on ranges of device's IP addresses

Would that be a better description?

html/pfappserver/lib/pfappserver/Form/Field/ProfileFilter.pm
```diff
@@ -49,6 +49,7 @@ has_field 'type' =>
                { value => 'vlan', label => 'VLAN' },
                { value => 'switch', label => 'SWITCH' },
                { value => 'uri', label => 'URI' },
+               { value => 'ip', label => 'NETWORK' },
```
jrouzierinverse: 'ip' should be change to either 'network' or 'cidr'

lib/pf/Portal/ProfileFactory.pm
```diff
@@ -33,7 +35,7 @@ Create a new pf::Portal::Profile instance based on parameters given.
 
 =cut
 
-our @MATCHES_TYPE = qw(uri ssid vlan switch);
+our @MATCHES_TYPE = qw(uri ssid vlan switch ip);
```
jrouzierinverse: 'ip' should be change to either 'network' or 'cidr'

lib/pf/Portal/ProfileFactory.pm
```diff
@@ -52,6 +54,16 @@ sub instantiate {
     my $node_info = node_view($mac) || {};
     $node_info = { %$options, %$node_info } ;
 
+    foreach my $filter (keys %Profile_Filters) {
+        if ($filter =~ /ip:(.*)/) {
```
jrouzierinverse: 'ip' should be change to either 'network' or 'cidr'

NEWS.asciidoc
fdurand: Or maybe something like : 
New portal profile based on Network range (ex: 192.168.0.0/24)

lib/pf/Portal/ProfileFactory.pm
fdurand: I vote for network, even if it's the ip address of the device we test, we change it on the fly by the network address.

PR #184, opened 2014-07-11
Add support for the Anyfi WIFI controller

julsemaan:
For review and merge

### DIFF
--- a/lib/pf/Switch/Anyfi.pm
+++ b/lib/pf/Switch/Anyfi.pm
@@ -2,7 +2,7 @@ package pf::Switch::Anyfi;
 
 =head1 NAME
 
-pf::Switch::WirelessModuleTemplate
+pf::Switch::Anyfi
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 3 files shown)


lib/pf/Switch/Anyfi.pm
```diff
@@ -0,0 +1,117 @@
+package pf::Switch::Anyfi;
+
+=head1 NAME
+
+pf::Switch::WirelessModuleTemplate
```
fdurand: pf::Switch::WirelessModuleTemplate

lib/pf/Switch/Anyfi.pm
```diff
@@ -0,0 +1,117 @@
...
+
+=head1 SYNOPSIS
+
+The pf::Switch::Anyfi module implements an object oriented interface to 
+manage the Anyfi gateway
+
+=head1 STATUS
+
+Developed and tested on Anyfi gateway running some version
```
fdurand: some version ?

PR #188, opened 2014-07-16
Fix issue where SOAP request were failing

jrouzierinverse:


cgx:
You can merge. It should also be merged into `maintenance/4.3`.

PR #189, opened 2014-07-17
Added option to set the timeout value for LDAP and AD sources.

louismunro:


cgx:
Looks good to me.

extrafu:
I don't think we should call the parameter "timeout".

We should call it "connection_timeout" since we could eventually have a "search_timeout".

PR #190, opened 2014-07-18
Renamed timeout into connection_timeout to distinguish from possible search timeout.

louismunro:
As requested by our dear Leader, I renamed the LDAP authentication source 'timeout' attribute into 'connection_timeout'. 
That leaves open the possibility to have a 'search_timeout' in the future.


extrafu:
Looks good to me.

PR #192, opened 2014-07-21
Return the error if exec fails

jrouzierinverse:


src/pfcmd.c
```diff
@@ -21,9 +21,8 @@
 #define REAL_PATH "/usr/local/pf/bin/pfcmd.pl"
 #endif
 
-int main(ac, av)
-char **av;
+int main(int argc,char** argv,char ** envp)
 {
-    execv(REAL_PATH, av);
-    return 0;
+    execve(REAL_PATH, argv, envp);
```
extrafu, line 26: I don't understand this change.

Why don't you simply do:

return execve(REAL_PATH, argv, envp); 

On success, execve never returns and error, it returns -1 and sets errno.

?

rouzier, line 26: I am using the return code for 'command not found' which is 127.
Which would be the only reason I would see execve failing.

PR #195, opened 2014-07-23
Modified default radiusd configuration to make it more resilient to

louismunro:
devices that would take too long to authenticate or never reply to a
retry message after ntlm authentiation failures.

Also, configured the mschap module to settings that make the most sense
for PF.


fdurand:
Ok for me, let´s merge

PR #196, opened 2014-07-23
Feature/vlan filter improvement

fdurand:
Added the radius request to be parsed in the vlan filter.

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -33,6 +33,7 @@ Enhancements
 * Added an option to set the timeout value for LDAP connections in authentication sources
 * FreeRADIUS default configuration should now be more scalable and resilient to misbehaving devices.
 * Added the possibility to create rules using the username in OAuth authentication sources
+* Added the radius request to the vlan filter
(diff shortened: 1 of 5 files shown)


NEWS.asciidoc
```diff
@@ -29,6 +29,7 @@ Enhancements
 * Improvement of Inline Layer 3 (Inline L3 can only be defined behind Inline Layer 2 network)
 * pfbandwidthd is now able to capture on all inline interfaces
 * Added an option to set the timeout value for LDAP connections in authentication sources
+* Added the radius request to the vlan filter
```
extrafu: radius -> RADIUS

lib/pf/vlan/filter.pm
```diff
@@ -138,6 +139,48 @@ sub node_info_parser {
     }
 }
 
+=item radius_parser
+
+Parse the radius request attribute and compare to the rule. If it matches then perform the action.
```
extrafu: radius -> RADIUS

PR #197, opened 2014-07-23
Feature/ntlm auth wrapper

louismunro:
Add a utility ("bin/ntlm_auth_wrapper") to log ntlm authentication times.

### DIFF
--- a/src/ntlm_auth_wrap.c
+++ b/src/ntlm_auth_wrap.c
@@ -4,8 +4,7 @@ WARNING: We cheat and do no bother to free memory allocated to strings here.
 The process is meant to be very short lived an never reused. */
 
 #define _POSIX_C_SOURCE 200809L 
-#define COMMAND "/bin/echo"
-//#define COMMAND "/usr/bin/ntlm_auth"
+#define COMMAND "/usr/bin/ntlm_auth"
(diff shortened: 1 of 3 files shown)


fdurand:
Arch dependent, cannot be merge like that.
You must add packaging change

src/ntlm_auth_wrap.c
```diff
@@ -0,0 +1,78 @@
+/* A wrapper around ntlm_auth to log arguments and
```
extrafu, line 1: Please add the copyright/license header like in perl files.

louismunro, line 1: Done. Added the header from pfcmd.c.

jrouzierinverse:
Please rebase
Add compile step to debian/rules and addons/packages/packetfence.spec

dwlfrth:
I think that should be merged and maybe even used by default in our call of the mschap module.
That adds some interesting logging.

dwlfrth:
What's my status ?

louismunro:
I am working on adding CLI options to it for things like log level and even StatsD support.

Also it needs to be compiled for each architecture so there's packaging involved.

---

LM

louismunro:
Closing. Will be part of the dashboard rework in 5.1.

PR #200, opened 2014-07-24
Removed the "magic" from pf::config::cached

jrouzierinverse:


louismunro:
That's it. 
The magic is gone. 
PF lost it's mojo.

PR #204, opened 2014-07-25
Feature/internal traps

louismunro:
For testing and inclusion in 4.4.

Adds the code changes to enable internal traps through the WebAPI. 
Adds the GUI setting to control the delay before VLAN reassignment or deassociation.

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -194,7 +194,7 @@ sub ReAssignVlan {
 
     my $switch = pf::SwitchFactory->getInstance()->instantiate( $postdata{'switch'} );
 
-    sleep $pf::config::Config{'vlan'}{'wait_for_redirect'}; 
+    sleep $pf::config::Config{'trapping'}{'wait_for_redirect'}; 
 
     # SNMP traps connections need to be handled specially to account for port-security etc.
     if ( ($postdata{'connection_type'} & $pf::config::WIRED_SNMP_TRAPS) == $pf::config::WIRED_SNMP_TRAPS ) {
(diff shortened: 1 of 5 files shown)


lib/pf/api.pm
```diff
@@ -216,6 +218,9 @@ sub desAssociate {
 
     my ($switchdeauthMethod, $deauthTechniques) = $switch->deauthTechniques($switch->{'_deauthMethod'});
 
+    # sleep long enought to give the device enough time to fetch the redirection page.
```
julsemaan: *sleep long enough

fdurand:
In api.pm you forgot to add sleep for firewall sub. (inline)

louismunro:
I did not forget.
Based on Derek's testing it seemed not to be necessary when n inline mode
snce the server is always reachable.

Testing will show whether that assumption is correct.

lib/pf/api.pm
```diff
@@ -194,6 +194,8 @@ sub ReAssignVlan {
 
     my $switch = pf::SwitchFactory->getInstance()->instantiate( $postdata{'switch'} );
 
+    sleep $pf::config::Config{'vlan'}{'wait_for_redirect'};
```
julsemaan: sleep $pf::config::Config{'trapping'}{'wait_for_redirect'};

It's in the trapping section

lib/pf/api.pm
```diff
@@ -216,6 +218,9 @@ sub desAssociate {
 
     my ($switchdeauthMethod, $deauthTechniques) = $switch->deauthTechniques($switch->{'_deauthMethod'});
 
+    # sleep long enought to give the device enough time to fetch the redirection page.
+    sleep $pf::config::Config{'vlan'}{'wait_for_redirect'};
```
julsemaan: same here

PR #205, opened 2014-07-28
Xirrus + AeroHIVE external portals + small ecwp refactor

julsemaan:
Adds :
- The external captive portal detection can now be done with a URL containing the name of the switch module to use (ex : http://1.1.1.1/Xirrus::AP_http)
- The redirect url is sent back from the external captive portal module so it can be set immediatly since external captive portals won't necessary go to /captive-portal anymore
- The CGI session ids are different from the catalyst captive portal so both can be used in parrallel
- The arguments initially sent to the captive portal are kept in the session (on lib/pf side) so they can be reused. They are kept in : 'ecwp-original-param-$key' = $value
- Adds support to release a user through the submision of html code through the captive portal (posting a form to the ap for example)

Needs a good review for the way of doing things.


NEWS.asciidoc
```diff
@@ -22,6 +22,7 @@ New Features
 * Show portal profiles directly on the admin gui
 * Added local account authentication for EAP
 * Added support for unreg date with dynamic year
+* Added external captive portal support for AeroHIVE
```
extrafu, line 27: and Xirrus?

extrafu:
Nice work! (tm) Ludo

lib/pf/Switch/AeroHIVE/AP_http.pm
```diff
@@ -0,0 +1,145 @@
...
+            <input type="hidden" name="password" value="$mac">
+            <input type="hidden" name="ssid" value="$last_ssid">
+            <input type="hidden" name="url" value="$destination_url">
+        </form>
+        <script language="JavaScript" type="text/javascript">
+        window.setTimeout('document.weblogin_form.submit();', 1000);
+        </script>
+    ];
+
+    $logger->info($html_form);
```
fdurand: debug

lib/pf/Switch/Xirrus/AP_http.pm
```diff
@@ -0,0 +1,170 @@
...
+=cut
+
+sub parseUrl {
+    my($this, $req, $r) = @_;
+    my $logger = Log::Log4perl::get_logger( ref($this) );
+    my $connection = $r->connection;
+    $this->synchronize_locationlog("0", "0", clean_mac($$req->param('mac')),
+        0, $WIRELESS_MAC_AUTH, clean_mac($$req->param('mac')), $$req->param('ssid')
+    );
+    return (clean_mac($$req->param('mac')),$$req->param('ssid'),$connection->remote_ip,$$req->param('userurl'),undef,"200");
```
fdurand, line 72: Not really agree with $connection->remote_ip, remote_ip is the source ip of the client so if you are behind a nat gateway this will not be the real ip address.  isn´t it possible to have the real device ip from the uri ? Or document  the pod that it´s not possible to use the AP behind a nat gateway.

julsemaan, line 72: That is a limitation since it's not in the URI. I'll add it to the doc

julsemaan, line 72: Fixed in 2509d1ced09e3088a470863453ae2052cf3ce3d6
Ready to merge ?

lib/pf/Switch/Xirrus/AP_http.pm
```diff
@@ -0,0 +1,170 @@
...
+grant url
+status code
+
+=cut
+
+sub parseUrl {
+    my($this, $req, $r) = @_;
+    my $logger = Log::Log4perl::get_logger( ref($this) );
+    my $connection = $r->connection;
+    $this->synchronize_locationlog("0", "0", clean_mac($$req->param('mac')),
```
fdurand, line 69: Can it be defined directly in externalportal.pm ?

julsemaan, line 69: Shouldn't be since it's not all the captive portals that work without RADIUS requests.
Doing it at the global level would open 2 location log entries. 1 for the RADIUS request + 1 during the external portal detection.

fdurand, line 69: Ok so ready to merge

PR #208, opened 2014-08-04
Pf client should not die sending a notify message

jrouzierinverse:


fdurand:
Tested and ready to merge

PR #209, opened 2014-08-04
Feature/net snmp perl

mullagain:
Added pf::snmp module to use SNMP.pm from net-snmp instead of Net::SNMP. SNMP.pm has huge performance gains over Net::SNMP. I have created an initial connectRead to replace the method in Switch.pm and get/get_tables methods to make read requests look the same as Net::SNMP.

This is also a Moose based class.

Future support will include set requests, traps, etc.


dwlfrth:
What's my status ?

mullagain:
What do you mean?

dwlfrth:
Asking if it is still accurate, regarding the comment in pull/210.

louismunro:
I actually like this stuff. 
While we aim to no longer use SNMP for authentication in  PF 6, the performance gains might be worth it. I will keep this around until we have some time to benchmark it.

louismunro:
Sorry. This was a good idea, but we went in a different direction in the end.

PR #210, opened 2014-08-04
Feature/redis

mullagain:
Added pf::ipc module to extend Redis::Fast. It is also using Sereal for data serialization.

I have implemented the following methods.
enqueue - add a string/hash/array to a redis list (perl push). Auto serialization
dequeue - pull something off of a redis list (perl shift). Auto serialization.
queueCount - get a count of items on a redis list.
pfpublish - redis publish with automatic serialization
pfSubscribeHander - de-serialization of a published method (used with subscribe)
cache - Similar to CHI's compute method only in Redis and very simple. (probably needs work).
lcache - Same as cache accept process specific. The "l" stands for local. It prepends the pid to the key.

I have also included sbin/pftrapp that uses the native perl extension features of snmptrapd. This script breaks a trap into a hash and publishes it. This could work similar to tailing a log file, but a lot less complex and very fast as it all happens in memory.

Redis should be configured to be persistent in this case, and if there is a standby server redis should be configured with a slave.

This should greatly improve any ipc in terms of complexity and speed. It should also allow us to move away from the very slow and inefficient threaded perl.

It should also help with the active/active initiative as we could move more of the config into redis (SQL is overkill and inefficient for config imho)

I probably need more documentation, logging and add support for redis sentinels.


lib/pf/ipc.pm
```diff
@@ -0,0 +1,193 @@
...
+
+sub enqueue {
+    my ($self,$q,$item) = @_;
+    my $txt;
+    if ( ref($item) ) {
+        $txt = $self->encoder->encode( $item );
+    }
+    else {
+        $txt = $item;
+    }
```
jrouzierinverse: Should always serialize the data going into the queue so it is easier to change the serializer later.

lib/pf/ipc.pm
```diff
@@ -0,0 +1,193 @@
...
+sub dequeue {
+    my ($self,$q) = @_;
+    my $txt = $self->lpop($q);
+    my $ret;
+    if ($txt =~ /(\{|\[)/) {
+        $ret = $self->decoder->decode($txt);
+    }
+    else {
+        $ret = $txt;
+    }
```
jrouzierinverse: Should always deserialize the data coming out of  the queue so it is easier to change the serializer later.

lib/pf/ipc.pm
```diff
@@ -0,0 +1,193 @@
...
+    if ($message =~ /(\{|\[)/) {
+        $ret = $self->decoder->decode($message);
+    }
+    else {
+        $ret = $message;
+    }
+    return $ret;
+}
+
+sub lcache {
```
jrouzierinverse: This is not needed because we already have a caching layer pf::CHI

lib/pf/ipc.pm
```diff
@@ -0,0 +1,193 @@
...
+    my ($self,$key,$ttl,$ref) = @_;
+    $self->cache($key,$ttl,$ref,$$);
+}
+
+sub cache {
```
jrouzierinverse: This is not needed because we already have a caching layer pf::CHI

lib/pf/ipc.pm
```diff
@@ -0,0 +1,193 @@
...
+
+use Sereal::Encoder;
+use Sereal::Decoder;
+
+has 'encoder' => (is => 'ro', default => sub { Sereal::Encoder->new; });
+has 'decoder' => (is => 'ro', default => sub { Sereal::Decoder->new; });
+
+sub FOREIGNBUILDARGS {
+    my $class = shift;
+    my %args = shift;
```
jrouzierinverse: shift only returns a single value
the rest of the arguments will not be returned

sbin/pftrapp
```diff
@@ -0,0 +1,36 @@
...
+        $x->[1] =~ s/^.+\:\s+//g;
+        $data->{varbinds}->{$x->[0]} = $x->[1];
+    }
+
+    # dump trap into a queue for a process to read.
+    # $ipc->enqueue('traps',$data) or warn $!;
+
+    # can use a publish/subscriber as well to allow multiple processes to read the messages.
+    # But the message is lost if there is nobody listening.
+    $ipc->pfpublish('sub_traps',$data);
```
jrouzierinverse: We are going to use list for the queue not the publish/subscribe model.
To avoid processing the same message twice.

mullagain:
Sounds good. I provided examples of both in the case that multiple processes needed to read the trap list. Digging through the code I found mulitple cases of the a process reading the trap log, which is the reason I provided the example. I can move it back to a simple queue.

mullagain:
I went ahead and updated everything. In other apps that I have written I used the list as well. I works very very well. I have a tool i have written to watch the queues. If i knew the queue names I could submit it as part of the toolset.

mullagain:
Caching methods have been removed.

dwlfrth:
What is my status ?
Also, is pull/209 still relevant since this one include the whole pf::snmp.

mullagain:
I will double check here shortly

mullagain:
Yes, you can ignore the net-snmp pull request since snmp.pm is included in
this one. I am only submitting with read capabilities to allow for further
testing. Once that is completed I will finished the write methods, etc.

mullagain:
snmp.pm in the net-snmp pull request is different. I did not intend to include snmp.pm in this request.

louismunro:
Hi Mullagain,
While I am closing this PR, it should not be considered a slight agains you.
We are moving towards implementing some of the same ideas in the coming releases.

I suggest you have a look at 5.5 when it comes out. 
We will be integrating Redis and increasingly making use of it in the coming releases.

Thank you for your contribution. I am sorry it could not be merged as is.

PR #211, opened 2014-08-05
New service iptables

jrouzierinverse:


lib/pf/services/manager/iptables.pm
```diff
@@ -0,0 +1,127 @@
...
+
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2013 Inverse inc.
```
extrafu: 2014

lib/pf/pfcmd/help.pm
```diff
@@ -155,6 +155,7 @@ Services managed by PacketFence:
   httpd.admin      | Apache Web admin
   httpd.portal     | Apache Captive Portal
   httpd.proxy      | Apache Proxy Interception
+  iptables         | firewall setting
```
extrafu: A better description should be given.

fdurand:
Look good for me, except that the NEWS.asciidoc doc is missing

PR #213, opened 2014-08-06
Added ability to search switches

jrouzierinverse:


html/pfappserver/lib/pfappserver/Controller/Configuration/Switch.pm
```diff
@@ -83,6 +87,38 @@ after list => sub {
...
+
+sub search : Local : AdminRole('SWITCHES_READ') {
+    my ($self, $c, $pageNum, $perPage) = @_;
+    $pageNum = 1 unless $pageNum;
+    $perPage = 25 unless $perPage;
+    my ($status, $status_msg, $result, $violations);
+    my %search_results;
+    my $model = $self->getModel($c);
+    my $form = $self->getForm($c);
+    # Store columns in the session
```
cgx: This comment is not relevant here.

julsemaan:
Looks good to me

html/pfappserver/lib/pfappserver/Model/Config/Switch.pm
```diff
@@ -16,6 +16,7 @@ use Moose;
 use namespace::autoclean;
 use pf::config::cached;
 use pf::config;
+use pf::log;
```
cgx: We don't log anything from this model. Remove this line.

html/pfappserver/root/configuration/switch/index.tt
```diff
@@ -14,10 +14,49 @@
...
+                        </div>
+                      </td>
+                    </tr>
+                  <tbody>
+                </table>
+              </form>
+            </div>
+          </div>
+        </div>
+      </div><!--/span-->
```
cgx, line 57: I don't like a tab panel with a single tab. I think we can merge the PR but I'll rework the layout afterwards.

cgx:
Please update the `NEWS` file.

jrouzierinverse:
NEWS file updated

mullagain:
Is there any movement on removing the Switch config file and moving it to the db? I have a network discovery/management system that I would love to integrate into pf.

louismunro:
I personally am not favorable to removing the config file.
It makes operations and support *much *easier.

You can ask people to post the relevant sections to the mailing list, you
can run a diff between two versions (easier than if it were in the DB at
any rate),
you can grep it, run sed scripts on it etc.

While some of that can also be done from the database (you can just run an
UPDATE table WHERE) I remain a big fan of flat files on unix.

I would say the benefits would have to outweigh the inconvenience and I
remain open to arguments, always keeping in mind that it's people like me
who will have to support clients with that.

Regards,
LM

mullagain:
I can see your points, but support tools could be written to gather the same information. Also automation becomes much more predictable and manageable when you move to a DB, not to mention data consistency. Plus with the goal of active/active flat files become less than ideal. 

I asked the question because I have written several network(switch/router/ap/controller) discovery engines over the years at different companies. It would be incredibly easy for me to create one for pf, but these systems use a database and managing/syncing the db to a flat file could become cumbersome. I am not even sure if there is an interest in a discovery engine, but it sure would make managing the switches easier... because it would be fully automated. 

If there is an interest in this, I contend that a single data source would be needed, A flat file is not the answer for network discovery or any type of automation.

Also, support tools could be created to gather troubleshooting information.

louismunro:
Speaking to others here, I may be the old fuddy-duddy unix greybeard
(without the beard, but still...).
My love of text files is not universally shared it would seem.

I wish other would speak up about this. I think it's worth discussing.
In any case, the support tools would have to be written and tested well
before we move everything to the database.
pfcmd is still a work in progress in that respect I believe.

To be frank, I am not sure this is the most critical thing keeping us from
having an active/active HA setup, which I think is the ultimate goal here.
We already know how to have an active/passive setup even though there may
be room for improvement there too.

What I see as things to fix first is the issue of services running on every
node of the cluster when they might conflict:
active/active dhcp (with shared list of leases between nodes),
active/active dhcplistener (currently may cause database issues if running
more than one simultaneously), SNMP traps, iptables and ipsets in inline
mode etc.

Some of them require rewriting a service (DHCP), while others require that
we decide how to deal with devices that may only send packets to a single
IP (SNMP in the case of some switches).
Do we use a virtual IP for the cluster? That's one more thing to manage,
one more thing that can fail.

Fixing some of these issues would at the same time make an active/passive
setup easier, so the benefits may be compounded and a valid strategy may be
to do this incrementally until full active/active HA is acheived.

None of these issues are unsolvable. I am not trying to belittle the idea,
in fact in think it's a worthwhile goal.
But we should not underestimate the tradeoffs and choices it will entail,
and operability should remain a primary concern.

Ok, rant off :-)
Please do chime in people...

L.

mullagain:
Adding devel list

mullagain:
Michael wrote:

> ```
> Speaking to others here, I may be the old fuddy-duddy unix greybeard
> (without the beard, but still...).
> My love of text files is not universally shared it would seem.
> ```

Test files are nice, but...  I think the biggest key to this is being
able to change settings at runtime.  Writing to a text file is nice
because, generally speaking, it's difficult to corrupt a text file.
Yes, I know it's possible, but you really have to work at it..  ;P

Databases also have that interesting drawback of "where do I put the
database?" ...  If it's on the node that fails, then what?  Sure you can
replicate, but cross replication with writes to both nodes can be quite
challenging..

> ```
> I wish other would speak up about this. I think it's worth discussing.
> In any case, the support tools would have to be written and tested well
> before we move everything to the database.
> pfcmd is still a work in progress in that respect I believe.
> ```

Properly written, it should be really simple to switch from
reading/writing text files to reading/writing to a database.  In either
case, the admin shouldn't have to be worried about where the config is
being stored.  Things should work well enough that this is irrelevant.

It's more advanced users that I think want the database functionality.
And that's primarily so we can manipulate things in a programmatic
fashion.  Since there isn't a robust API at this point, directly
manipulating configuration is the only way to do things.

Perhaps it may be even better to work on an API?  Given a robust API, I
would likely never try to directly access the database or text files.

> ```
> To be frank, I am not sure this is the most critical thing keeping
> us from
> having an active/active HA setup, which I think is the ultimate goal
> here.
> We already know how to have an active/passive setup even though
> there may
> be room for improvement there too.
> ```

Why would this prevent active/active?  It shouldn't be hard to trigger
an rsync when needed to sync the text files.  There are, of course,
drawbacks to this, but it can be done.

OR..  at least one product we currently use leverages a promise-theory
engine (CFEngine, Puppet, Chef, etc) to accomplish configuration changes
across multiple servers.  That _might_ be overkill for packetfence, but
maybe not..  Worth thinking about at any rate.

> ```
> What I see as things to fix first is the issue of services running
> on every
> node of the cluster when they might conflict:
> active/active dhcp (with shared list of leases between nodes),
> active/active dhcplistener (currently may cause database issues if
> running
> more than one simultaneously), SNMP traps, iptables and ipsets in
> inline
> mode etc.
> ```

active/active dhcp using ISC DHCP is actually really easy.  We run this
already.  The only real limitation here is that you can only have two
nodes, but I don't see that as an issue for packetfence at this stage.

> ```
> Some of them require rewriting a service (DHCP), while others
> require that
> we decide how to deal with devices that may only send packets to a
> single
> IP (SNMP in the case of some switches).
> Do we use a virtual IP for the cluster? That's one more thing to
> manage,
> one more thing that can fail.
> ```

This isn't strictly true.  I can only speak for Cisco here, but I can
configure the switches to SNMP traps to more than one IP pretty easily.
 In fact, we do this quite often.

The harder part (though I'm not sure it's _THAT_ hard) is to have only
one of the servers process the request.  Should that be done by
assigning switches to server?  Maybe it's done in a first-in basis?  I
think the answer may depend on what the admin is after.  Maybe it's load
balancing, maybe it's speed.  Or, maybe it doesn't matter either way and
the admin just wants an HA setup with zero downtime in the event of a
failure.

You'd still need a virtual IP, though, just for handling access to the
admin interface.

> ```
> Fixing some of these issues would at the same time make an
> active/passive
> setup easier, so the benefits may be compounded and a valid strategy
> may be
> to do this incrementally until full active/active HA is acheived.
> ```

Not a bad strategy.

> ```
> None of these issues are unsolvable. I am not trying to belittle the
> idea,
> in fact in think it's a worthwhile goal.
> But we should not underestimate the tradeoffs and choices it will
> entail,
> and operability should remain a primary concern.
> ```

Yes.  No breaking my servers, please.  Packetfence is a tad important.  :P

> ```
> Ok, rant off :-)
> Please do chime in people...
> ```

DING!

> ```
> L.
> ```

## 

---

Jason 'XenoPhage' Frisvold

## xenophage@godshell.com

"Any sufficiently advanced magic is indistinguishable from technology.\"
- Niven's Inverse of Clarke's Third Law

roadracer96:
Would totally fubar my active active setup.  Unless node specifics could be over ridden or maybe have system default configuration be node 0 and have duplicate options be node 1,2,3 or something.

My .02

Ps:  can still integrate plaintext configuration with other systems.  ;)

mullagain:
Has anyone's opinion changed on this with the move to MariaDB cluster?

PR #215, opened 2014-08-11
Added a cleanup of the posted username on /login.

julsemaan:
Cleans up the username that's sent during the post to /login

Removes leading + trailing whitespaces.

Had reports of tablets auto-completing the username with a space at the end causing this issue

Should we merge this or ignore it since it's caused by a bug on the client side ?

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -46,6 +46,7 @@ Bug Fixes
 * Fix previewing templates in the admin GUI
 * Fix bulk applying of roles and violations in the admin GUI
 * Fix importing of nodes when no pid is given
+* Added a cleanup of trailing and leading spaces of the posted username during the login
 
 Version 4.5.1 released on 2014-11-10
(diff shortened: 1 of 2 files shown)


julsemaan:
Can someone comment + review + discuss. 

This is an intrusive PR and brings a maybe unexpected feature for users

dwlfrth:
the _clean_username method doesn't belong in the Controller as per MVC way of thinking :)
Also, since we may want to use it elsewhere in the code, I'd put it in the util class (pf::util).

Oherwise, regex seems OK.
Would be great to test with a username containing spaces (ie: derek wuelfrath)

julsemaan:
My opinion would be that it belongs there because it's correcting the user input that comes from the view.

The cleanup is not needed by the backend but is necessary because of the view.

julsemaan:
This has been rebased. Can this be merged ?

PR #217, opened 2014-08-13
Reworked log format [] is the device mac address, () is the switchID "" for userID

fdurand:
Reworked log format [] is the device mac address, () is the switchID "" for userID


julsemaan:
Looks good, rebase and we can merge

PR #218, opened 2014-08-13
Fix case mismatch when updating a user in the database.

louismunro:
modified:   lib/pf/person.pm

Fixes case where a person does not get updated in the DB when the
username send over 802.1x or through the portal differs from the one in
the DB by case only.

E.g. Juser does not match juser.

### DIFF
--- a/lib/pf/person.pm
+++ b/lib/pf/person.pm
@@ -332,7 +332,7 @@ sub person_modify {
     my $new_pid   = $existing->{'pid'};
 
     # compare pid case insensitively to prevent juser from not matching Juser
-    if ( $pid !~ /$new_pid/i && person_exist($new_pid) ) {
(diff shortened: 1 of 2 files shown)


lib/pf/person.pm
```diff
@@ -331,7 +331,8 @@ sub person_modify {
     }
     my $new_pid   = $existing->{'pid'};
 
-    if ( $pid ne $new_pid && person_exist($new_pid) ) {
+    # compare pid case insensitively to prevent juser from not matching Juser
+    if ( $pid !~ /$new_pid/i && person_exist($new_pid) ) {
```
cgx: Is `$pid` always lowercased? To make sure, I would do `lc($pid) ne lc($new_pid)` instead.

PR #220, opened 2014-08-15
modified:   html/pfappserver/lib/pfappserver/Form/Authentication/Source/...

louismunro:
...LDAP.pm

```
Added a default value to prevent to form from choking on "undef" if
the value is not set manually.
```


cgx:
Looks good.

PR #221, opened 2014-08-19
Feature/local accounts external sources

dwlfrth:
Ability to create a local account in PacketFence for registration using external authentication sources.

Major rework needs to be done in whole guest workflow but that is another project.
For the moment, SMS registration is not able to create local account (will be integrated later after some discussion)

Ready to be merged... Please review

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Oauth2.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Oauth2.pm
@@ -2,6 +2,7 @@ package captiveportal::PacketFence::Controller::Oauth2;
 use Moose;
 use namespace::autoclean;
 use pf::config;
+use pf::util qw(isenabled);
 use pf::web;
 use Net::OAuth2::Client;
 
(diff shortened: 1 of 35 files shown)


fdurand:
The code look good, but it miss change in news.asciidoc file.
You probably have to update the traduction for the admin gui , like remove create_local_account_on_email_reg and create_local_account_on_email_sponsor and add the new one.

html/captive-portal/lib/captiveportal/PacketFence/Controller/Oauth2.pm
```diff
@@ -233,6 +233,28 @@ sub oauth2Result : Path : Args(1) {
         $c->forward('Authenticate' => 'postAuthentication');
         $c->forward('CaptivePortal' => 'webNodeRegister', [$pid, %{$c->stash->{info}}]);
         $c->forward('CaptivePortal' => 'endPortalSession');
+
+        # Create local account for external authentication sources (if configured to do so)
+        if ( isenabled($source->{create_local_account}) ) {
```
extrafu, line 239: This code block is the same in 3 different auth modules. Shouldn't it be moved to a utility module?

dwlfrth:
@fdurand 
Will update NEWS file once ready to merge... after test results
Regarding old parameters and traductions, already done... unless you see somewhere i forgot ?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Oauth2.pm
dwlfrth, line 239: Already done but in a refactor branch...
I was doing a lot of refactoring and didn't want to modify that much the actual codebase to make it easier to "read/merge"
Will put it in this branch if better.

The only reason why it is not in this branch is because some parameters are not the same but the refactoring made it the same.

fdurand:
Line 728 from html/pfappserver/lib/pfappserver/I18N/i_default.po (conf/documentation.conf guests_self_registration.create_local_account_on_email_reg) and after

dwlfrth:
@fdurand 
Done

fdurand:
It looks ok for me, ready to merge

PR #225, opened 2014-09-02
Do not override the controllerIp that is defined in switches.conf

jrouzierinverse:


fdurand:
Missing entry in the NEWS.asciidoc file

PR #227, opened 2014-09-02
Adjusted limits in httpd.portal to better handle heavy loads.

louismunro:
Lowered timeouts and raised maximum number of processes forked to handle heavy loads.


lib/pf/services/manager/httpd.pm
```diff
@@ -144,7 +144,7 @@ sub calculate_max_clients {
     my $max_clients = ceil(($total_ram - ( $total_ram * 0.25 + (300 * 1024) )) / (50 * 1024));
 
     # hard ceiling of MaxClients at 256
```
julsemaan, line 146: Should be hard ceiling to 512 now

PR #229, opened 2014-09-03
Removed commented code because no one maintains it and can lead to depre...

dwlfrth:
...cated code issue.


dwlfrth:
Ready to be merged...

julsemaan:
I would still leave the example methods in custom.pm with a simple example and a call to super at the end. That would remove the need to maintain the methods while showing how to use them.

dwlfrth:
@julsemaan: Good idea, done.

dwlfrth:
Can we merge ?

PR #230, opened 2014-09-04
Fix/local account creation

dwlfrth:


dwlfrth:
Ready to be reviewed and merged.
(Julien is probably the best placed to do this)

PR #232, opened 2014-09-08
Added the ability to lookup exclusive authentication rules when getting VLAN.

louismunro:
This allows for example an Eduroam SSID to use "NULL"auth and
still get more complex authorization rules, as well as using violations.


rouzier:
Looks good

PR #233, opened 2014-09-09
Reworked logrotate.d/packetfence.

louismunro:
Logrotate configuration does not support escaping NL.

This is a fix to a bug in devel that is not in stable, so no NEWS.asciidoc.


julsemaan:
That worked for me so we can merge

PR #235, opened 2014-09-11
New source blackhole that never succeeds in authorizations

jrouzierinverse:


extrafu:
What would be the purpose of that?

lib/pf/Authentication/Source/BlackholeSource.pm
```diff
@@ -0,0 +1,119 @@
...
+
+  my $super_attributes = $self->SUPER::available_attributes;
+  my $own_attributes = [{ value => "username", type => $Conditions::SUBSTRING }];
+
+  return [@$super_attributes, @$own_attributes];
+}
+
+=head2 available_actions
+
+For an Null source, we limit the available actions to B<set role>, B<set access duration>, and B<set unreg date>.
```
extrafu: Null?

lib/pf/Authentication/Source/BlackholeSource.pm
```diff
@@ -0,0 +1,119 @@
...
+
+=cut
+
+has '+class' => (default => 'exclusive');
+
+has '+type' => (default => 'Blackhole');
+
+=head2 available_attributes
+
+Allow to make a condition on the user's email address.
```
extrafu, line 35: Ugh?

lib/pf/Authentication/Source/BlackholeSource.pm
jrouzierinverse: This is a source that always fails it is the opposite of the Null source

extrafu: Sure, but that's not the Null source, but the Blackhole one.

extrafu:
Please merge it so we have that in v4.5.

PR #238, opened 2014-09-30
Fix timestamp solution to catch sub second changes in file

jrouzierinverse:


lib/pf/config/cached.pm
```diff
@@ -783,11 +785,13 @@ sub setControlFileTimestamp {
     $self->{_control_file_timestamp} = getControlFileTimestamp();
 }
 
-sub getControlFileTimestamp { int((stat($cache_control_file))[9] || 0) * 1000000000 }
+sub getControlFileTimestamp {
```
cgx, line 800: The pod header is missing for this function.

lib/pf/config/cached.pm
```diff
@@ -783,11 +785,13 @@ sub setControlFileTimestamp {
...
+    return pf::IniFiles::_getFileTimestamp($cache_control_file);
+}
 
 sub controlFileExpired {
```
cgx, line 810: Please add a pod header for this function.

jrouzierinverse, line 810: Done

PR #243, opened 2014-10-03
Added quick guide for Anyfi controller + gateway

julsemaan:
Not sure if I got all the pieces for creating the doc.

For review + merge


fdurand:
You need to change packetfence.spec (line 343)  to build your doc.

julsemaan:
Other guides are not build (like Cisco and Brocade), we should probably find a good prefix for these documents and add it in packetfence.spec instead of individual files. I'll include the refactor in this PR

julsemaan:
Added the new guide + other quick guides to the packetfence.spec

PR #244, opened 2014-10-03
Added portal profile filter based on switch port couple

fdurand:
New portal filter like Switch-ifindex

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -21,6 +21,7 @@ New Features
 * Added a "Blackhole" auth source 
 * Added new violation enforce provisioning
 * Violation can now be delayed
+* added portal profile filter based on switch-port couple
 
 Enhancements
 ++++++++++++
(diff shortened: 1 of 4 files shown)


NEWS.asciidoc
```diff
@@ -17,6 +17,7 @@ Version x.y.x released on yyyy-mm-dd
 New Features
 ++++++++++++
 * pfmon can now run tasks on different intervals
+* added portal profile filter based on switch-port couple
```
extrafu, line 24: shouldn't we use "pair" here instead of couple? same everywhere below.

louismunro, line 24: Tuple?

extrafu, line 24: That would work too - I just had in mind "key-value pair" which is common in programming. Couple is common is procreating ;-)

louismunro, line 24: I was only half serious.
Few people would know what I refer to. Pair is more intuitive.

julsemaan:
Can you rebase so we merge in devel. Thanks !

PR #253, opened 2014-10-17
Feature/firewall sso refactor

lzammit:
Thanks julien.


julsemaan:
the clone function still doesn't work

PR #257, opened 2014-10-21
Dedicated API / apache webservice for AAA

julsemaan:
Adds a dedicated apache service for RADIUS authorization and accouting.

For review + merge


fdurand:
Don´t forget iptables

julsemaan:
It shouldn't be necessary since it's meant to be used only locally by the running radius server.

Should it be made public ?

fdurand:
So why do you listen on the mamangement interface and not just 127.0.0.1 ?

julsemaan:
Pretty good point ! I'll remove the listening interface on the management IP and leave only 127.0.0.1

fdurand:
I prefer to be able to have a remote radius server.

julsemaan:
Done, I added it to iptables so the service stays public

julsemaan:
Merge me ?

jrouzierinverse:
The following files were missed.

lib/pf/radius/rpc.pm
raddb/packetfence.pm
raddb/packetfence-soh.pm

julsemaan:
Everything should have been fixed. _Ready_ to merge

jrouzierinverse:
This looks good from a code point of view

julsemaan:
Everything should be fixed.

Ready to merge if no more comments/reviews

PR #260, opened 2014-10-27
More precisly define which dhcp message types we are listening for

jrouzierinverse:


NEWS.asciidoc
```diff
@@ -21,6 +21,7 @@ New Features
 Enhancements
 ++++++++++++
 * Added httpd.aaa service as a new API service for the exclusive use of RADIUS
+* More precisly define which dhcp message types we are listening for
```
extrafu: typo -> precisely

html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
```diff
@@ -43,6 +43,8 @@ sub field_list {
             label => $doc_section_name,
           };
         my $type = $doc_section->{type} || "text";
+        #skip if hidden
```
extrafu, line 46: Remove if unused.

jrouzierinverse, line 46: It is used

lib/pf/util/dhcp.pm
```diff
@@ -289,6 +289,25 @@ sub _decode_dhcp_option82 {
...
+=cut
+
+sub make_pcap_filter {
+    my (@types) = @_;
+    #listen to all if no types are provided
+    return "udp and (port 67 or port 68)" unless @types;
+    for my $type (@types) {
+       die "Unknown message type $type" unless exists $MESSAGE_TYPE{$type} && defined $MESSAGE_TYPE{$type};
+    }
+    my $type_filter = join(" or ",map { sprintf("(udp[250:1] = 0x%x)",$MESSAGE_TYPE{$_}) } @types);
```
fdurand, line 306: If i have only one type, does the filter will look like : ( (port 67 or port 68) and ( or udp[250:1] = 0x5) ) ?

jrouzierinverse, line 306: Yes
perl -Ilib -Mpf::util::dhcp -e 'print make_pcap_filter(qw(DHCPDISCOVER)),"\n"'
(port 67 or port 68) and ( (udp[250:1] = 0x5) )

fdurand, line 306: Ok so ready to merge

lib/pf/config.pm
```diff
@@ -653,6 +653,7 @@ sub readPfConfigFiles {
                         )
                     ];
                 }
+                $Config{network}{dhcp_filter_by_message_types} = [split(/\s*,\s*/,$Config{network}{dhcp_filter_by_message_types} || '')],
```
jrouzierinverse, line 656: If we want to further fine tune the filtering.
We could remove the following conditionally 
- The offer message if rogue detection is disable.
- The request message if rogue detection is disable and there is no firewallsso configured and not inline mode

lib/pf/util/dhcp.pm
jrouzierinverse, line 306: Then push button

lib/pf/config.pm
fdurand, line 656: For setup with high dhcp traffic it can be very usefull

PR #261, opened 2014-10-27
OPSWAT compliance

julsemaan:
- Refactors the OPSWAT template
- Creates the basics for adding compliance configuration to the provisioners in the admin GUI
- Adds compliance configuration in the admin interface for OPSWAT


html/captive-portal/templates/provisioner/opswat.html
```diff
@@ -1,13 +1,15 @@
...
 [% INCLUDE header.html %]
 
-    <h3>To complete your network activation you need to install OPSWAT GEARS</h3>
-    <p>When you finish installing the application click next to activate your network connection</p>
+    <h3>To complete your network activation you need to install the OPSWAT GEARS client.</h3>
+    <p>When you finish installing the application click 'Continue' to activate your network connection</p>
 
     <p>
-    <a id="primary_download" href="[% provisioner.agent_download_uri %]">Click here to go to the download website</a>
+    <a id="primary_download" target="_blank" href="[% provisioner.agent_download_uri %]">Click here to go to the download website</a>
```
dwlfrth: Is the agent hosted on the PacketFence server or elsewhere ?
In second case, are we handling the passthrough to reach it ?

julsemaan: It's external.
The necessary passthroughs are documented in 4.5

lib/pf/provisioner/opswat.pm
```diff
@@ -203,10 +212,22 @@ sub validate_mac_in_opswat {
...
-        return $self->check_active($mac, $json_response);
+        return $json_response;
     } 
 }
 
+sub validate_mac_in_opswat {
+    my ($self, $mac) = @_;
+    my $logger = Log::Log4perl::get_logger( ref($self) );
+    my $info = $self->get_device_info($mac);
+    if($info != -1){
```
jrouzierinverse: Why are you checking for negative one?
get_device_info should return either a undef or the hash representation of the device,

lib/pf/provisioner/opswat.pm
```diff
@@ -233,6 +254,10 @@ sub check_active {
 sub authorize {
```
jrouzierinverse, line 257: Does it make sense to do multiple calls to the opswat service to get the device info for both verify_compliance and  validate_mac_in_opswat.
I would recommend to call get_device_info then pass the device info to verify_compliance and  validate_mac_in_opswat.
So to avoid the multiple calls to the opswat service.
Unless there is a compelling reason to do so call each time.

lib/pf/provisioner/opswat.pm
julsemaan: Because I made -1 be an error of communication so we don't trigger violations on bad communications with the API

lib/pf/provisioner/opswat.pm
julsemaan, line 257: I agree with you but that call is meant to be temporary.

We should create a separate flow for the compliance and the authorization which means the get_device_info will need to be called twice

lib/pf/provisioner/opswat.pm
extrafu: We should use constants for this, if possible.

julsemaan: Yeah I'll create a constant for the provisioners for failed communication. This way its clearer in the code

julsemaan:
Added documentation to be reviewed.

Then should be merged

cgx:
All strings of default templates should be localized (wrap them with `i18n()`).

docs/PacketFence_OPSWAT_Quick_Install_Guide.asciidoc
```diff
@@ -104,3 +104,44 @@ Connect a device to your test network and register like you normally would.
 At the end of the registration process you will be presented a page asking you to install the OPSWAT GEARS client on your device.
 After you install the client click continue. If your access is enabled than this means the connectivity between PacketFence and OPSWAT GEARS is good.
 
+Compliance enforcement
+----------------------
+
+Since version 4.5.1, PacketFence polls the OPSWAT GEARS API in order to trigger violations on non compliant devices.
```
extrafu: Avoid mentioning version numbers in the doc. v4.5 didn't have this capability, so obviously the doc wasn't there.

julsemaan: So should I completely remove it or reformulate without the version ?

extrafu: Just strip "Since version 4.5.1,"

cgx:
_Non compliance_ should be written as one word (_noncompliance_).

julsemaan:
Just added more refactoring to this PR so another review should be done before we merge

julsemaan:
Closing the PR to reopen since the purpose of the PR mostly changed to being provisioning version 2

PR #263, opened 2014-10-28
Firewall should allow 443 on management interface when pre-registration ...

dwlfrth:
## Description

Punch a hole in the iptables firewall of PacketFence to allow the httpd.portal server to be reach on the management ip, port 443, when pre-registration is enabled.
## Impacts

Shouldn't have any...

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -26,6 +26,7 @@ Enhancements
 Bug Fixes
 +++++++++
 * Fix old MAC addresses being left on port-security enabled ports in a RADIUS + port-security environment
+* Fix firewall rule that allows httpd.portal to be reach on management IP when pre-registration enabled
 
 Version 4.5.0 released on 2014-10-22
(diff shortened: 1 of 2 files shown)


julsemaan:
I honnestly think it should always be enabled.
Since that interface can now be used for the web auth portal, I don't see why we should still refuse access to it...

dwlfrth:
You are kind of right but for the moment it is not.
Maybe we can simply merge that one and work on getting iptables rules fixed with another branch

NEWS.asciidoc
```diff
@@ -26,6 +26,7 @@ Enhancements
 Bug Fixes
 +++++++++
 * Fix old MAC addresses being left on port-security enabled ports in a RADIUS + port-security environment
+* Fix firewall rule that allows httpd.portal to be reach on management IP when pre-registration enabled
```
extrafu: to be reached from the mangement IP

julsemaan:
Agreed,

Other than that, it looks good to me

dwlfrth:
Then you know what to do ;)

PR #264, opened 2014-10-29
Refactored check of configured FirewallSSO to pf::firewallsso

jrouzierinverse:


lib/pf/firewallsso.pm
```diff
@@ -2,13 +2,13 @@ package pf::firewallsso;
 
 =head1 NAME
 
-pf::web::firewallsso
+pf::firewallsso
 
 =cut
 
 =head1 DESCRIPTION
```
dwlfrth, line 9: Can we document the module rather than repeating it's name ?

sbin/pfdhcplistener
```diff
@@ -359,8 +359,8 @@ sub parse_dhcp_request {
         );
         update_iplog( $dhcp->{'chaddr'}, $client_ip, $lease_length );
         # if there is at least a firewall sso defined in the config && we have the lease length
```
dwlfrth: Comment is no longer accurate

lib/pf/firewallsso.pm
```diff
@@ -42,6 +43,7 @@ Send the firewall sso update request to the webapi.
 
 sub do_sso {
```
dwlfrth, line 46: Since we are in there, can we add some logging to the code ?

jrouzierinverse:
Please review

fdurand:
Except Derek´s comment it´s ok for me

jrouzierinverse:
I updated the pod doc as Derek suggested please let me know if there is anything else.

PR #265, opened 2014-10-29
Will only listen for dhcp ACK for non management interfaces.

jrouzierinverse:
Will listen for dhcp OFFER REQUEST DECLINE ACK RELEASE for management interfaces

### DIFF
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -121,8 +121,15 @@ if ( isenabled( $Config{'network'}{'dhcpdetector'} ) ) {
         $interface_ip = $interfaceData->{'ip'};
         my $interface_type = $interfaceData->{'type'};
         $interface_vlan = get_vlan_from_int($interface) || $NO_VLAN;
+        # Listening only for DHCPACK message type on every configured interface(s)
+        # that are not a 'management' one 
+        # (We can get the information we want from the DHCPACK)
         my @filter_types = qw(DHCPACK);
         if( any { $_ eq "management" } split /\s*,\s*/, $interface_type) {
+            # We need to listen for all DHCP message type on the 'management' interface(s) 
+            # since they are used for remote networks and in those cases, if PacketFence 
+            # is not the DHCP server on these networks, we only get the information we want
+            # from DHCPREQUEST (ACK from another DHCP server in layer 3 will be unicast so PacketFence won't see it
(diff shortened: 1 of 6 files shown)


sbin/pfdhcplistener
```diff
@@ -116,12 +116,18 @@ if ( isenabled( $Config{'network'}{'dhcpdetector'} ) ) {
                 $accessControl = new pf::inline::custom();
             }
         }
+        my $interfaceData = $Config{"interface $interface"};
 
-        $interface_ip = $Config{"interface $interface"}{'ip'};
+        $interface_ip = $interfaceData->{'ip'};
+        my $interface_type = $interfaceData->{'type'};
         $interface_vlan = get_vlan_from_int($interface) || $NO_VLAN;
+        my @filter_types = qw(DHCPACK);
```
dwlfrth, line 127: Add comments!
ie:
# Listening only for DHCPACK message type on every configured interface(s) that are not a 'management' one (We can get the information we want from the DHCPACK)
# We need to listen for all DHCP message type on the 'management' interface(s) since they are used for remote networks and in those cases, if PacketFence is not the DHCP server on these networks, we only get the information we want from DHCPREQUEST (ACK from another DHCP server in layer 3 will be unicast so PacketFence won't see it)

sbin/pfdhcplistener
```diff
@@ -116,12 +116,25 @@ if ( isenabled( $Config{'network'}{'dhcpdetector'} ) ) {
...
+        my $interfaceData = $Config{"interface $interface"};
 
-        $interface_ip = $Config{"interface $interface"}{'ip'};
+        $interface_ip = $interfaceData->{'ip'};
+        my $interface_type = $interfaceData->{'type'};
         $interface_vlan = get_vlan_from_int($interface) || $NO_VLAN;
+        # Listening only for DHCPACK message type on every configured interface(s)
+        # that are not a 'management' one 
+        # (We can get the information we want from the DHCPACK)
+        my @filter_types = qw(DHCPACK);
```
fdurand, line 127: We probably need to add DHCPREQUEST too for the dhcp fingerprints.

fdurand, line 127: To be more precise, if the interface is not management then we have to capture DHCPREQUEST to handle fingerprints but we don´t have to update iplog.

jrouzierinverse:
We need to schedule some time for this feature

dwlfrth:
We need to get this merged.
Status ?

PR #266, opened 2014-10-29
Use vlan filter in getNodeInfoForAutoReg and create/update person even i...

fdurand:
## Description

Added vlan filter in getNodeInfoForAutoReg (set a role to be able to test it in getNormalVlan)
Create or update a person even if the deviced has been registered (Like we did machine auth and after user auth but the pid didn´t changed) 
## Impacts

Autoregistration workflow.
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -452,7 +452,7 @@ If action is set, it will return the value of the action immediately.
 =cut
 
 sub match {
-    my ($source_id, $params, $action) = @_;
+    my ($source_id, $params, $action,$source_ref) = @_;
     my ($actions, @sources);
 
     $logger->debug("Match called with parameters ".join(", ", map { "$_ => $params->{$_}" } keys %$params));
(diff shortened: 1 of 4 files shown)


lib/pf/authentication.pm
```diff
@@ -452,7 +452,7 @@ If action is set, it will return the value of the action immediately.
 =cut
 
 sub match {
-    my ($source_id, $params, $action) = @_;
+    my ($source_id, $params, $action,$source_ref) = @_;
```
jrouzierinverse, line 456: Document the new parameter pf::authentication::match

lib/pf/vlan.pm
```diff
@@ -416,10 +417,25 @@ sub getNormalVlan {
                     'unregdate' => $value,
                     'category' => $role,
                     'autoreg' => 'yes',
+                    'pid' => $user_name,
                 );
                 if (defined $role) {
                     %info = (%info, (category => $role));
                 }
+                require pf::person;
+                require pf::lookup::person;
```
jrouzierinverse: Load these modules in the top of file
use pf::person;
use pf::lookup::person;

fdurand:
Can it be merge ?

julsemaan:
Looks good to me

PR #270, opened 2014-10-30
Feature/captiveportal rework

dwlfrth:
## Description
- Cleaning up dispatcher
- Cleaning up httpd.portal config
- Better workflow in dispatcher
- Better handling of mobile captive-portal detection mecanisms
- Better handling of destination URL
- New configuration parameter for captive-portal detection mecanism URLs
- New configuration parameter for 'portal' interface type
- New configuration parameter for mod_status listening port on httpd.portal
## Impacts
- Captive portal flow...
  -- Redirection when trying to reach an Internet resource while in registration
  -- Redirection when trying to reach an Internet resource while in registration from a mobile device with captive-portal detection mecanism
  -- Redirection when trying to reach an Internet resource while in registration using Google Chrome browser
- mod_status for httpd.portal
- httpd.portal vhost configuration depending on interface type
- More to come...


dwlfrth:
Can we review it a way to possibly integrate it

lib/pf/web/dispatcher.pm
```diff
@@ -82,31 +76,37 @@ sub handler {
         $r->pnotes(last_uri => $last_uri);
         return Apache2::Const::DECLINED;
     }
-    if ($r->uri =~ /\/apache_status/) {
-        $r->handler('server-status');
+
+    # httpd.portal mod_status
+    # We do not want to continue through the dispatcher if we ask for httpd diagnostic
+    if ( $r->uri =~ /\/server-status/ ) {
```
fdurand: I think this is useless since you define a new virtual host for mod_status

dwlfrth: Yes and no... Without this, the request will go all the way through the dispatcher and then exit to continue the Apache configuration...

lib/pf/web/dispatcher.pm
```diff
@@ -35,46 +36,39 @@ use pf::Portal::Session;
...
-    # be careful w/ performance here
-    # Warning: we might want to revisit the /o (compile Once) if we ever want
-    #          to reload Apache dynamically. pf::web::constants will need some
-    #          rework also
+    # We don't want to continue in the dispatcher if the requested URI is supposed to reach the captive-portal (Catalyst)
+    # - Captive-portal itself
+    # - Violation pages
+    # - Portal profile filters are handled by Catalyst (Will be part of captive-portal Catalyst app rework)
+    # See L<pf::web::constants::CAPTIVE_PORTAL_RESOURCES>
+    if ( $r->uri =~ /$WEB::CAPTIVE_PORTAL_RESOURCES/o ) {
```
fdurand, line 68: Can we add more test with $r->hostname , because if i go to http://acme.com/captive-portal i will see the portal.

lib/pf/web/dispatcher.pm
fdurand: Maybe you can set PerlTransHandler => '' in the mod_status vhost

fdurand:
Just a remark about portal interface, don´t we have to allow tcp/udp 53 and tcp 443 80 in this interface ?

lib/pf/web/dispatcher.pm
dwlfrth: Actually you are right... I forgot it was now part of a separate vhost..
Will remove

lib/pf/web/dispatcher.pm
dwlfrth, line 68: I agree but we need to keep support for external captive portal in which cases, the hostname may differ...
And, the redirect should do the job to ensure the hostname...

dwlfrth, line 68: Figured a way to handle it.
Will commit

julsemaan:
Need status on this. Is this finished work and ready to review ?

dwlfrth:
Forget about this one

PR #271, opened 2014-10-31
Restricts the roles, access levels and access durations for an admin users based on their admin role/access level

jrouzierinverse:
## Description

Restricts the roles, access levels and access durations for an admin users based on their admin role/access level
## Impacts

These affect user creation in the Admin GUI.
## Delete branch after merge

NO


dwlfrth:
No template ! Interesting ! ;)

html/pfappserver/lib/pfappserver/Base/Form/Authentication/Action.pm
```diff
@@ -154,16 +156,24 @@ Populate the select field for the roles template action.
...
 
-    my @roles;
+=head2 _get_allowed_options
 
-    # Build a list of existing roles
-    my ($status, $result) = $self->form->ctx->model('Roles')->list();
-    if (is_success($status)) {
-        @roles = map { $_->{name} => $_->{name} } @$result;
-    }
+TODO: documention
```
julsemaan, line 170: that virus is still spreading as i can see
at least correct the typo in documentation !

t/unittest/admin_roles.t
```diff
@@ -0,0 +1,62 @@
...
+use Test::NoWarnings;
+
+use_ok("pf::admin_roles");
+
+ok(admin_can(["User Manager"],'USERS_CREATE'),"User Manager can create a user");
+
+ok(!admin_can(["User Manager"],'NODES_READ'),"User Manager cannot read a node");
+
+ok(admin_can(["ALL"],'USERS_CREATE'),"ALL can create a user");
+
```
julsemaan, line 31: I'd propose you create tests for the new actions you created (set access duration, set role).

I would make sense to have those in here too since they're what led to this feature and PR

jrouzierinverse:
These changes are already in chained auth

PR #277, opened 2014-11-05
Fixed the logrotation.

louismunro:
Have had it up to here with the configuration format to logrotate?
How many times have you seen your disk space vanish because a file was not being rotated properly?


julsemaan:
+1
That will also future proof any new log so I completely agree with it.

cgx:
Go :+1:

PR #278, opened 2014-11-11
Retrieve the mac of an ip from the dhcpd server using omapi

jrouzierinverse:
CLOSED and LOCKED due to migration of the branch from James repo to inverse repo.
Kept for comments
Please refer to PR #421

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -128,6 +128,8 @@ Requires: perl(Time::HiRes)
 Requires: ipset, sudo
 Requires: perl(File::Which), perl(NetAddr::IP)
 Requires: perl(Net::LDAP)
+Requires: perl(Net::IP)
+Requires: perl(Digest::HMAC_MD5)
 # TODO: we should depend on perl modules not perl-libwww-perl package
 # find out what they are and specify them as perl(...::...) instead of perl-libwww-perl
 # LWP::Simple is one of them (required by inlined Net::MAC::Vendor and probably other stuff)
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -977,6 +977,37 @@ description=<<EOT
 Color of the success text for pfcmd
 EOT
 
+[advanced.use_omapi_to_lookup_mac]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Use omapi to query dhcpd for the mac address of an ip
+EOT
+
+[advanced.omapi_key_name]
+type=text
+description=<<EOT
+The omapi key name for signing messages
+EOT
+
+[advanced.omapi_key_base64]
+type=text
+description=<<EOT
+The omapi base64 key for signing messages
+EOT
+
+[advanced.omapi_port]
+type=text
+description=<<EOT
+The omapi port number
+EOT
+
+[advanced.omapi_host]
+type=text
+description=<<EOT
+The omapi host
+EOT
+
 [provisioning.autoconfig]
 type=toggle
 options=enabled|disabled
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -977,32 +977,32 @@ description=<<EOT
 Color of the success text for pfcmd
 EOT
 
-[advanced.use_omapi_to_lookup_mac]
+[omapi.ip2mac_lookup]
 type=toggle
 options=enabled|disabled
 description=<<EOT
 Use omapi to query dhcpd for the mac address of an ip
 EOT
 
-[advanced.omapi_key_name]
+[omapi.key_name]
 type=text
 description=<<EOT
 The omapi key name for signing messages
 EOT
 
-[advanced.omapi_key_base64]
+[omapi.key_base64]
 type=text
 description=<<EOT
 The omapi base64 key for signing messages
 EOT
 
-[advanced.omapi_port]
+[omapi.port]
 type=text
 description=<<EOT
 The omapi port number
 EOT
 
-[advanced.omapi_host]
+[omapi.host]
 type=text
 description=<<EOT
 The omapi host
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -757,31 +757,33 @@ pfcmd_warning_color=yellow
 #
 # Color of the success text for pfcmd
 pfcmd_success_color=green
+
+[omapi]
 #
-# advanced.use_omapi_to_lookup_mac
+# omapi.use_omapi_to_lookup_mac
 #
 # Use omapi to query dhcpd for the mac address of an ip
-use_omapi_to_lookup_mac=disabled
+ip2mac_lookup=enabled
 #
-# advanced.omapi_key_name
+# omapi.omapi_key_name
 #
 # The omapi key name for signing messages
-omapi_key_name=pf_omapi_key
+key_name=pf_omapi_key
 #
-# advanced.omapi_key_base64
+# omapi.omapi_key_base64
 #
 # The omapi base64 key for signing messages
-omapi_key_base64=
+key_base64=
 #
-# advanced.omapi_port
+# omapi.omapi_port
 #
 # Use omapi port number
-omapi_port=7911
+port=7911
 #
-# advanced.omapi_host
+# omapi.omapi_host
 #
 # The omapi host
-omapi_host=localhost
+host=localhost
 
 [provisioning]
 #
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -766,7 +766,7 @@ use_omapi_to_lookup_mac=disabled
 # advanced.omapi_key_name
 #
 # The omapi key name for signing messages
-omapi_key_name=
+omapi_key_name=pf_omapi_key
 #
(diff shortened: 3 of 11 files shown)


lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
...
+=head2 _build_sock
+
+Creates a socket
+
+=cut
+
+sub _build_sock {
+    my ($self) = @_;
+    my $sock = IO::Socket::INET->new(PeerAddr => $self->host, PeerPort => $self->port, Proto => 'tcp') || die "Can't bind : $@\n";
+    return $sock;
```
extrafu, line 365: When do we actually close the socket?

rouzier, line 365: When the object gets destroyed

fdurand:
We probably have to add documentation on how to enable omapi in iscdhcp.

jrouzierinverse:
Yes I will update the PF Guide

lib/pf/iplog.pm
```diff
@@ -344,6 +348,31 @@ sub ip2macinarp {
...
+    my ($host, $port, $keyname, $key_base64) =
+      @Config{qw( omapi_host omapi_port omapi_key_name omapi_key_base64 )};
+    my $omapi = pf::OMAPI->new(
+        omapi_host       => $host,
+        omapi_port       => $port,
+        omapi_key_name   => $keyname,
+        omapi_key_base64 => $key_base64
+    );
+    eval {
+        my $data = $omapi->lookup({type => 'lease'}, {'ip-address' => "172.32.100.185"});
```
fdurand: 172.32.100.185

conf/documentation.conf
```diff
@@ -955,6 +955,37 @@ description=<<EOT
...
+[advanced.omapi_key_name]
+type=text
+description=<<EOT
+The omapi key name for signing messages
+EOT
+
+[advanced.omapi_key_base64]
+type=text
+description=<<EOT
+The omapi base64 key for signing messages
```
fdurand, line 996: To generate a key:
dd if=/dev/urandom bs=16 count=1 2>/dev/null | openssl enc -e -base64

fdurand:
Work well on a big setup, i think we are able to merge it.

rouzier:
We still needs to document the feature

addons/packages/packetfence.spec
```diff
@@ -128,6 +128,8 @@ Requires: perl(Time::HiRes)
 Requires: ipset, sudo
 Requires: perl(File::Which), perl(NetAddr::IP)
 Requires: perl(Net::LDAP)
+Requires: perl(Net::IP)
+Requires: perl(Digest::HMAC_MD5)
```
fdurand, line 132: It miss the same dep for debian/ubuntu

jrouzierinverse:
Please review me

addons/packages/packetfence.spec
dwlfrth, line 132: Is this completed ? I mean, the debian/ubuntu part...

conf/documentation.conf
dwlfrth, line 996: Should be added to the .spec to be triggered on install and added a a default value

conf/pf.conf.defaults
```diff
@@ -757,6 +757,31 @@ pfcmd_warning_color=yellow
...
+#
+# advanced.use_omapi_to_lookup_mac
+#
+# Use omapi to query dhcpd for the mac address of an ip
+use_omapi_to_lookup_mac=disabled
+#
+# advanced.omapi_key_name
+#
+# The omapi key name for signing messages
+omapi_key_name=
```
dwlfrth: What is my default value. Can I work without a default one ? If not, how to get it ?
So many questions

addons/packages/packetfence.spec
fdurand, line 132: check debian/control, it's done

conf/documentation.conf
fdurand, line 996: Yes good idea

conf/pf.conf.defaults
jrouzierinverse: It will not work without a value set.
I will just create one.

conf/pf.conf.defaults
```diff
@@ -757,6 +757,31 @@ pfcmd_warning_color=yellow
 #
 # Color of the success text for pfcmd
 pfcmd_success_color=green
+#
+# advanced.use_omapi_to_lookup_mac
+#
+# Use omapi to query dhcpd for the mac address of an ip
+use_omapi_to_lookup_mac=disabled
```
dwlfrth: With the rework of iplog, we will enable this by default

dwlfrth:
With this amount of configuration parameters (5) shouldn't we add a new configuration section in pf.conf called omapi ?

jrouzierinverse:
Sure why not

conf/pf.conf.defaults
jrouzierinverse: We should be then document when to disable it.
For example a user use their own dhcp server that does not have omapi support or not configured.

dwlfrth: In this case it will simply silently fail without any impact I think... no ?

jrouzierinverse: In theory yes, but we should still document when to disable it.

dwlfrth: I agree

lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
+package pf::OMAPI;
+=head1 NAME
+
+pf::OMAPI add documentation
```
dwlfrth: Welcome back !

lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
...
+
+=head1 SYNOPSIS
+
+Quick summary of what the module does.
+
+Perhaps a little code snippet.
+
+    use pf::OMAPI;
+
+    my $omapi = pf::OMAPI->new( keyname => 'defomapi',key_base64 => 'xJviCHiQKcDu6hk7+Ffa3A==', host => 'localhost', port => 7911);
```
dwlfrth: Let's use our default key_name here

lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
...
+
+has connected => (is => 'rw' , default => sub { 0 } );
+
+=head2 keyname
+
+The name of the key 
+
+=cut
+
+has keyname => (is => 'rw');
```
dwlfrth: Configuration parameter is key_name, make it consistent ?

lib/pf/services/manager/dhcpd.pm
```diff
@@ -86,6 +87,31 @@ EOT
...
 
+
+=head2 omapi_section
+
+Generate the omapi section if it is defined
+
+=cut
+
+sub omapi_section {
+    return '' if isdisabled($Config{advanced}{use_omapi_to_lookup_mac});
```
dwlfrth: We could return "# OMAPI is not enabled on this server".

lib/pf/services/manager/dhcpd.pm
```diff
@@ -86,6 +87,31 @@ EOT
...
+    my $section = "omapi-port $Config{advanced}{omapi_port};\n";
+    my $keyname = $Config{advanced}{omapi_key_name};
```
dwlfrth: I looooooove consistency... s/keyname/key_name :)

lib/pf/services/manager/dhcpd.pm
rouzier: How would that be useful?
It just add a comment to a generated config file.
We already know it is not enabled by it being disabled in pf.conf

dwlfrth: It will help diagnose if something is not right.
We will not start to argue on the purpose of comments, I think you already know my position ;)

lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
...
+
+=head2 key
+
+The key base64 representation of the key
+
+=cut
+
+has key_base64 => (is => 'rw');
+
+our $OPEN    = 1;
```
dwlfrth, line 155: Describe us! Not each one but what are we.

lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
...
+our $REFRESH = 2;
+our $UPDATE  = 3;
+our $NOTIFY  = 4;
+our $ERROR   = 5;
+our $DELETE  = 6;
+
+
+our %FORMATLIST = (
```
dwlfrth, line 166: Same here

lib/pf/OMAPI.pm
```diff
@@ -0,0 +1,538 @@
...
+=head2 connect
+
+Will connect and authenticate to the omapi server
+
+=cut
+
+sub connect {
+    my ($self) = @_;
+    return 1 if $self->connected;
+    my ($recieved_startup_message,$len);
```
dwlfrth: recieved ?

PR #280, opened 2014-11-13
Refactor the vlan filter code to reduce code duplication

jrouzierinverse:
## Description

Refactor the vlan filter code to reduce code duplication
## Impacts

All places that use the vlan filters
## Delete branch after merge

NO


fdurand:
Just miss an entry in the NEWS file.

PR #283, opened 2014-11-16
Batch iplog and locationlog deletion

jrouzierinverse:
## Description

Batch the amount of rows that get are deleted from the iplog and location log to reduce the table locks.
## Impacts

This reduces the time spent deleting records in pfmon.
However this could lead to have more stale records.

An additional cronjob at night could also be used to cleanup tables.
There are two new scripts

addons/iplog-cleanup.pl
addons/locationlog-cleanup.pl

The options are the same for both

```
iplog-cleanup.pl [--expire SECONDS] [--batch BATCH] [--timeout SECONDS]

 Options:
  --expire   The expire time of the iplog in seconds     - default expire.iplog
  --batch    The number of iplog entries to batch delete - default maintenance.iplog_cleanup_batch
  --timeout  How long this job is allowed to run         - default maintenance.iplog_cleanup_timeout
  --help     Shows this help
```

This will run the job for an hour instead 10 seconds
addons/iplog-cleanup.pl --timeout 3600
## Delete branch after merge

NO


db/upgrade-4.5.0-4.x.y.sql
```diff
@@ -0,0 +1,17 @@
+--
+-- Alter to improve iplog clean up
+--
+
+ALTER TABLE iplog ADD INDEX iplog_end_time (end_time), DROP INDEX mac;
```
julsemaan: Why remove the index on mac ?

jrouzierinverse: Because we already have it in another index.
KEY `mac_view_open` (`mac`, `end_time`)

Since it is the first in the index it make no sense to index it twice.

http://dev.mysql.com/doc/refman/5.0/en/multiple-column-indexes.html

NEWS.asciidoc
```diff
@@ -26,6 +26,8 @@ Enhancements
 * More precisely define which dhcp message types we are listening for
 * Removed dead code referring to 'external' interface type which was, anyway, no longer supported
 * Added vlan filter in getNodeInfoForAutoReg and update/create person even if the device has been autoreg
+* Reduce deadlocks cause by the cleaning of the iplog table
```
extrafu: caused x 2

conf/documentation.conf
```diff
@@ -1097,6 +1097,18 @@ description=<<EOT
...
+[maintenance.iplog_cleanup_batch]
+type=numeric
+description=<<EOT
+How many iplog entries does pfmon batch delete
+EOT
+
+[maintenance.iplog_cleanup_timeout]
+type=time
+description=<<EOT
+How long is the iplog cleaiplog clean is allowed to run
```
extrafu: cleaiplog?

conf/documentation.conf
```diff
@@ -1109,6 +1121,18 @@ description=<<EOT
 Interval at which Packetfence runs locationlog cleanup
 EOT
 
+[maintenance.locationlog_cleanup_batch]
+type=numeric
+description=<<EOT
+How many locationlog entries does pfmon batch deletes
```
extrafu: delete?

jrouzierinverse:
Please review

fdurand:
It miss maintenance.iplog_cleanup_timeout ... in po file (for the admin gui).
It miss sql in pf-schema-4.6.0.sql file.

jrouzierinverse:
Done

PR #284, opened 2014-11-18
Cache the result of a rule for a LDAP/AD Source to reduce server load.

jrouzierinverse:
## Description

Cache the result of a rule for a LDAP/AD Source to reduce server load.
## Impacts

Since the match is cached time based rules will not work properly.
Rules changes can take up to the expire time of the cache to take affect for all users/lookup
## Delete branch after merge

NO

### DIFF
--- a/t/data/authentication.conf
+++ b/t/data/authentication.conf
@@ -47,3 +47,56 @@ description=
 match=all
 action0=set_role=guest
 action1=set_access_duration=1D
+
+[LDAP]
+description=pf-test
+password=
+scope=sub
+binddn=CN=test,DC=inverse,DC=ca
+basedn=DC=ldap,DC=inverse,DC=ca
+usernameattribute=user
+connection_timeout=5
+encryption=none
+port=33389
+type=AD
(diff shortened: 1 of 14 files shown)


jrouzierinverse:
Can someone review this?

t/data/authentication.conf
```diff
@@ -47,3 +47,56 @@ description=
...
+
+[LDAP rule Network_Team]
+description=Full Access
+match=any
+action0=set_role=default
+action1=set_access_level=ALL
+action2=set_unreg_date=2022-02-02
+condition0=memberOf,equals,CN=NOC Users,DC=ldap,DC=inverse,DC=ca
+
+[LDAP rule Barb_Iannucci-Web_Designer]
```
louismunro: You sure this should be in here?

jrouzierinverse: No

louismunro:
Tested. Ready to be merged.

jrouzierinverse:
Please Merge

PR #289, opened 2014-11-28
Reset any cached connections before forking

jrouzierinverse:
## Description

To avoid the avoid child processes using different the parents cached connections
## Impacts

Any service that forks multiple child process will be affected by this
pfmon and the any httpd webservice will be affected by this
## Delete branch after merge

NO


jrouzierinverse:
This functionality is in another branch

PR #295, opened 2014-12-05
New 'default' default user

dwlfrth:
## Description

Adding a new 'default' user that will be used to defaultly register node rather than using the 'admin' user
## Impacts

Make sure that doesn't change anything to role computing but that shouldn't.... it is actually the main purpose of the this feature.
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Added a 'default' default user in replacement of the admin one (we keep the admin default but we now defaultly register nodes using the 'default' one)

Bug Fixes
+++++++++
## Delete branch after merge

YES

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Node/Manager.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Node/Manager.pm
@@ -32,7 +32,7 @@ sub unreg :Local :Args(1) {
     my $node = node_view($mac);
     if ($username && $node) {
         $c->log->info("$username attempting to unregister $mac");
-        if (($username ne $default_pid || $username ne $admin_pid ) && $username eq $node->{pid}) {
+        if (($username ne $default_pid && $username ne $admin_pid ) && $username eq $node->{pid}) {
             node_deregister($mac, %$node);
             reevaluate_access($mac, "node_modify");
             $c->response->redirect("/status");
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Node/Manager.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Node/Manager.pm
@@ -32,7 +32,7 @@ sub unreg :Local :Args(1) {
     my $node = node_view($mac);
     if ($username && $node) {
         $c->log->info("$username attempting to unregister $mac");
-        if ($username ne $default_pid && $username eq $node->{pid}) {
+        if (($username ne $default_pid || $username ne $admin_pid ) && $username eq $node->{pid}) {
             node_deregister($mac, %$node);
             reevaluate_access($mac, "node_modify");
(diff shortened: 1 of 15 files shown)


dwlfrth:
Need to update the UPDATE document but first, it need to be templated for release purposes

jrouzierinverse:
In pf::node::is_max_reg_nodes_reached we consider the default_pid a special case with no limit for this user.
Does this behavior still make sense?

dwlfrth:
Review me

html/captive-portal/lib/captiveportal/PacketFence/Controller/Node/Manager.pm
```diff
@@ -31,7 +31,7 @@ sub unreg :Local :Args(1) {
     my $node = node_view($mac);
     if ($username && $node) {
         $c->log->info("$username attempting to unregister $mac");
-        if ($username ne $default_pid && $username eq $node->{pid}) {
+        if (($username ne $default_pid || $username ne $admin_pid ) && $username eq $node->{pid}) {
```
fdurand: Just to simplify ($username ne $default_pid || $username ne $admin_pid ) = 1 and i suppose that it´s wrong, it should probably be ($username ne $default_pid && $username ne $admin_pid ) instead

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -64,7 +64,7 @@ sub setupCurrentNodeInfo : Private {
     my ( $self, $c ) = @_;
     my $portalSession = $c->portalSession;
     my $node_info     = node_view( $portalSession->clientMac() );
-    if( $node_info && $node_info->{pid} ne $default_pid ) {
+    if( $node_info && ($node_info->{pid} ne $default_pid || $node_info->{pid} ne $admin_pid ) ) {
```
fdurand: Same here &&

html/captive-portal/lib/captiveportal/PacketFence/Controller/Node/Manager.pm
dwlfrth: I am actually testing the pull-request review! You pass the test ! ;)
Let me fix that for you!

dwlfrth:
Review me!

fdurand:
Look good to me.
 you will probably have to rebase on v5 branch

dwlfrth:
Based of 'devel'.
Don't think it should be merge in v5... it is kind of a minor feature... Let's talk about it tomorrow

dwlfrth:
Review me now that I'm rebased

julsemaan:
Shouldn't the existing nodes with the admin be migrated to the default user or that too risky ?

dwlfrth:
Added an SQL query to update existing unregistered nodes to re-assign them to default

PR #298, opened 2015-01-07
PacketFence FreeRADIUS should not deal with NAS login requests

dwlfrth:
## Description

In some cases, equipment vendors does not allow the configuration of multiple radius group to be used either for login request or NAS-login (login into a switch / console)
If you have MAB or 802.1x configured on your switch, you have the PacketFence IP as RADIUS server.
If you also use RADIUS as backend for authenticating console login, the request may end to the PacketFence FreeRADIUS and will be accepted, which means that any user / password will work.
## Impacts

No real impacts... unless some vendors are using the "NAS-Prompt-User" Service-Type for other purposes...
## Delete branch after merge

YES

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -47,6 +47,7 @@ Enhancements
 * Create a single ldap connection when matching against multiple rules
 * Reduce the numbers of entries in iplog table (update end_time instead of closing and inserting a new line)
 * Now matching on language and not only language/country combination for violation templates (See UPGRADE guide)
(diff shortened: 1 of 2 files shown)


julsemaan:
Looks good to me.

Would it be worth putting this in the maintenance also since it's a security issue ?

dwlfrth:
Need rework ?

dwlfrth:
Don't think it needs to be in maintenance branch :)

julsemaan:
Code looks good so rebase and merge. 

Let's leave it out of the maintenance.

julsemaan:
Rebase me ?

PR #299, opened 2015-01-09
Multi-domain binding + easing the binding

julsemaan:
## Description

This will allow PacketFence to be part of multiple domains. It binds with the new realms in PacketFence & freeradius to perform the ntlm_auth on multiple domains.
This also adds a manager script (in addons/AD) to manage the bind to the domain. This part is far from perfect and deserves discussion. Also, it's made for CentOS.
The multi domain configuration is generated outside of PacketFence and it may be worth managing the processes inside of PF. That would require we create a proper way to manage multi processes in the PacketFence service. For now it uses the classic init.d and rc.d directories to manage the winbind processes.
## Impacts

Adds a layer over the ntlm_auth (chroot + virtual network/ip namespace) which may reserve some surprises. Would deserve to be load tested.
Aims to deprecate the current way to bind to a domain (i.e manually and without chroot) although it's still 100% compatible with the current way of doing it. Configuring new domains may give unexpected results though.
Adds the configuration of the domains in PacketFence (domain.conf) and autogeneration of the samba and krb configuration for the PacketFence configuration
## Delete branch after merge

yes


addons/AD/winbind.init.tt
```diff
@@ -0,0 +1,124 @@
+#!/bin/sh
```
fdurand: What about debian/ubuntu ?

html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
```diff
@@ -0,0 +1,116 @@
...
+   messages => { required => 'Please specify the Active Directory server' },
+  );
+
+has_field 'bind_pass' =>
+  (
+   type => 'Password',
+   label => 'Password',
+   required => 1,
+   password => 0,
+   messages => { required => 'Please specify the password to bind to the Active directory' },
```
fdurand: Active Directory

html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
```diff
@@ -0,0 +1,116 @@
...
+  (
+   type => 'Text',
+   label => 'Username',
+   required => 1,
+   messages => { required => 'Please specify the user to bind to the Active Directory server' },
+  );
+
+has_field 'dns_server' =>
+  (
+   type => 'Text',
```
fdurand: Should be type => IPAddress or use another defiend type (https://github.com/inverse-inc/packetfence/blob/814249caa7d1ed12281af6868f06f6b0b6d4bcbe/html/pfappserver/lib/pfappserver/Form/Field/IPAddresses.pm)

addons/AD/winbind.init.tt
julsemaan: I think this file is now useless anyway so let me either fix it monday or just delete it

html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
julsemaan: You're right

julsemaan:
Is this code ok ?

Still a package missing but it would be good if the code was ready to be merged when the iproute package is made

lzammit:
This looks good to me !

conf/httpd.conf.d/httpd.admin
```diff
@@ -100,8 +100,8 @@ if (defined($management_network->{'Tip'}) && $management_network->{'Tip'} ne '')
     $vhost = "0.0.0.0:".$PfConfig->{'ports'}{'admin'};
 }
 
-$User = "pf";
-$Group = "pf";
+$User = "pf_admin";
+$Group = "pf_admin";
```
jrouzierinverse: Why are you changing this?

julsemaan: Because we need a user with all admin rights in the admin interface. The amount of commands we need sudo rights for is increasing a lot

rouzier: This will cause problem elsewhere.
We will not be able to modify the conf/*.conf files from the httpd.admin.
Since httpd.admin is not running as pf:pf anymore.

I think it is best to keep it a single user it is simpler.

However if you really want to go this route.
Do the following.
- Fix packaging there no where this pf_admin user is created
- Move to a separate sudoers file for each pf system user
- Make sure that conf/*.conf is read and writable from the httpd.admin
- Add the appropriate code to pfcmd fixpermissions
- Create the proper group permissions to segment who can do what

lib/pf/config.pm
```diff
@@ -109,6 +109,8 @@ our (
     %ConfigRealm,
 #provisioning.conf
     %ConfigProvisioning,
+#domain.conf
+    %ConfigDomain, $cached_domain,
```
rouzier: cached_domain should be removed

lib/pf/config.pm
```diff
@@ -155,6 +157,7 @@ BEGIN {
         %Doc_Config
         %ConfigRealm
         %ConfigProvisioning
+        %ConfigDomain $cached_domain
```
rouzier: cached_domain should be removed

lib/pf/file_paths.pm
```diff
@@ -112,6 +114,7 @@ $log_conf_dir  = catdir( $conf_dir,"log.conf.d" );
 $generated_conf_dir   = catdir( $var_dir,"conf");
 $tt_compile_cache_dir = catdir( $var_dir,"tt_compile_cache");
 $pfconfig_cache_dir = catdir( $var_dir,"cache/pfconfig");
+$domains_chroot_dir = catdir( "/chroots");
```
rouzier, line 121: Wouldn't
catdir( $var_dir,"chroots");
Be better?
Or are you afraid @louismunro may delete everything under var

louismunro, line 121: CooCooKatchoo, I am the Chaos Monkey.

lib/pfconfig/manager.pm
```diff
@@ -131,21 +131,7 @@ sub touch_cache {
...
-        close($fh);
-    }
-    if ( -e $filename ) {
-        sysopen( my $fh, $filename, O_RDWR | O_CREAT );
-        POSIX::2008::futimens( fileno $fh );
-        close($fh);
-    }
-    my ( undef, undef, $uid, $gid ) = getpwnam('pf');
-    chown( $uid, $gid, $filename );
+    `touch $filename`;
```
rouzier: This is an old commit you should get rid of touching the file use the touch binary

addons/AD/winbind.init.tt
rouzier: Have you deleted this or fix it?

addons/create_chroot.sh
```diff
@@ -0,0 +1,34 @@
...
+mount -o bind /proc           $BASE/$NS/proc/
+mount -o bind /etc            $BASE/$NS/etc
+mount -o bind /lib            $BASE/$NS/lib
+mount -o bind /lib64          $BASE/$NS/lib64
+mount -o bind /bin            $BASE/$NS/bin
+mount -o bind /usr            $BASE/$NS/usr
+mount -o bind /sbin           $BASE/$NS/sbin
+mount -o bind /sys            $BASE/$NS/sys
+mount -o bind /var/lib/samba/ $BASE/$NS/var/lib/samba/
+mount -o bind /dev/           $BASE/$NS/dev
```
rouzier: would a for loop be simpler?

lib/pf/file_paths.pm
julsemaan, line 121: No kidding, this is exactly why I'm leaving it there. Having it part of the pf directory means that when doing a full backup of that directory, we'd be backing up all the chroots (that have mount points to a lot of places in the filesystem)

julsemaan:
Fixes where made, please re-review

html/pfappserver/root/static/admin/config/items.js
```diff
@@ -45,6 +45,7 @@ Items.prototype.post = function(options) {
  * The ItemView class defines the DOM operations from the Web interface.
  */
 var ItemView = function(options) {
+    var that = this;
```
rouzier, line 48: This is not used

julsemaan, line 48: You're right but I usually like to have it always declared so it can always be used

julsemaan:
Please re-review !

extrafu:
No review here but I mentioned this to Louis, we should have a small script that migrates an existing  <= 5.0 configuration to a "multi-domain" one. I suggest the script to be called ad-domain-zammitator.pl

rouzier:
I suggest dinde-avec-zammit.pl

julsemaan:
I would suggest zammitbug-creator.pl but for some reason I think that it would have a tendency to reboot the server it's on

julsemaan:
Closed, refer to #523

PR #300, opened 2015-01-09
Profile filter all or any

jrouzierinverse:
## Description

Allow a profile to match all or any of its filters instead of just any
## Impacts

Changes how profile filters are matched
## Delete branch after merge

NO


extrafu:
Make sure this goes in v4.6.

lib/pf/config.pm
```diff
@@ -687,10 +688,16 @@ sub readProfileConfigFile {
...
-                    if ($profile_id ne 'default') {
-                        #Adding filters in profile order
+                    my $filters = $profile->{'filter'};
+                    if($profile_id ne 'default' && @$filters) {
+                        my @filterObjects;
                         foreach my $filter (@{$profile->{'filter'}}) {
-                            push @Profile_Filters, pf::factory::profile::filter->instantiate($profile_id,$filter);
+                            push @filterObjects, pf::factory::profile::filter->instantiate($profile_id,$filter);
+                        }
+                        if($profile->{filter_match_style} eq 'all') {
```
julsemaan: Should be 
if(defined($profile->{filter_match_style}) && $profile->{filter_match_style} eq 'all') 
because when upgrading from a previous version, that parameter won't be there and will generate undefined variable warnings

julsemaan:
The selection of the all or any in the profile, should be placed above of the profiles. Right now it's above the languages.

profile-screen

julsemaan:
Also missing the news entry.

jrouzierinverse:
Did you mean above the filters?

julsemaan:
Yes sorry, above the filters

lib/pf/config.pm
julsemaan: what about this ? do you think it should be changed ?

jrouzierinverse: Good point we should also add in the upgrade doc to add filter_match_style=any to the default profile

dwlfrth:
I think both the "Filter match style" and the "Filters" should be at the very bottom -or- at the very top...
Currently, they are in the middle of the different profile attributes...
IMO...

julsemaan:
I'd agree with that. I'd put them at the top since they're the base of the portal profile

jrouzierinverse:
No problem I suggest we sit together and decided which way is best.

jrouzierinverse:
The filters should not be on the very top since it's break consistency with how we display every other config entries.
All config entries always have the ID as the very first field (Followed by the description if supported).

In addition the filters can have multiple entries, so putting it on top makes the display of single value options inconsistent because it will pushes down the display of the single value options.

However the profile page is getting busy
I will create a new pull request for reorganization of the portal profile.
We will discuss the reorganization of the portal profile configuration page in that pull request.
Since it is out of scope for the feature.

cgx:
The any/all select field should be just above the list of filters, right beside the _Filter_ label. As it is when defining filters for authentication sources:
screen shot 2015-01-14 at 10 31 50 am
Nevertheless, it would a good idea to add at least one new tab in the portal profile editor.

jrouzierinverse:
I will make the filters look more like the rules

jrouzierinverse:
Please review

html/pfappserver/lib/pfappserver/Form/Config/Profile.pm
```diff
@@ -31,7 +31,7 @@ The main definition block
 
 has_block 'definition' =>
   (
-   render_list => [ qw(id description logo redirecturl always_use_redirecturl reuse_dot1x_credentials billing_engine nbregpages) ],
+   render_list => [ qw(id description logo redirecturl always_use_redirecturl reuse_dot1x_credentials billing_engine nbregpages ) ],
```
dwlfrth: Trailing whitespace

html/pfappserver/lib/pfappserver/Form/Config/Profile.pm
```diff
@@ -69,6 +69,25 @@ has_field 'filter.contains' =>
    widget_wrapper => 'DynamicTableRow',
   );
 
+=head2 filter_match_style
+
+The field for filter_match_style
```
dwlfrth: Shouldn't that be "more descriptive" :P
I know that it is pretty obvious, but still...

dwlfrth:
Can we have a screencap of how it looks like. (Only with this PR, not with reordering)

lib/pf/profile/filter/all.pm
```diff
@@ -0,0 +1,61 @@
...
+    my ($self,$data) = @_;
+    return all { $_->match($data) } @{$self->value};
+}
+
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+Minor parts of this file may have been contributed. See CREDITS.
```
dwlfrth: ...

lib/pf/profile/filter/all.pm
```diff
@@ -0,0 +1,61 @@
+package pf::profile::filter::profile::filter::all;
+=head1 NAME
+
+pf::profile::filter::profile::filter::all - Class for profile::filter::all
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::profile::filter::profile::filter::all
```
dwlfrth: Kind of redondant IMO... That would be great to have a one line description ;)

jrouzierinverse:
Screen shot
screen shot 2015-01-14 at 3 49 43 pm

lib/pf/profile/filter/all.pm
```diff
@@ -0,0 +1,59 @@
+package pf::profile::filter::profile::filter::all;
```
julsemaan: Shouldn't that be pf::profile::filter::all.

If not then I think we'll have a conversation on packages names :P

jrouzierinverse: That was a mistake

julsemaan: Now next question, I guess the code doesn't work with that mistake. So is this PR already tested or it needs testing before it's merged ?

jrouzierinverse:
Please review

lib/pf/profile/filter/all.pm
jrouzierinverse: I just added a unit test

html/pfappserver/lib/pfappserver/Form/Config/Profile.pm
dwlfrth: "Field defining how the configured filters will be applied for matching" would be better.
Otherwise, simply get rid of the comment... it is kind of a "joke" IMO

html/pfappserver/root/config/profile/tab-content.tt
```diff
@@ -1,4 +1,18 @@
           [% form.block('definition').render %]
+          [% IF form.field('filter') %]
+          <div class="control-group">
+            <label class="control-label" for="sources">[% l('Filters') %] <i class="icon-exclamation-sign"></i> </label>
+            <div class="controls">
+            <dl>
```
dwlfrth: Indentation inside a div

lib/pf/config.pm
```diff
@@ -687,10 +688,16 @@ sub readProfileConfigFile {
...
-                        #Adding filters in profile order
+                    my $filters = $profile->{'filter'};
+                    if($profile_id ne 'default' && @$filters) {
+                        my @filterObjects;
                         foreach my $filter (@{$profile->{'filter'}}) {
-                            push @Profile_Filters, pf::factory::profile::filter->instantiate($profile_id,$filter);
+                            push @filterObjects, pf::factory::profile::filter->instantiate($profile_id,$filter);
+                        }
+                        if(defined ($profile->{filter_match_style}) && $profile->{filter_match_style} eq 'all') {
+                            push @Profile_Filters,  pf::profile::filter::all->new(profile => $profile_id, value => \@filterObjects);
```
dwlfrth: Just ot of curiosity... why can't we stick with some code standard... I know that may be very picky but I'm having difficulty with the fact that on two different lines, that follows each others, there's spaces after the comma, the other one nothing...
I'm doing the comment here because it is quite obvious... but that applies to a bunch of code...

lib/pf/pfcmd/checkup.pm
```diff
@@ -965,23 +965,27 @@ Make sure only one external authentication source is selected for each type.
...
 
-        foreach my $key ( keys %{$Profiles_Config{$portal_profile}} ) {
+        foreach my $key ( keys %$data ) {
             add_problem( $WARN, "invalid parameter $key for profile $portal_profile" )
                 if ( $key !~ /$profile_params/ );
         }
 
         my %external;
-        foreach my $source ( grep { $_ && $_->class eq 'external' } map { pf::authentication::getAuthenticationSource($_) } @{$Profiles_Config{$portal_profile}{'sources'}} ) {
+        foreach my $source ( grep { $_ && $_->class eq 'external' } map { pf::authentication::getAuthenticationSource($_) } @{$data->{'sources'}} ) {
```
dwlfrth, line 989: I think that a condition like this one would require just a single line of comment from the developper.
It will save someone else some time to understand what we're dealing with and that single line of comment only takes 10 seconds to write.

t/unittest/profile/filter/all.t
```diff
@@ -0,0 +1,67 @@
+=head1 NAME
+
+profile/filter/all.t
+
+=cut
+
+=head1 DESCRIPTION
+
+value
```
dwlfrth: value! excellent

t/unittest/profile/filter/all.t
```diff
@@ -0,0 +1,67 @@
...
+ 
+ok(!$filter->match({ test2 => 'test2'}),"fails to match when only one filter matches");
+ 
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
dwlfrth: 2015

jrouzierinverse:
Please review

html/pfappserver/lib/pfappserver/Form/Config/Profile.pm
dwlfrth: I'm still there...

jrouzierinverse:
Please review

NEWS.asciidoc
```diff
@@ -26,6 +26,7 @@ New Features
 * Added call to the web API from the vlan filters
 * Added a way to retrieve user information after the first registration.
 * Added the ability to filter profiles by connection type
+* Profiles can be matched by all or any of it's filters
```
extrafu, line 29: it's -> its

PR #301, opened 2015-01-09
fix portal uri profiles

julsemaan:
Once approved and merged, should hit the maintenance branch.
## Description

Fixes the URI portal profiles by caching the portal profile only when using a uri based portal profile.
Switching between two profiles is still working except when using the URI based filter
## Impacts

Fixes the URI portals. Doesn't break the current behavior
## Delete branch after merge

Yes.

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
@@ -92,6 +92,8 @@ has [qw(forwardedFor guestNodeMac)] => ( is => 'rw', );
 sub ACCEPT_CONTEXT {
     my ( $self, $c, @args ) = @_;
     my $class = ref $self || $self;
+    my $previous_model = $c->session->{$class};
+    return $previous_model if(defined($previous_model) && $previous_model->{options}->{in_uri_portal});
     my $model;
     my $request       = $c->request;
     my $r = $request->{'env'}->{'psgi.input'};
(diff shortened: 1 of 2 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
```diff
@@ -104,9 +106,10 @@ sub ACCEPT_CONTEXT {
     my $mgmt_ip = $management_network->{'Tvip'} || $management_network->{'Tip'} if $management_network;
     $destination_url = $request->param('destination_url') if defined($request->param('destination_url'));
 
-    if( $r->isa('Apache2::Request') &&  defined ( my $last_uri = $r->pnotes('last_uri') )) {
+    if( defined ( my $last_uri = $r->pnotes('last_uri') )) {
```
rouzier, line 109: Why was $r->isa('Apache2::Request') removed?

julsemaan, line 109: Because it wasn't an Apache2::Request

rouzier, line 109: Ok that only causes a problem in the catalyst server script so I will look into this later.
Since it does not cause a problem with the apache based server.

rouzier:
Looks good to me

dwlfrth:
Rebase and will be good to merge.

html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
```diff
@@ -92,6 +92,8 @@ has [qw(forwardedFor guestNodeMac)] => ( is => 'rw', );
 sub ACCEPT_CONTEXT {
     my ( $self, $c, @args ) = @_;
     my $class = ref $self || $self;
+    my $previous_model = $c->session->{$class};
+    return $previous_model if(defined($previous_model) && $previous_model->{options}->{in_uri_portal});
```
cgx, line 96: How about simply testing for the existence of `last_uri` in `$previous_model->{options}` instead of adding another value to the _options_ hash?

julsemaan, line 96: Because in the event that we use last_uri as really the last_uri the user visited then that would break the code. Would seem like a misusage to my eyes.

PR #302, opened 2015-01-09
Honor order of profiles when matching profile filter

jrouzierinverse:
## Description

Honor order of profiles when matching profile filter
## Impacts

Impacts the profiles matching
## Delete branch after merge

NO


fdurand:
Tested and works

PR #303, opened 2015-01-11
The filter a profile by connection type

jrouzierinverse:
## Description

New profile filter by connection type
## Impacts

Filtering the profile
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/profile/filter/connection_type.pm
+++ b/lib/pf/profile/filter/connection_type.pm
@@ -30,8 +30,6 @@ has '+key' => ( default => sub { 'last_connection_type' } );
 
 Inverse inc. <info@inverse.ca>
 
-Minor parts of this file may have been contributed. See CREDITS.
-
 =head1 COPYRIGHT
 
 Copyright (C) 2005-2015 Inverse inc.
--- a/lib/pf/profile/filter/connection_type.pm
+++ b/lib/pf/profile/filter/connection_type.pm
@@ -1,0 +1,58 @@
+package pf::profile::filter::connection_type;
+=head1 NAME
+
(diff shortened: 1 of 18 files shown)


julsemaan:
Looks good to me. It's a pretty simple PR

lib/pf/profile/filter/connection_type.pm
```diff
@@ -0,0 +1,58 @@
...
+
+=cut
+
+has '+key' => ( default => sub { 'last_connection_type' } );
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+Minor parts of this file may have been contributed. See CREDITS.
```
dwlfrth: Should that be there ?

jrouzierinverse: It say "may have been contributed."
I would prefer it to be there since if we have to add credits later.
We do not have to add it back later.

dwlfrth: There's one line of code in the file...

julsemaan: Actually there are 5 with the imports. jk

Since we don't usually put it in the other new files, I'd agree to remove it.

extrafu: New files MUST NOT have this mention.

dwlfrth: That was exactly my point

jrouzierinverse: I updated my copyright template. And fixed all new files that have this information.

jrouzierinverse:
Please Review

PR #305, opened 2015-01-15
Change how we strip username when reuse_dot1x_credentials is enabled

jrouzierinverse:
## Description

Change how we strip username when reuse_dot1x_credentials is enabled 
## Impacts

We are changing the existing workflow for authentication when reuse_dot1x_credentials is enabled
Should we do the one or a mix of the following
- Only strip name when using machine auth
- Make it configurable user/machine auth
- Extend the node/person to add a field with a stripped username
## Delete branch after merge

YES | NO

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -463,6 +463,7 @@ sub authenticationLogin : Private {
         #TODO Do we config stripping the username for machine and user auth
         if ( $username =~ /^[^\/]+\/(.*)$/ ) {
             $username = $1;
+            $logger->warn("Reuse Dot1x has been enabled on this portal, here the username from 802.1x connection ".$node_info->{'last_dot1x_username'}." , here the stripped username $username");
         }
         $c->session(
             "username"  => $username,
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -459,7 +459,9 @@ sub authenticationLogin : Private {
         my $mac       = $portalSession->clientMac;
         my $node_info = node_view($mac);
         my $username = $node_info->{'last_dot1x_username'};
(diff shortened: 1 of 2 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
```diff
@@ -459,7 +459,9 @@ sub authenticationLogin : Private {
         my $mac       = $portalSession->clientMac;
         my $node_info = node_view($mac);
         my $username = $node_info->{'last_dot1x_username'};
-        if ($username =~ /^(.*)@/ || $username =~ /^[^\/]+\/(.*)$/ ) {
+        #Strip the username when using machine auth
+        #TODO Do we config stripping the username for machine and user auth
+        if ( $username =~ /^[^\/]+\/(.*)$/ ) {
             $username = $1;
         }
```
fdurand, line 467: We can probably add a warn in the code like : $logger->warn("Reuse Dot1x has been enabled on this portal, here the username from 802.1x connection ".$node_info->{'last_dot1x_username'}." , here the stripped username $username");

extrafu, line 467: In fact, we MUST add a warning like this.

jrouzierinverse, line 467: Done

html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
```diff
@@ -459,8 +459,11 @@ sub authenticationLogin : Private {
         my $mac       = $portalSession->clientMac;
         my $node_info = node_view($mac);
         my $username = $node_info->{'last_dot1x_username'};
```
fdurand, line 461: In the case of status/login and if packetfence is not able to have the mac address based on the ip address then the username will be empty.

louismunro:
I think we need to see if this is still relevant at this point.
At the minimum it would require a rebase but it's debatable whether we still want to do it this way.

fdurand:
We have to split the logic, if status page ... else ...

PR #306, opened 2015-01-19
Preload all pf::Switch modules

jrouzierinverse:
## Description

Preload pf::Switch modules to avoid runtime compilation
## Impacts

Creation of pf::Switch and the display Switch module in the admin gui
## Delete branch after merge

NO


jrouzierinverse:
Please review

lib/pf/SwitchFactory.pm
```diff
@@ -165,6 +179,36 @@ sub config {
...
+
+=cut
+
+sub buildVendorsList {
+    for my $module (@MODULES) {
+        my $switch = $module;
+        $switch =~ s/^pf::Switch:://;
+        my @p = split /::/, $switch;
+        my $vendor = shift @p;
+        if ($module->can('description')) {
```
julsemaan, line 217: I'm not sure that's a good assumption. Assuming a module is a switch module because it has a description sub seems a bit weird

lib/pf/SwitchFactory.pm
```diff
@@ -28,9 +28,23 @@ use Benchmark qw(:all);
...
+
+#Loading all the switch modules ahead of time
+use Module::Pluggable
+  search_path => [qw(pf::Switch)],
+  'require' => 1,
+  sub_name    => 'modules';
+
+our @MODULES = __PACKAGE__->modules;
+
+our %TYPE_TO_MODULE = map { my $type = $_; $type =~ s/^pf::Switch:://; $type => $_ } grep { $_->can('description') } @MODULES;
```
julsemaan: Same for the description assumption here

Also, this seems like a pretty long line that is not easily readable.

I think this should be rewritten into a sub that builds this variable or at the very least have a comment above that describes what it does and how it does it.

lib/pf/SwitchFactory.pm
```diff
@@ -165,6 +179,36 @@ sub config {
     return \%temp;
 }
 
+sub getModule {
```
julsemaan, line 188: Missing POD doc for that sub

lib/pf/SwitchFactory.pm
```diff
@@ -165,6 +179,36 @@ sub config {
...
+    my ($type) = @_;
+    return (exists $TYPE_TO_MODULE{$type} ) ? $TYPE_TO_MODULE{$type} : undef;;
+}
+
+=item buildVendorsList
```
julsemaan, line 204: I think we should discuss the new sub headers all of us together before we start putting them in the codebase because then we may end up with (an even more) heterogenous pod documentation.

So I'd remove the usage + return for now.

lib/pf/SwitchFactory.pm
cgx, line 217: It's a way to ignore base modules such as `pf::Switch::Cisco` or `pf::Switch::Cisco::Aironet`. We test for the same method in `pfappserver::Form::Config::Switch` to only display the appropriate switch models in the admin interface.

julsemaan, line 217: I'm not saying it doesn't accomplish the goal, it's more a question of 'is it clean ?'

cgx, line 217: The question would be more: _Is it documented?_

jrouzierinverse:
I incorporated your suggestions in the module.
Let me know if there is anything else.

lib/pf/SwitchFactory.pm
julsemaan: I see you added the builder but this is stll here.

Remove the assignment and just declare it ?

lib/pf/SwitchFactory.pm
```diff
@@ -134,23 +149,23 @@ sub instantiate {
...
-    $type = untaint_chain($type);
-    # load the module to instantiate
-    if ( !(eval "$type->require()" ) ) {
+    unless ($module) {
         $logger->error("Can not load perl module for switch $requestedSwitch, type: $type. "
-            . "Either the type is unknown or the perl module has compilation errors. "
-            . "Read the following message for details: $@");
-        return 0;
+                  . "The type is unknown or the perl module has compilation errors. ");
+        return undef;
```
julsemaan: Really not sure about this one. When we instanciate we then test the result with if(!$switch) which will warn for undefined value. That's what I think but I may be wrong.

lib/pf/SwitchFactory.pm
julsemaan, line 204: I don't want to harass you with this but I really think you should stick to the old headers until we come up with the official header

lib/pf/SwitchFactory.pm
```diff
@@ -165,6 +180,57 @@ sub config {
     return \%temp;
 }
 
+=item getModule
```
julsemaan, line 182: Same here for the doc header

lib/pf/SwitchFactory.pm
```diff
@@ -165,6 +180,57 @@ sub config {
...
+    }
+}
+
+=item buildTypeToModuleMap
+
+builds the type to module map
+
+=cut
+
+sub buildTypeToModuleMap {
```
julsemaan, line 245: Beautiful !

jrouzierinverse:
Please Review

julsemaan:
Code looks good, I'll test then merge.

julsemaan:
Please add the news entry and then I'll merge

lib/pf/SwitchFactory.pm
```diff
@@ -165,6 +179,53 @@ sub config {
...
 
+=item getModule
+
+Get the module from the type
+
+=cut
+
+sub getModule {
+    my ($type) = @_;
+    return (exists $TYPE_TO_MODULE{$type} ) ? $TYPE_TO_MODULE{$type} : undef;;
```
louismunro: <nitpick>double semi-colon at the end of the line? </nitpick>

jrouzierinverse:
Please merge

louismunro:
Code has not been tested to show a significant improvement in runtime.

julsemaan:
Do we keep this for 4.6 ?

extrafu:
No.

julsemaan:
Still relevent @louismunro ?

louismunro:
Relevant but not ready to merge.
We need to see a real performance improvement.

louismunro:
Let's sit down  soon enough and see if we can measure the performance gains here.

jrouzierinverse:
Please review

louismunro:
Please rebase.

rouzier:
Review

PR #307, opened 2015-01-19
Ensure that the reloading of a config file only happenes once

jrouzierinverse:
## DESCRIPTION

Ensure that configuration files are only reloaded once to avoid reparsing
## Impacts

This impact all services that use configuration
bin/pfcmd configreload
bin/pfcmd configreload hard
## NEWS file entries

Bug Fixes
+++++++++
Fixed issue with cache reloading increasing memory usage while reloading the cache
## Delete branch after merge

NO

### DIFF
--- a/addons/dev-helpers/config-reload-loop.pl
+++ b/addons/dev-helpers/config-reload-loop.pl
@@ -2,13 +2,11 @@
 
 =head1 NAME
 
-loop.pl - a script to help test reloading config
-
-=cut
+config-reload-loop.pl - a script to help test reloading config
 
 =head1 DESCRIPTION
 
-loop
+Stress test by refreshing configuration by 30 different processes every 2 seconds
 
 =cut
 
--- a/addons/dev-helpers/config-reload-loop.pl
+++ b/addons/dev-helpers/config-reload-loop.pl
@@ -1,0 +1,103 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+loop.pl - a script to help test reloading config
+
+=cut
+
+=head1 DESCRIPTION
+
+loop
+
+=cut
+
+use strict;
+use warnings;
+use lib qw(/usr/local/pf/lib);
+use pf::config::cached;
+use pf::ConfigStore::Switch;
+use Cache::Memcached;
+use pf::db;
+
+use POSIX qw(:sys_wait_h pause);
+our %KIDS;
+$SIG{CHLD} = sub { };
+$SIG{INT} = $SIG{TERM} = sub {kill INT => keys %KIDS};
+our $RUNNING = 1;
+our $reloaded = 0;
+$pf::ConfigStore::Switch::switches_cached_config->addFileReloadCallbacks(
+    onfilereload => sub {print STDERR "$$ onfilereload " . time . "\n"});
+$pf::ConfigStore::Switch::switches_cached_config->addCacheReloadCallbacks(
+    oncachereload => sub {print STDERR "$$ oncachereload " . time . "\n"});
+
+{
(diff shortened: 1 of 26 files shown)


julsemaan:
Minor detail but can you change the title, it's DESCRIPTION

This way it's easier to refer to it or search it when it's closed.

jrouzierinverse:
Thanks

julsemaan:
@fdurand we'll need the packages built before we can merge this.

I'll test it in the meantime

jrouzierinverse:
The updated CHI is built already
File::FcntlLock just needs to be built 
You can get it from epel in the mean time perl-File-FcntlLock

addons/dev-helpers/config-reload-loop.pl
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+loop.pl - a script to help test reloading config
```
cgx: It doesn't match the exact name.

addons/dev-helpers/config-reload-loop.pl
```diff
@@ -0,0 +1,103 @@
...
+
+=cut
+
+=head1 DESCRIPTION
+
+loop
```
cgx: Can you be more precise?

lib/pf/FileLocker.pm
```diff
@@ -0,0 +1,229 @@
+package pf::FileLocker;
+=head1 NAME
+
+pf::FileLocker add documentation
```
cgx: Fix NAME and DESCRIPTION sections of the POD.

lib/pf/config/cached.pm
```diff
@@ -835,6 +916,23 @@ sub cacheForData {
...
+sub RefreshConfigs {
+    my $logger = get_logger();
+    $logger->trace("Started Reloading all configs");
+    foreach my $config (@LOADED_CONFIGS{@LOADED_CONFIGS_FILE}) {
+        next unless $config;
+        next unless $config->LockFileHasChanged;
+        $config->ReloadConfig();
+    }
+    $logger->trace("Finished Reloading all configs");
+}
```
cgx, line 940: Your log entries for this sub are the same as for `ReloadConfigs`. You should reword them.

julsemaan:
Tested with https://gist.github.com/julsemaan/302783b46c098e8fcb34 
- 2 processes updating an entry in the switches.conf file
- tested with 2915 switches
- Both are able to interact with the config store and load is not escalating too much.

Admin interface is slow but everything is still working.

julsemaan:
the chi version on the repo is 0.58-2 and not 0.59

You'll need to adjust the spec file

jrouzierinverse:
Check the packetfence-devel repo

julsemaan:
I did

julsemaan:
Confirmed : combining a switch modification with radius stress testing makes the load go high and almost no radius request are answered (using 2000 switches)

julsemaan:
What is everybody's opinion on this PR ?

Should we merge it or leave the code as is for now ?

jrouzierinverse:
Please Review or Merge

julsemaan:
No relevent anymore IMO. Close ?

louismunro:
Is the performance issue still there? 
If it's still there let's close this and rework it.

julsemaan:
The performance issue is still there but we're rewriting the config access to I'm not sure we should be adding additionnal layers to the 'old' config

louismunro:
Agreed.
Let's close until we know better where we are going with this.
We could always reopen if needed.

PR #308, opened 2015-01-19
Feature/stats d integration

louismunro:
Just a stub for the final pull request. 

This will add StatsD integration to PF so that we can simply call e.g. $statsd->timing( $stat, $time ) to log timing information.


lib/Etsy/StatsD.pm
```diff
@@ -0,0 +1,140 @@
...
+  CORE::send( $sock, $msg, 0 );
+}
+
+=head1 SEE ALSO
+
+L<http://codeascraft.etsy.com/2011/02/15/measure-anything-measure-everything/>
+
+=head1 AUTHOR
+
+Steve Sanbeg L<http://www.buzzfeed.com/stv>
```
extrafu, line 136: What's the copyright and license of that file/code?

lib/pf/StatsD.pm
```diff
@@ -0,0 +1,73 @@
...
+    initStatsd;
+}
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
extrafu: 2015

lib/Etsy/StatsD.pm
louismunro, line 136: Good point.

License is extremely liberal.
https://github.com/etsy/statsd/blob/master/LICENSE

Not sure if that's BSD or MIT.

extrafu, line 136: Perfect, please include it in the file itself.

Thanks!

julsemaan:
Will that be in PF5, if yes then I guess an associated issue should be opened and the proper milestone associated to it.

This also needs rebasing

louismunro:
It will be in 5.0.
Do we really want to use use issues for new features?
I thought it was for bugs/improvements.

julsemaan:
I'd put it in issues to track it outside of the code. Not absolutelly necessary but it gives us a central point to see the projects and in which version they belong

louismunro:
I am closing this one.
Will reopen another as the branches had diverged too much to my taste and so I created a new feature branch rather than go to the trouble to rebase.

PR #311, opened 2015-01-21
Sms retry limit

jrouzierinverse:
## Description

Will not allow a person to retry a sms activation code more than three times
## Impacts

Changes the workflow of sms activation
## Delete branch after merge

NO

### DIFF
--- a/html/captive-portal/lib/captiveportal.pm
+++ b/html/captive-portal/lib/captiveportal.pm
@@ -114,14 +114,17 @@ sub loadCustomStatic {
 
 =head2 user_cache
 
-Returns the user specific cache
+Returns the user/mac specific cache
 
 =cut
 
 sub user_cache {
     my ($c) = @_;
-    #Return a CHI::Driver that prefixes each
-    return CHI->new( driver => 'SubNamespace', chi_object => pf::CHI->new(namespace => 'httpd.portal'), namespace => $c->portalSession->clientMac );
+    return CHI->new(
+        driver     => 'SubNamespace',
+        chi_object => pf::CHI->new(namespace => 'httpd.portal'),
+        namespace  => $c->portalSession->clientMac
+    );
 }
 
 has portalSession => (
--- a/html/captive-portal/lib/captiveportal.pm
+++ b/html/captive-portal/lib/captiveportal.pm
@@ -43,6 +43,7 @@ use captiveportal::Role::Request;
 use pf::config::cached;
 use pf::file_paths;
 use pf::CHI;
+use CHI::Driver::SubNamespace;
 
 extends 'Catalyst';
 
@@ -111,8 +112,16 @@ sub loadCustomStatic {
     return $dirs;
 }
 
+=head2 user_cache
+
+Returns the user specific cache
+
+=cut
+
 sub user_cache {
-    return pf::CHI->new( namespace => 'httpd.portal');
+    my ($c) = @_;
+    #Return a CHI::Driver that prefixes each
+    return CHI->new( driver => 'SubNamespace', chi_object => pf::CHI->new(namespace => 'httpd.portal'), namespace => $c->portalSession->clientMac );
 }
 
 has portalSession => (
--- a/html/captive-portal/lib/captiveportal/Base/Controller.pm
+++ b/html/captive-portal/lib/captiveportal/Base/Controller.pm
@@ -53,7 +53,7 @@ If the max is undef or 0 then it will return true
 
 sub reached_retry_limit {
     my ( $self, $c, $retry_key, $max ) = @_;
-    return 1 unless $max;
+    return 0 unless $max;
     my $cache = $c->user_cache;
     my $retries = $cache->get($retry_key) || 1;
     if($retries > $max ) {
--- a/html/captive-portal/lib/captiveportal/Base/Controller.pm
+++ b/html/captive-portal/lib/captiveportal/Base/Controller.pm
@@ -52,12 +52,13 @@ Test if the retry limit has been reached for a session key
 
 sub reached_retry_limit {
     my ( $self, $c, $retry_key, $max ) = @_;
-    my $retries = $c->session->{$retry_key} || 0;
+    my $cache = $c->user_cache;
+    my $retries = $cache->get($retry_key) || 0;
     $retries++;
-    $c->session->{$retry_key}  = $retries;
     if($retries > $max ) {
         return 1;
     }
+    $cache->set($retry_key,$retries);
     return 0;
 }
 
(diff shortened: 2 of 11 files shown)


html/captive-portal/lib/captiveportal.pm
```diff
@@ -111,8 +112,16 @@ sub loadCustomStatic {
...
+=head2 user_cache
+
+Returns the user specific cache
+
+=cut
+
 sub user_cache {
-    return pf::CHI->new( namespace => 'httpd.portal');
+    my ($c) = @_;
+    #Return a CHI::Driver that prefixes each
```
dwlfrth: each... ? :P

html/captive-portal/lib/captiveportal/Base/Controller.pm
```diff
@@ -44,6 +44,24 @@ sub showError {
...
+
+=cut
+
+sub reached_retry_limit {
+    my ( $self, $c, $retry_key, $max ) = @_;
+    my $cache = $c->user_cache;
+    my $retries = $cache->get($retry_key) || 0;
+    $retries++;
+    if($retries > $max ) {
+        return 1;
```
dwlfrth, line 60: In these case, I'd prefer using $TRUE and $FALSE...

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Sms.pm
```diff
@@ -119,22 +119,30 @@ sub showSmsConfirmation : Private {
...
     # no form was submitted, assume first time
     my $pin = $c->request->param("pin");
     if ($pin) {
-        $c->log->info("Mobile phone number validation attempt");
-        if ( pf::activation::validate_code($pin) ) {
-            return ( $TRUE, 0 );
-        } else {
-            return ( $FALSE, $GUEST::ERROR_INVALID_PIN );
+        $logger->info("Mobile phone number validation attempt");
+        if ($self->reached_retry_limit($c, 'sms_retries', 3)) {
```
dwlfrth: Should be a configuration parameter (of the source)
- Do we enforce limiting
- Limit

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Sms.pm
```diff
@@ -119,22 +119,30 @@ sub showSmsConfirmation : Private {
...
+            my $mac = $c->portalSession->clientMac;
+            $logger->info("Max tries reached invalidating code for $mac");
```
dwlfrth, line 133: I'd add the limit (in this case, 3) in the comment

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Sms.pm
extrafu: s/Should be/Must be/

We must never hardcode things like that in the code, unless if a default and safe value if no configuration is found.

jrouzierinverse:
Please Review

html/captive-portal/lib/captiveportal/Base/Controller.pm
```diff
@@ -44,6 +44,26 @@ sub showError {
...
+=head2 reached_retry_limit
+
+Test if the retry limit has been reached for a session key
+If the max is undef or 0 then it will return true
+
+=cut
+
+sub reached_retry_limit {
+    my ( $self, $c, $retry_key, $max ) = @_;
+    return 1 unless $max;
```
cgx: If you set _sms_pin_retry_limit_ to 0 (as in `conf/profiles.conf.example`), the limit will always be reached and the user will never be able to validate her phone number. Is this what we want? If the limit is set to 0, I would expect the limit to be disabled.

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Sms.pm
```diff
@@ -119,22 +119,32 @@ sub showSmsConfirmation : Private {
...
 
     # no form was submitted, assume first time
     my $pin = $c->request->param("pin");
     if ($pin) {
-        $c->log->info("Mobile phone number validation attempt");
-        if ( pf::activation::validate_code($pin) ) {
-            return ( $TRUE, 0 );
-        } else {
-            return ( $FALSE, $GUEST::ERROR_INVALID_PIN );
+        $logger->info("Mobile phone number validation attempt");
```
cgx: I would categorize this log entry into _debug_.

html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
```diff
@@ -241,6 +241,19 @@ has_field 'nbregpages' =>
...
+
+=cut
+
+has_field 'sms_pin_retry_limit' =>
+  (
+    type => 'PosInteger',
+    label => 'SMS Pin Retry Limit',
+    default => 0,
+    tags => { after_element => \&help,
+             help => 'How many times a user can retry an sms pin having to request anothor pin. A value of 0 mean disables retries check.' },
```
cgx: I would rephrase this: 

_Maximum number of times a user can retry a SMS PIN before having to request another PIN. A value of 0 disables the limit._

lib/pf/activation.pm
```diff
@@ -303,6 +307,39 @@ sub invalidate_codes {
     return;
 }
 
+=item invalidate_codes_for_mac
```
cgx: Should be `=head2`, no?

lib/pf/web/guest.pm
```diff
@@ -490,6 +491,7 @@ Readonly::Hash our %ERRORS => (
     $ERROR_SPONSOR_NOT_ALLOWED  => 'Email %s is not allowed to sponsor guest access',
     $ERROR_PREREG_NOT_ALLOWED  => 'Guest pre-registration is not allowed by policy',
     $ERROR_INVALID_PIN => 'PIN is Invalid!',
+    $ERROR_MAX_RETRIES => 'Maximum amount of retries attempted',
```
cgx, line 494: Please update `conf/locale/en/LC_MESSAGES/packetfence.po`. I will eventually extend `extract_i18n_strings.pl` to cover the captive portal and not only the Web admin.

lib/pf/activation.pm
rouzier: Yes I will update the entire pod

lib/pf/web/guest.pm
rouzier, line 494: Done

PR #313, opened 2015-01-22
Will only create a single ldap connection when matching against multiple rules

jrouzierinverse:
## Description

Will only create a single ldap connection when matching against multiple rules
## Impacts

This impacts matching rule against an LDAP source 
## Delete branch after merge

NO


jrouzierinverse:
@louismunro  has tested this already.

cgx:
The code looks good!

louismunro:
Tested as charged.

PR #314, opened 2015-01-22
Renamed FR template files so they can be used by the code

dwlfrth:
## Description

FR violation template files were not usable by the code since it was trying to load fr_FR instead of FR
## Impacts

None... it was not working ...
## Delete branch after merge

YES


dwlfrth:
Review me please

julsemaan:
If you use fr_CA then I guess these templates don't work, right ?

cgx:
The problem is in `pf::Portal::Session->getLanguages` where we should include the 2-character languages from the profile's locales. I'll fix it.

julsemaan:
Good idea because I don't think that it should have been fr_FR anyway. 
French people would have felt too important with that.

dwlfrth:
:+1: What @cgx said
:+1: :+1: What @julsemaan said

PR #317, opened 2015-01-25
Initial work for pf::constants

jrouzierinverse:
## Description

A new module pf::constants
It help avoid using pf::config for simple $TRUE $FALSE
## Impacts

Any module that uses pf::config for $TRUE $FALSE $YES $NO $default_id
## NEWS file entries

Enhancements
++++++++++++
- Refactoring and code cleanup
## Delete branch after merge

NO


jrouzierinverse:
No longer needed

PR #321, opened 2015-01-28
Authentication sources rule classes

dwlfrth:
# Description

Add the ability to specify if a rule in an authentication source is of "admin" or "auth" type, and filters actions accordingly
# Impacts

Authentications sources
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Add the ability to create "administrative" or "authentication" purposes rules in authentication sources
## Enhancements
- Based on rule classes, will no longer match a "set_access_level" rule when trying to "set_role" (actions given for example...) Filter the applicable rules for any given authentication source based on the action requested
# UPGRADE file entry

Will need an entry to review rules...


dwlfrth:
Review me

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Rule.pm
```diff
@@ -327,6 +328,13 @@ sub validate {
         $self->field('actions')->add_error("You must at least set a role, mark the user as a sponsor, or set an access level.");
     }
 
+    @actions = grep { $_->{type} eq $Actions::SET_ACCESS_LEVEL } @{$self->value->{actions}};
+    if ( scalar @actions > 0 ) {
+        if ( @{$self->value->{actions}} > 1 ) {
+            $self->field('actions')->add_error("'set access level' should be an exclusive action to a rule.");
+        }
+    }
+
```
jrouzierinverse: Add this check in pf/pfcmd/checkup.pm

lib/pf/authentication.pm
```diff
@@ -367,6 +367,46 @@ sub getExternalAuthenticationSources {
...
+                if ( $action->{'type'} eq 'set_access_level' ) {
+                    push (@rules, $rule);
+                }
+            }
+        }
+
+        # If we have @rules defined and not empty, we consider this source as an admin authentication source and we 
+        # recreate it with only specific rules
+        if ( @rules ) {
+            @{$source->{'rules'}} = ();
```
jrouzierinverse: You should create a new source object not modify the current source object.

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Rule.pm
dwlfrth: I'm not sure I will add it to the checkup since none of the current constraints in the Form validation are checked by checkup.
Also, we may want... just for the beginning, override that behavior using the flat file.

julsemaan:
Need status

extrafu:
@fdurand Review and eventually merge.

extrafu:
@julsemaan Review too.

dwlfrth:
Should we go forward with that one ?

julsemaan:
Yes we discussed it last Friday and it is indeed useful

dwlfrth:
Review me

PR #329, opened 2015-02-04
Make OS selection explicit in the provisioners

julsemaan:
## Description

Fixes the OS detection in the provisioners
## Impacts

Fixes the OS detection in the provisioners
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
- Make provisioner OS selection always explicit to prevent bad overiding of the default OS for mobileconfig and Android
## Delete branch after merge

YES

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Provisioning/mobileconfig.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Provisioning/mobileconfig.pm
@@ -26,7 +26,7 @@ has_field 'ca_cert_path' =>
 
 has_block definition =>
   (
-   render_list => [ qw(id type description category ssid ca_cert_path) ],
+   render_list => [ qw(id type description oses category ssid ca_cert_path) ],
   );
 
 =head1 COPYRIGHT
--- a/lib/pf/provisioner.pm
+++ b/lib/pf/provisioner.pm
@@ -85,7 +85,7 @@ The oses to match against
 
 =cut
 
-#has oses => (is => 'rw', default => sub { [] } );
+has oses => (is => 'rw', default => sub { [] } );
 
 =head2 enforce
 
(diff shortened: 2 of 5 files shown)


jrouzierinverse:
Can you setup an environment that allow us to test this?
It do not see how this fixes the root cause

julsemaan:
Antoine has an environment with provisioners setup.
Other than that you'll have to create your test environment

julsemaan:
To explain what this fixes :
Use this provisioning configuration (in devel)

```
[android]
type=mobileconfig
template=provisioner/android.html
ssid=SANDWICH-AU-PAIN
hidden_ssid=FacketSecure
category=default
description=Android devices

[ios]
type=mobileconfig
#oses=
template=provisioner/mobileconfig.html
category=gaming
ssid=FacketSecure
description=Apple iOS devices
```

Then run : 

```
perl -Ilib -e 'use pf::factory::provisioner; use Data::Dumper; print Dumper(pf::factory::provisioner->new("android")->{oses})'
```

And it'll give you 'Apple ipod, iphone,....' while it should give you Android.

This seems to be caused because it was designed using the default value as a way to set what the module uses as a list of OS.

This breaks because android inherits from mobileconfig and mobileconfig declares it's list of OS in the default section so the default section of android is ignored. This way of assigning the OS seemed like a bad idea to me because if you put an empty list of OS in the configuration then running : 

```
perl -Ilib -e 'use pf::factory::provisioner; use Data::Dumper; print Dumper(pf::factory::provisioner->new("android")->{oses})'
```

will give you a list with an empty value.

Instead of fighing with all this code so close to the release I thought that making it explicit wouldn't hurt. This way if the user wants to add more OS than we initially wanted to he can, and it makes it clearer to debug instead of digging though multiple levels of inheritance that all define a default value.

jrouzierinverse:
This is no longer needed

PR #330, opened 2015-02-04
Reworked vlan.pm for autoregistration and added a new configuration parameter on the portal to recompute or no the role in case of 802.1x connection

fdurand:
## Description

Moved code from getNormalVlan to getNodeInfoForAutoreg.
If we configure autoregistration then the role and access duration will be compute in getNodeInfoForAutoreg instead of in getNormalVlan.  With that we will be able to create vlan filter rules in getNormalVlan based on the role of the device. 
Added a new parameter in the portal profile to compute or not the role based on the 802.1x userid (registration has been made on the portal)
Added a test to be sure that even if the node has been autoregistered on a 802.1x connection it pass in autoreg section (if no then we compute the role on the portal profile) 
## Impacts

Simplify getNormalVlan 
## NEWS file entries

Refactored autoregistration workflow in getNormalVlan and getNodeInfoForAutoreg
## Delete branch after merge

Yes

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
@@ -35,7 +35,7 @@ The main definition block
 
 has_block 'definition' =>
   (
-    render_list => [qw(id description reuse_dot1x_credentials billing_engine)],
+    render_list => [qw(id description reuse_dot1x_credentials dot1x_recompute_role_from_portal billing_engine)],
   );
 
 =head2 captive_portal
--- a/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
@@ -232,6 +232,21 @@ has_field 'reuse_dot1x_credentials' =>
     unchecked_value => 'disabled',
   );
 
+=head2 dot1x_recompute_role_from_portal
+
+=cut
+
+has_field 'dot1x_recompute_role_from_portal' =>
+  (
+    type => 'Checkbox',
+    checkbox_value => 'enabled',
+    unchecked_value => 'disabled',
+    default => 'enabled',
+    tags => { after_element => \&help,
+             help => 'When enabled PacketFence will not use the role initialy computed on the portal but will use the dot1x username to recompute the role.' },
+  );
+
+
 =head2 nbregpages
 
 =cut
--- a/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
@@ -243,7 +243,7 @@ has_field 'dot1x_recompute_role_from_portal' =>
     unchecked_value => 'disabled',
     default => 'enabled',
     tags => { after_element => \&help,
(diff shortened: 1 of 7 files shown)


fdurand:
Review please

html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
```diff
@@ -232,6 +232,21 @@ has_field 'reuse_dot1x_credentials' =>
...
+=cut
+
+has_field 'dot1x_recompute_role_from_portal' =>
+  (
+    type => 'Checkbox',
+    checkbox_value => 'enabled',
+    unchecked_value => 'disabled',
+    default => 'enabled',
+    tags => { after_element => \&help,
+             help => 'When enabled PacketFence will not use the role initialy computed on the portal but will use the dot1x username to recompute the role.' },
```
julsemaan: *When enabled, PacketFence
(Missing a comma)

lib/pf/vlan.pm
```diff
@@ -392,53 +401,42 @@ sub getNormalVlan {
...
-        if (isenabled($node_info->{'autoreg'})) {
-            my $value = &pf::authentication::match([@sources], $params, $Actions::SET_ACCESS_DURATION);
-            if (defined $value) {
-                $logger->trace("No unregdate found - computing it from access duration");
-                $value = access_duration($value);
-            }
-            else {
-                $value = &pf::authentication::match([@sources], $params, $Actions::SET_UNREG_DATE);
+        # Attributes has been computed in getNodeInfoForAutoReg
+        if ((isenabled($node_info->{'autoreg'}) && $autoreg) || !(isenabled($profile->dot1xRecomputeRoleFromPortal))) {
```
julsemaan: !isenabled would be clearer if written using isdisabled

lib/pf/vlan.pm
```diff
@@ -392,53 +401,42 @@ sub getNormalVlan {
...
-                    'source'  => $source,
-                    'portal'  => $profile->getName,
-                );
-                node_modify($mac,%info);
+            pf::person::person_modify($user_name,
+                'source'  => $source,
+                'portal'  => $profile->getName,
+            );
+            my %info = (
+                'autoreg' => 'no',
```
julsemaan, line 433: We should not set autoreg no since we may be in here because we enabled dot1xRecomputeRoleFromPortal

lib/pf/vlan.pm
```diff
@@ -590,11 +621,11 @@ sub shouldAutoRegister {
     my $node_info = node_attributes($mac);
     my $filter = new pf::vlan::filter;
     my ($result,$role) = $filter->test('AutoRegister',$switch, $port, $mac, $node_info, $conn_type, $user_name, $ssid, $radius_request);
-    if ($role) {
-        if ($switch->getVlanByName($role) eq -1) {
+    if ($result) {
+        if ($switch->getVlanByName($role) eq '-1') {
```
julsemaan: in radius.pm it's -1 as an integer. Which one is good or why do they need to be different

lib/pf/vlan.pm
fdurand, line 433: (isenabled($node_info->{'autoreg'}) && $autoreg) mean that we are not in autoreg workflow so i have to set autoreg to no

fdurand:
This code has been applied on 2 setup and work as expected

louismunro:
Reviewed and accepted.

PR #333, opened 2015-02-05
remove delete button from node view

julsemaan:
## Description

Removes the delete button from node view since it can't be pressed unless there is no locationlog entry opened for that node which practically never happens. Nodes should be delete/expired using pfmon.
## Impacts

Removes the option of deleting a node (how much was it used ?)
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
- Removed the node delete button.

### DIFF
--- a/html/pfappserver/root/node/view.tt
+++ b/html/pfappserver/root/node/view.tt
@@ -183,7 +183,6 @@
         </div><!--modal-body-->
 
         <div class="modal-footer">
-          [% IF can_access("NODES_DELETE") %]<a href="[% c.uri_for(c.controller('Node').action_for('delete'), [ node.mac ]) %]" id="deleteNode" class="btn btn-danger pull-left"><i class="icon-trash icon-white"></i> [% l('Delete') %]</a>[% END %]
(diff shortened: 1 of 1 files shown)


cgx:
Instead of removing it, how about allowing the user to _force_ the deletion of a node? Would it make sense?

julsemaan:
I know users would press it and then contact us because the code is broken. Force deletion would break the deauthentication and other post-connection flows.

It really doesn't seem like a good idea to me to have that accessible.

jrouzierinverse:
How about only showing it when a device has no locationlog entry?
Also If the decision is to remove it permanently remove the delete button, then we should remove the code in the controller and model to delete also.

julsemaan:
We could, but then I see the questions coming in on why the button is there sometimes and not there some other times. 

I agree we should remove the backend code if this idea is approved

jrouzierinverse:
Instead not show the button. We could just disable the button with a note on why we cannot not delete it.

julsemaan:
That would be pretty much what it's doing now. It brings back the error 'Can't delete the node because it's still active'

jrouzierinverse:
Yes the difference we say it in the beginning

PR #334, opened 2015-02-05
Added yum command + changed SQL commands to absolute paths

tristanbob:
Added the most important step, running "yum update  --enablerepo=packetfence".

Also changed relative paths for MYSQL upgrades to absolute path.  If someone did a default install or used Packetfence ZEN appliance, these commands will now work.   If they installed PacketFence in a custom location, they will likely be skilled enough to change to the path of these commands.

### DIFF
--- a/UPGRADE.asciidoc
+++ b/UPGRADE.asciidoc
@@ -9,7 +9,7 @@ For RedHat-based systems, run the following command:
 
   yum update --enablerepo=packetfence
   
-Packetfence should now be upgraded.  However, there may be extra steps required depending on the version you are upgrading from.  Please review the following notes about upgrading from an older release.
+PacketFence should now be upgraded.  However, there may be extra steps required depending on the version you are upgrading from.  Please review the following notes about upgrading from an older release.
 
 Upgrading from a version prior to 4.6.0
 ---------------------------------------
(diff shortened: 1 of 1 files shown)


UPGRADE.asciidoc
```diff
@@ -3,7 +3,13 @@ PacketFence Upgrade Guide
...
 http://www.packetfence.org/
 
-Notes on upgrading from an older release.
+Upgrade procedure
+-----------------
+For RedHat-based systems, run the following command:
+
+  yum update --enablerepo=packetfence
+  
+Packetfence should now be upgraded.  However, there may be extra steps required depending on the version you are upgrading from.  Please review the following notes about upgrading from an older release.
```
cgx: _PacketFence_ (with a capital F)

julsemaan:
Looks good to me. I see the point in having the absolute paths.

Good to merge for me

dwlfrth:
Will actually modify it. We shouldn't instruct the end-user the update the whole system by simply running 'yum update'. We should 'yum update packetfence...'. We are maintaining an UPGRADE guide for PacketFence and not the whole OS. Yes it is a good practice to update the OS but it can lead to some issues too.

PR #336, opened 2015-02-06
Generating the same password multiple times

jrouzierinverse:
## Description

When generating a password each apache process has the same random seed
Which would cause every apache process to create the exact same password
## Impacts

Creating users in the httpd.admin
## NEWS file entries

Bug Fixes
+++++++++
Fixed issue where the same password can be generated multiple times
## Delete branch after merge

NO


extrafu:
Nice fix!

jrouzierinverse:
Please review

fdurand:
Ok to merge

extrafu:
Should go in 4.6's maint

PR #337, opened 2015-02-06
Time based profile filters

jrouzierinverse:
## Description

Allow a profile to be filtered by time of day or days of the week
## Impacts

The captive portal profile filtering
## NEWS file entries

New Features
++++++++++++
- Added the ability to filter a profile by the time of the day
- Added the ability to filter a profile by the days of the week
## Delete branch after merge

NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/I18N/i_default.po
+++ b/html/pfappserver/lib/pfappserver/I18N/i_default.po
@@ -6040,6 +6040,10 @@ msgid "pf::profile::filter::connection_type"
 msgstr "Connection Type"
 
 # pf::filter (Portal Profile Filters)
+msgid "pf::profile::filter::day_of_week"
+msgstr "Day Of Week"
+
+# pf::filter (Portal Profile Filters)
 msgid "pf::profile::filter::network"
 msgstr "Network"
 
--- a/html/pfappserver/lib/pfappserver/I18N/i_default.po
+++ b/html/pfappserver/lib/pfappserver/I18N/i_default.po
@@ -6041,7 +6041,7 @@ msgstr "Connection Type"
 
 # pf::filter (Portal Profile Filters)
 msgid "pf::profile::filter::day_of_week"
-msgstr "Day Of Week"
+msgstr "Day of week"
 
 # pf::filter (Portal Profile Filters)
 msgid "pf::profile::filter::network"
--- a/html/pfappserver/lib/pfappserver/I18N/i_default.po
+++ b/html/pfappserver/lib/pfappserver/I18N/i_default.po
@@ -6068,6 +6068,10 @@ msgid "pf::profile::filter::switch_port"
 msgstr "Switch Port"
 
 # pf::filter (Portal Profile Filters)
+msgid "pf::profile::filter::time_of_day"
+msgstr "Time of day"
+
+# pf::filter (Portal Profile Filters)
 msgid "pf::profile::filter::uri"
 msgstr "URI"
 
--- a/lib/pf/profile/filter/day_of_week.pm
+++ b/lib/pf/profile/filter/day_of_week.pm
@@ -1,0 +1,101 @@
+package pf::profile::filter::day_of_week;
+=head1 NAME
+
+pf::profile::filter::day_of_week add documentation
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::profile::filter::day_of_week
+
+=cut
+
+use strict;
+use warnings;
+use Date::Format;
+use Moo;
+extends 'pf::profile::filter';
+
+=head1 ATTRIBUTES
+
+=head2 allowed_days
+
+The allowed days of the week
+
+=cut
+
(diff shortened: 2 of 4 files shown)


jrouzierinverse:
Please Review

lib/pf/profile/filter/day_of_week.pm
```diff
@@ -0,0 +1,101 @@
...
+
+add a trigger to the value to create update allowed_days
+
+=cut
+
+has '+value' => ( trigger => 1 );
+
+=head1 METHODS
+
+=head2 match
```
julsemaan, line 40: +1 for that documentation header.

This is a perfect example of something that needed additional info and that header covers it

lib/pf/profile/filter/day_of_week.pm
```diff
@@ -0,0 +1,101 @@
...
+
+=head2 _trigger_value
+
+Set allowed_days from the value
+
+=cut
+
+sub _trigger_value {
+    my ($self) = @_;
+    my %allowed_days = map { $_ => undef} split /\s*,\s*/ ,$self->value;
```
julsemaan: Could we maybe create that hash with values of 1. Just to make it explicit that it's a truth value.

html/pfappserver/lib/pfappserver/I18N/i_default.po
```diff
@@ -5935,6 +5935,10 @@ msgid "pf::profile::filter::connection_type"
 msgstr "Connection Type"
 
 # pf::filter (Portal Profile Filters)
+msgid "pf::profile::filter::day_of_week"
+msgstr "Day Of Week"
```
julsemaan: Any special reason why it shouldn't be written 'Day of week' (without capital letters at the beginning of each word)

lib/pf/profile/filter/day_of_week.pm
```diff
@@ -0,0 +1,101 @@
...
+    $self->allowed_days(\%allowed_days);
+}
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
julsemaan: 2015

lib/pf/profile/filter/time_of_day.pm
```diff
@@ -0,0 +1,98 @@
...
+    $self->end_time($end);
+}
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
julsemaan: 2015

lib/pf/profile/filter/time_of_day.pm
```diff
@@ -0,0 +1,98 @@
+package pf::profile::filter::time_of_day;
+=head1 NAME
+
+pf::profile::filter::time_of_day add documentation
```
julsemaan: add documentation ebola virus still happening

lib/pf/profile/filter/day_of_week.pm
extrafu, line 40: Comments submitted from Giorgo aren't tolerated here!

lib/pf/profile/filter/time_of_day.pm
```diff
@@ -0,0 +1,98 @@
...
+ 
+=head2 _trigger_value
+
+Set start_time and end_time from the value
+
+=cut
+
+sub _trigger_value {
+    my ($self) = @_;
+    my ($start,$end) = split(/-/,$self->value);
```
julsemaan, line 65: This would clearly benefit from validation + error logging.

julsemaan:
You should add validation though pfcmd for these new triggers as they need to follow a specific syntax

lib/pf/profile/filter/day_of_week.pm
```diff
@@ -0,0 +1,101 @@
+package pf::profile::filter::day_of_week;
+=head1 NAME
+
+pf::profile::filter::day_of_week add documentation
```
julsemaan: add or remove documentation

lib/pf/profile/filter/day_of_week.pm
julsemaan, line 40: Except from the Giorgio in laval. As it welcomes the most jetset customers, it is exempt from any rules governing the lower end society.

jrouzierinverse:
I agree with adding validation. 
However I prefer to do this in a separate pull request since we should also add validation for the other filters.

I would prefer a general solution for all the filters instead just for these two.

julsemaan:
Then I'll suggest we pause this PR. Wait for the validation PR, then adapt this PR to have validation. I find that making an error in these filters is too easy for an end user and having no form of validation will create had to debug cases

jrouzierinverse:
This is going to be replaced by the filter engine

PR #340, opened 2015-02-11
Roaming support based on accounting data for Aerohive

fdurand:
## Description

We update the locationlog based on accounting data.
## Impacts

Must activate accounting on the access point.
pfsetvlan and snmptrap will be useless for roaming
## NEWS file entries

Update locationlog from the accounting data (Aerohive access point)
## Delete branch after merge

Yes

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -81,6 +81,21 @@ sub radius_accounting : Public {
     return $return;
 }
 
+sub radius_update_locationlog : Public {
+    my ($class, %radius_request) = @_;
+    my $logger = pf::log::get_logger();
+
+    my $radius = new pf::radius::custom();
+    my $return;
+    eval {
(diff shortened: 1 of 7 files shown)


lib/pf/api.pm
```diff
@@ -81,6 +81,21 @@ sub radius_accounting : Public {
...
+    my ($class, %radius_request) = @_;
+    my $logger = pf::log::get_logger();
+
+    my $radius = new pf::radius::custom();
+    my $return;
+    eval {
+        $return = $radius->update_locationlog_accounting(\%radius_request);
+    };
+    if ($@) {
+        $logger->logdie("radius update locationlog accounting failed with error: $@");
```
extrafu: Should we really die?

fdurand:
Tested with a client, look like working as expected

extrafu:
This should go in v4.7 in my opinion.

PR #341, opened 2015-02-13
Added nested groups support for Active Directory Authentication Source

fdurand:
## Description

Added memberOf:1.2.840.113556.1.4.1941: as a filter attribute to be able to search in the chain of group ancestry
## Impacts

No
## NEWS file entries

Added nested groups support for Active Directory
## Delete branch after merge

Yes

### DIFF
--- a/lib/pf/Authentication/Source/ADSource.pm
+++ b/lib/pf/Authentication/Source/ADSource.pm
@@ -16,6 +16,13 @@ extends 'pf::Authentication::Source::LDAPSource';
 
 has '+type' => ( default => 'AD' );
 
+=head2 available_attributes
+
+Available ldap search attributes for Active Directory
+memberOf:1.2.840.113556.1.4.1941: attribute is for nested group, see https://msdn.microsoft.com/en-us/library/aa746475%28v=vs.85%29.aspx
+
+=cut
+
 sub available_attributes {
   my $self = shift;
   my $super_attributes = $self->SUPER::available_attributes;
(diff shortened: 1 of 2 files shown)


julsemaan:
I think that should that be documented somewhere. When you look at it, it looks more like a visual bug than a nice feature.

fdurand:
Yes i am completly ok with you, i asked to have something like that in packetfence to be able to have the documentation directly in the admin gui (http://getbootstrap.com/2.3.2/javascript.html#popovers)

fdurand:
I want to be merge

lib/pf/Authentication/Source/ADSource.pm
```diff
@@ -24,6 +24,7 @@ sub available_attributes {
      { value => "sAMAccountName", type => $Conditions::SUBSTRING },
      { value => "sAMAccountType", type => $Conditions::SUBSTRING },
      { value => "userAccountControl", type => $Conditions::SUBSTRING },
+     { value => "memberOf:1.2.840.113556.1.4.1941:", type => $Conditions::SUBSTRING },
```
louismunro, line 34: Too much magic number makes my brain hurt. 
Can we put that into a named constant at least?

fdurand, line 34: Doing that:
our $NESTED_GROUP = "memberOf:1.2.840.113556.1.4.1941:";
...
 { value => "memberOf:1.2.840.113556.1.4.1941:", type => $Conditions::SUBSTRING }
kill the translation.
So what do you prefer, having nested group or having memberOf:1.2.840.113556.1.4.1941: in the admin gui ?
I can also just add a pod with  this link https://msdn.microsoft.com/en-us/library/aa746475%28v=vs.85%29.aspx

PR #342, opened 2015-02-15
Avoid creating the same switch twice

jrouzierinverse:
## Description

Within the scope of a function call the same switch object will be reused until the switch goes out of scope.
This is done to avoid creating multiple switch object for the same switch id
## Impacts

Any code the uses the pf::SwitchFactory
## NEWS file entries

Enhancements
++++++++++++
Optimize switch creation
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/SwitchFactory.pm
+++ b/lib/pf/SwitchFactory.pm
@@ -28,6 +28,9 @@ use Benchmark qw(:all);
 use List::Util qw(first);
 use pf::ConfigStore::Switch;
 use pf::CHI;
+use Scalar::Util qw(weaken);
+
+our %WEAKEN_REFS;
 
 our ($singleton);
 
@@ -100,6 +103,10 @@ sub instantiate {
             push @requestedSwitches,$switch_ip;
         }
     } else {
+        #Check to see if the switch was already requested before
+        if( exists $WEAKEN_REFS{$switchRequest} && defined (my $object = $WEAKEN_REFS{$switchRequest} ) ) {
+            return $object;
+        }
         @requestedSwitches = ($switchRequest);
         if(valid_ip($switchRequest)) {
             $switch_ip = $switchRequest;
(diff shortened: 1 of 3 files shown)


julsemaan:
Not that I want to disencourage that type of work but like the Switch modules preloading, I don't know if we're going to see a real performance improvement, meaning that it wouldn't be useful to add another layer of complexity to the code.

rouzier:
The only way to know is to test

jrouzierinverse:
Here is the results of the timing script.

Old code
timethis for 3:  3 wallclock secs ( 3.17 usr +  0.02 sys =  3.19 CPU) @ 3669.28/s (n=11705)
New code
timethis for 3:  3 wallclock secs ( 3.04 usr +  0.00 sys =  3.04 CPU) @ 83717.76/s (n=254502)

jrouzierinverse:
Julien this pull request is based of your idea of caching the switch creation.

julsemaan:
More talking about what it will give in our codebase. It will obviously be faster but bf we save 0.01 millisecond per request then we can't say that it has a real impact on the code. Temp caching something that will be reused once isn't that much of a gain. I'm not saying we should stop optimising but we should profile our code to see what needs optimization so we don't optimize parts of the program that don't need it. Integration testing will tell for that one.

julsemaan:
The switch (to my knowlodge) and SwitchFactory always goes out of scope before it's called again. That is mainly my point. The fact that there is a need for Singleton + optimization in the SwitchFactory was unanimous when we discussed it. Only thing I'm questionning here is to know if this is the correct solution to our problem. Using a more global cache could be a better solution so the cached objects live through the whole request or through multiple requests.

jrouzierinverse:
When you say Singleton what object will be the singleton the pf::SwitchFactory or the pf::Switch

julsemaan:
Should be the switches the factory creates. The SwitchFactory is currently a singleton but it would make much more sense for the Switches it creates to be the singletons. That would need more elaborate caching though since these singleton objects need to be expired with config changes.

jrouzierinverse:
After some thoughts about this or any code that caches switch objects has the possibility of breaking something.
First is the overlay where we can have additional data added to the switch on creation.
Since this data is dynamic by nature it cannot be cached it must be looked up every time.

Second is the switch can change state during so if cache the singleton switch object.
Old data may be available which may cause problems with the workflow later down the line.

Since there is no way to guarantee that the object will not change or the changes made will not cause problems when reusing the object.

We should not attempt to make switch objects singletons.

julsemaan:
I agree with that. The overlay adds too much dynamism to the switches for them to become singletons. There are ways that come to my mind to do it differently though but I think we're going out of the scope of this PR. In the end, this deserves to be stress tested so we can see what we are gaining with this new caching layer.

PR #350, opened 2015-02-16
Add pfconfig service for the SwitchFactory

julsemaan:
## Description

This brings a new service (pfconfig) that lives outside of PacketFence. It is a possible replacement for the global configuration hashes. This uses the tying functionnality of Perl to proxy calls of the elements to an external service in charge of building and maintaining the configuration objects.

This central service uses a raw memory cache, a backend (could be anything) for cross process sharing and has the ability to build any configuration object from scratch to it's final form (usually the config hash).

This supports currently hashes and arrays but could be extended to support objects.

This service is directly tied to the resources it controls, meaning that the caching system knows exactly how to recreate the resources it caches without any external code. They are part of the service.

Important elements missing ? : 
- A factory to create the tied resources
- Many more I'm sure...

This description needs some more love...

This code needs some cleanup. There are scripts that shouldn't be included in the release but that would be needed for testing. 
## Impacts

Right now, only the switch configuration uses the new tied interface (in readonly) and since the code is centralized in the SwitchFactory, the impacts should be contained to that module.

See t/configstore-2-pfconfig.t for comparison of the ConfigStore generated hashes with the pfconfig generated hashes.
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Add the pfconfig service as the central interface to the configuration.

Bug Fixes
+++++++++
- Fix the case where the RADIUS requests would stop processing when a modification was made to a big switches.conf file
## Delete branch after merge

NO (for patches)


julsemaan:
Closing to reopen with new branch

PR #361, opened 2015-02-18
No autocorrect / uppercasing on login field

dwlfrth:
## Description

We should tell the device (iPad per example) not to autocorrect / uppercasing login field
## Impacts

None... Except if not supported by some browsers
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Device will avoid autocorrect / uppercasing the login field in the captive portal

Bug Fixes
+++++++++
## Delete branch after merge

YES

### DIFF
--- a/html/captive-portal/templates/device-login.html
+++ b/html/captive-portal/templates/device-login.html
@@ -61,7 +61,7 @@ document.observe("dom:loaded", function() {
           [%# User / Pass %]
           <div class="input">
             <span>[% i18n("Username") %]</span>
-	    <input class="field" name="username" id="username" type="text" value="[% username %]" />
+	    <input class="field" name="username" id="username" type="text" autocorrect="off" autocapitalize="off" value="[% username %]" />
           </div>
           <div class="input">
             <span>[% i18n("Password") %]</span><input class="field" name="password" type="password" />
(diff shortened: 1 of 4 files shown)


dwlfrth:
Review me

julsemaan:
Tested with Chrome on Android + IE8.

Android device doesn't autocap + autocorrect and IE8 doesn't break (surprise surprise)

If devel is ready to accept merging then we can go ahead and merge this one

cgx:
Technically, this is not supported in XHTML Transitional. I suggest we move to HTML5.

cgx:
I insist: please change the doctype.

dwlfrth:
What should be the new doctype ?
Will change it in templates/header.html

cgx:
`<!DOCTYPE html>`

dwlfrth:
So we should remove
<!DOCTYPE html
     PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
     "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
and replace it by
<!DOCTYPE html>

?

cgx:
Yes!

dwlfrth:
https://github.com/inverse-inc/packetfence/commit/8f645c428ab7b4a1110aa228c2227397852a4488

PR #373, opened 2015-02-20
Add filter category on portal profile

lzammit:
## Description

Now you can create a portal profile based the category (role) of a node.
## NEWS file entries

Add category filter on portal profile

Enhancements
++++++++++++
Add category filter on portal profile
## Delete branch after merge

No


html/pfappserver/lib/pfappserver/I18N/i_default.po
```diff
@@ -5931,6 +5931,10 @@ msgid "pending"
 msgstr "pending"
 
 # pf::filter (Portal Profile Filters)
+msgid "pf::profile::filter::category"
+msgstr "Category - Role"
```
jrouzierinverse: I think this should be
"Node category/role" or just 
"Node role"

cgx:
Could we display a select input field with values from the `node_category` database table?

jrouzierinverse:
We could but I would want to keep that as a separate pull request.
Since I am planning to rework the display of the filters.
I would prefer to have the mini-framework there instead of doing a one off just for this filter

cgx:
@jrouzierinverse All right.

@lzammit Even though the table field is named `category`, I would tend to use the new nomenclature and name the module `pf::profile::filter::role`.

jrouzierinverse:
Would the name pf::profile::filter::node_role be better?

cgx:
Sure!

lzammit:
done @louismunro

lzammit:
please merge @julsemaan

julsemaan:
Why is there two new filters added. Also you changed the module name (which is good) but you changed the matching attribute at the same time which to my eyes looks like it's going to break the filter. Please fix these

lzammit:
please review @julsemaan

PR #377, opened 2015-02-23
Feature/bypass role

louismunro:
## Description

Allows nodes to have a bypass_role (same as bypass_vlan) that allows unconditional role override on a node by node basis. 
## Impacts

Should return the bypass_role when node is registered regardless of authentication rules.
## NEWS file entries

New Features
++++++++++++
Bypass Role: allows bypassing all authorization rules for a registered node. 
## Delete branch after merge

YES


julsemaan:
I would maybe rephrased the news file entry as : 
- Added a bypass role to the nodes that bypasses all authorization rules for a registered node.

julsemaan:
That PR also needs rebasing against devel

louismunro:
It's not really finished anyway. 
It needs a few changes in the admin.

jrouzierinverse:
Closing

PR #387, opened 2015-02-24
Verify if a user has sponsor rights before allowing them to sponsor a user

jrouzierinverse:
## Description

Fixes issue where a non sponsor user can validate a sponsored email
## Impacts

Impacts the sponsored user workflow.
May require an additional rule in the admin source to mark the username as a sponsor
## NEWS file entries

Bug Fixes
+++++++++
Fixes allowing a non-sponsored user to verify a sponsored email address
## Delete branch after merge

NO

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
@@ -104,6 +104,10 @@ TODO: documention
 
 sub login : Private {
     my ( $self, $c ) = @_;
+    if ( $c->has_errors ) {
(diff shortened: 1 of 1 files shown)


julsemaan:
To clarify your questionning in the description. i dont think theres a need to create additionnal rules since we first lookup the object using the mail field and then pass it through the rules so I think it's good as is.

rouzier:
We would need two rules for in order for this to work.
The first to match the user's email address to see if the email address is a sponsor's email address (this is what we do now)
The second rule is to match the user's name that they enter.
Unless there is another way to figure out if the user is sponsor before allow them to let the user in.

julsemaan:
right now we dont create a rule for the email. we lookup the ldap info from the email. Unless im misunderstanding it. Ill probably just test it to be sure

fdurand:
Tested and it works

PR #390, opened 2015-02-25
Fix locales detection on the portal and use the closest one

fdurand:
## Description

Use the closest language of the portal that match the browser language.
Fix email subject with the correct charset
## Impacts

No
## NEWS file entries

Detect the closest language to use on the portal.
Fix email subject encoding
## Delete branch after merge

Yes

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
@@ -194,6 +194,8 @@ sub getLanguages :Private {
         }
     }
 
+    # 4. Check the closest language that match the browser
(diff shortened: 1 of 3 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
```diff
@@ -192,14 +194,28 @@ sub getLanguages :Private {
         }
     }
 
+    foreach my $browser_language (@$browser_languages) {
```
cgx, line 198: Please comment this block. It should be labeled as the fourth check.

fdurand:
This code has been applied on a setup to fix locale on the portal

PR #392, opened 2015-02-26
Fix javascript escaping issue with the profile file browser

jrouzierinverse:
## Description

When ever there is a single quote in a translation it can break the javascript
## Impacts

View files in the portal profile file browser
## NEWS file entries

Bug Fixes
+++++++++
Fix portal profile file browser in french
## Delete branch after merge

NO


cgx:
Please use a tt filter instead, as defined in _pfappserver_ in [pfappserver::View::HTML](https://github.com/inverse-inc/packetfence/blob/devel/html/pfappserver/lib/pfappserver/View/HTML.pm#L51).

rouzier:
ok

PR #395, opened 2015-02-26
Show better error messages when auth doesn't give proper role or unregdate

### LINKED ISSUE #380: Max number of devices reached.
Correct that message please.

whitx:
## Description

Show accurate error message while trying to auth on a source that doesn't have role or unregdate matching your login username(instead of max devices reached).
fixes #380
## Impacts

Authentication with a source that does not have role or unregdate setup. 
## NEWS file entries

Bug Fixes
+++++++++

Fix error messages.
## Delete branch after merge

NO

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
@@ -312,7 +312,8 @@ sub doSponsorRegistration : Private {
 
             $c->session->{"username"} = $pid;
             $c->session->{source_id} = $source->{id};
-            $c->stash->{info}=\%info;
+            $c->session->{source_match} = undef;
+            $c->stash->{info}=\%info; 
             $c->forward('Authenticate' => 'postAuthentication');
             $c->forward('Authenticate' => 'createLocalAccount', [$auth_params]) if ( isenabled($source->{create_local_account}) );
             $c->forward('CaptivePortal' => 'webNodeRegister', [$pid, %{$c->stash->{info}}]);
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -319,7 +319,8 @@ sub setRole : Private {
         $logger->debug("Got role '$value' for username \"$pid\"");
         $info->{category} = $value;
     } else {
-        $logger->debug("Got no role for username \"$pid\"");
+        $logger->info("Got no role for username \"$pid\"");
+        $self->showError($c, "The username you have used does not match any configured role.");
     }
 
 }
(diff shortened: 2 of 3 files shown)


cgx:
Does the error really mean anything to the user? Since the remediation involves an administrator, shouldn't we return an error such as:

> Your username has currently no access to the network. Please contact your network administrator.

And we should provide the phone number and/or email address of the helpdesk/administrator.

whitx:
Then shouldn't it be

> the error message + "Please contact your network administrator" 

With mail/number, this way the admin know straight why he doesn't have access(w/o looking through logs).

cgx:
That's fine!

louismunro:
I believe this does not solve the problem.
See maxRegNodesReached in html/captive-portal/lib/captiveportal/PacketFence/Controller/CaptivePortal.pm

louismunro:
I could be wrong. Have you tested this?

whitx:
I tested, while logging through the source, The right message popup, anyhow there is way more to do if we want to improve the message, since we will need this for every other messages.

whitx:
Worked with AD, it doesnt with Email, trying other sources now

louismunro:
Ok, then let's do it please.

louismunro:
Would also require translation and i18n.

julsemaan:
I wouldn't say to contact the administrator personally. 

The thing is you could match no auth rule in a source but bind to it. ex : my AD username + password is ok but I'm not part of the group that can have access.

Then get role will return nothing, but not because of a misconfiguration, but because I'm not aloud to connect to the network.

Asking to contact the administrator would suggest that the user deserves network access but that a problem has happened.

julsemaan:
Same thing with max devices actually. If I reached the maximum number of devices then I shouldn't contact my admin. I should unregister devices.

julsemaan:
Also from what I remember showError I18N's the errors directly so it just needs to be added to transifex

cgx:
So back to my initial comment: let's make the error message meaningful to the user and give her a remedial procedure if any. Examples of sentences that could be appended or prefixed to the error message:
- You don't have access to the network.
- Please unregister a device before registering a new one.
- You've been lost in an `else clause` of PacketFence.

julsemaan:
I really like the last one.

Jokes apart I'll rework those tomorrow

extrafu:
We could also let users know to call the 24/7 hotline and use Giorgio's number in there.

julsemaan:
I've made the error message better. Didn't go into the whole contact administrator thing since this is clearly out of the scope of 4.7.

The max node reached has to be reworked to bring the user to a page to unregister devices anyway.

julsemaan:
wait before merging, still needs guests to be fixed

louismunro:
Looks good to me.
"Le mieux est l'ennemi du bien."
Merge away!

whitx:
Still not owrking on email/sms reg

julsemaan:
Fixed and tested. Waiting for @whitx to test also and we'll merge.

PR #404, opened 2015-03-04
Fix broken registration VLAN handling when registrationVLAN is not integer.

### LINKED ISSUE #405: Fix broken registration VLAN handling when registrationVLAN is not integer.
See PR #404 . 

VLAN ids can be any string. 
We should have more tests for that case.

louismunro:
## Description

Fixes a case where unregistered nodes are sent to getNormalVlan when the registration VLAN is not an integer.
fixes #405 .
## Impacts

VLAN ids can be any string, not just integers (as per the RFC). 
The old code breaks if registrationVLAN is set to anything other than an integer. 
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
- Fixes a case where nodes are not sent to registration VLAN when that VLAN id is not numeric.
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -117,7 +117,7 @@ sub fetchVlanForNode {
 
     # there were no violation, now onto registration handling
     ($registration,$role) = $this->getRegistrationVlan($switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $realm, $stripped_user_name, $autoreg);
-    if (defined($registration) && $registration ne 0) {
+    if (defined($registration) && $registration ne "0") {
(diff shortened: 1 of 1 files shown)


lib/pf/vlan.pm
```diff
@@ -108,7 +108,7 @@ sub fetchVlanForNode {
 
     # there were no violation, now onto registration handling
     ($registration,$role) = $this->getRegistrationVlan($switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $realm, $stripped_user_name, $autoreg);
-    if (defined($registration) && $registration != 0) {
+    if (defined($registration) && $registration ne 0) {
```
jrouzierinverse: I would change it to $registration ne "0" so it is clearer that we are comparing a string.

PR #410, opened 2015-03-05
Restricts the allowed ciphers to prevent FREAK SSL attack.

### LINKED ISSUE #411: Prevent SSL FREAK Attack.
Our current httpd SSL config may be vulnerable to CVE-2015-0204.
Fix the allowed ciphers to prevent a downgrade to vulnerable ciphers. 

See PR #410 .

louismunro:
## Description

Restricts the allowed ciphers to prevent FREAK SSL attack (CVE-2015-0204).
Turns on HSTS.

Fixes #411 .
## Impacts

This restricts what ciphers can be used on TLS webservers.
Note that it will prevent legacy browsers from working (i.e. IE 7 and older). 

Please test that all TLS using services still work. 
## NEWS file entries

Enhancements
++++++++++++
- Http server will no longer allow downgrade to legacy "export" cipher suites and will not accept SSLv3.
## Delete branch after merge

YES

### DIFF
--- a/conf/httpd.conf.d/httpd.aaa
+++ b/conf/httpd.conf.d/httpd.aaa
@@ -29,6 +29,9 @@
   <IfModule !mod_status.c>
     LoadModule status_module /usr/lib/apache2/modules/mod_status.so
   </IfModule>
+  <IfModule !mod_headers.c>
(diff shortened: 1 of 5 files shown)


louismunro:
Will rebase and merge in v4.7 considering the impacts.

PR #413, opened 2015-03-09
Add new admin role USERS_CREATE_MULTIPLE to restrict creating multiple users from the admin gui

jrouzierinverse:
## Description

Add new admin role USERS_CREATE_MULTIPLE to restrict creating multiple users from the admin gui
## Impacts

The admin gui user creation
## NEWS file entries

Enhancements
++++++++++++
- New admin role action USERS_CREATE_MULTIPLE for more fine grain control of the admin GUI
## Delete branch after merge

NO


jrouzierinverse:
Please Review

html/pfappserver/lib/pfappserver/Form/Config/AdminRoles.pm
```diff
@@ -90,7 +90,7 @@ sub options_actions {
     my @options;
 
     map {
```
louismunro: Please use a foreach instead of map.
It's easier to read and less prone to breaking when we later modify the code.

Or better yet, stick it into a subroutine. See PBP 6.13.

louismunro:
Other than that one comment about map, it looks good to me.

jrouzierinverse:
Please review

jrouzierinverse:
Please review

html/pfappserver/lib/pfappserver/Form/Config/AdminRoles.pm
```diff
@@ -89,12 +91,11 @@ sub options_actions {
 
     my %groups;
     my @options;
-
-    map {
```
julsemaan, line 93: You really want this pull request merged from what I see.
1 less ! yay !

julsemaan:
Looks good merging

html/pfappserver/lib/pfappserver/Form/Config/AdminRoles.pm
jrouzierinverse, line 93: There is no problem with using map.
It was not used properly in this case

PR #422, opened 2015-03-13
ConfigStore cleanup in lib/

julsemaan:
## Description

This reworks the ConfigStore access in lib/ for modules that only do read access. They will instead use the configuration hashes.
## Impacts

None for the flow. Mostly internal reworking
## NEWS file entries

Not required since it's internal reworking
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/iptables.pm
+++ b/lib/pf/iptables.pm
@@ -638,32 +638,32 @@ sub generate_interception_rules {
 
 sub generate_provisioning_passthroughs {
     my $logger = Log::Log4perl::get_logger('pf::iptables');
-    foreach my $config (search_hash(\%ConfigProvisioning, type => 'sepm')) {
+    foreach my $config (tied(%ConfigProvisioning)->search(type => 'sepm')) {
         $logger->info("Adding passthrough for Symantec Endpoint Manager");
-        my $cmd = "LANG=C sudo ipset --add pfsession_passthrough $config->{'host'},8014 2>&1";
+        my $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{'host'},8014 2>&1");
         my @lines  = pf_run($cmd); 
     }
 
-    foreach my $config (search_hash(\%ConfigProvisioning, type => 'mobileiron')) {
+    foreach my $config (tied(%ConfigProvisioning)->search(type => 'mobileiron')) {
         $logger->info("Adding passthrough for MobileIron");
         # Allow the host for the onboarding of devices
-        my $cmd = "LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},$config->{boarding_port} 2>&1";
+        my $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},$config->{boarding_port} 2>&1");
         my @lines  = pf_run($cmd); 
         # Allow http communication with the MobileIron server
-        $cmd = "LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},80 2>&1";
+        $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},80 2>&1");
         @lines  = pf_run($cmd); 
         # Allow https communication with the MobileIron server
-        $cmd = "LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},443 2>&1";
+        $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},443 2>&1");
         @lines  = pf_run($cmd); 
     }
 
-    foreach my $config (search_hash(\%ConfigProvisioning, type => 'opswat')) {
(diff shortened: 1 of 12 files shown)


lib/pf/iptables.pm
```diff
@@ -639,32 +638,32 @@ sub generate_interception_rules {
 
 sub generate_provisioning_passthroughs {
     my $logger = Log::Log4perl::get_logger('pf::iptables');
-    foreach my $config (pf::ConfigStore::Provisioning->new->search(type => 'sepm')) {
+    foreach my $config (tied(%ConfigProvisioning)->search(type => 'sepm')) {
```
jrouzierinverse, line 641: This looks like it will fail pfconfig::cached_hash has no search function

lib/pf/iptables.pm
```diff
@@ -639,32 +638,32 @@ sub generate_interception_rules {
...
         $logger->info("Adding passthrough for Symantec Endpoint Manager");
-        my $cmd = "LANG=C sudo ipset --add pfsession_passthrough $config->{'host'},8014 2>&1";
+        my $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{'host'},8014 2>&1");
         my @lines  = pf_run($cmd); 
     }
 
-    foreach my $config (pf::ConfigStore::Provisioning->new->search(type => 'mobileiron')) {
+    foreach my $config (tied(%ConfigProvisioning)->search(type => 'mobileiron')) {
```
jrouzierinverse, line 647: This looks like it will fail pfconfig::cached_hash has no search function

lib/pf/iptables.pm
```diff
@@ -639,32 +638,32 @@ sub generate_interception_rules {
...
+        $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},80 2>&1");
         @lines  = pf_run($cmd); 
         # Allow https communication with the MobileIron server
-        $cmd = "LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},443 2>&1";
+        $cmd = untaint_chain("LANG=C sudo ipset --add pfsession_passthrough $config->{boarding_host},443 2>&1");
         @lines  = pf_run($cmd); 
     }
 
-    foreach my $config (pf::ConfigStore::Provisioning->new->search(type => 'opswat')) {
+    foreach my $config (tied(%ConfigProvisioning)->search(type => 'opswat')) {
```
jrouzierinverse, line 660: This looks like it will fail pfconfig::cached_hash has no search function

lib/pf/iptables.pm
jrouzierinverse, line 641: Ignore I was wrong did not get that far in the commit

lib/pf/iptables.pm
jrouzierinverse, line 660: Ignore I was wrong did not get that far in the commit

lib/pf/iptables.pm
jrouzierinverse, line 647: Ignore I was wrong did not get that far in the commit

lib/pfconfig/cached_hash.pm
```diff
@@ -177,6 +177,46 @@ sub EXISTS {
...
 
+=head2 values
+
+Added method that can be called on the underlying object of the tied hash
+Will return all the values of the hash. Mostly for internal use
+Call it using tied(%hash)->values
+
+=cut
+
+sub values {
```
jrouzierinverse, line 188: You should push this to the pfconfig it will be faster there

lib/pf/util.pm
```diff
@@ -928,6 +929,27 @@ sub normalize_time {
     }
 }
 
+=item search_hash
```
jrouzierinverse, line 932: Is this even being used?

jrouzierinverse:
Looks good to me

lib/pf/util.pm
julsemaan, line 932: No but I taught it's a useful function. It could be removed

lib/pfconfig/cached_hash.pm
julsemaan, line 188: You're right but the thing is using FETCH uses the subcache. This means that next calls to values will be done without poking pfconfig. Also, the subcache may have already been built so it'll be faster this way IMO. I could rewrite it so it pokes pfconfig and builds the subcache with the result but i don't think it'll do a big difference for the code it adds

lib/pf/iptables.pm
julsemaan, line 660: Zammit, did you hack the Github account of James ?

lib/pfconfig/cached_hash.pm
rouzier, line 188: You are right also I realized sending all values could be very expensive as the config size grows.
And since pfconfig is only a single process at the moment it will block all other processes

lib/pf/util.pm
rouzier, line 932: taught who were you teaching?

julsemaan, line 932: Zammit, who else ?

PR #433, opened 2015-03-26
Active/active clustering

julsemaan:
## Description

Adds active/active clustering using haproxy, keepalive, freeradius load balancing and the isc dhcp failover feature.
## Impacts

So much, this makes one of the servers run pfmon and pfdhcplistener. The other ones don't. 
It load balances the http connection
Add to list....
## NEWS file entries

New Features
++++++++++++
- Added active/active clustering support
## Delete branch after merge

NO


bin/cluster/management_update
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+Script to call when there a change in the management state
```
extrafu: there is a change

conf/documentation.conf
```diff
@@ -1274,6 +1308,24 @@ description=<<EOT
...
+[maintenance.cluster_sync_interval]
+type=time
+description=<<EOT
+Interval at which PacketFence runs the sync between all the members of the cluster.
+EOT
+
+[active_active.members]
+type=text
+description=<<EOT
+List of ip separate by a comma of all the cluster members.
```
extrafu: of IP separated

conf/documentation.conf
```diff
@@ -1274,6 +1308,24 @@ description=<<EOT
...
+EOT
+
+[active_active.password]
+type=text
+description=<<EOT
+Shared KEY for vrrp protocol (Must be the same on all members).
```
extrafu: VRRP

extrafu: must be the

conf/iptables.conf.example
```diff
@@ -34,11 +37,22 @@
 -A input-management-if --protocol tcp --match tcp --dport 9392 --jump ACCEPT
 # Nessus Administration Interface
 -A input-management-if --protocol tcp --match tcp --dport 8834 --jump ACCEPT
+# Galera cluster
```
extrafu: Useful?

conf/pf.conf.defaults
```diff
@@ -1000,6 +1025,23 @@ provisioning_compliance_poll_interval=300s
...
+# maintenance.cluster_sync_interval
+#
+# Interval at which PacketFence runs the sync between all the members of the cluster
+cluster_sync_interval=60s
+
+[active_active]
+#
+# active_active.members
+#
+# List of ip separate by a comma of all the cluster members
```
extrafu: Same typo

conf/pf.conf.defaults
```diff
@@ -1000,6 +1025,23 @@ provisioning_compliance_poll_interval=300s
...
+members=
+#
+# active_active.password
+#
+# Shared KEY for vrrp protocol (Must be the same on all members).
```
extrafu, line 1032: Same typos

html/pfappserver/lib/pfappserver/Form/Field/IPAddresses.pm
```diff
@@ -0,0 +1,77 @@
...
+         my ( $value, $field ) = @_;
+         return $field->get_message('ipv4');
+     },
+    }
+   ]
+  );
+
+=head1 COPYRIGHT
+
+Copyright (C) 2013 Inverse inc.
```
extrafu: You weren't even born!

lib/pf/api.pm
```diff
@@ -439,6 +451,130 @@ sub node_information : Public {
...
+        my %data = (
+            namespace => $postdata{namespace},
+            light => 1
+        );
+        eval {
+            $apiclient->call('expire', %data ); 
+        };
+
+        if($@){
+            $logger->error("Can't connect to $server->{management_ip} to expire the configuration. $@")
```
extrafu: why continue if we can't connect?

lib/pf/api.pm
```diff
@@ -439,6 +451,130 @@ sub node_information : Public {
...
+    else {
+        my $all = $postdata{namespace} eq "__all__" ? 1 : 0;
+        if($all){
+            pfconfig::manager->new->expire_all();
+        }
+        else{
+            pfconfig::manager->new->expire($postdata{namespace});
+        }
+    }
+    # refactor me, it's most probably not always a success
```
extrafu: Do it!

lib/pf/clustermgmt.pm
```diff
@@ -0,0 +1,230 @@
...
+}
+
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
extrafu: You were in your mom's belly.

lib/pf/services/manager/haproxy.pm
```diff
@@ -0,0 +1,163 @@
...
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
extrafu: Almost out!

lib/pf/services/manager/keepalived.pm
```diff
@@ -0,0 +1,141 @@
...
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2014 Inverse inc.
```
extrafu: Push, push, harder!

lib/pfconfig/namespaces/config/Cluster.pm
```diff
@@ -0,0 +1,97 @@
...
+
+=back
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2015 Inverse inc.
```
extrafu, line 71: _POP_  Yay!

conf/iptables.conf.example
julsemaan: Well we're still not sure about what we'll use. If we drop Galera for sure then it can be removed

lib/pf/api.pm
julsemaan: That logging message was bad. This means an error occured during the request so the next step may work if the issue is not network related. I'll fix the logging

bin/pfupdate
```diff
@@ -0,0 +1,23 @@
...
+my $mode;
+my $logger = get_logger;
+
+exit;
+
+$logger->info("Got a notification from keepalived");
+
+$logger->info("Refreshing services for management mode");
+$logger->info(`/usr/local/pf/bin/pfcmd service pfdhcplistener restart`);
+$logger->info(`/usr/local/pf/bin/pfcmd service pfmon restart`);
```
jrouzierinverse: What does this do?

html/pfappserver/lib/pfappserver/Form/Field/IPAddresses.pm
```diff
@@ -0,0 +1,77 @@
+package pfappserver::Form::Field::IPAddresses;
```
jrouzierinverse, line 1: Why?

fdurand, line 1: because we want multiples ip address not just one

lib/pf/cluster.pm
```diff
@@ -0,0 +1,126 @@
+package pf::cluster;
```
jrouzierinverse, line 1: Pod docs??

bin/cluster/management_update
```diff
@@ -0,0 +1,75 @@
...
+
+$logger->info("Refreshing services for management mode");
+
+my $vrrp_ip = $ARGV[1];
+my $state = $ARGV[2];
+my $priority = $ARGV[3];
+
+if($state eq "MASTER"){
+    $logger->info("Transition to master state, starting management services.");
+    `/usr/local/pf/bin/pfcmd service pfdhcplistener start`;
```
fdurand, line 32: Just a question, before we sent a kill -HUP to pfdhcplistener to reload his config, the goal was to keep pfdhcplistener running on all the member and just change on the fly if the pfdhcplistener must capture traffic or no.
I just prefer to have all the services running on all the members instead of trying to start/stop each service.
The other question, is what happen if i stop the reg interface on the master server, will it stop the pfdhcplistener for all the interfaces ?

julsemaan, line 32: I agree it's more multi-master to do it this way but it adds complexity to the code especially in pfmon. It required to add the pfmon timers to the configuration overlay specific to each server and would call the expiration on them to sync on each cluster change. This makes it longer I agree but it gives a general way of handling all the services (through pf::services::manager::isEnabled) instead of adapting each service to temporarly stop while it's not master. Answer to your other question in the next comment

julsemaan, line 32: And then in the implementation I adapted, the master server is the master of all interfaces in keepalive so it will own all the registration interfaces. This means all pfdhcplistener is listening on all interfaces.

lib/pf/cluster.pm
julsemaan, line 1: Done

bin/pfupdate
julsemaan: This restarts pfmon and pfdhcplistener but it not useful anymore since it's managed by bin/cluster/management_update. I'll remove it

PR #446, opened 2015-04-07
pfmon should unregister nodes in pending state when unregdate is reached.

louismunro:
## Description

pfmon needs to check nodes that are in pending state to prevent them from staying in that state indefinitely if the user never actually registers.
## Impacts

Changes the maintenance sql run by pfmon so that pending nodes are also unregistered if their time is up.
## NEWS file entries

Bug Fixes
+++++++++

Devices can no longer remain in "pending" state indefinitely.
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/node.pm
+++ b/lib/pf/node.pm
@@ -296,7 +296,9 @@ sub node_db_prepare {
     ];
 
     $node_statements->{'node_expire_unreg_field_sql'} = get_db_handle()->prepare(
-        qq [ select mac from node where status="reg" and unregdate != 0 and unregdate < now() ]);
+        qq [ select mac from node where ( 
+                status="reg" and unregdate != 0 and unregdate < now() ) or (
+                status="pending" and unregdate < now() ) ]);
 
     $node_statements->{'node_expire_lastarp_sql'} = get_db_handle()->prepare(
         qq [ select mac from node where unix_timestamp(last_arp) < (unix_timestamp(now()) - ?) and last_arp!=0 ]);
--- a/lib/pf/node.pm
+++ b/lib/pf/node.pm
@@ -296,9 +296,9 @@ sub node_db_prepare {
     ];
 
     $node_statements->{'node_expire_unreg_field_sql'} = get_db_handle()->prepare(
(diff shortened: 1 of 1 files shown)


lib/pf/node.pm
```diff
@@ -296,7 +296,9 @@ sub node_db_prepare {
     ];
 
     $node_statements->{'node_expire_unreg_field_sql'} = get_db_handle()->prepare(
-        qq [ select mac from node where status="reg" and unregdate != 0 and unregdate < now() ]);
+        qq [ select mac from node where ( 
+                status="reg" and unregdate != 0 and unregdate < now() ) or (
```
julsemaan: i feel very picky but can you put the opening parenthesis on the line below and same for the line above. That would make it clearer to read IMO

lib/pf/node.pm
```diff
@@ -296,7 +296,9 @@ sub node_db_prepare {
     ];
 
     $node_statements->{'node_expire_unreg_field_sql'} = get_db_handle()->prepare(
-        qq [ select mac from node where status="reg" and unregdate != 0 and unregdate < now() ]);
```
jrouzierinverse, line 299: would'nt be easier to just say status != 'unreg'

lib/pf/node.pm
louismunro: Done.  

## 

LM

lib/pf/node.pm
louismunro, line 299: It would be easier, but it may not be the same thing.  

## 

LM

louismunro:
Review me pleeeeease....

sinusoidal:
How often is this pfmon cleanup run?  I still seem to have devices stuck in the pending state on 5.1 several days after requesting a pin (but never using it).  I have a workaround script running at the moment which switches any devices not registered within an hour back to unreg (as in my text message based system I assume anyone requesting a pin who has not registered after an hour isn't going to) but not ideal.

sinusoidal:
Ah, has this not actually been pushed to stable yet? Apologies if so.

PR #449, opened 2015-04-07
Changed random source to  Bytes::Random::Secure.

louismunro:
For testing only.


fdurand:
Tested and prevent  core dump with httpd.admin and httpd.portal

PR #450, opened 2015-04-08
Allow HTTP and HTTPS in iptables on management interface

julsemaan:
## Description

Allows access to the captive portal on the management interface by default for web authentication, status page, etc
## Impacts

Allows access to the captive portal from the production network by default
## NEWS file entries

None required I think
## Delete branch after merge

YES


julsemaan:
Discussed this with @fdurand and it seems to make sense.
I'd very like this to be in pf5
Anybody has objections ?

dwlfrth:
Looks good to me.

PR #456, opened 2015-04-09
Will only display the print and email password buttons if password is plaintext or right after resetting the passwords.

jrouzierinverse:
## Description

Will only display the print and email password buttons if password is plaintext or right after resetting the passwords.
## Impacts

Httpd admin
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/password.pm
+++ b/lib/pf/password.pm
@@ -473,7 +473,7 @@ extract the type of hash for the password
 sub password_get_hash_type {
     my ($passwd) = @_;
     my $type  = 'plaintext';
-    if ($passwd =~ /\{([^{}]*)\}/ ) {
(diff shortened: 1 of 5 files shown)


lib/pf/password.pm
```diff
@@ -464,6 +464,21 @@ sub _check_password {
...
+=item password_get_hash_type
+
+extract the type of hash for the password
+
+=cut
+
+sub password_get_hash_type {
+    my ($passwd) = @_;
+    my $type  = 'plaintext';
+    if ($passwd =~ /\{([^{}]*)\}/ ) {
```
louismunro: Should anchor this. I.e. /^{ .../

jrouzierinverse: Done

louismunro:
Works as advertised.
Merging.

PR #457, opened 2015-04-09
Checkup should only warn that there are no internal networks

julsemaan:
## Description

There are cases where we do not want internal networks like in a pure web auth setup. Adding fake devices makes clustering configuration a pain
## Impacts

Allows PacketFence to start without any internal networks
## NEWS file entries

Enhancements
++++++++++++
- PacketFence will accept to start even if there are no internal networks.
## Delete branch after merge

YES


julsemaan:
Anybody has objections ? I'd like this to be in PF5 so clustering doesn't become a mess with web auth customers

louismunro:
Forgiveness > Permission.

## 

LM

dwlfrth:
Let's do this!

PR #481, opened 2015-04-16
Will now match on both source and action instead of just source

jrouzierinverse:
## Description

Will now match on both source and action instead of just source
## Impacts

Matching rules in sources
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Allow rules to be match on both source and action

Bug Fixes
+++++++++
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -259,29 +259,26 @@ sub match {
             @sources = ($source);
         }
     }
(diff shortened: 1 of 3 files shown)


jrouzierinverse:
Need Unit tests

fdurand:
I like this logic

jrouzierinverse:
Review test are in t/authentication.t

PR #484, opened 2015-04-17
Use the host header if define to redirect the device after the registration instead of the fqdn of packetfence

fdurand:
## Description

At the end of the portal profile if we use https portal then the http redirection will use the host header instead of packetfence fqdn. (Usefull for external captive portal)  
## Impacts

NO
## Delete branch after merge

YES
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Adapt the end process redirection on the portal


whitx:
Tested with Win 7 and Android phone, worked fine on both.

PR #499, opened 2015-04-22
Stopped pfsetvlan+snmptrapd by default

### LINKED ISSUE #489: pfsetvlan should not start by default.
It should be started only if required.
Most people don't need it.

julsemaan:
## Description

Stops pfsetvlan and snmptrapd by default in PacketFence
## Impacts

Will stop eating RAM for no reason
## Issue

fixes #489 
## Delete branch after merge

YES
## NEWS file entries

Enhancements
++++++++++++
- pfsetvlan and snmptrapd are now stopped by default


extrafu:
Added zammitbug label, because rebooting usually clears up the RAM usage.

PR #505, opened 2015-04-23
Show mandatory fields when you select them on the portal profile

fdurand:
## Description

Show mandatory fields when they has been selected on the portal profile
## Impacts

No impacts
## Delete branch after merge

YES
## NEWS file entries

Enhancements
++++++++++++
- Show mandatory fields on the captive portal

### DIFF
--- a/html/captive-portal/templates/login.html
+++ b/html/captive-portal/templates/login.html
@@ -70,8 +70,15 @@ document.observe("dom:loaded", function() {
           <div class="input">
             <span>[% i18n("Password") %]</span><input class="field" name="password" type="password" />
           </div>
+
           [% END %]
-          
+          [% FOREACH field IN mandatory_fields %]
+          <div class="input">
+            <span>[% i18n("$field") %]</span>
+            <input class="field" name="[%field%]" type="text" value="" /><br/>
+          </div>
+          [% END %]
+
           [%# submit %]
           <div class="input">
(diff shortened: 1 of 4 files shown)


html/captive-portal/templates/login.html
```diff
@@ -70,8 +70,15 @@
           <div class="input">
             <span>[% i18n("Password") %]</span><input class="field" name="password" type="password" />
           </div>
+
           [% END %]
-          
+          [% FOREACH field IN mandatory_fields %]
+          <div class="input">
+            <span>[% i18n("$field") %]</span>
+            <input class="field" name="[%field%]" type="text" value="" /><br/>
```
julsemaan, line 78: Does this mean the user registering with user/password has to fill in this form ?

fdurand, line 78: Yes but you know what i forgot to add the code behind that, so it just appear in the form but do nothing ;-)

julsemaan, line 78: Not sure it's worth having that though. If you're using guest + AD on the same profile then your AD users have to fill-in this form (which is not that useful)

fdurand, line 78: Yes you are true, i just have a case with a local auth source and i need to have mandatory field email.
So i have probably have to add a checkbox in the auth source configuration to ask if we have to take care of mandatory fields on the portal profile or no.

PR #506, opened 2015-04-24
Use a binary protocol for receiving sereal data from pfconfig

jrouzierinverse:
## Description

Use a binary protocol for receiving sereal data for pfconfig.
This was done to improve the speed in which we send and retrieve data
## Impacts

pfconfig
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Improve the pfconfig protocol

Bug Fixes
+++++++++
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/util/networking.pm
+++ b/lib/pf/util/networking.pm
@@ -1,0 +1,101 @@
+package pf::util::networking;
+=head1 NAME
+
+pf::util::networking
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::util::networking
+
+=cut
+
+use strict;
+use warnings;
+use Errno qw(EINTR EAGAIN);
+
+use base qw(Exporter);
+
+our @EXPORT_OK = qw(syswrite_all sysread_all);
+
+=head2 syswrite_all
+
+A wrapper around syswrite to ensure all bytes are written to a socket
+
+=cut
+
+sub syswrite_all {
+    my ($socket,$buffer) = @_;
+    my $bytes_to_send = length $buffer;
+    my $offset = 0;
+    my $total_written = 0;
+    while ($bytes_to_send) {
+        my $bytes_sent = syswrite($socket, $buffer, $bytes_to_send, $offset);
+        unless (defined $bytes_sent) {
+            next if $! == EINTR;
+            last;
+        }
+        return $offset unless $bytes_sent;
+        $offset += $bytes_sent;
+        $bytes_to_send -= $bytes_sent;
+    }
+    return $offset;
+}
+
+=head2 sysread_all
+
+A wrapper around sysread to ensure all bytes are read from a socket
+
+=cut
+
+
+sub sysread_all {
+    my $socket = $_[0];
+    our $buf;
+    local *buf = \$_[1];    # alias
+    my $bytes_to_read = $_[2];
+    my $offset = 0;
+    $buf = '';
+    while($bytes_to_read) {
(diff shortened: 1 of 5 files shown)


louismunro:
This makes debugging significantly harder. 
There are operational issues to be addressed here. 
I would like to discuss tools to be able to read the pfconfig requests. 
At the moment strace is our (only) friend.

rouzier:
Right now debugging pfconfig kind of hard because send a request using json (ok this is easy to read) but then the response is a string with the number of lines followed by sereal data (which is binary).

I would prefer to have a simple binary protocol 4 byte length followed by data.

rouzier:
If a text based is prefered I would prefer the following format.
BYTES_IN_ASCII NEWLINE
binary_data (length of data is BYTES_IN_ASCII)

louismunro:
Actually the issue really is one of tooling.
If we had tools to read input/output to pfconfig the issue of text vs binary would be moot...

jrouzierinverse:
Please review

louismunro:
Please give us some debugging tools and we'll merge it.

jrouzierinverse:
Would a script to parse the packet data be good enough?

louismunro:
@julsemaan What do you think of the proposed tool?

t/unittest/util/networking.t
```diff
@@ -0,0 +1,116 @@
+=head1 NAME
+
+example pf test
```
julsemaan: Not an example

t/unittest/util/networking.t
```diff
@@ -0,0 +1,116 @@
...
+
+=cut
+
+=head1 DESCRIPTION
+
+example pf test script
```
julsemaan: here too

t/unittest/util/networking.t
```diff
@@ -0,0 +1,116 @@
...
+
+=cut
+
+use strict;
+use warnings;
+#
```
julsemaan: useful ?

lib/pfconfig/util.pm
```diff
@@ -34,55 +35,18 @@ our @EXPORT_OK = qw(
...
-    # We workaround this by deactivating warnings when we read though the socket
-    no warnings;
-    while ( $line_read < $count ) {
-        chomp( $line = <$socket> );
-        $response .= $line . "\n";
-        $line_read += 1;
-    }
-    use warnings;
-    return $response;
+    $payload .= "\n";
```
julsemaan, line 46: remove this, payload doesn't need to include this, it was only because it was reading lines

lib/pf/util/networking.pm
```diff
@@ -0,0 +1,129 @@
...
+=cut
+
+sub send_data_with_length {
+    my ($socket,$data) = @_;
+    #Packing data with 32 bit little endian int as it length
+    my $packed_data = pack("V/a*",$data);
+    my $bytes_to_send = length $packed_data;
+    my $bytes_sent = syswrite_all($socket, $packed_data);
+    if(defined $bytes_sent && $bytes_sent > 4) {
+        #substracting the four bytes appened to the begining
```
julsemaan: appended

lib/pfconfig/util.pm
julsemaan, line 46: my bad, read this too quick. Forget about it

jrouzierinverse:
Please re-review

PR #516, opened 2015-04-28
Show mandatory fields when you select them on the portal profile

fdurand:
## Description

Show mandatory fields when they have been selected on the portal profile
Added an config parameter in authentication sources (LDAP, AD, Htpasswd, Kerberos,Radius) to make the parameter mandatory or no. (if you have a guest auth source per example)
Update the person table with mandatory field.
## Impacts

No impacts
## Delete branch after merge

YES
## NEWS file entries

Enhancements
++++++++++++
- Show mandatory fields on the captive portal and update the person table with them.

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -10,7 +10,7 @@ use pf::util;
 use pf::locationlog;
 use pf::authentication;
 use HTML::Entities;
-use List::MoreUtils qw(any);
+use List::MoreUtils qw(any uniq);
 use pf::config;
 use pf::person qw(person_modify);
 
@@ -582,7 +582,7 @@ sub showLogin : Private {
     # Make sure mandatory fields are unique
     @mandatory_fields = uniq @mandatory_fields;
 
-    $c->stash( mandatory_fields => @mandatory_fields;
+    $c->stash( mandatory_fields => \@mandatory_fields );
 }
 
 sub _clean_username {
@@ -618,6 +618,7 @@ sub validateMandatoryFields : Private {
     # Make sure mandatory fields are unique
     @mandatory_fields = uniq @mandatory_fields;
 
+    my %mandatory_fields = map { $_ => undef } @mandatory_fields;
     my @missing_fields = grep { !$request->param($_) } @mandatory_fields;
 
     if (@missing_fields) {
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -565,7 +565,7 @@ sub showLogin : Private {
         oauth2_win_live => is_in_list( $SELFREG_MODE_WIN_LIVE, $guestModes ),
         guest_allowed   => $guest_allowed,
     );
-    $c->stash( mandatory_fields => $profile->getMandatoryFields) if (defined(@sources) && grep { $_->{use_mandatory_fields} eq 'yes' } @sources);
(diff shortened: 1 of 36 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
```diff
@@ -12,7 +12,9 @@ use pf::authentication;
 use HTML::Entities;
 use List::MoreUtils qw(any);
 use pf::config;
+use pf::person qw(person_modify @FIELDS);
```
dwlfrth: There's no need to import '@FIELDS' since you use the complete name in line 17 and it is the right way to do :)

html/captive-portal/lib/captiveportal/PacketFence/Controller/Signup.pm
```diff
@@ -22,7 +22,7 @@ use POSIX;
 use URI::Escape::XS qw(uri_escape);
 use pf::iplog;
 use pf::node;
-use pf::person qw(person_modify);
+use pf::person qw(person_modify @FIELDS);
```
dwlfrth: There's no need to import '@FIELDS' since you use the complete name in line 34 and it is the right way to do :)

lib/pf/person.pm
```diff
@@ -38,7 +38,7 @@ BEGIN {
         person_violations
         person_custom_search
     );
-    @EXPORT_OK = qw( $PID_RE );
+    @EXPORT_OK = qw( $PID_RE @FIELDS );
```
dwlfrth: No need to export '@FIELDS' since you use the complete package name each time you use it.
We also try to avoid exporting only for exporting :)

dwlfrth:
'use_mandatory_fields' option should be configured for all the authentication sources (internal, external, exclusives) except for OAuth (at the moment (see issue #524))

dwlfrth:
Should remove anything related to SQL (for the moment... we need to talk about this) because we can't edit that source (as I said, need to talk about this.)
Hint: lib/pf/Authentication/Source/SQLSource.pm

dwlfrth:
Shouldn't the 'use_mandatory_field' authentication source object attribute be a BOOLEAN with a default of '0' rather than a 'STR' ?

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/SponsorEmail.pm
```diff
@@ -40,6 +40,18 @@ has_field 'create_local_account' => (
     },
 );
 
+has_field 'use_mandatory_fields' => (
+    type => 'Toggle',
+    checkbox_value => 'yes',
+    unchecked_value => 'no',
+    label => 'Use mandatory fields',
+    default => pf::Authentication::Source::EmailSource->meta->get_attribute('use_mandatory_fields')->default,
```
dwlfrth: Shouldn't it be 
default => pf::Authentication::Source::SponsorEmailSource->meta ...

fdurand: Yes it's probably true and we probably have to change it in all this file.

dwlfrth: Let's do it, there's 3 instances.

html/captive-portal/templates/login.html
```diff
@@ -70,8 +73,15 @@
           <div class="input">
             <span>[% i18n("Password") %]</span><input class="field" name="password" type="password" />
           </div>
+
           [% END %]
-          
+          [% FOREACH field IN mandatory_fields %]
+          <div class="input">
+            <span>[% i18n("$field") %]</span>
+            <input class="field" name="[% field %]" type="text" value="" /><br/>
```
dwlfrth: value should be the value of the field. This way, if there's an error, the field will not be blank and will be populated by the value entered

dwlfrth:
Review me... Works "as is" for the moment..

PR #517, opened 2015-04-28
Switch range definition in switch configuration

fdurand:
## Description

It's now possible to add a switch range (cird format) for the switch ip address configuration parameter.
It's usefull if on the same network you have multiples switch/ap with the same configuration, you just have to define one.
## Impacts

No impacts
## Delete branch after merge

YES
## NEWS file entries

New Features
+++++++++++
- Define range of network switches in switch configuration

### DIFF
--- a/lib/pf/SwitchFactory.pm
+++ b/lib/pf/SwitchFactory.pm
@@ -87,7 +87,7 @@ sub instantiate {
             last;
         }
     }
-    if ($requestedSwitch) {
+    if (!$requestedSwitch) {
         foreach my $search (@requestedSwitches){
             next if (valid_mac($search));
             foreach my $switch ( keys (%SwitchConfig) ) {
--- a/lib/pf/SwitchFactory.pm
+++ b/lib/pf/SwitchFactory.pm
@@ -80,16 +80,14 @@ sub instantiate {
     });
 
     my $switch_data;
-    my $i = 1;
     foreach my $search (@requestedSwitches){
         if($SwitchConfig{$search}){
             $requestedSwitch = $search;
             $switch_data = $SwitchConfig{$search};
-            $i = 0;
             last;
         }
     }
-    if ($i) {
+    if ($requestedSwitch) {
         foreach my $search (@requestedSwitches){
             next if (valid_mac($search));
             foreach my $switch ( keys (%SwitchConfig) ) {
--- a/lib/pf/SwitchFactory.pm
+++ b/lib/pf/SwitchFactory.pm
@@ -90,10 +90,11 @@ sub instantiate {
         }
     }
     if (!$requestedSwitch) {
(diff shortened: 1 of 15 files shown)


lib/pf/SwitchFactory.pm
```diff
@@ -79,13 +80,30 @@ sub instantiate {
     });
 
     my $switch_data;
+    my $i = 1;
```
rouzier: Remove

lib/pf/SwitchFactory.pm
```diff
@@ -79,13 +80,30 @@ sub instantiate {
...
     foreach my $search (@requestedSwitches){
         if($SwitchConfig{$search}){
             $requestedSwitch = $search;
             $switch_data = $SwitchConfig{$search};
+            $i = 0;
```
jrouzierinverse: Remove

lib/pf/SwitchFactory.pm
```diff
@@ -79,13 +80,30 @@ sub instantiate {
...
             last;
         }
     }
+    if ($i) {
```
jrouzierinverse: Just check if $requestedSwitch is set

lib/pf/util.pm
```diff
@@ -870,10 +872,20 @@ sub valid_mac_or_ip {
         my ($mac) = clean_mac($mac_or_ip);
         return 1 if($mac && $mac !~ $NON_VALID_MAC_REGEX && $mac =~ $VALID_PF_MAC_REGEX);
     }
-    get_logger()->error("invalid MAC or IP: $mac_or_ip");
+    get_logger()->warn("invalid MAC or IP: $mac_or_ip");
```
jrouzierinverse: This should be debug to reduce log noise

julsemaan:
An example in conf/switches.conf would be necessary IMO and maybe a note in the admin guide.

_Insert some joke that includes the word zammit here_

jrouzierinverse:
Add a test case in t/SwitchFactory.t

julsemaan:
Code looks good but as discussed it's missing the freeradius part

t/SwitchFactory.t
```diff
@@ -99,6 +99,15 @@ $switch = pf::SwitchFactory->instantiate({ switch_mac => "ff:01:01:01:01:04", sw
 isa_ok( $switch, 'pf::Switch::Cisco::Catalyst_2900XL' );
 is($switch->{_id}, '192.168.0.1', "Proper id is set");
 
+#Test using ip address in a range
+$switch = pf::SwitchFactory->instantiate('192.168.3.1');
+isa_ok($switch, 'pf::Switch::Cisco::Catalyst_2960');
+is($switch->{_id}, '192.168.3.1', "Proper id is set 192.168.3.1");
```
dwlfrth: Proper ID is set for

jrouzierinverse:
Please Review and test if dynamic controller is working

whitx:
Tested, the priority apply properly @mac ->@ip -> range IP.

Issue #509, opened 2015-04-27
Convert Email templates to HTML


cgx:
We have various [email templates](https://github.com/inverse-inc/packetfence/tree/devel/conf/templates) in plain text. We should update them to proper HTML emails.

As they are a PITA to style, we could get some inspiration from [these templates](http://blog.mailgun.com/transactional-html-email-templates/).

PR #527, opened 2015-05-05
Add email html template FR/EN

whitx:
# Description

Redefine email template, to use HTML emails instead of plaintext.
# Impacts

This should not have impact on the PacketFence,  try to open the html email with different client/webmail/device would be the best testing.
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Email are now in HTML format
## Enhancements
- Redefine email to be HTML instead of plaintext, email are now responsive and better looking.
- Added French version of the templates

Used #509 as a ref for the templates.

### DIFF
--- a/conf/templates/emails-billing_confirmation-fr.html
+++ b/conf/templates/emails-billing_confirmation-fr.html
@@ -1,0 +1,31 @@
+<link rel="stylesheet" type="text/css" href="/usr/local/pf/html/pfappserver/root/static/css/email.css">
+<span class="preheader">Informations de payement pour l'enregistrement.</span>
+<table class="main">
+        <tr>
+                <td class="wrapper">
+                        <table>
+                                <tr>
+                                        <td>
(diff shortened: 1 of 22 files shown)


extrafu:
I guess you should use a tab-stop of 2 spaces, 8 is pure insanity!

louismunro:
That's cause he's french. 
8 is metric for 2.

whitx:
biling
preregadmin
preregsponsor

Email preview.

dwlfrth:
I think we should avoid any "personal' formula in these kind of email. 
No "we", "nous" i mean...
ie: We have created you an account to access our network  -> An account has been created so you can access the network
ie: Nous vous avons créé un compte... -> Un compte a été créé pour vous permettre....

louismunro:
Mistakes were made…
(It’s called the impersonal voice. Great way to absolve anyone from anything).  

## 

LM

jrouzierinverse:
Mistakes were made but not by me

PR #539, opened 2015-05-12
DB schema version checkup modification

### LINKED ISSUE #538: DB schema version checkup should check for minor version
Currently, DB schema version checkup check the schema version and the pf-release file for the whole (major.minor.subminor) version.
That conflict with the release process that we discussed and with the "standard" thinking in which we avoid schema upgrade on sub-minors.

DB schema version should then check only the minor version (major.minor)

dwlfrth:
# Description

Modify the way we check for DB schema version. Checking for minor rather than sub-minor
# Impacts

Shouldn't have any since this fixes new code not yet released (Reason why there's no NEWS file entry)
# Issue

fixes #538 
# Delete branch after merge

YES

### DIFF
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -4,25 +4,13 @@
 
 SET @MAJOR_VERSION = 5;
 SET @MINOR_VERSION = 1;
-SET @PATCH_LEVEL = 0;
+SET @SUBMINOR_VERSION = 0;
 
 --
 -- The VERSION_INT to ensure proper ordering of the version in queries
 --
 
-SET @VERSION_INT = @MAJOR_VERSION << 16 | @MINOR_VERSION << 8 | @PATCH_LEVEL;
-
---
--- Table structure for table 'pf_version'
---
-
-CREATE TABLE pf_version ( `id` INT NOT NULL PRIMARY KEY, `version` VARCHAR(11) NOT NULL UNIQUE KEY);
-
---
--- Updating to current version
---
-
-INSERT INTO pf_version (id, version) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION, @PATCH_LEVEL));
+SET @VERSION_INT = @MAJOR_VERSION << 16 | @MINOR_VERSION << 8 | @SUBMINOR_VERSION;
 
 --
 -- Table structure for table `class`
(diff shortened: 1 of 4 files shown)


dwlfrth:
Review me please

db/pf-schema-X.Y.Z.sql
```diff
@@ -936,3 +924,15 @@ CREATE TABLE keyed (
...
+-- Table structure for table 'pf_version'
+--
+
+CREATE TABLE pf_version ( `id` INT NOT NULL PRIMARY KEY, `version` VARCHAR(11) NOT NULL UNIQUE KEY);
+
+--
+-- Updating to current version
+--
+
+INSERT INTO pf_version (id, version) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION, @SUBMINOR_VERSION));
```
jrouzierinverse, line 941: We should only store X.Y

lib/pf/version.pm
```diff
@@ -40,7 +40,7 @@ Initialize database prepared statements
 sub version_db_prepare {
 
     $version_statements->{'version_check_db_sql'} = get_db_handle()->prepare(qq[
-        SELECT version FROM pf_version WHERE version = ?;
+        SELECT version FROM pf_version WHERE version LIKE ?'%';
```
jrouzierinverse: The like allows the following to be true.
5.10 LIKE '5.1%'

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -412,3 +400,15 @@ CREATE TABLE `tagging_taggeditem` (
...
+-- Table structure for table 'pf_version'
+--
+
+CREATE TABLE pf_version ( `id` INT NOT NULL PRIMARY KEY, `version` VARCHAR(11) NOT NULL UNIQUE KEY);
+
+--
+-- Updating to current version
+--
+
+INSERT INTO pf_version (id, version) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION, @SUBMINOR_VERION));
```
jrouzierinverse: We should only store X.Y

jrouzierinverse: The will be created in the wrong database pf_graphite

dwlfrth:
Review me

jrouzierinverse:
Rebase and I will merge

db/pf-schema-X.Y.Z.sql
julsemaan, line 941: IMO I think we should store all but use what interests us. This way it will stay consistent with pf-release, currently-at, ...

PR #553, opened 2015-05-15
adjust httpd processes on startup

### LINKED ISSUE #547: RAM usage of new dashboard services is excessive
Needs to be optimized

julsemaan:
# Description

Adjust the started and spares servers for our httpd processes
# Impacts

Consumes less RAM
# Issue

fixes #547 
# Delete branch after merge

YES
# NEWS file entries

None required

### DIFF
--- a/conf/httpd.conf.d/httpd.graphite
+++ b/conf/httpd.conf.d/httpd.graphite
@@ -101,10 +101,10 @@ $KeepAliveTimeout = "10";
 $KeepAlive = "On";
 $MaxKeepAliveRequests = "100";
 
-$MaxClients = "56";
-$StartServers = "4";
-$MinSpareServers = "4";
-$MaxSpareServers = "20";
+$MaxClients = "2";
+$StartServers = "1";
+$MinSpareServers = "1";
+$MaxSpareServers = "1";
 
 $HostnameLookups = "off";
 $MaxRequestsPerChild = "1000";
--- a/conf/httpd.conf.d/httpd.aaa
+++ b/conf/httpd.conf.d/httpd.aaa
@@ -109,8 +109,8 @@ $Timeout = "50";
 $KeepAliveTimeout = "10";
 
 $MaxClients = "256";
-$StartServers = "10";
-$MinSpareServers = "3";
+$StartServers = "4";
+$MinSpareServers = "2";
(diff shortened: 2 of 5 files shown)


conf/httpd.conf.d/httpd.graphite
```diff
@@ -101,10 +101,10 @@ $KeepAliveTimeout = "10";
...
 $MaxKeepAliveRequests = "100";
 
-$MaxClients = "56";
-$StartServers = "4";
-$MinSpareServers = "4";
-$MaxSpareServers = "20";
+$MaxClients = "2";
+$StartServers = "1";
+$MinSpareServers = "1";
+$MaxSpareServers = "1";
```
louismunro, line 107: I would set the max higher than that. Something around 10.
The service is not only used in our dashboard but also by the other servers in active/active setups.

A low minimum is fine OTOH.

louismunro, line 107: I mean the $MaxClients though.
Not the $MaxSpareServers.

louismunro:
Poor Debian.

conf/httpd.conf.d/httpd.graphite
julsemaan, line 107: I just remembered why I kept it low.
Since we have ~10 graphs on the PF dashboard, then additionnal httpd processes start whenever we access the status page on the admin since these requests are done at the same time. I agree we could put it a bit higher (like 4) but if we give him 10 processes it will use almost all of them.

Another solution would be to have graphite and httpd.admin share the same process pool and have them use VirtualHosts. If they fight together to process requests it has much less impact then if the portal and webservices where sharing the same processes.

louismunro, line 107: It could make things worse. Admin is a mod perl app, while graphite is django. That would mean loading both mod perl and python in the httpd process.

julsemaan, line 107: Yes but they would share the same pool so it could reduce the amount of processes that they're using even though each process would use a bit more memory

louismunro, line 107: Ok, it's a bit late to change that. 
Let's go with your proposal to lower that number of processes and see if we can merge the pools for 5.2.

PR #559, opened 2015-05-20
fixed destinationUrl when it's PacketFence itself

julsemaan:
# Description

Prevents the destination URL from being the PF domain name, empty or any IP the portal can listen on
# Impacts

Prevents redirection to the PacketFence portal at the end of the reg process
No more need to force redirection URL when using external portal :D
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fixed bad redirection to the portal at the end of the registration process

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
@@ -146,7 +146,7 @@ sub _build_destinationUrl {
 
     my $host = URI::URL->new($url)->host();
 
-    get_logger->info("User requested URL $url on host $host");
+    get_logger->debug("User requested URL $url on host $host");
     # Return portal profile's redirection URL if destination_url is not set or if redirection URL is forced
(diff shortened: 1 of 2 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
```diff
@@ -141,8 +143,19 @@ sub ACCEPT_CONTEXT {
...
+    get_logger->info("User requested URL $url on host $host");
     # Return portal profile's redirection URL if destination_url is not set or if redirection URL is forced
-    if (!defined($url) || isenabled($self->profile->forceRedirectURL)) {
+    if (!defined($url) || !$url || isenabled($self->profile->forceRedirectURL)) {
+        return $self->profile->getRedirectURL;
+    }
+
+    my @portal_hosts = portal_hosts();
+    # if the destination URL points to the portal, we put the default URL of the portal profile
+    if (URI::URL->new($url)->host() ~~ @portal_hosts) {
```
jrouzierinverse: Why are you calling URI::URL->new($url) again? Why not just use $host

julsemaan:
Please re-review

PR #575, opened 2015-05-22
Hostname

whitx:
# Description

Added a assumptions taht cause me some issue while building the cluster.
# Impacts

None
# Delete branch after merge

YES 
## Enhancements
- Added assumption

### DIFF
--- a/docs/PacketFence_Clustering_Guide.asciidoc
+++ b/docs/PacketFence_Clustering_Guide.asciidoc
@@ -16,7 +16,8 @@ Assumptions
 * Both servers are identical copies for the network interfaces
 * Both servers are running CentOS 6
 * Both servers have access to the same layer 2 network on all their network interfaces
-* Both servers have an empty, unformatted partition for the database.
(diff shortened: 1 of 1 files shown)


whitx:
We ask later to use the hostname in config, but i feel like this need to be said in assumptions since it is mandatory to the cluster and should be the case in a company network.

docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -16,7 +16,8 @@ Assumptions
 * Both servers are identical copies for the network interfaces
 * Both servers are running CentOS 6
 * Both servers have access to the same layer 2 network on all their network interfaces
-* Both servers have an empty, unformatted partition for the database.
+* Both servers have an empty, unformatted partition for the database
+* Both servers hostname must be reachable via a DNS resolution.
```
louismunro: s/reachable/resolvable/

cgx: Remove the ending period.

PR #584, opened 2015-05-22
Add way to skip checkup

### LINKED ISSUE #577: Should be able to bypass checkup
There should be an option to not stop on unsuccessful checkup. 

This way advanced users could start PF with a bad DB version if they really feel the need for it (like in a complex upgrade)

Suggestion : 
pfcmd service pf start --ignore-checkup

julsemaan:
# Description

Allows to start the requested services even if the checkup fails with a fatal error
# Impacts

Prevents frustrations during an upgrade 
# Issue

fixes #577
# Delete branch after merge

(REQUIRED)
YES
# NEWS file entries
## New Features
- Added an option to pfcmd service to ignore a fatal checkup


lib/pf/cmd/pf/service.pm
```diff
@@ -154,6 +159,7 @@ sub startService {
 }
 
 sub checkup {
+    my ($self) = @_;
```
jrouzierinverse: Delete this

PR #593, opened 2015-05-25
Feature/ipset roles

dwlfrth:
# Description

Now categorizing registered nodes (in inline) by their roles and assigning them a CLASS for traffic shaping.
# Impacts

Inline flow
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Inline registered nodes are being classified in iptables to allow traffic shaping based on their role id


dwlfrth:
Review and merge me

fdurand:
Tested and works.
Also the code look ok for me.

PR #601, opened 2015-05-27
Meraki web auth

julsemaan:
# Description

Adds web authentication support with the Meraki access points
# Impacts

Another switch module
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added support for web authentication on the Meraki cloud controller


docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -3524,6 +3524,32 @@ Hostapd (software)
...
 To configure Hostapd software you can use the same configuration parameters above in the configuration file.
 
+Meraki
+~~~~~~
+
+In this section, we will cover the configuration of the Meraki controller to use Web authentication.
+
+CAUTION: There is currently no way to reevaluate the access of the device using PacketFence. This means that there is no real-time action when trying to isolate a device. The RADIUS reauthentication delay will determine the maximum response time of the access reevaluation.
+
+First create a new SSID and then edit it's settings.
```
extrafu: its settings

docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -3524,6 +3524,32 @@ Hostapd (software)
...
+
+Configure it as seen in the screenshot below.
+
+image::docs/images/meraki-radius-config.png[scaledwidth="100%",alt="RADIUS configuration"]
+
+NOTE: Make sure to configure your RADIUS server so it's accessible externally under a domain name or a public IP address. Replace `pf.domain.net` by what you will configure. Refer to the Meraki documentation for more details.
```
extrafu: Perhaps we could also say that only the ports XYZ need to be open.

lib/pf/Switch/Meraki/AP_http.pm
```diff
@@ -0,0 +1,165 @@
...
+pf::Switch::Meraki::AP_http
+
+=head1 SYNOPSIS
+
+The pf::Switch::Meraki::AP_http module implements an object oriented interface to 
+manage the external captive portal on Meraki access points
+
+=head1 STATUS
+
+Developed and tested on an MR12 access point
```
extrafu: on a MR12

extrafu:
@whitx Please test with the doc as-is. Also check with @julsemaan for the firewall rules, and @lzammit if the firewall needs a reboot.

whitx:
Screenshots missing

whitx:
Need to add a DHCP helper on the switch to make it work or have the AP on the same subnet than PacketFence. Meraki send a natted IP(of the client) to PF where it should just send the one that was attributed by any dhcp server to the client.
Try to use last firmware update but didn't change anything.

julsemaan:
Someone has last thoughts about this PR before it gets merged ?

PR #605, opened 2015-05-27
Rework of actual IDs and add a script to unregister node if an accoun…

lzammit:
# Description

Added script to act on the microsoft Event ID 4726 which is a disabled account in Active Directory. Packetfence will unregister node with the PID of that sAMAccountName.
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Abilitie to unreg nodes that belong to a disabled account in Active Directory

### DIFF
--- a/docs/PacketFence_Administration_Guide.asciidoc
+++ b/docs/PacketFence_Administration_Guide.asciidoc
@@ -2719,6 +2719,82 @@ At the bottom, select in the list "Run a new instance in parallel" in order to u
 
 Validate with Ok and give the account who will run this task. (Usually _DOMAIN\Administrator_)
 
+Disabled Account
+^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2729,7 +2805,7 @@ Create the script `unreg_node_locked_account.ps1` on the Windows Server with the
 #Powershell script to unregister locked Active Directory account based on the UserName.#
 #########################################################################################
 
-Get-EventLog -LogName Security -InstanceId 4725 |
+Get-EventLog -LogName Security -InstanceId 4740 |
```
julsemaan, line 2808: You put 4740 here and 4726 in the script above. Which one is right ?

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2776,7 +2852,7 @@ Triggers -> New
 Begin on the task: On an event
 Log: Security
 Source: Microsoft Windows security auditing.
-Event ID: 4726
+Event ID: 4740
```
julsemaan, line 2855: Same here

PR #607, opened 2015-05-27
Proper external portal detection in the catalyst portal

julsemaan:
# Description

Allows access to the dispatcher session from the Catalyst portal
Also fixes the management message showing in unknown state when using the external portal
# Impacts

Modifies the Catalyst session and external portal release flow
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fixed bad error message when an external portal authenticated device hits the unknown state


louismunro:
LGTM.

PR #616, opened 2015-05-29
Vlan pool support for vlan by id

fdurand:
# Description

Adds a way to define a pool of vlan to use in a role.
# Impacts

No
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added support for vlan pool in role by vlan id

### DIFF
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -497,7 +497,21 @@ sub getNormalVlan {
     }
     $vlan = $switch->getVlanByName($role);
     my $vlanpool = new pf::vlan::pool;
-    $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $role);
+
+    my $pool_args = {
+        vlan => $vlan,
+        switch => $switch,
+        ifIndex => $ifIndex,
+        mac => $mac,
+        node_info => $node_info,
+        connection_type => $connection_type,
+        user_name => $user_name,
+        ssid => $ssid,
+        radius_request => $radius_request,
+        role => $role,
+    };
+    $vlan = $vlanpool->getVlanFromPool($pool_args);
+
     $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.05 );
     return ($vlan, $role);
 }
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -497,7 +497,7 @@ sub getNormalVlan {
     }
     $vlan = $switch->getVlanByName($role);
     my $vlanpool = new pf::vlan::pool;
-    $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request);
+    $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $role);
     $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.05 );
     return ($vlan, $role);
 }
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -496,9 +496,9 @@ sub getNormalVlan {
         $logger->info("[$mac] Username was NOT defined or unable to match a role - returning node based role '$role'");
     }
     $vlan = $switch->getVlanByName($role);
-    $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.05 );
     my $vlanpool = new pf::vlan::pool;
     $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request);
+    $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.05 );
     return ($vlan, $role);
 }
 
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -31,6 +31,7 @@ use pf::authentication;
 use pf::Authentication::constants;
 use pf::Portal::ProfileFactory;
 use pf::vlan::filter;
+use pf::vlan::pool;
 use pf::person;
 use pf::lookup::person;
 use Time::HiRes;
@@ -496,6 +497,8 @@ sub getNormalVlan {
     }
     $vlan = $switch->getVlanByName($role);
     $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.05 );
+    my $vlanpool = new pf::vlan::pool;
+    $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request);
     return ($vlan, $role);
 }
 
--- a/lib/pf/vlan/pool.pm
+++ b/lib/pf/vlan/pool.pm
@@ -44,13 +44,13 @@ sub new {
 }
 
 sub getVlanPool {
-    my ($self, $vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request) = @_;
+    my ($self, $vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $role) = @_;
     my $logger =  pf::log::get_logger();
 
     return $vlan if $self->rangeValidator($vlan);
     my $range = Number::Range->new($vlan);
 
-    $vlan = $self->getRoundRobin($mac, $node_info, $range);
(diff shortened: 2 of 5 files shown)


lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+}
+
+
+=head2 getRoundRobbin
+
+Return the vlan id based on round robbin
+
+=cut
+
+sub getRoundRobbin {
```
julsemaan: *getRoundRobin

It's spelled with only 1 b

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+         $rangesep$
+         )/x;
+    return 1 if ($range =~ m/$validation/g);
+    return 0;
+}
+
+
+=head2 getRoundRobbin
+
+Return the vlan id based on round robbin
```
julsemaan: *robin

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+         ^$rangesep|
+         $sectsep$|
+         $rangesep$
+         )/x;
+    return 1 if ($range =~ m/$validation/g);
+    return 0;
+}
+
+
+=head2 getRoundRobbin
```
julsemaan: getRoundRobin

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+}
+
+sub getVlanPool {
+    my ($self, $vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request) = @_;
+    my $logger =  pf::log::get_logger();
+
+    return $vlan if $self->rangeValidator($vlan);
+    my $range = Number::Range->new($vlan);
+
+    $vlan = $self->getRoundRobbin($mac, $node_info, $range);
```
julsemaan: getRoundRobin

lib/pf/vlan.pm
```diff
@@ -496,6 +497,8 @@ sub getNormalVlan {
     }
     $vlan = $switch->getVlanByName($role);
     $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.05 );
+    my $vlanpool = new pf::vlan::pool;
+    $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request);
```
julsemaan: These two lines should go above the timing as now there's not timed at all

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+
+sub new {
+    my $logger =  pf::log::get_logger();
+    $logger->debug("instantiating new pf::vlan::pool object");
+    my ( $class, %argv ) = @_;
+    my $this = bless {}, $class;
+    return $this;
+}
+
+sub getVlanPool {
```
louismunro: getVlanFromPool? 
Otherwise it sounds like it is returning the whole pool.

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+
+Validate the range definition
+Should be something like that 20..23 or 20..23,27..30
+
+=cut
+
+sub rangeValidator {
+    my ($self, $range) =@_;
+    my $rangesep = qr/(?:\.\.)/;
+    my $sectsep  = qr/(?:\s|,)/;
```
louismunro, line 67: I am not sure we should accept spaces as section separators.
It may be much easier in the end to just ignore spaces.
That could prevent a whole class of hard to debug errors (typos etc.)

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+    return 0;
+}
+
+
+=head2 getRoundRobin
+
+Return the vlan id based on round robin
+
+=cut
+
```
louismunro, line 90: This code needs some explanations. It is a bit hard to follow.

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+    my ($self, $mac, $role, $range) = @_;
+    my $logger =  pf::log::get_logger();
+
+    my $vlan_count = $range->size;
+    my $node_info_complete = node_view($mac);
+    if ( defined($node_info_complete->{'last_vlan'}) && $range->inrange($node_info_complete->{'last_vlan'}) ) {
+        $logger->debug("NODE LAST VLAN ".$node_info_complete->{'last_vlan'});
+        return ($node_info_complete->{'last_vlan'});
+    }
+    my $last_reg_mac = node_last_reg_non_inline_on_category($mac, $role);
```
louismunro, line 101: Performance worries me here.
Joins tend to be bad when it comes time to scale.

extrafu, line 101: Why not simply use a robust hash function? The bucket size could be configurable.

lib/pf/vlan/pool.pm
```diff
@@ -0,0 +1,154 @@
...
+    return $this;
+}
+
+sub getVlanPool {
+    my ($self, $vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $role) = @_;
+    my $logger =  pf::log::get_logger();
+
+    return $vlan if $self->rangeValidator($vlan);
+    my $range = Number::Range->new($vlan);
+
```
louismunro, line 52: Not sure I understand this part.
If the rangeValidator returns 1 we will never hit the getRoundRobin() ?

lib/pf/vlan.pm
```diff
@@ -495,6 +496,8 @@ sub getNormalVlan {
         $logger->info("[$mac] Username was NOT defined or unable to match a role - returning node based role '$role'");
     }
     $vlan = $switch->getVlanByName($role);
+    my $vlanpool = new pf::vlan::pool;
+    $vlan = $vlanpool->getVlanPool($vlan, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $role);
```
jrouzierinverse: Change the call to submit a single hash ref instead of the long list of parameters.
$vlanpool->getVlanPool({vlan => $vlan, switch =>  $switch, ifIndex => $ifIndex, mac => $mac,  node_info => $node_info, connection_type => $connection_type, user_name => $user_name, ssid => $ssid, radius_request => $radius_request, role => $role});

dwlfrth: I'd build the hash separately to keep it clean and then pass the reference in the call :) :nail_care:

jrouzierinverse: my $pool_args = {vlan => $vlan, switch => $switch, ifIndex => $ifIndex, mac => $mac, node_info => $node_info, connection_type => $connection_type, user_name => $user_name, ssid => $ssid, radius_request => $radius_request, role => $role};
$vlan = $vlanpool->getVlanPool($pool_args);

It makes it easier to add new parameters later

dwlfrth: Exaaaaccccttlllyyyy !!

``` perl
my $pool_args = {
    vlan => $vlan,
    switch => $switch,
    ifIndex => $ifIndex,
    mac => $mac,
    node_info => $node_info,
    connection_type => $connection_type,
    user_name => $user_name,
    ssid => $ssid,
    radius_request => $radius_request,
    role => $role,
};
$vlan = $vlanpool->getVlanPool($pool_args);
```

"I'm sexy and I know it"

lib/pf/vlan/pool.pm
fdurand, line 52: if it return 1 it mean that it's not a range

lib/pf/vlan/pool.pm
fdurand, line 101: The last reg non inline on category value must be shared between all the radius process.
The better way is to take it from the database

lib/pf/vlan/pool.pm
fdurand, line 67: This code is based on the Number::Range module http://cpansearch.perl.org/src/LARRYSH/Number-Range-0.12/lib/Number/Range.pm it's use to detect that it's a range of number and not only a number or string

lib/pf/vlan.pm
fdurand: i don't like hash

dwlfrth: And I don't like having to modify all the code to add a parameter :P ahah

lib/pf/vlan/pool.pm
fdurand, line 101: Btw it will only happen one time on registration. If the device is reg then we use the last_vlan instead.

fdurand:
I want to rework it because i refactor the vlan.pm in another branch.

louismunro:
Ok, we really should sit down together sometime this week and see that we can merge both of our implementations.

louismunro:
Closing for now.
We will reopen.

PR #664, opened 2015-06-05
pf.conf merged lists

### LINKED ISSUE #598: "overloadable" configuration field type
Introduce a new configuration field type (let's call it overloadable for the moment) that will act as the following (alway as a list context):
- In pf.conf.default, we have some value for a configuration parameter:
  Let's say we have captive_portal.detection_url:http://steak1.com, http://steak2.com,...
- Currently, when configuring the value for a configuration field in pf.conf, it is overwritting the defaults value... in that case, it would ADD to it... 

Possible usage: 
- Incoming captive-portal detection URLs
- Proxy passthrought
- CRL
- ...

That would let the user configure some without removing the ones we consider "necessary" to PacketFence and remove the hardcoding part actually used.

julsemaan:
# Description

Adds the possibility to create fields that are a list merged from pf.conf.defaults to pf.conf
Would allow a progressive move of user related constants to pf.conf.defaults so they can be shown cleanly in the admin
# Impacts

Changes proxy_passthroughs creation for now. Other configurations are not affected
# Issue

ie: fixes #598
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added merged_list configuration type which is the result of a configuration list merge between pf.conf.defaults and pf.conf

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
@@ -78,6 +78,7 @@ sub field_list {
                 last;
             };
             $type eq 'merged_list' && do {
+                delete $field->{element_attr}->{placeholder};
                 $field->{tags}  = { before_element => \&defaults_list,
               defaults => $defaults->{$name} },
 
--- a/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
@@ -13,6 +13,7 @@ Form definition to create or update a section of pf.conf.
 use HTML::FormHandler::Moose;
 extends 'pfappserver::Base::Form';
 with 'pfappserver::Base::Form::Role::Help';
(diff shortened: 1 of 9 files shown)


jrouzierinverse:
Why are there commits for the twitterSource here?

dwlfrth:
:heart_eyes:

html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
```diff
@@ -76,6 +77,15 @@ sub field_list {
                 #   if $field->{element_attr}->{placeholder};
                 last;
             };
+            $type eq 'merged_list' && do {
+                delete $field->{element_attr}->{placeholder};
+                $field->{tags}  = { before_element => \&defaults_list,
+              defaults => $defaults->{$name} },
```
dwlfrth, line 83: Indentation

dwlfrth:
Tested and unit tested. Works beautifully :nail_care: 
:rocket:

PR #669, opened 2015-06-08
Force sending JSON if accept header is requesting it

julsemaan:
# Description

If a client is explicitally requesting JSON through the Accept header, we should enforce that since it it removes code duplication by creating new actions just for that
# Impacts

Shouldn't be any for existing code.
# Delete branch after merge

NO (required for cloud)
# NEWS file entries
## Enhancements
- Force a JSON response if the Accept header is set to 'application/json'


julsemaan:
Well for now I think that this simple piece of code will do it.

I don't think we're at the stage where we'll handle more than JSON and HTML

louismunro:
The Hammer says it's good.

PR #677, opened 2015-06-10
Allow to strip sensible infos from logs on pf_run failure

julsemaan:
# Description

Allow to strip sensible infos from logs on pf_run failure
# Impacts

Passwords won't be logged....
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added way to not log sensible information when failing to execute a command through pf_run
## Bug Fixes
- Stopped logging database password on schema import failure

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/DB.pm
+++ b/html/pfappserver/lib/pfappserver/Model/DB.pm
@@ -210,7 +210,7 @@ sub schema {
     $db = quotemeta ($db);
     my $mysql_cmd = "/usr/bin/mysql -u $root_user -p$root_password $db";
     my $cmd = "$mysql_cmd < $install_dir/db/pf-schema.sql";
-    eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ])) };
+    eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ]), log_strip => "-p$root_password") };
     if ( $@ || !defined($result) ) {
         $status_msg = ["Error applying the schema to the database [_1]",$db ];
         $logger->warn("$@: $result");
@@ -220,7 +220,7 @@ sub schema {
     @custom_schemas = sort @custom_schemas;
     foreach my $custom_schema (@custom_schemas) {
         my $cmd = "$mysql_cmd < $custom_schema";
-        eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ])) };
+        eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ], log_strip => "-p$root_password")) };
         if ( $@ || !defined($result) ) {
             $status_msg = ["Error applying the custom schema $custom_schema to the database [_1]",$db ];
             $logger->warn("$@: $result");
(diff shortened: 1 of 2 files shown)


lib/pf/util.pm
```diff
@@ -751,16 +751,18 @@ sub pf_run {
     my $caller = ( caller(1) )[3] || basename($0);
     $caller =~ s/^(pf::\w+|main):://;
 
+    my $loggable_command = $command;
+    $loggable_command =~ s/$options{log_strip}/*obfuscated-information*/g;
```
jrouzierinverse, line 755: Change s/$options{log_strip}/_obfuscated-information_/g to s/\Q$options{log_strip}\E/_obfuscated-information_/g

So perl does not try to interpret  log_strip as a regex

lzammit, line 755: It's better to have it as a regex I think since it allows to do finer replacement in the command.

I love dinde

jrouzierinverse, line 755: No

julsemaan, line 755: That was me talking anyway. 
But for example if I'd like to do a common regex to strip any MySQL password instead of knowing the password to strip I could. That's mainly why I left it as an interpretable regex

jrouzierinverse, line 755: That make sense

html/pfappserver/lib/pfappserver/Model/DB.pm
```diff
@@ -210,7 +210,7 @@ sub schema {
     $db = quotemeta ($db);
     my $mysql_cmd = "/usr/bin/mysql -u $root_user -p$root_password $db";
     my $cmd = "$mysql_cmd < $install_dir/db/pf-schema.sql";
-    eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ])) };
+    eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ]), log_strip => "-p$root_password") };
```
jrouzierinverse: Replace 
log_strip => "-p$root_password"
With
log_strip => quotemeta("-p$root_password")

html/pfappserver/lib/pfappserver/Model/DB.pm
```diff
@@ -220,7 +220,7 @@ sub schema {
     @custom_schemas = sort @custom_schemas;
     foreach my $custom_schema (@custom_schemas) {
         my $cmd = "$mysql_cmd < $custom_schema";
-        eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ])) };
+        eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ], log_strip => "-p$root_password")) };
```
jrouzierinverse: Same as above

html/pfappserver/lib/pfappserver/Model/DB.pm
```diff
@@ -235,7 +235,7 @@ sub schema {
     $db = quotemeta ($db);
     $mysql_cmd = "/usr/bin/mysql -u $root_user -p$root_password $db";
     $cmd = "$mysql_cmd < $install_dir/db/pf_graphite-schema.sql";
-    eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ])) };
+    eval { $result = pf_run($cmd, (accepted_exit_status => [ 0 ], log_strip => "-p$root_password")) };
```
jrouzierinverse: Same as above

julsemaan:
Fixed, re-review

PR #679, opened 2015-06-11
Use external portal session to compute IP in catalyst

julsemaan:
# Description

Allows NATed clients to use the portal if they are using web authentication
# Impacts

Allows NATed clients to use the portal if they are using web authentication
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added support for web authentication for NATed clients


fdurand:
No comment , looks good

PR #682, opened 2015-06-11
Fixes 681

### LINKED ISSUE #681: 'portal' interface type is not handled in the GUI
- Can't set a 'portal' interface type from the GUI
- When saving existing configuration including 'portal' interface, that existing configuration is being overwritten and 'portal' type is removed...

dwlfrth:
# Description

Add a new field in interface configuration modal window to allow configure additional listening daemon (httpd...)
For the current case, only 'portal' is available

IT DOES NEED SOME POLISHING BUT FOR THE MOMENT, IT DOES WHAT IT NEED TO BE DONE
# Impacts
- Admin GUI
- Interfaces configuration
- webauth
# Issue

fixes #681 
# Delete branch after merge

YES
# NEWS file entries

Nothing since it "fixes" a "bug" which has been merged but not released yet.

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Interface.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Interface.pm
@@ -444,7 +444,7 @@ sub setType {
                                     $self->_prepare_interface_for_pfconf($interface, $interface_ref, $type));
 
         # Update networks.conf
-        if ( $type =~ /^management$|^portal$/ ) {
+        if ( $type =~ /management|portal/ ) {
             # management interfaces must not appear in networks.conf
             $models->{network}->remove($interface_ref->{network}) if ($interface_ref->{network});
         }
--- a/html/pfappserver/lib/pfappserver/Model/Interface.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Interface.pm
@@ -403,6 +403,7 @@ sub getType {
         else {
             $type = $interface->{type};
             $type = ($type =~ /management|managed/i) ? 'management' : 'other';
+            $type .= ($interface->{type} =~ /portal/i) ? ',portal' : '';
         }
     }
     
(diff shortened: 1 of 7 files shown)


dwlfrth:
Please review and test since it "fixes a bug" introduced by the new 'portal' interface type.

dwlfrth:
There an issue with the "network.conf" file when assigning new portal type. I"ll have a look into that

dwlfrth:
Latest comment should now be fixed

julsemaan:
Interface name is not sexy when enabling portal on it

I think it should show 'Management' and then we could add another column for additionnal interfaces roles

image

html/pfappserver/lib/pfappserver/Model/Interface.pm
```diff
@@ -403,6 +403,7 @@ sub getType {
         else {
             $type = $interface->{type};
             $type = ($type =~ /management|managed/i) ? 'management' : 'other';
+            $type .= ($interface->{type} =~ /portal/i) ? ',portal' : '';
```
julsemaan, line 406: The comment about display sexyness should be fixed here

julsemaan:
Code looks good other than the issue raised above

html/pfappserver/lib/pfappserver/Model/Interface.pm
dwlfrth, line 406: Actually no... since it mess up with something else...
I fixed it in the template for the moment... again, it NEEDS polishing... will be some other work

julsemaan:
When in the configurator, you can't get past the interfaces if you select web auth since it fails to detect the portal interface

image

image

julsemaan:
And then if you have management and portal setup on your interface you can't get past vlan isolation flow

image

Interface config is the same as above

dwlfrth:
Fixed the configurator issue.

julsemaan:
Works fine. Merging so it can be tested directly in devel

PR #686, opened 2015-06-12
fingerbank processing in pfdhcplistener should be async

dwlfrth:
# Description

Using webservices to async fingerbank querying in pfdhcplistener
# Impacts

pfdhcplistener fingerbank integration
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Fingerbank processing in pfdhcplistener is now async using webservices

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -23,6 +23,7 @@ use pf::config::cached;
 use pf::ConfigStore::Interface();
 use pf::ConfigStore::Pf();
 use pf::iplog();
+use pf::fingerbank;
 use pf::Portal::ProfileFactory();
 use pf::radius::custom();
 use pf::violation();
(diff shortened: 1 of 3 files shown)


dwlfrth:
Please review

lib/pf/api.pm
```diff
@@ -23,6 +23,7 @@ use pf::config::cached;
 use pf::ConfigStore::Interface();
 use pf::ConfigStore::Pf();
 use pf::iplog();
+use pf::fingerbank;
```
fdurand, line 26: use pf::fingerbank();

dwlfrth, line 26: Is there a difference ? Just asking :) (i'm okay with making it standard to the others but wondering why we start using parenthesis after use)

rouzier, line 26: This was done for pf::api to avoid exporting methods into the pf::api namespace.
To avoid people from calling methods that were not meant to be exported.
However it is not needed anymore since we have the Public attribute for all functions that are exported.

dwlfrth, line 26: Ok! Thanks for explanation! I was thinking about the export thing too.
I will add it just to make it "standard" but since I usually avoid EXPORT in modules, that shouldn't change anything

PR #698, opened 2015-06-15
PacketFence will no longer try to start all services after upgrade or install

### LINKED ISSUE #622: Do not start PacketFence service on install / upgrade
No playing around with if install / if upgrade / if nomoremustardinzammitwich....
Simply stop starting packetfence service after installation / upgrade and replace it by httpd.admin...

louismunro:
# Description

Prevent PacketFence from starting all services post installation or upgrade.
# Impacts

Changes to the spec file.
# Issue

fixes #622 
# Delete branch after merge

YES 
# NEWS file entries
## Bug Fixes
- PacketFence no longer starts all services following an upgrade (#622).


lzammit:
Tested on debian during an upgrade, it's working as expected.

PR #727, opened 2015-06-25
Feature/db maintenance

julsemaan:
# Description

Reworks the way we do the database maintenance so it doesn't lock the database for hours
# Impacts

Removes the archiving part of the maintenance script.
Archiving will have to be handled in another script.
I tried to make the script fit the majority of the users and they do not use the archiving tables
# Delete branch after merge

NO
# NEWS file entries
## New Features
- Reworked maintenance script so it doesn't lock the table it cleans for too much time.

### DIFF
--- a/addons/database-backup-and-maintenance.sh
+++ b/addons/database-backup-and-maintenance.sh
@@ -18,7 +18,7 @@
 NB_DAYS_TO_KEEP_DB=30
 NB_DAYS_TO_KEEP_FILES=30
 DB_USER='pf';
-DB_PWD='';
+DB_PWD='pf';
 DB_NAME='pf';
 PF_DIRECTORY='/usr/local/pf/'
 PF_DIRECTORY_EXCLUDED='/usr/local/pf/logs'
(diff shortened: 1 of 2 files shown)


addons/database-cleaner
```diff
@@ -0,0 +1,175 @@
...
+
+
+
+use strict;
+use warnings;
+
+use lib '/usr/local/pf/lib';
+use pf::db;
+use DBI;
+use Data::Dumper;
```
fdurand, line 37: Useless

addons/database-cleaner
```diff
@@ -0,0 +1,175 @@
...
+use Getopt::Long;
+use pf::log;
```
fdurand, line 39: Useless

addons/database-cleaner
```diff
@@ -0,0 +1,175 @@
...
+
+=cut
+sub seed_data {
+    my ($self) = @_;
+
+    my $sth = $self->{dbh}->prepare("INSERT INTO `radacct` (`acctsessionid`, `acctuniqueid`, `acctstarttime`) VALUES (?,?,?)");
+
+    foreach my $i (1..10000) {
+        $sth->bind_param(1, int(rand(10000))); 
+        $sth->bind_param(2, "dinde");
```
fdurand, line 93: Is it possible to replace by jambe de dinde ?

addons/database-backup-and-maintenance.sh
```diff
@@ -95,13 +91,10 @@ if [ -f /var/run/mysqld/mysqld.pid ]; then
     if [ `/bin/date +%d` -eq '01' ]; then
```
louismunro, line 91: This line has given us trouble in the past (on Jan. 1st for example).
It would be better if we ran this every night in smaller increments.

louismunro:
We need to think carefully about this.
While I don't see any problem with the code as is, it may be a hack to fix a problem that should really be fixed with a more radical solution.
The root of our troubles is that some tables are not properly indexed.
Whenever we delete entries from a table using a query for which there are no indexes mysql has to lock the whole table for the duration.
If there were proper indexes it would only lock the entries it is about to delete, which would let other queries run concurrently.
Can we sit down and talk about it?

julsemaan:
In my opinion, indexing is completely another issue.

We had a previous issue that made the maintenance hold an abusive lock on the tables when a big amount of cleanup was necessary. This fixes it.

Now even if the tables are properly indexed, any abusive number of deletions will lock that tables for a big amount of time which we do not want.

louismunro:
Not really. 
Tables are locked for deletion only if there are no indexes matching the rows to delete.
Otherwise only the rows in question will lock (assuming innodb of course).

So I do see indexes as a cause of this problem.

julsemaan:
We've been talking about these indexes for 3 releases so I think it's time to have them in 5.3 and see if our issues are really gone. 

We can then fallback to this in case it doesn't work, but we need a solution for the next release as big environments die during 10 minutes every night...

louismunro:
Agreed. 
Let's try to get that out the door in 5.3. 
I will look into it.

louismunro:
Let's go with it as is. We are too close to release to mess with the indexes now.

PR #733, opened 2015-07-03
Allow a chained source to continue processing it's authentication source

jrouzierinverse:
## Description

Allow a chained source to continue processing it's authentication source
## Impacts

The portal workflow for a chained source.
## NEWS file entries

New Features
++++++++++++
- Chained sources can now continue processing it's authentication source in addition to it's chained source
## Delete branch after merge

NO

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -182,6 +182,7 @@ sub login : Local : Args(0) {
         $c->forward('authenticationLogin');
         $c->detach('showLogin') if $c->has_errors;
         $c->forward('validateMandatoryFields');
+        $c->forward("checkIfChainedAuth");
         $c->forward('postAuthentication');
         $c->forward( 'CaptivePortal' => 'webNodeRegister', [$c->stash->{info}->{pid}, %{$c->stash->{info}}] );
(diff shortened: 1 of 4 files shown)


html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Chained.pm
```diff
@@ -33,6 +33,11 @@ has_field 'authentication_source' =>
    options_method => \&options_authentication_source,
   );
 
+has_field authentication_source_can_continue => (
```
julsemaan: I'm really not sure what this does...
A help message would be really useful here

jrouzierinverse: If it is enabled then the source defined by the authentication_source attribute can be used in addition to the registered user.

html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
```diff
@@ -259,6 +260,32 @@ sub checkIfChainedAuth : Private {
     }
 }
 
+=head2 continue_chained_auth
```
julsemaan: skipChainedSource would be a better name I think

jrouzierinverse:
Please review

julsemaan:
I'm not sure about what this does and if it's really useful for a use case.

Is this used somewhere or requested by a user ?

PR #737, opened 2015-07-06
Cleanup of the old unused script in addons.

### LINKED ISSUE #490: cleanup old useless scripts in addons
Some of the addons don't make sense anymore.
Remove them.

louismunro:
# Description

Cleanup of unused scripts in addons.
# Impacts

This deletes a number of scripts from the addons directory that are old, unused or otherwise unsupported. Have a look and make sure there is nothing removed that you can't live without.
Remember that these scripts are part of the packages and thus have to be supported.
The less code we have to support the better our lives will be. So don't just ask to keep them around if you are not willing to answer questions about them on the mailing list.
# Issue

fixes #490 
# Delete branch after merge

YES 
# NEWS file entries
## Enhancements
- Removed obsolete and otherwise unsupported scripts from the addons directory.


dwlfrth:
We have at least one client using 'addons/network-save-configs.pl'

louismunro:
He can keep using it.
Does not mean we have to support it for everyone else.

## 

LM

julsemaan:
Looks good to me, as discussed with @louismunro we'll keep those somewhere in the wiki in case we need them but no need to package them

PR #741, opened 2015-07-07
Set the max message size to 4096 octets instead of the default 1472 (…

### LINKED ISSUE #738: issue if snmp get answer is greater than 1472 octets
Not able to detect voip if the answer is greatter than 1472 octets.
http://search.cpan.org/~dtown/Net-SNMP-5.2.0/lib/Net/SNMP.pm#max_msg_size()_-_set_or_get_the_current_maxMsgSize_for_the_object

fdurand:
# Description

Define the max message size because sometimes if we request something on a switch stack the size can be greatter than the default and packetfence is not able to detect voip phones.
# Impacts

No
# Fixes issue

Fixes #738 
# Delete branch after merge

Yes
# NEWS file entries
## Bug Fixes
- Define the max messages size that snmp get can return (fix voip lldp/cdp detection on switch stack)

### DIFF
--- a/bin/pfcmd_vlan
+++ b/bin/pfcmd_vlan
@@ -318,7 +318,8 @@ if ($reevaluateAccess) {
                 -authprotocol => $switch->{_SNMPAuthProtocolRead},
                 -authpassword => $switch->{_SNMPAuthPasswordRead},
                 -privprotocol => $switch->{_SNMPPrivProtocolRead},
-                -privpassword => $switch->{_SNMPPrivPasswordRead}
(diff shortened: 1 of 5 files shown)


julsemaan:
Looks good to me, could even get into 5.3 IMO

jrouzierinverse:
Also update the following files
bin/pfcmd_vlan
lib/pf/Switch/PacketFence.pm
lib/pf/Switch/Extricom.pm
lib/pf/Switch/Cisco.pm

PR #751, opened 2015-07-11
Added LOGGING of MAC and IP address in addition to the USERNAME, after succesfull authentication

verzulli:
I've added three lines of code to fetch IP-address and MAC-address from PortalSession within the authentication.pm module, so to being able to LOG them (IP/MAC) in addition to what is currently logged (USERNAME).
This is relevent, expecially in "inline L2" deployments, as thanks to this it will be very easy to keep an historical mapping between USERNAME<=>MAC address, regardless of the historical data stored in MySQL, thanks to what's reported in LOG-files.
Probably it will be also useful in other "vlan-enforcement" deployments.

### DIFF
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -218,6 +218,11 @@ sub authenticate {
 
     $logger->debug(sub {"Authenticating '$display_username' from source(s) ".join(', ', map { $_->id } @sources) });
 
+    # Fetching IP and MAC from Portal Session, as to properly LOG them together with the USERNAME 
+	my $portalSession = pf::Portal::Session->new();
+	my $ip = $portalSession->getClientIp();
+	my $mac = $portalSession->getClientMac();
+
     foreach my $current_source (@sources) {
         my ($result, $message);
         $logger->trace("Trying to authenticate '$display_username' with source '".$current_source->id."'");
@@ -226,7 +231,7 @@ sub authenticate {
         };
         # First match wins!
         if ($result) {
-            $logger->info("Authentication successful for $display_username in source ".$current_source->id." (".$current_source->type.")");
+            $logger->info("Authentication successful for ".$display_username."/".$ip."/".$mac." in source ".$current_source->id." (".$current_source->type.")");
             return ($result, $message, $current_source->id);
         }
     }


lib/pf/authentication.pm
```diff
@@ -218,6 +218,11 @@ sub authenticate {
 
     $logger->debug(sub {"Authenticating '$display_username' from source(s) ".join(', ', map { $_->id } @sources) });
 
+    # Fetching IP and MAC from Portal Session, as to properly LOG them together with the USERNAME 
+	my $portalSession = pf::Portal::Session->new();
+	my $ip = $portalSession->getClientIp();
+	my $mac = $portalSession->getClientMac();
```
dwlfrth, line 224: Indentation is bogus

lib/pf/authentication.pm
```diff
@@ -226,7 +231,7 @@ sub authenticate {
         };
         # First match wins!
         if ($result) {
-            $logger->info("Authentication successful for $display_username in source ".$current_source->id." (".$current_source->type.")");
+            $logger->info("Authentication successful for ".$display_username."/".$ip."/".$mac." in source ".$current_source->id." (".$current_source->type.")");
```
dwlfrth, line 234: Use spaces :)
s/".$dis.../" . $dis ...

dwlfrth:
It is in fact useful since I do have couple of request for this :)

fdurand:
In my opinion it will probably be better to add this log in the function authenticationLogin in html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm since we already have  the portal session  and all the information needed

julsemaan:
Agree with @fdurand as this seems to add a cost and (i hope) we are phasing out pf::Portal::Session so the Authenticate controller seems like the best place to put it

dwlfrth:
Agree with @julsemaan who agrees with @fdurand on using the Controller to achieve this purpose.

dwlfrth:
Comments are still valid tho (except for indentation since no longer needed)

jrouzierinverse:
+1

jrouzierinverse:
This will cause a bug in other parts of the code where `pf::authentication::authenticate` is used outside of the context of the captive portal.

I am closing this pull request.

dwlfrth:
Even it is closed, proceed with @fdurand comment and open a new one :)

jrouzierinverse:
I have a better approach will discuss tomorrow

PR #753, opened 2015-07-13
Always log the ip and mac for the captive portal

jrouzierinverse:
## Description

Always log the ip and mac in the captive portal for everry log entry
## Impacts

Captive portal logging
## NEWS file entries

Enhancements
++++++++++++
- The client's mac and ip address is always logged for every log entry
## Delete branch after merge

NO


jrouzierinverse:
This will add the mac and ip to every line in httpd.portal log entries

You can see the difference here between these two lines.

Jul 12 22:48:44 httpd.portal(3928) ERROR: Can't bind : IO::Socket::INET: connect: Connection refused

Jul 12 22:48:44 httpd.portal(3928) ERROR: [mac:0a:00:27:00:00:00 ip:192.168.56.1] Can't bind : IO::Socket::INET: connect: Connection refused

jrouzierinverse:
Please Review

jrouzierinverse:
We may want to consider this for 5.3.1 but most definitely 5.4.0

This improves logging for the captive portal by allowing use to following a particular mac or ip.

You want to see the activity for a node in the captive portal then just do the following.

```
tail log/packetfence.log | grep mac:01:02:03:04:05:06
```

As this will be done for every log entry there is no longer the need to always add the node or ip in the log entry itself.

lzammit:
Looks good to me, need to be tested.

jrouzierinverse:
Then test it

James Rouzier

julsemaan:
image

PR #760, opened 2015-07-15
allow pfdns to run only on vip in cluster

### LINKED ISSUE #820: Passthroughs broken in clustered mode depending on the dns used
We return both servers IPs as nameservers in clustered mode.
That does not play well with passthroughs since pfdns may not see the request and add the passthrough if the dns request goes to the other server.

There's a proposed fix in #821

julsemaan:
# Description

Allows to push only the VIP in the DHCP options when using active/active.

Defaults to the current behavior
# Issue

fixes #820 
# Impacts

active/active clustering
# Delete branch after merge

YES
# NEWS file entries
## Bug fixes
- Made dhcp point DNS only on cluster IP if passthroughs are enabled in active/active clusters (#820)
## New Features
- Added the option to use only the management node as the DNS server in Active/Active clustering

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1397,6 +1397,13 @@ description=<<EOT
 The virtual router id for keepalive. Leave untouched unless you have another keepalive cluster in this network.
 EOT
 
+[active_active.dns_on_vip_only]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Activate so the clients in registration and isolation are pointed to the VIP of the server instead of all the servers.
+EOT
+
 [monitoring.graphite_host]
 type=text
 description=<<EOT
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1401,7 +1401,7 @@ EOT
 type=toggle
 options=enabled|disabled
 description=<<EOT
-Activate so the clients in registration and isolation are pointed to the VIP of the server instead of all the servers.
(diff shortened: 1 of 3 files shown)


conf/documentation.conf
```diff
@@ -1397,6 +1397,13 @@ description=<<EOT
 The virtual router id for keepalive. Leave untouched unless you have another keepalive cluster in this network.
 EOT
 
+[active_active.dns_on_vip_only]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Activate so the clients in registration and isolation are pointed to the VIP of the server instead of all the servers.
```
louismunro: I would make this more explicit. E.g. set the name server  option in dhcp replies to point only to the VIP in cluster mode rather than to all servers in the cluster.

conf/pf.conf.defaults
```diff
@@ -1093,6 +1093,11 @@ password=1234
 #
 # Shared KEY for vrrp protocol (Must be the same on all members).
 virtual_router_id=50
+#
+# active_active.dns_on_vip_only
+#
+# Activate so the clients in registration and isolation are pointed to the VIP of the server instead of all the servers.
```
louismunro: Same here.

louismunro:
LGTM other than minor nitpicking about comments and documentation.

julsemaan:
Fixed doc, re-review

PR #788, opened 2015-07-21
RADIUS-based IP address tracking

borepstein:
Added the following to raddb/packetfence.pm as line 323 (per discussion list recommendations):

```
$data = send_rpc_request($config, "update_iplog", {mac => $mac, ip => $RAD_REQUEST{'Framed-IP-Address'}}) if ($RAD_REQUEST{'Framed-IP-Address'} );
```

Tested this and ascertained that it did indeed add IP address to the node data whereas none was recorded previously.

Affects RADIUS-only configurations.


julsemaan:
You should be rebasing your work with our devel branch instead of merging it as it creates useless commits.

## Next time, here is a quick how-to for your case :

```
# git fetch upstream
# git rebase upstream/devel
# git push YOUR_REPO devel --force (you need to force it as we rewrote your history)
```

Follow the instructions in case of conflicts.

## Now to fix this, please do

```
# git checkout -b devel-backup (so if this goes wrong you don't lose everything)
# git checkout devel (going back to our original point)
# git reset upstream/devel --hard (we reset to our pointer)
# git cherry-pick cb84a3fb29b3f1159e8c1f6ab9bdcdee3a037a46
# git cherry-pick f86a377eb33a8ba5602573c270d899de35ce17d7
# git push YOUR_REPO devel --force (you need to force it as we rewrote your history)
The two lines above are your two commits so your devel is now ours + your 2 commits (no more dirty merges)
```

Then if you need to fast-forward your work to our branch, just follow the rebasing instructions.

BTW, love this, don't know why we didn't do it before

PR #789, opened 2015-07-21
Deny devices that hammering the portal

fdurand:
# Description

Detect devices that asking for the same uri on the portal and add them in an ipset session to deny access.
We use mod_evasive to configure the threshold and add it in ipset session with a timeout
# Impacts

Kick out portal's hammering devices
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Deny access on the portal interface if the device hit to much the captive portal.


extrafu:
What a great idea! Good work.

conf/pf.conf.defaults
```diff
@@ -817,6 +807,71 @@ secure_redirect=enabled
...
+#
+# captive_portal.httpd_mod_evasive_email_notify
+#
+# The email address where the notification will be sent
+httpd_mod_evasive_email_notify=root@localhost
+#
+# captive_portal.httpd_mod_evasive_system_command
+#
+# The command that will be executed when a ip becomes blacklisted
+httpd_mod_evasive_system_command="sudo ipset add portal_deny %s timeout 10 ; rm -f /usr/local/pf/logs/dos*"
```
julsemaan: Why do we need to rm -f the logs when blocking one IP

conf/documentation.conf
```diff
@@ -1059,6 +1045,87 @@ When enabled the /status page will only be available on the
...
+
+[captive_portal.httpd_mod_evasive]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Enables mod_evasive for the portal
+EOT
+
+[captive_portal.httpd_mod_evasive_white_list]
+type=text
```
julsemaan: Shouldn't that be a list ?

conf/documentation.conf
```diff
@@ -1059,6 +1045,87 @@ When enabled the /status page will only be available on the
...
+The intervale for the site count in second
+EOT
+
+[captive_portal.httpd_mod_evasive_blocking_period]
+type=numeric
+description=<<EOT
+The amount of time a client will be blocked
+EOT
+
+[captive_portal.httpd_mod_evasive_email_notify]
```
julsemaan, line 1138: I think that it should be sent to the alerting email address so it lowers the amount of config necessary

conf/documentation.conf
```diff
@@ -1059,6 +1045,87 @@ When enabled the /status page will only be available on the
...
+The email address where the notification will be send
+EOT
+
+[captive_portal.httpd_mod_evasive_system_command]
+type=text
+description=<<EOT
+The command that will be executed when a ip becomes blacklisted
+EOT
+
+[captive_portal.httpd_mod_evasive_log_dir]
```
julsemaan: Overconfiguration ?

conf/documentation.conf
fdurand: Nop, just an ip or a cidr format

conf/documentation.conf
fdurand, line 1138: Yes you are right

conf/pf.conf.defaults
fdurand: Because the device will never be cached again until the file is removed.
I will probably rework this system_command stuff to use a perl script like pfbloquinator to have more logic inside.

conf/documentation.conf
fdurand, line 1138: After reflection do we want to receive alert on the alerting email address each time a device is deny on the portal ?

julsemaan, line 1138: That could be configurable. Since we might want to monitor actively a portal for some time then turn it off when we the situation is more stable

fdurand:
It has been rebased and i removed the logdir configuration parameter.
For the httpd_mod_evasive_email_notify since it's a apache configuration parameter and to update it we need to restart the portal, i would like to keep it like that for the moment.

conf/pf.conf.defaults
dwlfrth: pfbloquinator -_-
#notenoughdaemons

dwlfrth:
It needs an upgrade procedure since we are modifying configuration parameters.

conf/pf.conf.defaults
fdurand: not a daemon, but just a script that do the logic

fdurand: btw pfbloquinator, i like it

conf/documentation.conf
dwlfrth: As @julsemaan says, not sure to understand that config parameter... also, it is not being used (not present in this PR except that for config files and translation file...)

conf/documentation.conf
```diff
@@ -1080,6 +1066,81 @@ description=<<EOT
...
+Enables mod_evasive for the portal
+EOT
+
+[captive_portal.httpd_mod_evasive_white_list]
+type=text
+description=<<EOT
+Whitelisted ip/network
+EOT
+
+[captive_portal.httpd_mod_evasive_hash_table_size]
```
dwlfrth, line 1102: Should this really be a configuration parameter ?

conf/documentation.conf
```diff
@@ -1080,6 +1066,81 @@ description=<<EOT
...
+[captive_portal.httpd_mod_evasive_page_count]
+type=numeric
+description=<<EOT
+Limit the number of request on the same page
+EOT
+
+[captive_portal.httpd_mod_evasive_page_interval]
+type=numeric
+description=<<EOT
+The intervale for the page count in second
```
dwlfrth: interval
Also, should explain interval between what and what

conf/documentation.conf
```diff
@@ -1080,6 +1066,81 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive_site_count]
```
dwlfrth, line 1120: I think it should be better documented

conf/documentation.conf
```diff
@@ -1080,6 +1066,81 @@ description=<<EOT
...
+type=numeric
+description=<<EOT
+Limit the number of request on the same listener
+EOT
+
+[captive_portal.httpd_mod_evasive_site_interval]
+type=numeric
+description=<<EOT
+The intervale for the site count in second
```
dwlfrth: interval
Also, should explain interval between what and what

conf/documentation.conf
```diff
@@ -1080,6 +1066,81 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive_blocking_period]
+type=numeric
+description=<<EOT
+The amount of time a client will be blocked
```
dwlfrth: time unit should be documented

conf/documentation.conf
```diff
@@ -1080,6 +1066,81 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive_email_notify]
+type=text
+description=<<EOT
+The email address where the notification will be send
+EOT
+
+[captive_portal.httpd_mod_evasive_system_command]
```
dwlfrth, line 1144: Should this really be a configuration parameter ?

lib/pf/ipset.pm
```diff
@@ -59,6 +59,10 @@ sub iptables_generate {
     my $cmd = "LANG=C sudo ipset --destroy";
     my @lines = pf_run($cmd);
     my @roles = pf::nodecategory::nodecategory_view_all;
+
+    $cmd = "LANG=C sudo ipset --create portal_deny hash:ip timeout 300 2>&1";
```
dwlfrth, line 63: Shouldn't we use the block time period parameter for the timeout ?

conf/pf.conf.defaults
dwlfrth: Time is "hardcoded" in that command... if we specify a timeout on set creation (which we are... not just honoring the config parameter), we don't need to specify a timeout for each single entry of the set

fdurand: Let's say the default timeout is 300s but the admin want to change this timeout to 3600s so it just have to change this timeout value. Timeout is not mandatory here but it's just an example.
As i said i want in the futur to change this by a script with more logic indise, like the first time you have been catched it's for 10s but after an hundred of time it will probably be for a moment or i will probably disconnect the device and deny the radius request ....

lib/pf/ipset.pm
fdurand, line 63: We can, btw time x is just to create a timeout ipset hash, without the timeout there is no way to change the ipset config.

conf/documentation.conf
fdurand, line 1102: Yes, it must be configurable, it must be increase if we have a busy portal.

conf/documentation.conf
fdurand, line 1144: Yes why not ?

UPGRADE.asciidoc
```diff
@@ -111,6 +111,12 @@ Also in violations.conf, the parameter whitelisted_categories has been renamed i
 
 Once completed, update the file /usr/local/pf/conf/currently-at to match the new release number (PacketFence X.Y.Z).
 
+Mod_qos configuration changes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Mod_qos configuration has been moved from services to captive_portal section.
```
dwlfrth: the "services" section to the "captive_portal" section

UPGRADE.asciidoc
```diff
@@ -111,6 +111,12 @@ Also in violations.conf, the parameter whitelisted_categories has been renamed i
...
+If you used it before make sure to enable it in captive portal configuration menu.
```
dwlfrth: Make sure to apply the appropriate changes if needed.

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
 Enable or disable WISPr redirection capabilities on the captive-portal
 EOT
 
+[captive_portal.httpd_mod_qos]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Enables mod_qos for the portal
```
dwlfrth: s/portal/captive-portal

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_qos_maximum_connections_per_device]
+type=numeric
+options=enabled|disabled
```
dwlfrth: options on a numeric type parameter ?

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+description=<<EOT
+The maximum connections per device
```
dwlfrth: maximum number of connnections

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Enables mod_evasive for the portal
```
dwlfrth: s/portal/captive-portal

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive_white_list]
+type=list
+description=<<EOT
+Using this option, comma delimited IP addresses or Networks CIDR format of trusted clients can be added to the whitelist to ensure they are never denied.
```
dwlfrth: Remove "Using this options, "
s/Networks/networks

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+[captive_portal.httpd_mod_evasive_page_count]
+type=numeric
+description=<<EOT
+This is the limit for the number of requests for the same page per page interval by an IP address. Once the threshold for that interval has been exceeded, the IP address of the client will be added to the blocked list.
+EOT
+
+[captive_portal.httpd_mod_evasive_page_interval]
+type=numeric
+description=<<EOT
+The interval in second for the page count threshold.
```
dwlfrth: s/second/seconds

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+[captive_portal.httpd_mod_evasive_hash_table_size]
+type=numeric
+description=<<EOT
+The hash table size defines the number of top-level nodes for each child's hash table. Increasing this number will provide faster performance by decreasing the number of iterations required to get to the record, but consume more memory for table space. You should increase this if you have a busy web server.
+EOT
+
+[captive_portal.httpd_mod_evasive_page_count]
+type=numeric
+description=<<EOT
+This is the limit for the number of requests for the same page per page interval by an IP address. Once the threshold for that interval has been exceeded, the IP address of the client will be added to the blocked list.
```
dwlfrth: Remove "This is"

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+[captive_portal.httpd_mod_evasive_page_interval]
+type=numeric
+description=<<EOT
+The interval in second for the page count threshold.
+EOT
+
+[captive_portal.httpd_mod_evasive_site_count]
+type=numeric
+description=<<EOT
+This is the limit for the total number of requests for the portal by an IP address per site interval.
```
dwlfrth: Remove "This is"

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive_site_interval]
+type=numeric
+description=<<EOT
+The interval in second for the site count threshold.
```
dwlfrth: s/second/seconds

conf/documentation.conf
```diff
@@ -1087,6 +1073,83 @@ description=<<EOT
...
+EOT
+
+[captive_portal.httpd_mod_evasive_blocking_period]
+type=numeric
+description=<<EOT
+The blocking period is the amount of time (in seconds) that a client will be blocked for if they are added to the blocking list. During this time, all subsequent requests from the client will result in a 403 (Forbidden) and the timer being reset (e.g. another 10 seconds). Since the timer is reset for every subsequent request, it is not necessary to have a long blocking period; in the event of a DoS attack, this timer will keep getting reset.
```
dwlfrth: Remove "The blocking period is "

conf/httpd.conf.d/httpd.portal
```diff
@@ -201,10 +201,32 @@ $MaxClients = pf::services::manager::httpd::calculate_max_clients(pf::services::
...
-if( pf::config::isenabled ($PfConfig->{services}{httpd_mod_qos})) {
+if( pf::config::isenabled ($PfConfig->{captive_portal}{httpd_mod_qos})) {
     my $qos = $MaxClients * .7;
     $QS_SrvMaxConnClose = $qos;
-    $QS_SrvMaxConnPerIP = $PfConfig->{services}{httpd_mod_qos_maximum_connections_per_device};
+    $QS_SrvMaxConnPerIP = $PfConfig->{captive_portal}{httpd_mod_qos_maximum_connections_per_device};
+}
+
+if( pf::config::isenabled ($PfConfig->{captive_portal}{httpd_mod_evasive})) {
+    if( pf::config::os_detection() eq 'debian' ) {
```
dwlfrth, line 211: Shouldn't the LoadModule part be part of the LoadModule section at the beginning of this file ?

conf/pf.conf.defaults
```diff
@@ -840,6 +830,66 @@ detection_mecanism_urls = http://www.gstatic.com/generate_204,http://clients3.go
 # Enable or disable WISPr redirection capabilities on the captive-portal
 wispr_redirection = enabled
 
+# captive_portal.httpd_mod_qos
```
dwlfrth, line 833: Adjust pf.conf.defaults to match documentation.conf

conf/httpd.conf.d/httpd.portal
fdurand, line 211: In fact if i load this module in general LoadModule it will come with default values and handle request even if we disabled it.

dwlfrth, line 211: I agree, but can it be loaded but have no effect without any configuration ?
Thinking about mod_qos, which I think, is kind of the same thing no ?

fdurand, line 211: It looks like no, we have that in the source code of mod_evasive:
#define DEFAULT_HASH_TBL_SIZE   3079ul  // Default hash table size
#define DEFAULT_PAGE_COUNT      2       // Default max page hit count/interval
#define DEFAULT_SITE_COUNT      50      // Default max site hit count/interval
#define DEFAULT_PAGE_INTERVAL   1       // Default 1 second page interval
#define DEFAULT_SITE_INTERVAL   1       // Default 1 second site interval
#define DEFAULT_BLOCKING_PERIOD 10      // Default block time (Seconds)
#define DEFAULT_LOG_DIR         "/tmp

dwlfrth, line 211: Allright then :)

dwlfrth:
Remove the extra "n" from connections where I commented and we are GOLDEN to merge.

fdurand:
So we are GOLDEN

PR #799, opened 2015-07-23
Use client mac and ip in portal help instead of pregen messages

julsemaan:
# Description

Refactored portal stash to make IP and MAC available in templates instead of using a pregenerated line to show them
# Impacts

Display on the portal
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Client IP and MAC address are now available though direct variables in the captive portal templates
# UPDATE NOTES

The help message that contains the IP and MAC of the device on the captive portal has been reworked.
Make sure you adapt the changes in footer.html on each portal profile you have


lzammit:
Looks good to me, need to be tested.

PR #805, opened 2015-07-24
Added external HTTP source

julsemaan:
# Description

Allows to use an external API as an authentication source
# Impacts

Adds another authentication source
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added an HTTP authentication source that does API calls to authenticate and authorize users


jrouzierinverse:
+1

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/HTTP.pm
```diff
@@ -0,0 +1,97 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+USA.
+
+=cut
+
+__PACKAGE__->meta->make_immutable;
+1;
```
jrouzierinverse, line 101: Get rid of all the copy and paste LDAP stuff

addons/external_auth/README.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+</FilesMatch>
+
+<VirtualHost *:10000>
+    DocumentRoot /usr/local/pf/addons/external_auth
+</VirtualHost>
+```
+
+Then restart the httpd.admin process and this will be available on port 10000
+
+Note that this should only be used for testing purposes and IS NOT made for production purposes
```
jrouzierinverse: The directory could be renamed to example_external_auth

conf/httpd.conf.d/httpd.admin
```diff
@@ -191,3 +191,13 @@ push @{ $VirtualHost{$vhost.":".$PfConfig->{'ports'}{'pf_status'}} },
...
+Listen 10000
+LoadModule php5_module modules/libphp5.so
+
+<FilesMatch \.php$>
+    SetHandler application/x-httpd-php
+</FilesMatch>
+
+<VirtualHost *:10000>
+    DocumentRoot /usr/local/pf/addons/external_auth
+</VirtualHost>
```
jrouzierinverse: We should not have this started by default

html/pfappserver/root/authentication/source/type/HTTP.tt
```diff
@@ -0,0 +1,10 @@
+<div class="control-group">
+  <label class="control-label">[% l('Host') %]</label>
+  <div class="controls">
+    [% form.field('encryption').render_element | none %] [% form.field('host').render_element | none %]:[% form.field('port').render_element | none %]  
+  </div>
+</div>
+[% form.field('username').render | none %]
+[% form.field('password').render | none %]
+[% form.field('authentication_url').render | none %]
+[% form.field('authorization_url').render | none %]
```
jrouzierinverse, line 10: Place everything else in a definition block in the form
[% form.block('definition').render | none %]

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,280 @@
...
+Readonly our $API_PORT => '10000';
+
+# If your API uses HTTP basic authentication, set these to the username/password
+Readonly our $API_USERNAME => undef;
+Readonly our $API_PASSWORD => undef;
+
+# These are the two URLs necessary for this module to work
+# See the authenticate and match subroutines for the expected replies
+Readonly our $AUTHENTICATION_URL => '/authenticate.php';
+Readonly our $AUTHORIZATION_URL => '/authorize.php';
```
jrouzierinverse: Remove the use of the global variables API_USERNAME, API_PASSWORD, AUTHENTICATION_URL, AUTHORIZATION_URL

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,280 @@
...
+
+    $uri = $API_PROTO."://".$API_HOST.":".$API_PORT.$uri;
+
+    my $curl = WWW::Curl::Easy->new;
+    my $request = $post_fields;
+
+    if(defined($API_USERNAME) && defined($API_PASSWORD)){
+        $curl->setopt(CURLOPT_USERNAME, $API_USERNAME);
+        $curl->setopt(CURLOPT_PASSWORD, $API_PASSWORD);
+    }
```
jrouzierinverse, line 61: No global variables

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,280 @@
...
+  result => 1,
+  message => "Valid username and password",
+}
+
+=cut
+
+sub authenticate {
+    my ( $self, $username, $password ) = @_;
+
+    my $uri = $AUTHENTICATION_URL;
```
jrouzierinverse: No global variables

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,280 @@
...
+  category => 'default'
+}
+
+=cut
+
+sub match {
+    my ($self, $params) = @_;
+    my $common_attributes = $self->common_attributes();
+
+    my $uri = $AUTHORIZATION_URL;
```
jrouzierinverse: No global variables

jrouzierinverse:
The code should be as generic as possible.

conf/httpd.conf.d/httpd.admin
jrouzierinverse: Should provide an example in perl not php because that adds an additional dependency of php5

lib/pf/Authentication/Source/HTTPSource.pm
julsemaan: Oups my bad, forgot to push the new code

html/pfappserver/root/authentication/source/read.tt
```diff
@@ -21,7 +21,7 @@
         [% form.field('id').render | none %]
         [% form.field('description').render | none %]
         [% INCLUDE "authentication/source/type/${source.type}.tt" %]
-        [% IF source.type != 'Chained' %]
+        [% IF source.type != 'Chained' && source.type != 'HTTP' %]
```
jrouzierinverse: Maybe we should have a source attribute or method ignore rules? Because I have similar code in billing as a source

html/pfappserver/root/authentication/source/type/HTTP.tt
julsemaan, line 10: Will not work as the host configuration doesn't use default rendering and puts it in one line.

I could put the rest of the attributes in a definition but it wouldn't be reprensentative of the whole and could lead to confusion.

Unless there is a way to accomplish it using definition

conf/httpd.conf.d/httpd.admin
julsemaan: I chose PHP as it is the most common web language and is the one we see most often.

Putting it in the admin is an error. It's a leftover from my dev environment.

addons/external_auth/README.asciidoc
julsemaan: agreed

html/pfappserver/root/authentication/source/read.tt
julsemaan: agreed also

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/HTTP.pm
julsemaan, line 101: Done

html/pfappserver/root/authentication/source/type/HTTP.tt
jrouzierinverse, line 10: There may be a way but it would probably require too much digging into HTML::FormHandler for an easy work around so just leave it as is

julsemaan:
Good to re-review

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,266 @@
...
+
+Method used to build a basic curl object
+
+=cut
+
+sub _post_curl {
+    my ($self, $uri, $post_fields) = @_;
+    my $logger = get_logger;
+
+    $uri = $self->protocol."://".$self->host.":".$self->port.$uri;
```
jrouzierinverse: If the Authentication or Authorization URL does not have a '/' in the front then it will create a invalid url.
http://host:8080auth
Instead of 
http://host:8080/auth

So either inform the user or prepend the '/'

jrouzierinverse:
Outside of my comments about Authentication or Authorization URL it works.
I attached the test service I used to test

```
starman --port 10000 test.psgi
```

test.psgi

```
use JSON::XS;
my $app = sub {
    my $env = shift;
    if ($env->{PATH_INFO} eq '/authentication') {
        return [
            '200',
            ['Content-Type' => 'text/json'],
            [encode_json({result => 1, message => "Valid user name and password"})],    # or IO::Handle-like object
        ];
    }
    my $data = {
        access_duration => '1D',
        access_level    => 'ALL',
        sponsor         => 1,
        unregdate       => '2030-01-01',
        category        => 'default'
    };
    return [
        '200',
        ['Content-Type' => 'text/json'],
        [encode_json($data)],    # or IO::Handle-like object
    ];
};
```

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,275 @@
...
+
+result should be 1 for success, 0 for failure
+message should be the reason it succeeded or failed
+
+Example response hash :
+{
+  result => 1,
+  message => "Valid username and password",
+}
+
```
jrouzierinverse, line 127: A json example might be better

lib/pf/Authentication/Source/HTTPSource.pm
```diff
@@ -0,0 +1,275 @@
...
+The actions are defined by the API through it's JSON response.
+
+Sample response hash, note that not all are necessary, only send back what you need.
+{
+  access_duration => '1D',
+  access_level => 'ALL',
+  sponsor => 1,
+  unregdate => '2030-01-01',
+  category => 'default'
+}
```
jrouzierinverse: A json example might be better

lib/pf/Authentication/Source/HTTPSource.pm
julsemaan: Agreed, I'll fix

julsemaan:
Re-re-review :)

jrouzierinverse:
They are valid but not expected behavior.
I think we should treat 'path' and '/path' as the same or we document the format.

With that said If someone puts '//path' then we should allow them to do so.

jrouzierinverse:
I also tested this against the admin gui and it works as expected.

julsemaan:
Everything should be ok now.

Doc needs reviewing though

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/HTTP.pm
```diff
@@ -0,0 +1,97 @@
...
+   required => 1,
+  );
+has_field 'authorization_url' =>
+  (
+   type => 'Text',
+   label => 'Authorization URL',
+   required => 1,
+  );
+=head1 COPYRIGHT
+
```
jrouzierinverse, line 78: Add help for both Authentication and Authorization url to explain about the '/'.

jrouzierinverse, line 78: Add a space between each field definition.

julsemaan, line 78: We talked about it to say it belongs in the admin guide and it is there now.
I find putting it here is giving too much details for contextual help

jrouzierinverse, line 78: - Note the uri is automatically prefixed by a slash.

julsemaan, line 78: I know it's not much putting it there. I'm mostly saying it belongs in documentation rather than in the frontend

jrouzierinverse, line 78: The problem is most of the time most people will not have the documentation in front of them when they are configuring PacketFence.

Providing a little help when the user needs it in this case the Admin GUI is a big plus+.

What do we lose by adding the help in the GUI?

julsemaan, line 78: I agree at 95% with you but this is a special case that requires to read the documentation as nobody will be able to configure that http source without looking at the documentation.

As it's a corner-case that double slashes cause issues as normal web servers just don't care if they are there, I don't see the point of documenting it in the front-end

julsemaan:
Re-review

PR #806, opened 2015-07-27
complete removal of statsd config

julsemaan:
# Description

Complete removal of statsd host and port configuration in pf.conf as it shouldn't be used anymore
# Impacts

statsd
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Moved statsd host and port configuration in a constants file instead of the configuration
# UPGRADE

`monitoring.statsd_host` and `monitoring.statsd_port` have been removed from pf.conf. If you have specified a specific host or port, remove them from your configuration and change them in `/usr/local/pf/lib/pf/StatsD.pm`

### DIFF
--- a/lib/pf/services/manager/statsd.pm
+++ b/lib/pf/services/manager/statsd.pm
@@ -32,13 +32,6 @@ sub generateConfig {
     my %tags;
     $tags{'template'}      = "$conf_dir/monitoring/statsd_config.js";
     $tags{'pid_file'}      = "$install_dir/var/run/statsd.pid";
-    $tags{'graphite_host'} = "$Config{'monitoring'}{'graphite_host'}";
-    $tags{'graphite_port'} = "$Config{'monitoring'}{'graphite_port'}";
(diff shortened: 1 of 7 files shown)


lib/pf/services/manager/statsd.pm
```diff
@@ -32,13 +32,6 @@ sub generateConfig {
     my %tags;
     $tags{'template'}      = "$conf_dir/monitoring/statsd_config.js";
     $tags{'pid_file'}      = "$install_dir/var/run/statsd.pid";
-    $tags{'graphite_host'} = "$Config{'monitoring'}{'graphite_host'}";
```
fdurand, line 35: statd ok but why graphite ?

julsemaan, line 35: because that is not even used in the template. thought I might as well cleanup useless stuff at the same time

fdurand, line 35: Yes i've seen that after my comment.

louismunro:
Let's do it.

PR #808, opened 2015-07-27
New module pf::condition_parser for parsing conditions

jrouzierinverse:
## Description

Function to parse the following BNF

EXPR = OR || OR
EXPR = OR
OR   = FACT && FACT
OR   = FACT
FACT = '(' EXPR ')'
FACT = ID
ID   = /a-zA-Z0-9_/+

my $array = parse_condition_string('(a || b) && (c || d)');

Where $array would be structure

$array = [
          'AND',
          [
            'OR',
            'a',
            [
              'AND',
              'b',
              'd'
            ]
          ],
          [
            'OR',
            'c',
            'd'
          ]
        ];
## Impacts

Nothing
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
## Delete branch after merge

NO


lib/pf/condition_parser.pm
```diff
@@ -0,0 +1,169 @@
...
+}
+
+=head2 _parse_expr
+
+Handle an 'expr' expression
+
+=cut
+
+sub _parse_expr {
+# EXPR = OR || OR
```
julsemaan: Fix tabulation

lib/pf/condition_parser.pm
```diff
@@ -0,0 +1,169 @@
...
+}
+
+=head2 _parse_or
+
+Handle an 'or' expression
+
+=cut
+
+sub _parse_or {
+# OR   = FACT && FACT
```
julsemaan: Tabulation

lib/pf/condition_parser.pm
```diff
@@ -0,0 +1,169 @@
...
+}
+
+=head2 _parse_fact
+
+Handle a 'fact' expression
+
+=cut
+
+sub _parse_fact {
+# FACT = '(' EXPR ')'
```
julsemaan: Tabulation

julsemaan:
Other than the tabulation that looks good to me

jrouzierinverse:
Please re-review

lib/pf/condition_parser.pm
```diff
@@ -0,0 +1,169 @@
...
+                ]
+              ],
+              [
+                'OR',
+                'c',
+                'd'
+              ]
+            ];
+
+If an invalid string is passed then it will die
```
julsemaan: Shouldn't it return undef instead ?

We discussed the fact that we should reduce dying and this seems a bit overkill here

jrouzierinverse: It easier to report the parsing error this way.
Also returning undef will make the code a bit more complicated for something this simple.

jrouzierinverse: I thought about what you said and I reworked the interface let me know what you think

jrouzierinverse:
RE-review

PR #819, opened 2015-07-30
Combine the matching set_unreg_date and set_access_duration action against sources

### LINKED ISSUE #816: unreg_date over access_duration
Currently, when matching authentication sources/rules from the portal, we are asking for specific actions in the following order:
- set_role
- set_access_duration
- set_unreg_date

Every action requested is a call to match and the first rule containing the requested action returns it.

The problem is, in this flow, access_duration is considered as "higher" than unreg_date, when in fact, both have the same priority.
We can then see an issue where, in a source being matched with a role and also containing an unreg_date, the second call for access duration will go take it from another rule which is not the expected behavior.
Considering that a rule containing a role MUST contain either an access_duration OR an unreg_date, that may lead to some issue as previously mentionned.

jrouzierinverse:
## Description

Combine the matching set_unreg_date and set_access_duration action against sources
## Impacts

Setting the unreg date of a node using the source actions
## NEWS file entries

Bug Fixes
+++++++++
- set_unreg_date and set_access_duration actions now have the same priority when matching rule and actions

Fixes #816
## Delete branch after merge

NO


lib/pf/authentication.pm
```diff
@@ -245,15 +246,33 @@ If source_id_ref is defined then it will be set to the matching source_id
 
 =cut
 
+our %ALLOW_ACTIONS_TYPES = (
```
dwlfrth: Shouldn't that goes in pf::Authentication::constants ?

lib/pf/authentication.pm
```diff
@@ -245,15 +246,33 @@ If source_id_ref is defined then it will be set to the matching source_id
...
+    $Actions::MARK_AS_SPONSOR  => {$Actions::MARK_AS_SPONSOR  => 1},
+    $Actions::SET_ACCESS_LEVEL => {$Actions::SET_ACCESS_LEVEL => 1},
+    $Actions::SET_ROLE         => {$Actions::SET_ROLE         => 1},
+    $Actions::SET_UNREG_DATE   => {
+        $Actions::SET_UNREG_DATE      => 1,
+        $Actions::SET_ACCESS_DURATION => 1,
+    }
+);
+
+our %ACTION_VALUE_FILTERS = (
```
dwlfrth, line 249: Shouldn't that goes in pf::Authentication::constants ?

jrouzierinverse, line 249: No because it would mean that pf::Authentication::constants would have to import pf::config

lib/pf/authentication.pm
jrouzierinverse: Yes

lib/pf/authentication.pm
```diff
@@ -245,15 +246,33 @@ If source_id_ref is defined then it will be set to the matching source_id
...
+
 sub match {
     my ($source_id, $params, $action, $source_id_ref) = @_;
     my ($actions, @sources);
-
-    $logger->debug("Match called with parameters ".join(", ", map { "$_ => $params->{$_}" } keys %$params));
-
+    $logger->debug( sub { "Match called with parameters ".join(", ", map { "$_ => $params->{$_}" } keys %$params) });
+    if( defined $action && !exists $ALLOW_ACTIONS_TYPES{$action}) {
+        $logger->warn("Calling match with an invalid $action");
```
dwlfrth: Calling match with an invalid action '$action'

dwlfrth:
Tested and working as expected. :rocket: 
Ready to merge once comments are 'resolved'
Also, little mention if we could 'stick' to a coding convention... 
- (pf::Authentication) line 274, 288
- multiple spaces at multiple places...

jrouzierinverse:
For line 274 the preference is an uncuddled else (unless I am wrong) so I changed it since I was modify the code.

For line 288 I am not sure what is the issue is with the change since you did not exactly explain exactly your complaint.

multiple spaces at multiple places ?? Can please what you mean so I can fix it.

dwlfrth:
Uncuddled else is usually used when need to comment the behavior of that 'else'. Outside than that, we are generally using standard cuddled else...

Line 288: why is there a new line. We were trying to enforce 120 characters per row a long time ago but that's been kind of... deprecated

jrouzierinverse:
Please re-review

PR #825, opened 2015-08-03
returning source error message instead of generic one

julsemaan:
# Description

Return the error message from the auth source instead of a generic 'wrong username/password' if the source provides one
# Impacts

login on the portal
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Allow precise error messages from the authentication source when providing invalid credentials on the captive portal

### DIFF
--- a/lib/pf/Authentication/Source/LDAPSource.pm
+++ b/lib/pf/Authentication/Source/LDAPSource.pm
@@ -9,6 +9,7 @@ pf::Authentication::Source::LDAPSource
 =cut
 
 use pf::constants qw($TRUE $FALSE);
+use pf::constants::authentication::messages;
 use pf::Authentication::constants;
 use pf::Authentication::Condition;
 use pf::CHI;
@@ -75,13 +76,13 @@ sub authenticate {
   my ($connection, $LDAPServer, $LDAPServerPort ) = $self->_connect();
 
   if (!defined($connection)) {
-    return ($FALSE, 'Unable to validate credentials at the moment');
+    return ($FALSE, $COMMUNICATION_ERROR_MSG);
   }
   my $result = $self->bind_with_credentials($connection);
 
   if ($result->is_error) {
     $logger->error("[$self->{'id'}] Unable to bind with $self->{'binddn'} on $LDAPServer:$LDAPServerPort");
-    return ($FALSE, 'Unable to validate credentials at the moment');
+    return ($FALSE, $COMMUNICATION_ERROR_MSG);
   }
 
   my $filter = "($self->{'usernameattribute'}=$username)";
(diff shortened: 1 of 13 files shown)


jrouzierinverse:
If there are multiple sources and all of them failed which error message should we use?

julsemaan:
Right now, it uses the last one and I don't see how we could go around that as our circular search through all the sources will not always show the right message of the array

Coupling this with #839 will show the right error message though

If you have a good idea, I'm open for it

jrouzierinverse:
The only idea I would have is to show them the error message.

What problem are you trying to solve with the pull request?

julsemaan:
Well right now, when you're doing null auth and you enter an invalid e-mail, it uses the same generic message instead of using the one of the null source. Meaning an honnest mistake in an e-mail will show you 'Invalid username/password' which is not the case at all.

jrouzierinverse:
The work I did in the branch feature/null-auth-non-exclusive will solve that problem by treating a null source like an external source

julsemaan:
Doesn't solve the fact that we're using the same error message everywhere though.

Let's say there is an LDAP connection error. We should be showing that with along with a 'Please try again later' instead of showing 'Invalid username/password' when it was actually right.

We do have the info messages populated in authentication so why not leverage them to make the portal more user friendly...

jrouzierinverse:
That make sense.
However at the moment each source have different error message for the same type of event.

All sources should return the same message for the same type of event.

Invalid credentials should say.

'Wrong username or password.'

When there is a connection problem.

'Please try again later'

julsemaan:
Correct, this part is still to-do on my side before this hits devel

julsemaan:
Messages reworked, re-review

lib/pf/authentication.pm
```diff
@@ -255,8 +256,8 @@ sub authenticate {
         }
     }
 
-    $logger->trace("Authentication failed for '$username' for all ".scalar(@sources)." sources");
-    return ($FALSE, 'Wrong username or password.');
+    $logger->trace("Authentication failed for '$display_username' for all ".scalar(@sources)." sources");
+    return ($FALSE, $message ? $message : 'Wrong username or password.');
 }
```
jrouzierinverse, line 261: All success/error messages should be in constants.

julsemaan:
Re-review

lib/pf/constants/authentication.pm
```diff
@@ -26,6 +26,15 @@ our @SOURCES = __PACKAGE__->sources();
...
 
 
+=head2 Messages
+
+Replies for authentication success/error
+
+=cut
+
+Readonly our $INVALID_CREDENTIALS => ;
+Readonly our $AUTH_SUCESS
```
jrouzierinverse: This looks incomplete

jrouzierinverse:
where is the module for pf::constants::authentication::messages

julsemaan:
Fixed

lib/pf/authentication.pm
```diff
@@ -242,6 +242,7 @@ sub authenticate {
 
     $logger->debug(sub {"Authenticating '$username' from source(s) " . join( ', ', map { $_->id } @sources ) });
 
+    my $message;
     foreach my $current_source (@sources) {
         my ($result, $message);
```
jrouzierinverse: This message overrides the one above

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -93,15 +94,15 @@ sub authenticate {
...
   if ($result->is_error) {
     $logger->error("[$self->{'id'}] Unable to execute search $filter from $self->{'basedn'} on $LDAPServer:$LDAPServerPort");
-    return ($FALSE, 'Unable to validate credentials at the moment');
+    return ($FALSE, $COMMUNICATION_ERROR_MSG);
   }
 
   if ($result->count == 0) {
     $logger->warn("[$self->{'id'}] No entries found (". $result->count .") with filter $filter from $self->{'basedn'} on $LDAPServer:$LDAPServerPort");
-    return ($FALSE, 'Invalid login or password');
+    return ($FALSE, $COMMUNICATION_ERROR_MSG);
```
jrouzierinverse: This mean there are no user not a communication error

PR #853, opened 2015-08-20
Fix/851 omapi config is bogus

### LINKED ISSUE #851: OMAPI configuration is bogus
OMAPI configuration is wrongly built.
If enabled (ip2mac and/or mac2ip with OMAPI), the configuration will be built (only the omapi-port section) even if there's no key.
That actually means it will work but OMAPI will listen on 0.0.0.0 without being secured.

dwlfrth:
# Description

Do not configure OMAPI dhcpd side if no key is provided, even if the feature is enabled
# Impacts

iplog (ip2mac, mac2ip)
# Issue

fixes #851 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Now generates default OMAPI key on install time
## Bug Fixes
- Issue with dhcpd side OMAPI configuration being unsecure (#851)

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -585,7 +585,7 @@ fi
 
 # Create OMAPI key
 if [ ! -f /usr/local/pf/conf/pf_omapi_key ]; then
-    echo $(date) $HOSTNAME | base64 > /usr/local/pf/conf/pf_omapi_key
+    /usr/bin/openssl rand -base64 -out /usr/local/pf/conf/pf_omapi_key 32
 fi
 
 for service in snortd httpd snmptrapd memcached portreserve
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -583,6 +583,10 @@ if [ ! -f /usr/local/pf/conf/ssl/server.crt ]; then
     cat /usr/local/pf/conf/ssl/server.crt /usr/local/pf/conf/ssl/server.key > /usr/local/pf/conf/ssl/server.pem
 fi
 
+# Create OMAPI key
+if [ ! -f /usr/local/pf/conf/pf_omapi_key ]; then
+    echo $(date) $HOSTNAME | base64 > /usr/local/pf/conf/pf_omapi_key
+fi
 
 for service in snortd httpd snmptrapd memcached portreserve
 do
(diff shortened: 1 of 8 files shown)


julsemaan:
I even think this is maintenance

dwlfrth:
Please review

lib/pf/config.pm
```diff
@@ -830,6 +830,23 @@ sub is_omapi_enabled {
...
+=cut
+
+sub is_omapi_configured {
+    return $FALSE unless $Config{'omapi'}{'host'} eq "localhost";
+
+    if ( ($Config{'omapi'}{'key_name'} && $Config{'omapi'}{'key_name'} ne '') && ($Config{'omapi'}{'key_base64'} && $Config{'omapi'}{'key_base64'} ne '') ) {
+        return $TRUE;
+    }
+
+    $logger->warn("OMAPI lookup is locally enabled but missing required configuration parameters 'key_name' and/or 'key_base64'");
```
fdurand, line 846: A detail but key_base64will always be set

dwlfrth, line 846: Yep but it is a pretty generic message... It is covering up if someone messes with the default one... and since it is the values we are checking I output them

addons/packages/packetfence.spec
```diff
@@ -583,6 +583,10 @@ if [ ! -f /usr/local/pf/conf/ssl/server.crt ]; then
     cat /usr/local/pf/conf/ssl/server.crt /usr/local/pf/conf/ssl/server.key > /usr/local/pf/conf/ssl/server.pem
 fi
 
+# Create OMAPI key
+if [ ! -f /usr/local/pf/conf/pf_omapi_key ]; then
+    echo $(date) $HOSTNAME | base64 > /usr/local/pf/conf/pf_omapi_key
```
fdurand: Can we use something a little bit random.
Because if i want to crash the dhcpd service i just have do a brute force attack with a hash of a date and the hostname (it can be found).

fdurand: Something like that:
dd if=/dev/urandom bs=16 count=1 2>/dev/null | openssl enc -e -base64

extrafu: I would suggest something like that:

/usr/bin/openssl rand -base64 -out /usr/local/pf/conf/pf_omapi_key 32

Much more random, cleaner, faster, zammitier.

dwlfrth:
Re-Review now that it has been Zammitified :muscle:

PR #855, opened 2015-08-21
Feature/oauth lookup

julsemaan:
# Description

Adds back e-mail address lookup into Facebook source
Adds e-mail address lookup into Github source
Makes sure that e-mail that is fetched during oauth transaction gets recorded in the email field of the person table instead of just the pid
# Impacts

OAUTH2 workflow, facebook source
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added e-mail address lookup for Github and Facebook sources
## Enhancements
- E-mail addresses obtained through OAuth sources (Google, Facebook, LinkedIn, Twitter, Github, Windows Live) will be now recorded in the email field.
# Upgrade guide

The Facebook API now requires to specify the fields to be specified in the query.
In all your facebook sources, change the parameter `protected_resource_url` to `https://graph.facebook.com/me?fields=id,name,email,first_name,last_name`

Change the parameter `scope` to `user,user:email` in all your Github sources as PacketFence is now fetching the email address of the user when registering with Github

### DIFF
--- a/lib/pf/Authentication/Source/LinkedInSource.pm
+++ b/lib/pf/Authentication/Source/LinkedInSource.pm
@@ -11,6 +11,7 @@ pf::Authentication::Source::LinkedInSource
 use WWW::Curl::Easy;
 use JSON qw( decode_json );
 use Moose;
+use pf::person;
 extends 'pf::Authentication::Source::OAuthSource';
 
 has '+type' => (default => 'LinkedIn');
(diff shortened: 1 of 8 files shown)


lzammit:
Doesn't bring the right email. Into packetfence I've email@facebook.com.

lzammit:
Works!

lib/pf/Authentication/Source/LinkedInSource.pm
```diff
@@ -27,6 +28,17 @@ has 'redirect_url' => (isa => 'Str', is => 'rw', required => 1, default => 'http
 has 'domains' => (isa => 'Str', is => 'rw', required => 1, default => 'www.linkedin.com,api.linkedin.com,static.licdn.com');
 has 'create_local_account' => (isa => 'Str', is => 'rw', default => 'no');
 
+=head2 lookup_from_provider_info
+
+Lookup the person information from the authentication hash received during the OAuth process
+
+=cut
+
+sub lookup_from_provider_info {
```
louismunro, line 37: Not sure the name is well chosen here.
lookup does not imply modify in my mind.

julsemaan, line 37: I know, I put it this way because of pf::lookup::person which populates from LDAP for consistency

I have nothing against changing it though

louismunro, line 37: I think pf::lookup::person is wrong then.
Let's not dig ourselves in even deeper...

julsemaan, line 37: No problem, what do you suggest as a name ?

jrouzierinverse, line 37: update_person_from_provider_info

louismunro, line 37: maybe just update_from_provider_info? 
I am not trying to be picky, just to make it explicit that this updates the DB.

PR #859, opened 2015-08-27
Feature/filter from ap

whitx:
# Description

Allow to create a filter from AP, useful if you want to have different portal profile from the same controller
# Impacts

Profile Filter
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Filter from AP
## Bug Fixes
- Correct some typo in other filter

### DIFF
--- a/lib/pf/factory/condition/profile.pm
+++ b/lib/pf/factory/condition/profile.pm
@@ -34,6 +34,7 @@ our %PROFILE_FILTER_TYPE_TO_CONDITION_TYPE = (
     'realm'               => {type => 'equals',        key  => 'realm'},
     'ssid'                => {type => 'equals',        key  => 'last_ssid'},
     'switch'              => {type => 'exists_in',     key  => 'last_switch'},
+    'switch_mac'          => {type => 'equals',        key  => 'last_switch_mac'},
     'switch_port'         => {type => 'couple_equals', key1 => 'last_switch', key2 => 'last_port'},
     'uri'                 => {type => 'equals',        key  => 'last_uri'},
     'vlan'                => {type => 'equals',        key  => 'last_vlan'},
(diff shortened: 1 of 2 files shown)


whitx:
I might have add some unnecessary code /discuss

jrouzierinverse:
As far as I know in the locationlog of when last_ap (switch_mac) is defined is should be equal to last_switch.

How is the ap profile filter different than the switch profile filter?

whitx:
If I am not mistaken, this allow you to create a filter from the AP without having to add all the AP to the switch.conf, since the last_switch bring you the switch you connect to, and last_ap just the mac you connected to.
In the case of a controller with multiple AP, you can easily create profile filter from those AP.

jrouzierinverse:
Since it is possible for the ip address to be dynamic I think filtering by the switches ip address would be more useful.

julsemaan:
I know you just changed it so I feel bad but I'd make it last_switch_mac and not last_ap_mac to match the column in the locationlog which is switch_mac

:nail_care: \* 1000

But it's a good idea to have it, me like it

whitx:
The issue raised, after trying to use it, is that this is really heavy maintenance wise, since you have to maintain a list of all mac addresses of AP you want to filter. The use would be specific in the case when you have a wireless controller, and AP connects to it.

julsemaan:
I'm not sure why this was closed...

This seems like a reasonable addition to the filters and is requested right now by another user

julsemaan:
I will take the ownership and complete it

whitx:
I think it was ready, might need some correction with the rebase..

jrouzierinverse:
I think is a candidate for the exists_in filter type like switch in the `pf::factory::condition::profile`

fdurand:
Why don't we merge that ?

julsemaan:
I agree, will rebase and we'll merge

julsemaan:
@fdurand 
A last quick review and we merge ?

PR #870, opened 2015-09-02
Don't enter multi-domain unless it's configured

### LINKED ISSUE #868: Don't do multi-domain unless there is a domain configured
Also add the possibility to disable it even if there are domains configured so that when domains config in PacketFence will be used for something else than FreeRADIUS, it still won't do it

Idea is to make it conditional in the template generation on service start.

Also need to add message to restart FreeRADIUS after a change

julsemaan:
# Description

Doesn't activate the multi domain FreeRADIUS module unless there are domains configured in PacketFence

Also adds the possibility to disable it even if there are domains configured so that when domains config in PacketFence will be used for something else than FreeRADIUS, it still won't do it

Remove the call to packetfence-multi-domain in the outer tunnel
# Impacts

FreeRADIUS domain authentication
# Issue

fixes #868 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Will not activate the packetfence multi-domain FreeRADIUS module unless there are domains configured in PacketFence
- Added option to disable the packetfence multi-domain FreeRADIUS module from the admin interface

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1103,6 +1103,14 @@ The cost factor to apply to the password hashing if applicable.
 Currently only applies to bcrypt.
 EOT
 
+[advanced.disable_pf_domain_auth] 
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Disables authentification on the domains defined in PacketFence.
+This will make it so the authentication only happens on the domain binded to the OS.
+EOT
+
 [omapi.ip2mac_lookup]
 type=toggle
 options=enabled|disabled
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1107,8 +1107,7 @@ EOT
 type=toggle
 options=enabled|disabled
 description=<<EOT
-Disables authentification on the domains defined in PacketFence.
-This will make it so the authentication only happens on the domain binded to the OS.
+Disable Active-Directory domain configuration from within PacketFence
 EOT
 
 [omapi.ip2mac_lookup]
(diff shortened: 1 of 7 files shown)


conf/documentation.conf
```diff
@@ -1103,6 +1103,14 @@ The cost factor to apply to the password hashing if applicable.
 Currently only applies to bcrypt.
 EOT
 
+[advanced.disable_pf_domain_auth] 
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Disables authentification on the domains defined in PacketFence.
+This will make it so the authentication only happens on the domain binded to the OS.
```
louismunro: Not sure this message is clear enough for a user that is not familiar with the FreeRADIUS/PacketFence/Winbind internals.
Maybe we should say something like 'disable Active-Directory domain configuration from within PacketFence'.

conf/pf.conf.defaults
```diff
@@ -851,6 +851,12 @@ hash_passwords=bcrypt
 # The cost factor to apply to the password hashing if applicable.
 # Currently only applies to bcrypt.
 hashing_cost=8
+#
+# advanced.disable_pf_domain_auth
+#
+# Disables authentification on the domains defined in PacketFence
+# This will make it so the authentication only happens on the domain binded to the OS.
```
louismunro: Same here.

louismunro:
LGTM apart from minor comments and documentation nitpicking.

louismunro:
Please fix the (minor) comment and documentation issues.
I will merge as soon as that is done.

julsemaan:
Sorry forgot about this one, I'll fix in the next minutes

julsemaan:
Fixaide

PR #871, opened 2015-09-02
Allow for pfconfig to fork

julsemaan:
# Description

Allows for pfconfig to fork and serve concurent request in order to make configuration access faster when having a lot of disposable threads/processes
# Impacts

Configuration access, basically pretty much PacketFence as a whole
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- pfconfig is now able to process multiple requests concurently via process forking

### DIFF
--- a/conf/pfconfig.conf.example
+++ b/conf/pfconfig.conf.example
@@ -1,3 +1,6 @@
+[general]
+childs=2
+
 [mysql]
 host=localhost
 user=pf
--- a/conf/pfconfig.conf.example
+++ b/conf/pfconfig.conf.example
@@ -1,5 +1,5 @@
 [general]
-childs=2
+children=2
 
 [mysql]
 host=localhost
--- a/sbin/pfconfig
+++ b/sbin/pfconfig
@@ -128,44 +128,100 @@ our %DISPATCH = (
     'sleep'              => \&server_sleep,
 );
 
-while($RUNNING) {
-    my $socket;
-    my $line;
-    eval {
-        $socket = $listner->accept();
-        #Check if a signal was caught
-        unless (defined $socket || $! == EINTR) {
-            die("Can't accept connection: $!\n");
-        }
-        if($socket) {
-            chomp( $line = <$socket> );
 
-            if($line eq "exit") {exit}
+my $configuration = pfconfig::config->new;
 
-            my $query = decode_json($line);
-            #use Data::Dumper;
-            #print Dumper($query);
+our $PARENT_PID = $$;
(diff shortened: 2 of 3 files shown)


conf/pfconfig.conf.example
```diff
@@ -1,3 +1,6 @@
+[general]
+childs=2
```
extrafu: children

jrouzierinverse:
Expiration is a problem.

The following scenario

Master
- Child 1
- Child 2

A Process sends the command to expire the cache.
Child1 gets the request.
Child1 then updates his cache

Child2 still has the older data.

It would be best to have the master process expire the cache.
Then fork off new children then kill off the old ones.

julsemaan:
It doesn't matter because the control files are touched so all processes binded to the namespaces get the signal of the expiration and reload their memory.

Like when you do configreload, it doesn't even send a signal the pfconfig process, it just expires the control files and that makes the pfconfig process reload it's memory cache

sbin/pfconfig
```diff
@@ -128,44 +128,100 @@ our %DISPATCH = (
...
 }
 
-$logger->info("Stop running\n");
+sub is_parent_alive {
+    kill (0,$PARENT_PID)
+}
+
+start_children();
+while($RUNNING){
+    sleep 1;
```
jrouzierinverse, line 216: A nice to have would be for this to is not sleep but pause (using POSIX::pause).
This would avoid unnecessary work since it will only checks for dead children if a signal arrived.
But this also means handling the child signal instead of just ignoring it.

You can look at pfmon for example code

julsemaan, line 216: I personally prefer active checking over signals for the childs as we can execute more actions during checkups should we need to like spawn more childs dynamically

Don't know what the others think about each implementation but we could discuss both to see which one to use

jrouzierinverse, line 216: Do you mean increase the number of child process running?
For example If previously the number of child process was 2 and you want to increase to 3 without restarting pfconfig.

julsemaan, line 216: Correct

jrouzierinverse, line 216: I think we should sit down and discuss the different options to obtain that goal using the best practices for multi process daemons.

julsemaan, line 216: Fully agree with that

julsemaan, line 216: After discussion, this will be reworked with the process forking superclass.
So stays as is.

See #894

julsemaan:
Should be good to merge I think

PR #874, opened 2015-09-03
pfconfig subcache amelioration

julsemaan:
# Description

Use the subcache in all relevant pfconfig::cached subclasses calls
Also do not constantly call pfconfig on undef key
Use Time::HiRes for the control file timestamp instead of POSIX
Added tests for subcache + subcache computing
# Impacts

Configuration in the processes (pfconfig clients), doesn't change the pfconfig server part
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added additional in-process caching for pfconfig proxied configuration

### DIFF
--- a/lib/pfconfig/cached.pm
+++ b/lib/pfconfig/cached.pm
@@ -145,6 +145,12 @@ sub set_in_subcache {
 
 }
 
+=head2 compute_from_subcache
+
+Get an element of the subcache or compute it's value and store it in the subcache
+
+=cut
+
 sub compute_from_subcache {
     my ($self, $key, $on_miss) = @_;
 
--- a/lib/pfconfig/cached.pm
+++ b/lib/pfconfig/cached.pm
@@ -34,9 +34,8 @@ use IO::Socket::UNIX qw( SOCK_STREAM );
 use JSON;
 use pfconfig::timeme;
 use pfconfig::log;
-use pfconfig::util;
+use pfconfig::util qw($undef_element);
 use pfconfig::constants;
-use pfconfig::undef_element;
 use Sereal::Encoder;
 use Sereal::Decoder;
 use Time::HiRes qw(stat time);
@@ -167,7 +166,7 @@ sub compute_from_subcache {
         $self->set_in_subcache($key,$result);
     }
     else {
-        $self->set_in_subcache($key, pfconfig::undef_element->new);
+        $self->set_in_subcache($key, $undef_element);
     }
 
     return $result;
--- a/lib/pfconfig/cached.pm
+++ b/lib/pfconfig/cached.pm
@@ -36,6 +36,7 @@ use pfconfig::timeme;
 use pfconfig::log;
 use pfconfig::util;
 use pfconfig::constants;
+use pfconfig::undef_element;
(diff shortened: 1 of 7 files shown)


lib/pfconfig/cached.pm
```diff
@@ -144,6 +145,35 @@ sub set_in_subcache {
...
+    elsif(defined($subcache_value)){
+        return $subcache_value;
+    }
+
+    my $result = $on_miss->();
+    if(defined($result)){
+        $self->set_in_subcache($key,$result);
+    }
+    else {
+        $self->set_in_subcache($key, pfconfig::undef_element->new);
```
jrouzierinverse: This does not need to created every time.
This could be a singleton.

lib/pfconfig/cached_array.pm
```diff
@@ -82,13 +82,11 @@ sub FETCH {
...
-
-    my $reply = $self->_get_from_socket("$self->{_namespace};$index");
-    my $result = defined($reply) ? $self->_get_from_socket("$self->{_namespace};$index")->{element} : undef;
-
-    $self->set_in_subcache( $index, $result );
+    my $result = $self->compute_from_subcache($index, sub {
+      my $reply = $self->_get_from_socket("$self->{_namespace};$index");
+      my $result = defined($reply) ? $reply->{element} : undef;
+      return $result;
+    });
```
jrouzierinverse, line 89: I like the compute_from_subcache

lib/pfconfig/manager.pm
```diff
@@ -189,9 +189,8 @@ sub touch_cache {
         close($fh);
     }
     if ( -e $filename ) {
-        sysopen( my $fh, $filename, O_RDWR | O_CREAT );
-        POSIX::2008::futimens( fileno $fh );
-        close($fh);
+        my $command = 'touch --date=@'.time.' '."'".$filename."'";
+        `$command`;
```
louismunro, line 193: What was wrong with the old way?

t/unittest/pfconfig/cached.t
```diff
@@ -0,0 +1,138 @@
...
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+USA.
+
+=cut
+
+1;
+
+
```
louismunro, line 138: +1 for including unit tests.

louismunro:
LGTM. Let's merge it.

lib/pfconfig/manager.pm
extrafu, line 193: Why not use Perl's utime? (http://perldoc.perl.org/functions/utime.html)

julsemaan, line 193: The old way with POSIX doesn't seem to use the same source as Time::HiRes as it was nanoseconds late of the latest.

That meant the processes were being expired a few nanoseconds after which made unit tests fail.

I'll try utime and see if it works with subseconds

julsemaan, line 193: Tested utime and it doesn't handle the microseconds.

PR #879, opened 2015-09-09
Feature/pfconfig sync pp

julsemaan:
# Description

Adds better error propagation on sync failure in bin/cluster/sync
Admin will show a failure when saving and the cluster fails to sync completely
Adds an option to fix permissions of specific files in `pfcmd fixpermissions`
Fix the permissions of a file before synching so pf can write to it
# Impacts

Active/active sync
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added better error message propagation during a cluster synchronization
- pfcmd fixpermissions can not be used on specific files
## Bug Fixes
- Fix bad permissions before synching a file in a cluster

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -480,7 +480,8 @@ sub notify_configfile_changed : Public {
 
     my $apiclient = pf::api::jsonrpcclient->new(proto => 'https', host => $master_server->{management_ip});
 
-    open(my $fh, '>', $postdata{conf_file}) or die "Cannot open file $postdata{conf_file} for writing. This is excessively bad. Run '/usr/local/pf/bin/pfcmd fixpermissions'";
+    pf::util::pf_run("sudo /usr/local/pf/bin/pfcmd fixpermissions file $postdata{conf_file}");
+    open(my $fh, '>', $postdata{conf_file}) or die "Cannot open file $postdata{conf_file} for writing.'";
     eval {
         my %data = ( conf_file => $postdata{conf_file} );
         my ($result) = $apiclient->call( 'download_configfile', %data );
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -480,9 +480,9 @@ sub notify_configfile_changed : Public {
 
     my $apiclient = pf::api::jsonrpcclient->new(proto => 'https', host => $master_server->{management_ip});
 
-    pf::util::pf_run("sudo /usr/local/pf/bin/pfcmd fixpermissions file $postdata{conf_file}");
-    open(my $fh, '>', $postdata{conf_file}) or die "Cannot open file $postdata{conf_file} for writing.'";
     eval {
+        pf::util::pf_run("sudo /usr/local/pf/bin/pfcmd fixpermissions file $postdata{conf_file}");
+        open(my $fh, '>', $postdata{conf_file}) or die "Cannot open file $postdata{conf_file} for writing.'";
         my %data = ( conf_file => $postdata{conf_file} );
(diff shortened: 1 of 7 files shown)


lib/pf/cmd/pf/fixpermissions.pm
```diff
@@ -18,19 +18,20 @@ pf::cmd::pf::fixpermissions
 use strict;
```
rouzier, line 21: Update pod doc to reflect new options

lib/pf/cmd/pf/fixpermissions.pm
```diff
@@ -39,6 +40,16 @@ sub _run {
     chmod(0664, @stored_config_files);
     chmod(02775, $conf_dir, $var_dir, $log_dir);
     _fingerbank();
+    print "Fixed permissions.\n";
+    return $EXIT_SUCCESS;
+}
+
+sub action_file {
```
jrouzierinverse, line 79: Create a parse_file method to check if the file exists

lib/pf/api.pm
```diff
@@ -480,10 +480,11 @@ sub notify_configfile_changed : Public {
 
     my $apiclient = pf::api::jsonrpcclient->new(proto => 'https', host => $master_server->{management_ip});
 
+    pf::util::pf_run("sudo /usr/local/pf/bin/pfcmd fixpermissions file $postdata{conf_file}");
+    open(my $fh, '>', $postdata{conf_file}) or die "Cannot open file $postdata{conf_file} for writing.'";
```
jrouzierinverse: Should re-throw error below

jrouzierinverse:
As a safety measure we should only allow the api to modify/retrieve allowed files in conf directories.

lib/pf/cmd/pf/fixpermissions.pm
jrouzierinverse, line 79: It would be nice to accept multiple files.

lib/pf/cmd/pf/fixpermissions.pm
```diff
@@ -39,6 +40,16 @@ sub _run {
...
+    my ($self) = @_;
+    my ($file) = $self->action_args;
+    $file = untaint_chain($file);
```
jrouzierinverse: Check to see if the file path is in /usr/local/pf or /usr/local/fingerbank

lib/pf/api.pm
julsemaan: Why ? It's already going up the chain and generating a 500

lib/pf/cmd/pf/fixpermissions.pm
```diff
@@ -39,6 +43,52 @@ sub _run {
...
+        }
+        unless($file =~ /\/usr\/local\/pf\// || $file =~ /\/usr\/local\/fingerbank\//){
+            print STDERR "File $file is not in an allowed directory \n";
+            return 0;
+        }
+    }
+    return 1;
+}
+
+sub action_file {
```
jrouzierinverse, line 79: doc

lib/pf/cmd/pf/fixpermissions.pm
```diff
@@ -18,19 +21,20 @@ pf::cmd::pf::fixpermissions
...
 
 use fingerbank::FilePath;
 
 use File::Spec::Functions qw(catfile);
 
-sub parseArgs { 1 }
+sub default_action { 'all' }
 
-sub _run {
+sub action_all {
```
jrouzierinverse, line 43: doc

julsemaan:
Re-review

PR #880, opened 2015-09-10
add api call to release violations for a node

julsemaan:
# Description

Adds API call to release all violations for a node
# Impacts

Nothing if the code compiles I guess...
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added API call to release all violations of a node

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -388,6 +388,25 @@ sub trigger_violation : Public {
     return (pf::violation::violation_trigger($postdata{'mac'}, $postdata{'tid'}, $postdata{'type'}));
 }
 
+=head2 release_all_violations
+
+Release all violations for a node
+
+=cut
+
+sub release_all_violations : Public {
+    my ($class, $mac) = @_;
+    my $logger = pf::log::get_logger;
+    die "Missing MAC address" unless($mac);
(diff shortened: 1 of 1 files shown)


lib/pf/api.pm
```diff
@@ -388,6 +388,26 @@ sub trigger_violation : Public {
...
+
+sub release_all_violations : Public {
+    my ($class, $mac) = @_;
+    my $logger = pf::log::get_logger;
+    $mac = pf::util::clean_mac($mac);
+    die "Missing MAC address" unless($mac);
+    my $closed_violation = 0;
+    foreach my $violation (pf::violation::violation_view_open($mac)){
+        $logger->info("Releasing violation $violation->{vid} for $mac though release_all_violations");
+        pf::violation::violation_force_close($mac,$violation->{vid});
```
jrouzierinverse: Should check the value of violation_force_close to see if it was successful

lib/pf/api.pm
```diff
@@ -388,6 +388,26 @@ sub trigger_violation : Public {
...
+        $closed_violation = 1;
```
jrouzierinverse: should increment on each successful close

julsemaan:
Re-review

PR #881, opened 2015-09-10
add cluster option to switch dhcp roles

### LINKED ISSUE #817: Add option to make management node primary dhcp server
Will help in routed networks where ip helpers point only to the management IP

julsemaan:
# Description

add cluster option to switch dhcp roles
# Impacts

dhcpd in clustering
# Issue

ie: fixes #817 
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added an option to make the management node the primary DHCP server in an active/active cluster

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1397,6 +1397,13 @@ description=<<EOT
 The virtual router id for keepalive. Leave untouched unless you have another keepalive cluster in this network.
 EOT
 
+[active_active.switch_dhcpd_roles]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Switch dhcpd roles so that server #1 will be primary and server #2 secondary
+EOT
+
 [monitoring.graphite_host]
 type=text
 description=<<EOT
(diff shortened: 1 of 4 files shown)


lzammit:
Works! Tested!

louismunro:
Ready to merge as far as I can tell. 
I wonder what is the real impact though, since ISC dhcpd is a true active/active setup.
None of the servers is more primary than the other.

julsemaan:
From what I recall I had to do it because only the secondary was seeing the DHCP traffic and wasn't acting on it while it was secondary.

When he was primary and the only one to see it, it started responding.

louismunro:
This is odd. 
Either that dhcpd was misconfigured or something else was going on. 
Normally each server should handle half the requests, with a timeout after which it should reply instead of the other if the peer was not responding.

julsemaan:
maybe related to the fact that both were online and healthy on 647 but only one was seeing the traffic

louismunro:
This one looks stale. 
Closing. I don't believe this was truly required in the first place anyway.

PR #882, opened 2015-09-11
Natural vlan filter order + defined test

fdurand:
# Description

Order the vlan filters by natural order and added a test "defined" to test if a value is defined
# Impacts

No
# Delete branch after merge

YES
# NEWS file entries
## New Features
- new defined test for vlan filter
## Bug Fixes
- Natural order for Vlan Filter

### DIFF
--- a/lib/pf/vlan/filter.pm
+++ b/lib/pf/vlan/filter.pm
@@ -50,7 +50,7 @@ sub test {
     my ($self, $scope, $switch,$ifIndex,$mac,$node_info,$connection_type,$user_name,$ssid,$radius_request) = @_;
     my $logger = Log::Log4perl::get_logger( ref($self) );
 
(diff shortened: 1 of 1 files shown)


dwlfrth:
I agree with @julsemaan. Hash is not properly built to honor file ordering.
Also, would be great to have differentiation between action and condition when building the hash.
I have to work related to that on the side, I'll go ahead and push that stuff to override this PR (if it's ok with you) but probably won't be part of 5.4

jrouzierinverse:
Derek any work you on this may be wasted since we will be moving to use the filter engine for the vlan filters here is the PR #889

dwlfrth:
Totally aware of that (we actually discussed that together ;)) and that's why my work is "on the side" and "does not have that much stuff in it".
I'll check once the PR #889 will be merged

PR #883, opened 2015-09-11
Instantiate the correct portal based on the current connection attributes

fdurand:
# Description

Location is updated after we try to compute the vlan so a portal profile based on the node will not be instantiate on the first connection.
# Impacts

No
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix wrong portal profile instantiate based on the first connection.

### DIFF
--- a/lib/pf/vlan.pm
+++ b/lib/pf/vlan.pm
@@ -387,7 +387,7 @@ sub getNormalVlan {
     my $logger = Log::Log4perl->get_logger(__PACKAGE__);
     my $start = Time::HiRes::gettimeofday();
 
-    my $options;
+    my $options = {};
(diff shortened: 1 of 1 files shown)


lib/pf/vlan.pm
```diff
@@ -386,7 +386,16 @@ sub getNormalVlan {
     my ($this, $switch, $ifIndex, $mac, $node_info, $connection_type, $user_name, $ssid, $radius_request, $realm, $stripped_user_name, $autoreg) = @_;
     my $logger = Log::Log4perl->get_logger(__PACKAGE__);
     my $start = Time::HiRes::gettimeofday();
-    my $profile = pf::Portal::ProfileFactory->instantiate($mac);
+
+    my $options;
```
jrouzierinverse: my $options = {};

PR #884, opened 2015-09-11
Safely update the contents of a file using rename in the pf::api

jrouzierinverse:
## Description

To avoid file corruption when updating configuration files through the api.
Use file rename instead of writing directly to the file.
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
- Avoid the possibility of file corruption when updating configuration files through the api 
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -38,6 +38,7 @@ use pfconfig::manager;
 use pf::api::jsonrpcclient;
 use pf::cluster;
 use JSON;
+use File::Slurp;
 use pf::file_paths;
 
 use List::MoreUtils qw(uniq);
@@ -409,7 +410,7 @@ sub release_all_violations : Public {
         else {
             $logger->error("Cannot close violation $violation->{vid} for $mac");
         }
-    } 
+    }
     return $closed_violation;
 }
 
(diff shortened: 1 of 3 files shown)


lib/pf/api.pm
```diff
@@ -522,13 +519,32 @@ sub distant_download_configfile : Public {
...
     return 1;
+}
+
+=head2 _safe_file_update
 
+This safely modifies the contents of a file using a rename
+
+=cut
+
+sub _safe_file_update {
```
julsemaan: Belongs in pf::util much more than here

lib/pf/api.pm
```diff
@@ -522,13 +519,32 @@ sub distant_download_configfile : Public {
...
+    syswrite $temp, $contents;
+    $temp->flush;
+    close $temp;
+    unless( rename ($temp->filename, $file) ) {
+        my $logger = pf::log::get_logger;
+        $logger->error("cannot save contents to $file '$!'");
+        die "cannot save contents to $file";
+    }
+    $temp->unlink_on_destroy(0);
+    `chown pf.pf $file`;
```
julsemaan: That will not work as it also sync fingerbank files
You will need to use the reworked pfcmd fixpermissions

jrouzierinverse: That feature depends on the pull request #879.

lib/pf/api.pm
jrouzierinverse: I will move into pf::util

lib/pf/api.pm
julsemaan: Yes, I would suggest to base off that PR and cite it as a dependency as both will be part of the same release

jrouzierinverse:
Please review

louismunro:
LGTM.

PR #889, opened 2015-09-11
Vlan filter based off the filter engine

jrouzierinverse:
## Description

Vlan filters based off the filter engine depends on feature/filter-engine
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
## Delete branch after merge

NO


jrouzierinverse:
This depends on the filter engine

lib/pf/factory/condition/vlanfilter.pm
```diff
@@ -0,0 +1,107 @@
...
+
+our @MODULES;
+
+sub factory_for {'pf::condition'}
+
+our %VLAN_FILTER_TYPE_TO_CONDITION_TYPE = (
+    'is'        => 'pf::condition::equals',
+    'is_not'    => 'pf::condition::not_equals',
+    'match'     => 'pf::condition::matches',
+    'match_not' => 'pf::condition::not_matches',
```
fdurand: if we can have a defined condition

lib/pf/factory/condition/vlanfilter.pm
```diff
@@ -0,0 +1,107 @@
...
+);
+
+our %VLAN_FILTER_KEY_TYPES = (
```
fdurand: ssid, connection_type, username ... ?

rouzier: This is for types that are hashes not simple values.
{
node_info => {...},
ssid => 'BigBob',
connection_type => 'pigeon'
}

jrouzierinverse:
Please Review

addons/dev-helpers/dump.pl
```diff
@@ -70,14 +70,14 @@ sub _run {
...
 }
 
-package pf::dump::vlan_filters;
+package pf::dump::access_filters;
 use base qw(pf::dump::cmd);
 use Data::Dumper;
 __PACKAGE__->mark_as_loaded();
 
 sub _run {
-    require pf::vlan::filter;
```
fdurand, line 79: Why don't you keep pf::vlan::filter ? I like the way to have that explicit and use the same logic for radius per example (pf::radius::filter)

PR #891, opened 2015-09-14
Person lookup amelioration

### LINKED ISSUE #888: pf::person::lookup should always use stripped username
Will have to go through the code to make sure it's this way or LDAP queries will fail

julsemaan:
# Description

rework person lookup to strip username if needed
Also makes it more abstract so we can use it with other sources than LDAP
# Impacts

Authentication attributes lookup
# Issue

Fixes #888 
# Delete branch after merge

YES
# NEWS file entries
## New Features
- 
## Enhancements
- The person attributes lookup 
## Bug Fixes
- The person attributes lookup will now be made on the stripped username if needed (#888)

### DIFF
--- a/lib/pf/Authentication/Source.pm
+++ b/lib/pf/Authentication/Source.pm
@@ -13,6 +13,7 @@ We must at least always have one rule defined, the fallback one.
 use pf::config;
 use pf::constants;
 use Moose;
+use pf::util;
 use pf::Authentication::constants;
 use pf::Authentication::Action;
 
@@ -238,8 +239,24 @@ sub match_condition {
 =cut
 
 sub search_attributes {
+    my ($self,$username) = @_;
     my $logger = Log::Log4perl->get_logger( __PACKAGE__ );
-    $logger->debug("Search_attributes is not supported on this source.");
+    my $realm;
+    ($username,$realm) = strip_username($username) if isenabled($self->{'stripped_user_name'});
+    return $self->search_attributes_in_subclass($username);
+}
+
+=head2 search_attributes_in_subclass
+
+Search for the attributes of a user
+
+Returns a hashref that will be injected in pf::person::modify or 0 ($FALSE) if it fails
+
+=cut
+
+sub search_attributes_in_subclass {
+    my $logger = Log::Log4perl->get_logger( __PACKAGE__ );
+    $logger->debug("search_attributes_in_subclass is not supported on this source.");
     return $FALSE;
 }
 
--- a/lib/pf/util.pm
+++ b/lib/pf/util.pm
@@ -1060,13 +1060,22 @@ Returns ($user,$realm) if found or ($user) if not matching any realm pattern
 
 sub strip_username {
     my ($username) = @_;
+    return $username unless(defined($username));
+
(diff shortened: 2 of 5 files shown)


lib/pf/util.pm
```diff
@@ -1011,6 +1012,25 @@ sub cert_has_expired {
...
 
+=item strip_username
+
+Will strip a username matching pattern user@realm or \\realm\user
+
+Returns ($user,$realm) if found or ($user) if not matching any realm pattern
+
+=cut
+
+sub strip_username {
```
fdurand, line 1061: I think you will probably have to add $username =~ /(._)\%(._)/ to be close to radius default config

julsemaan, line 1061: Done

lib/pf/util.pm
```diff
@@ -1011,6 +1012,28 @@ sub cert_has_expired {
...
+    }
+    elsif($username =~ /(.*)\%(.*)/){
+        return ($1,$2);
+    }
+    elsif($username =~ /\\\\(.*)\\(.*)/) {
+        return ($2,$1);
+    }
+    return $username;
+}
+
```
louismunro, line 1083: I will be an ass and say we should use named regexes here.
E.g. $suffix_rx = qr/(._)\@(._)/;

louismunro:
Unit tests would be really nice.

lib/pf/util.pm
dwlfrth, line 1083: Don't think we need named regex but having a simple line comment describing which use case we are matching would be great.

louismunro, line 1083: Well, I would argue that self documenting code is better than comments.
But that's me. I just ate a pigsnout sandwich for lunch, so...

lib/pf/Authentication/Source.pm
```diff
@@ -217,8 +218,24 @@ sub match_condition {
 =cut
 
 sub search_attributes {
+    my ($self,$username) = @_;
     my $logger = Log::Log4perl->get_logger( __PACKAGE__ );
-    $logger->debug("Search_attributes is not supported on this source.");
+    my $realm;
+    ($username,$realm) = strip_username($username) if isenabled($self->{'stripped_user_name'});
```
dwlfrth, line 245: :heart_eyes:

lib/pf/util.pm
dwlfrth, line 1083: I would agree with you if the regex was used more than once :) Since it is a pretty simple sub, almost self-described, I think just highlighting the use cases with comment would be enough.
Anyway.. one way or another... (https://www.youtube.com/watch?v=1VFuHj9_Tgw)
I think that comment would be necessary even in the case of a named regex.

julsemaan:
Please re-review

t/util.t
```diff
@@ -17,6 +17,22 @@ BEGIN {
     use_ok('pf::util::apache');
 }
 
+my @info = ("lzammit","turkeycorp");
```
dwlfrth, line 20: :heart_eyes:

dwlfrth:
:rocket:

PR #901, opened 2015-09-23
Separate radiusd load balancer from main AAA process

julsemaan:
# Description

Separates RADIUS accounting, authentication and load balancing into three different processes
# Impacts

RADIUS authentication and proxy
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Authentication and accounting are now in two different RADIUS processes
- Load balancing in an active/active process now uses a dedicated process

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -528,7 +528,6 @@ cd $RPM_BUILD_ROOT/usr/local/pf/conf
 #radius sites-enabled symlinks
 #We standardize the way to use site-available/sites-enabled for the RADIUS server
 cd $RPM_BUILD_ROOT/usr/local/pf/raddb/sites-enabled
-ln -s ../sites-available/control-socket control-socket
 ln -s ../sites-available/default default
 ln -s ../sites-available/packetfence-soh packetfence-soh
 ln -s ../sites-available/dynamic-clients dynamic-clients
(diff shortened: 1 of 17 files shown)


extrafu:
Shouldn't we also have a configuration for accounting? This would prevent bogus setups where accounting information is sent by thousands of spartans from congesting RADIUS authentication and authorization threads.

julsemaan:
I like the idea yes.

julsemaan:
Accounting has been separated as well

addons/packages/packetfence.spec
```diff
@@ -516,9 +516,7 @@ cd $RPM_BUILD_ROOT/usr/local/pf/conf
 #radius sites-enabled symlinks
 #We standardize the way to use site-available/sites-enabled for the RADIUS server
 cd $RPM_BUILD_ROOT/usr/local/pf/raddb/sites-enabled
-ln -s ../sites-available/control-socket control-socket
```
louismunro, line 531: why? these are useful.

addons/packages/packetfence.spec
```diff
@@ -516,9 +516,7 @@ cd $RPM_BUILD_ROOT/usr/local/pf/conf
...
 ln -s ../sites-available/default default
```
louismunro, line 531: On the other hand this one is unused anyway.

addons/packages/packetfence.spec
julsemaan, line 531: The different processes cannot share the same socket so it is specified in the individual files

addons/packages/packetfence.spec
julsemaan, line 531: I'd prefer not to remove it from this PR as it's unrelated to what this does.

louismunro:
Do you want some help finishing this feature?

julsemaan:
Its completed, I just wanted to check if it would work in a real environment and if the benefits would be there.

And they are :)

julsemaan:
Ready for review

louismunro:
Good enough. Merging.

PR #904, opened 2015-09-23
add update action and caching to firewallsso to reduce duplicate requests

julsemaan:
# Description

Add caching to firewall SSO update calls to reduce duplicate calls when a device renews DHCP often due to roaming
# Impacts

pfdhcplistener, firewall SSO
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added option to cache firewall SSO requests to reduce duplicate calls to the firewall

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -40,6 +40,7 @@ use pf::cluster;
 use JSON;
 use File::Slurp;
 use pf::file_paths;
+use pf::CHI;
 
 use List::MoreUtils qw(uniq);
 use NetAddr::IP;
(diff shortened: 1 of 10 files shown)


julsemaan:
Ready for review

lib/pf/api.pm
```diff
@@ -222,9 +223,29 @@ sub firewallsso : Public {
...
-    foreach my $firewall_conf ( sort keys %pf::config::ConfigFirewallSSO ) {
-        my $firewall = pf::factory::firewallsso->new($firewall_conf);
-        $firewall->action($firewall_conf,$postdata{'method'},$postdata{'mac'},$postdata{'ip'},$postdata{'timeout'});
+    foreach my $firewall_id ( sort keys %pf::config::ConfigFirewallSSO ) {
+        my $firewall = pf::factory::firewallsso->new($firewall_id);
+
+        if($postdata{method} eq "Update"){
+            if( pf::util::isenabled($pf::config::ConfigFirewallSSO{$firewall_id}{cache_updates}) ){
+                my $cache = pf::CHI->new( namespace => 'firewall_sso' );
+                my $cache_timeout = $pf::config::ConfigFirewallSSO{$firewall_id}{cache_timeout} || $postdata{timeout} / 2;
```
rouzier, line 232: Why divide by two?

fdurand, line 232: Why not multiply by 2 ? and instead of just caching $firewall_id -$postdata{ip} key why don't you cache $firewall_id -$postdata{ip} - $postdata{mac}

So if we receive a new dhcp request after the dhcp lease timeout  then the device will be still in the cache.

julsemaan, line 232: Divide by two is to get close to the renewal.

I want to get the next DHCP packet available to redo the request to handle the expiration on the firewall side.
Basically with this I want it to be the amount of DHCP requests that occur when you don't roam (so one every time the renewal kicks in) 

If I use the lease length then it means I will do it every 2 DHCP packets and multiplying by 2 will mean I will do it every 4.

This is the sweet spot I found in the use case I have so I made it default.
There is still the ability to override if defaults don't apply to the setup

I like the MAC address idea, I will add it

fdurand, line 232: By multiplying by 2 i was thinking on the registration/isolation vlan were we have a dhcp request every 30s.
So maybe we have to imagine to have a threshold on $postdata{timeout} like if < 300s then use the default $DEFAULT_LEASE_LENGTH. So the first request on the reg/isol vlan will trigger the firewall sso and not the hundred other request after.

louismunro:
So, is this stuff ready? 
Should it be moved to 5.6?

julsemaan:
No it's all good. Me and @fdurand talked about it this morning.

Last news is he wanted to check some code on his side and we're good to merge

PR #912, opened 2015-09-25
Feature/unreg onstop accouting

whitx:
# Description

Allow to unreg a node after receiving an accounting stop request, added a timing so it won't unreg on the deauth 
# Impacts

RADIUS accounting
# Delete branch after merge

NO
# NEWS file entries

Radius disconnect on stop accounting request
# N.B

This is designed as a custom code for a client, it could be adapt to general use

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1130,6 +1130,12 @@ description=<<EOT
 List of LDAP attributes that can be used in the sources configuration
 EOT
 
+[advanced.unreg_onstop_timing]                 
+type=numeric
+description=<<EOT
+Define the wait timing before unreg node on accounting stop request.
+EOT
+
 [omapi.ip2mac_lookup]
 type=toggle
 options=enabled|disabled
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1130,10 +1130,16 @@ description=<<EOT
 List of LDAP attributes that can be used in the sources configuration
 EOT
 
-[advanced.unreg_onstop_timing]                 
+[advanced.unreg_onstop_timing]
 type=numeric
 description=<<EOT
 Define the wait timing before unreg node on accounting stop request.
+EOT
+
+[advanced.unreg_onstop_ssid]
+type=text
+description=<<EOT
+Define the ssid on which apply the unreg on accounting stop request.
 EOT
 
 [omapi.ip2mac_lookup]
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1133,13 +1133,13 @@ EOT
 [advanced.unreg_onstop_timing]
 type=numeric
 description=<<EOT
-Define the wait timing before unreg node on accounting stop request.
+Define the wait timing before unreg node when receiving an accounting stop request.
 EOT
 
 [advanced.unreg_onstop_ssid]
 type=text
 description=<<EOT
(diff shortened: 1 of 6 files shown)


whitx:
NOT READY, need to add the ssid param in conf

lib/pf/radius.pm
```diff
@@ -281,6 +281,22 @@ sub accounting {
...
             $this->_handleAccountingFloatingDevices($switch, $mac, $port);
+            #unreg on disconnect
+            use Date::Parse;
+            my $node_info = node_view($mac);
+            my $time = time();
+            my $regdate = $node_info->{'regdate'};
+            my $convertregdate = str2time($regdate);
+            my $interval = $Config{'advanced'}{'unreg_onstop_timing'};
+            my $limitunreg = $convertregdate + $interval;
+            if ($node_info->{'last_ssid'} eq 'aa-t' && $time > $limitunreg){
```
extrafu: SSID equals 'aa-t' ? generalize.

lib/pf/radius.pm
```diff
@@ -281,6 +281,22 @@ sub accounting {
...
+                use pf::api::jsonrpcclient;
+                my $apiclient = pf::api::jsonrpcclient->new;
+                my %options;
+                $options{'mac'} = $node_info->{'mac'};
+                $apiclient->notify('deregister_node', %options );
+                $logger->info("Unregistred node $mac since we receive his stop accounting request.");
```
extrafu: "Unregistered node $mac since we received an account stop request."

conf/documentation.conf
```diff
@@ -1130,6 +1130,18 @@ description=<<EOT
 List of LDAP attributes that can be used in the sources configuration
 EOT
 
+[advanced.unreg_onstop_timing]
+type=numeric
+description=<<EOT
+Define the wait timing before unregister a node when receiving an accounting stop request.
+EOT
+
+[advanced.unreg_onstop_ssid]
```
julsemaan: That's not dynamic at all...
We need to be able to define more specific rules other than the SSID or at least multiple SSIDs or no filtering at all.

I suggest you create a new scope in the VLAN filters and use it for that.

lib/pf/radius.pm
```diff
@@ -281,6 +281,28 @@ sub accounting {
...
+            #unreg on disconnect
+            use Date::Parse;
+            my $node_info = node_view($mac);
+            my $time = time();
+            my $regdate = $node_info->{'regdate'};
+            my $convertregdate = str2time($regdate);
+            my $interval = $Config{'advanced'}{'unreg_on_disconnect_timing'};
+            my $limitunreg = $convertregdate + $interval;
+            my $filter = new pf::vlan::filter;
+            my $switch = pf::SwitchFactory->instantiate($node_info->{'last_switch'});
```
fdurand, line 293: Use the current value in the radius request because $node_info->{'last_switch'} come from the request before.
Same for ssid, connection_type,ifIndex

fdurand:
In my opinion we can probably handle that feature in the radius filters since it's really custom stuff.
Also if we merge that as is then each device with the default role will be unreg if we receive an accounting stop after 30s.

whitx:
Well that was for custom case, but since it is not necessary anymore I don't think we should merge it.

whitx:
Guess we could just closed it, since it is of no use and has a better alternative.

PR #915, opened 2015-09-25
Feature/better captiveportal detection mecanism handling

dwlfrth:
# Description

Improving the detection of captive-portal detection mecanism as well as handling self-signed SSL cert
# Impacts

Dispatcher flow (captive-portal detection mecanisms)
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Detecting when using a self-signed SSL certificate for device with captive-portal detection mecanisms

### DIFF
--- a/lib/pf/web/dispatcher.pm
+++ b/lib/pf/web/dispatcher.pm
@@ -138,12 +138,19 @@ sub handler {
 =cut
 
 sub html_redirect {
-    my ($r) = @_;
+    my ($r, $params) = @_;
     my $logger = Log::Log4perl->get_logger(__PACKAGE__);
 
     $logger->debug("hitting html_redirect with URI '" . $r->uri . "' (URL: " . $r->construct_url . ")");
+    $logger->debug(sub { "'html_redirect' called with the following parameters: " . join(", ", map { "$_ => $params->{$_}" } keys %$params) }) if defined($params);
 
-    my $proto = isenabled($Config{'captive_portal'}{'secure_redirect'}) ? $HTTPS : $HTTP;
+    my $proto;
+    if ( defined($params->{'secure'}) && $params->{'secure'} ne "" ) {
+        $proto = isenabled($params->{'secure'}) ? $HTTPS : $HTTP;
+    } else {
+        $proto = isenabled($Config{'captive_portal'}{'secure_redirect'}) ? $HTTPS : $HTTP;
+    }
+
     my $captive_portal_domain = $Config{'general'}{'hostname'}.".".$Config{'general'}{'domain'};
     my $user_agent = $r->headers_in->{'User-Agent'};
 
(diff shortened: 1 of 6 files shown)


dwlfrth:
Please review

lib/pf/web/dispatcher.pm
```diff
@@ -74,11 +74,18 @@ sub handler {
...
-    if ( ($url =~ /$WEB::CAPTIVE_PORTAL_DETECTION_MECANISM_URLS/o || $user_agent =~ /CaptiveNetworkSupport/s) && isenabled($Config{'captive_portal'}{'detection_mecanism_bypass'}) ) {
+    # Captive-portal detection mecanism
+    # Ability to bypass captive-portal detection mecanism
+    my $captive_portal_detection_mecanism_urls = pf::web::util::build_captive_portal_detection_mecanisms_regex;
+    if ( ($url =~ /$captive_portal_detection_mecanism_urls/o || $user_agent =~ /CaptiveNetworkSupport/s) && isenabled($Config{'captive_portal'}{'detection_mecanism_bypass'}) ) {
         $logger->info("Dealing with a endpoint / browser with captive-portal detection capabilities while having captive-portal detection mecanism bypass enabled. Proxying");
         return proxy_redirect($r);
     }
+    # Enforce HTTP instead of HTTPS when dealing with captive-portal detection mecanism and having a self-signed SSL certificate
+    elsif ( ($url =~ /$captive_portal_detection_mecanism_urls/o || $user_agent =~ /CaptiveNetworkSupport/s) && pf::web::util::is_certificate_self_signed ) {
```
jrouzierinverse: Does captive_portal.detection_mecanism_bypass need to be check for self signed also
If so it would be better rewritten like so

```
if ( ($url =~ /$captive_portal_detection_mecanism_urls/o || $user_agent =~ /CaptiveNetworkSupport/s) && isenabled($Config{'captive_portal'}{'detection_mecanism_bypass'}) ) {
    if(pf::web::util::is_certificate_self_signed) {
        $logger->info("Dealing with a endpoint / browser with captive-portal detection capabilities while having a self-signed SSL certificate. Using HTTP instead of HTTPS");
        return html_redirect($r, { secure => $FALSE });  
    } 
    $logger->info("Dealing with a endpoint / browser with captive-portal detection capabilities while having captive-portal detection mecanism bypass enabled. Proxying");
    return proxy_redirect($r);
}
```

dwlfrth: It "does not need" but would be better to handle it, so, you are right.
I will add the condition.

dwlfrth: Actually, your condition is wrong, it would require that the "bypass" is enabled to check if self-signed, which we don't want... it should be two "distinct" flows

fdurand: && isenabled($Config{'captive_portal'}{'secure_redirect'})

dwlfrth: Is it THAT important to have the condition ? No matter what, we want to force HTTP, so I'm just asking do we want to add more conditions to the actually one because it will anyway, take the same path at the end? Just asking

fdurand: This is just because of the log message

fdurand:
Tested and works. Ready to merge

PR #918, opened 2015-09-28
Feature/remove useless log

fdurand:
# Description

Removed useless logs in apache and packetfence.log
Bypass catalyst on static content
# Impacts

Less logs
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Removed useless logs in Apache and packetfence.log

### DIFF
--- a/lib/pf/web/dispatcher.pm
+++ b/lib/pf/web/dispatcher.pm
@@ -88,6 +88,7 @@ sub handler {
     # See L<pf::web::constants::CAPTIVE_PORTAL_STATIC_RESOURCES>
     if ( $uri =~ /$WEB::CAPTIVE_PORTAL_STATIC_RESOURCES/o ) {
         $logger->debug("URI '$uri' (URL: $url) is a captive-portal static resource");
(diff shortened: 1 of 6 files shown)


jrouzierinverse:
Less logging @dwlfrth will not be pleased

dwlfrth:
So happy with it that I'll merge it myself

lib/pf/web/dispatcher.pm
```diff
@@ -88,6 +88,7 @@ sub handler {
     # See L<pf::web::constants::CAPTIVE_PORTAL_STATIC_RESOURCES>
     if ( $uri =~ /$WEB::CAPTIVE_PORTAL_STATIC_RESOURCES/o ) {
         $logger->debug("URI '$uri' (URL: $url) is a captive-portal static resource");
+        $r->set_handlers( PerlResponseHandler => ['pf::web::static'] );
```
jrouzierinverse, line 91: This is a good thing.
We should make this a configuration item so users can choose if they want to use the dynamic asset feature.

PR #927, opened 2015-09-30
Speed up the creation of switches

jrouzierinverse:
## Description

During a benchmark session switch creation was shown to be slow.
One of the factors was the fact we passed the hash by value not by reference

Below is the benchmark of the speed difference by passing by value vs reference

```
Benchmark: running by value, by ref for at least 3 CPU seconds...
by value:  4 wallclock secs ( 3.37 usr +  0.00 sys =  3.37 CPU) @ 100499.70/s (n=338684)
by ref:    4 wallclock secs ( 3.17 usr +  0.00 sys =  3.17 CPU) @ 2260450.79/s (n=7165629)

```

| Benchmark | Rate | by value | by ref |
| --- | --- | --- | --- |
| **by value** | 100500/s | -- | -96% |
| **by ref** | 2260451/s | 2149% | -- |
## Impacts

Switch creation
## NEWS file entries

Enhancements
++++++++++++
- Speed up switch creation
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -254,7 +254,7 @@ sub supportsMABFloatingDevices {
 }
 
 sub new {
-    my ( $class, %argv ) = @_;
+    my ($class, $argv) = @_;
     my $this = bless {
         '_error'                    => undef,
         '_id'                       => undef,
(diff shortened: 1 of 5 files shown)


jrouzierinverse:
Review me

lib/pf/Switch.pm
```diff
@@ -307,114 +307,8 @@ sub new {
...
-        } elsif (/^-?access_lists$/i) {
-            $this->{_access_lists} = $argv{$_};
-        }
-        # customVlan members are now dynamically generated. 0 to 99 supported.
-        elsif (/^-?(\w+)Vlan$/i) {
-            $this->{'_'.$1.'Vlan'} = $argv{$_};
-        }
-
-    }
     return $this;
```
louismunro, line 312: Yes. One thousand times yes.
About time we did this.

lib/pf/Switch.pm
```diff
@@ -687,14 +581,14 @@ sub setVlan {
     # VLAN -1 handling
     # TODO at some point we should create a new blackhole / blacklist API
```
louismunro, line 582: Please go ahead and remove all the TODOs as you go along.
Most of them don't make sense anymore.

PR #941, opened 2015-10-02
Feature/tcp fingerprints

julsemaan:
# Description

Adds TCP fingerprinting to PacketFence via the Fingerbank library and p0f
# Impacts

Device recognition
# Dependencies

Requires the TCP fingerprinting in the Fingerbank library (https://github.com/inverse-inc/fingerbank/pull/24)
# NEWS file entries

(REQUIRED)
## New Features
- Added device detection through TCP fingerprinting

### DIFF
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -625,9 +625,9 @@ collectd_binary=/usr/sbin/collectd
 # Location of the radsniff3 binary. Only necessary to change if you are not running the pre-packaged version.
 radsniff3_binary=/usr/bin/radsniff3
 # 
-# services.radsniff_binary
+# services.p0f_binary
 # 
-# Location of the radsniff3 binary. Only necessary to change if you are not running the pre-packaged version.
+# Location of the p0f binary. Only necessary to change if you are not running the pre-packaged version.
 p0f_binary=/usr/local/bin/p0f
 # 
 # services.iptables_binary
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -555,6 +555,11 @@ statsd=enabled
 # Should radsniff be started?
 radsniff3=enabled
 #
+# services.p0f
+#
+# Should p0f be started?
+p0f=enabled
+#
 # services.snort_binary
 #
 # Location of the snort binary. Only necessary to change if you are not running the RPMed version. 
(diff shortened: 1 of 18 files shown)


julsemaan:
Please review

julsemaan:
@dwlfrth can you review since DHCPv6 depends on that one

conf/pf.conf.defaults
```diff
@@ -620,6 +625,11 @@ collectd_binary=/usr/sbin/collectd
 # Location of the radsniff3 binary. Only necessary to change if you are not running the pre-packaged version.
 radsniff3_binary=/usr/bin/radsniff3
 # 
+# services.radsniff_binary
```
dwlfrth: services.p0f_binary

conf/pf.conf.defaults
```diff
@@ -620,6 +625,11 @@ collectd_binary=/usr/sbin/collectd
...
+# 
+# Location of the radsniff3 binary. Only necessary to change if you are not running the pre-packaged version.
```
dwlfrth: s/radsniff3/p0f

html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Fingerbank/Settings.pm
```diff
@@ -128,6 +129,17 @@ sub index :Path :Args(0) :AdminRole('FINGERBANK_READ') {
     $c->response->status($status);
 }
 
+sub update_p0f_map :Local :Args(0) :AdminRole('FINGERBANK_UPDATE') {
```
dwlfrth, line 132: Shouldn't we use the same logic as for the database operations ? (using async API calls to avoid slow update timing out on the GUI ?)
https://github.com/julsemaan/packetfence/blob/feature/tcp-fingerprints/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Fingerbank/DB.pm#L26

Which make me think that moving subs of that class (DB) here could be a good idea too.

dwlfrth:
Should not merge before having https://github.com/inverse-inc/fingerbank/pull/24 merged and new Fingerbank packages ready.

html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Fingerbank/Settings.pm
julsemaan, line 132: The map is a few kilobytes so I'm not sure.

It adds weight to the code for a case that won't happen

dwlfrth, line 132: Should we move the DB class subs under there ?

dwlfrth:
Will be ready to merge once dependancies will be taken care of. I'll take care of it

PR #949, opened 2015-10-08
Fix/943 inline enforcement

### LINKED ISSUE #943: Inline: Node not being removed from 'reg' set if iplog entry is closed
Example:
A node is set to be unreg by pfmon at midnight. That node gets an ip during the day and the lease is of 3 hours. The node does not renew the lease so the iplog entry is closed at the end of lease time.

On pfmon run, it 'isInlineEnforcementRequired' calls 'get_mangle_mark_for_node' which check if there's an active iplog entry. If not (which is the case in the example), 'get_mangle_mark_for_node' return 'UNREG' and 'isInlineEnforcementRequired' is supposed to mark the node UNREG. 

It creates a true condition where the new status is equals to the expected status, then, no action taken.

dwlfrth:
# Description

Reworked inline enforcement evaluation logic
# Impacts

Inline flow
# Issue

fixes #943 
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Inline: Node not being removed from 'reg' set if iplog entry is closed (#943)

### DIFF
--- a/lib/pf/enforcement.pm
+++ b/lib/pf/enforcement.pm
@@ -86,19 +86,8 @@ sub reevaluate_access {
 
         my $conn_type = str_to_connection_type( $locationlog_entry->{'connection_type'} );
         if ( $conn_type == $INLINE ) {
-
             my $client = pf::client::getClient();
-            my $inline = new pf::inline::custom();
-            my %data = (
-                'switch'           => '127.0.0.1',
-                'mac'              => $mac,
-            );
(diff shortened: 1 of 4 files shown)


dwlfrth:
Review me

louismunro:
What is missing for this to be ready for review again? 
When will it be ready?

louismunro:
Closed, won't fix.

PR #953, opened 2015-10-13
Feature/switch info node view

julsemaan:
# Description

Adds switch IP and description in the available columns in the node list view
# Impacts

Node list view in admin interface
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added switch IP and description in the available columns in the node list view


julsemaan:
Ready for review

dwlfrth:
Would be great to add the DESC to the detailed node view too.

PR #965, opened 2015-10-19
Feature/auth log

julsemaan:
not ready


louismunro:
Adding grist to the mill, I thing we should have audit capabilities in PF 6 at the latest.
We should be able to tell why a device changed status or role as well as who changed any configuration item manually including role changes or bypass.

julsemaan:
Agreed. This is a sponsored feature so I kept it limited to the scope that was asked.

I don't think we should use this for the audit log as an SQL table doesn't seem the right data structure.

An RRD would be much better IMO

This can then be integrated in our new audit log.

louismunro:
An RRD would be terrible for what I have in mind.
You can't really do any reporting or  complex queries on it.

louismunro:
Is this PR still relevant with the latest auditing work?

julsemaan:
It is because this records authentication requests and the flow a user goes through (sms, email, oauth)

Like the current implementation of the radius log it should be replaced by a better approach in pf6

I still think this is useful for users (and us too) to know what flow a user went through and it can help with troubleshooting

louismunro:
ok, let me rephrase then.
Should this still be in 5.6 ? Or should it be moved to a later release or even PF6?

julsemaan:
5.6 in my biased opinion :) since I feel it would be useful to follow a weird error a client may have got by going through e-mail, sms, back to e-mail, then sms and then sponsor.

Will also log authentication failures so it would help troubleshooting that side too.

Code is there and tested working 100% on a production setup and scales so there is technically no downside to it.

Worst case we rework it or remove it depending on the final form of the audit log

fdurand:
On my side i am not against adding it to 5.6.
As julien said it will interesting to be able to troubleshoot authentication issue.
Just one thing is missing for me is the cleanup.

julsemaan:
Rebase + reworked for comments.

So its there and ready

fdurand:
You forgot to add auth_log in the pf database schema

julsemaan:
You are right.

Maybe we should review the way we deal with the schema because I will likely conflict with #1069 

What do you think about putting the schema dump in the release process so basically when we have the 'final' version, we dump the schema at that time.

Will prevent conflicts during development phase and should mistakes like these not have been seen, prevent bugs.

fdurand:
Ok, i will test it tomorrow and if it's ok then i will merge it.

PR #971, opened 2015-10-21
last_connection_sub_type filter for portal profile

fdurand:
## Description

Added new portal filter to be able to define the subconnection type of the radius request
## Impacts

The portal profile
## NEWS file entries
# NEWS file entries
## New Features
- New portal profile filter (sub connection type)
## Delete branch after merge

YES


julsemaan:
The connection subtype is persisted in the database so it would be cleaner to add it to the node_view so you don't have to add it manually to the node info.

Then it can be used in the portal as well and is not localized to the RADIUS process.

Because now if you create a 'Portal profile' that filters on the connection subtype it can't even be used in a portal.

fdurand:
My bad, i forgot to update node.pm

Issue #978, opened 2015-10-23
pf::CHI compute method does not support computing undef values


julsemaan:
This means that a full LDAP miss using the cache ends up not being cached.

Will need support for undef values (when explicitly requiring it)

PR #979, opened 2015-10-23
Feature/chi compute with undef

julsemaan:
# Description

Add the ability to cache undef values when using cache computing
# Impacts

LDAP caching as its the only one which uses it
# Issue
#978
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Non-found users will now be cached when using the caching in an LDAP source (#978)

### DIFF
--- a/lib/pf/CHI.pm
+++ b/lib/pf/CHI.pm
@@ -27,7 +27,6 @@ use DBI;
 use Scalar::Util qw(tainted reftype);
 use pf::log;
 use Log::Any::Adapter;
-use pfconfig::util qw($undef_element);
 Log::Any::Adapter->set('Log4perl');
 
 my @PRELOADED_CHI_DRIVERS;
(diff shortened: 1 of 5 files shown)


julsemaan:
Ready for review

lib/pf/CHI.pm
```diff
@@ -62,7 +62,7 @@ Hash::Merge::specify_behavior(
     'PF_CHI_MERGE'
 );
 
-our @CACHE_NAMESPACES = qw(configfilesdata configfiles httpd.admin httpd.portal pfdns switch.overlay ldap_auth omapi fingerbank);
+our @CACHE_NAMESPACES = qw(configfilesdata configfiles httpd.admin httpd.portal pfdns switch.overlay ldap_auth omapi fingerbank t);
```
jrouzierinverse: No need to add a test namespace.
Add the following in the t/data/chi.conf

[namespace t]
driver=File
root_dir=/tmp/chi

jrouzierinverse: Better yet you should do the following.

[namespace t]
storage=t

[storage t]
storage=file

julsemaan: Fixed

PR #986, opened 2015-10-26
run pfdhcplistener on each cluster members

fdurand:
# Description

Run all the pfdhcplistener process on all cluster's members and update the status with keepalived
# Impacts

No
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Manage pfdhcplistener status with keepalived and run pfdhcplistener on all cluster's members.

### DIFF
--- a/lib/pf/cluster.pm
+++ b/lib/pf/cluster.pm
@@ -288,14 +288,14 @@ sub is_vip_running {
                 my @addresses = $inf->address(AF_INET());
                 for my $i (0 .. $#masks) {
                     if (inet_ntoa($addresses[$i]) eq cluster_ip($int)) {
-                        return 'master';
+                        return $TRUE;
                     }
                 }
-                return 'slave';
+                return $FALSE;
             }
         }
     }
-    return 0;
+    return $FALSE;
 }
 
 =head1 AUTHOR
--- a/lib/pf/cluster.pm
+++ b/lib/pf/cluster.pm
@@ -272,7 +272,31 @@ sub sync_storages {
     }
 }
 
+=head2 is_vip_running
 
+=cut
+
+sub is_vip_running {
+    my ($int) = @_;
+
+    if ( defined($ConfigCluster{$CLUSTER}->{"interface $int"}) ) {
+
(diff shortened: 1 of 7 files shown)


lib/pf/cluster.pm
```diff
@@ -272,7 +272,31 @@ sub sync_storages {
...
+    if ( defined($ConfigCluster{$CLUSTER}->{"interface $int"}) ) {
+
+        my @all_ifs = Net::Interface->interfaces();
+        foreach my $inf (@all_ifs) {
+            if ($inf->name eq $int) {
+                my @masks = $inf->netmask(AF_INET());
+                my @addresses = $inf->address(AF_INET());
+                for my $i (0 .. $#masks) {
+                    if (inet_ntoa($addresses[$i]) eq cluster_ip($int)) {
+                        return 'master';
```
julsemaan: return $TRUE instead

It checks if the VIP is running not what is the state name of the cluster

lib/pf/cluster.pm
```diff
@@ -272,7 +272,31 @@ sub sync_storages {
...
+                    }
+                }
+                return 'slave';
```
julsemaan: $FALSE

lib/pf/cluster.pm
```diff
@@ -272,7 +272,31 @@ sub sync_storages {
...
+            }
+        }
+    }
+    return 0;
```
julsemaan: $FALSE

sbin/pfdhcplistener
```diff
@@ -179,7 +187,7 @@ sub dhcp_detector {
 
 sub process_pkt {
     my ( $user_data, $hdr, $pkt ) = @_;
-    listen_dhcp( $pkt, $user_data );
+    listen_dhcp( $pkt, $user_data ) if (($process eq '0') || ($process eq 'master'));
```
julsemaan: now can be put to if($process)

julsemaan: Should also check if it's in cluster

so should be : 
if($process || !$cluster_enabled)

PR #1008, opened 2015-11-09
Update Config::Inifiles

jrouzierinverse:
## Description

Use the latest Config::Inifiles to avoid the cost of parsing a huge config file
## Impacts

Configuration
## NEWS file entries

Enhancements
++++++++++++
- Upgrade Config::IniFiles to allow faster loading of configuration files
## Delete branch after merge

NO


addons/packages/packetfence.spec
```diff
@@ -111,7 +111,8 @@ Requires: perl(Class::Accessor)
 Requires: perl(Class::Accessor::Fast::Contained)
 Requires: perl(Class::Data::Inheritable)
 Requires: perl(Class::Gomor)
-Requires: perl(Config::IniFiles) < 2.88
+Requires: perl(Config::IniFiles) >= 2.88
+Requires: perl-Scalar-List-Utils >= 1.42
```
julsemaan: I thought you said it wasn't necessary to force the version.

If it is, we should adjust devel...

rouzier: It is still not required. This is not packaged by centos upstream

julsemaan: So can you remove it from there if its not needed ?

julsemaan:
Not much to review

I'll merge and build a package.

PR #1025, opened 2015-11-24
Avoid database lookup when filtering for arguments that are never used.

jrouzierinverse:
## Description

Avoid database lookup when filtering for arguments that are never used.
## Impacts

The filter engine
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
- Avoid unneccesary calls to the database when using vlan filtering
- Avoid unneccesary calls to the database when using radius filtering
- Avoid unneccesary calls to the database for the profile filters

Bug Fixes
+++++++++
## Delete branch after merge

NO

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -261,6 +261,7 @@ Requires: perl(Hash::Merge)
 Requires: perl(IO::Socket::INET6)
 Requires: perl(IO::Interface)
 Requires: perl(Time::Period)
+Requires: perl(Data::Thunk);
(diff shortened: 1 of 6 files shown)


fdurand:
Can we use it there https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/access_filter/radius.pm#L41 ?

jrouzierinverse:
Yes

fdurand:
Debian package libdata-thunk-perl done. Let's merge

PR #1029, opened 2015-11-26
make local dhcp detection packet dependent

julsemaan:
# Description

Process requests properly when running a pfdhcplistener on an interface that has networks with and without dhcpd activated
# Impacts

pfdhcplistener
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Process requests properly when running a pfdhcplistener on an interface that has networks with and without dhcpd activated

### DIFF
--- a/lib/pf/dhcp/processor.pm
+++ b/lib/pf/dhcp/processor.pm
@@ -79,7 +79,29 @@ sub new {
         $self->{accessControl} = new pf::inline::custom();
     }
     $self->{api_client} = pf::client::getClient();
+    $self->_build_DHCP_networks();
     return $self;
+}
+
+=head2 _build_DHCP_networks
+
+Builds the list of networks on which PacketFence is the DHCP server
+
+=cut
+
+sub _build_DHCP_networks {
+    my ($self) = @_;
+
+    my @dhcp_networks;
+    foreach my $network (keys %ConfigNetworks) {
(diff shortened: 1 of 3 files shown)


lib/pf/dhcp/processor.pm
```diff
@@ -274,7 +296,7 @@ sub parse_dhcp_request {
 
     # We check if we are running without dhcpd
     # This means we don't see ACK so we need to act on requests
-    if((!$self->{running_w_dhcpd} && !isenabled($Config{network}{force_listener_update_on_ack})) && (defined($client_ip) && defined($client_mac))){
+    if( (defined($client_ip) && defined($client_mac)) && (!$self->pf_is_dhcp($client_ip) && !isenabled($Config{network}{force_listener_update_on_ack})) ){
```
fdurand, line 299: Don't we have to remove !isenabled($Config{network}{force_listener_update_on_ack} condition ?, if packetfence is not the dhcp for this client then we update the iplog since we are not sure that we will receive the ACK packet.

julsemaan, line 299: The parameter is disabled by default

Enabling it is for case where we wouldn't be the DHCP server but would have the reflector installed.

In that case we wouldn't want to act on the REQUEST

fdurand, line 299: Ok perfect, let's merge

PR #1030, opened 2015-11-26
Force violation add

### LINKED ISSUE #1028: Violation trigger from web admin should bypass grace period
Trying to trigger a violation from the web admin will fail if there is still a grace period for that same violation.
Triggering a violation from the web admin should bypass the grace period (force a violation trigger)

dwlfrth:
# Description

See #1028 
# Impacts

Violation triggering
# Issue

fixes #1028 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Violation triggering from web admin will add even if there is still a grace period
## Bug Fixes
- Violation trigger from web admin should bypass grace period (#1028)

### DIFF
--- a/lib/pf/violation.pm
+++ b/lib/pf/violation.pm
@@ -482,6 +482,9 @@ sub violation_add {
             $logger->info($msg);
             $pf::StatsD::statsd->end(called() . ".timing" , $start, 0.25 );
             return (-1);
+        } elsif ( $remaining_time > 0 && $data{'force'} eq $TRUE ) {
+            my $msg = "Force violation $vid for node $mac even if $remaining_time grace remaining";
+            $logger->info($msg);
         } else {
             my $msg = "grace expired on violation $vid for node $mac";
             $logger->info($msg);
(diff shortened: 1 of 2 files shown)


lib/pf/violation.pm
```diff
@@ -476,7 +476,7 @@ sub violation_add {
 
         # check if we are under the grace period of a previous violation
         my ($remaining_time) = violation_grace( $mac, $vid );
-        if ( $remaining_time > 0 ) {
+        if ( $remaining_time > 0 && $data{'force'} ne $TRUE ) {
```
fdurand, line 479: Maybe a detail but something like that should be usefull:
} elsif ( $remaining_time > 0 && $data{'force'} eq $TRUE ) {
            my $msg = "Force violation $vid for node $mac";
            $logger->info($msg);
} else { ...

dwlfrth, line 479: Is the elsif meant to be there or simply a typo ?

dwlfrth, line 479: But for the logging, I like your style ;) (moooaaaarrr logging)

PR #1031, opened 2015-11-27
Feature/dhcp filter

fdurand:
# Description

Added dhcp filters rules to bypass fingerbank check if a rule match.
# Impacts

No
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Dhcp filter to bypass fingerbank check if a rule match and trigger an action (api call)

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -919,6 +919,7 @@ sub fingerbank_process : Public {
         delete $args->{'computer_name'};
         return (pf::fingerbank::process($args));
     }
+    return undef;
 }
 
 =head2 fingerbank_update_upstream_db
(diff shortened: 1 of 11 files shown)


lib/pf/api.pm
```diff
@@ -912,8 +913,12 @@ sub dynamic_register_node : Public {
...
 sub fingerbank_process : Public {
     my ( $class, $args ) = @_;
-
-    return (pf::fingerbank::process($args));
+    my $filter = pf::access_filter::dhcp->new;
+    my $rule = $filter->filter('DhcpFingerbank', $args);
+    if (!$rule) {
+        delete $args->{'computer_name'};
+        return (pf::fingerbank::process($args));
+    }
```
louismunro, line 921: I think we should set a return if it does not match.
We may not check for it now but who knows what we will do with it someday.
We could just return undef for now.

PR #1037, opened 2015-11-30
Fix the task counters not in sync when a task expires

jrouzierinverse:
## Description

Task counters are not in sync when a task expires
## Impacts

pfqueue
## NEWS file entries

Bug Fixes
+++++++++
- Fix task counters out of sync when a task expires
## Delete branch after merge

NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Pfqueue.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Pfqueue.pm
@@ -13,7 +13,7 @@ Catalyst Model.
 use Moose;
 use namespace::autoclean;
 use pf::util::pfqueue;
-use pf::constants::pfqueue qw($PFQUEUE_COUNTER $PFQUEUE_QUEUE_PREFIX);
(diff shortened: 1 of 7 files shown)


julsemaan:
The miss count should be shown in the admin so all the infos is there when we troubleshoot

jrouzierinverse:
Already does

julsemaan:
I don't see it.

Is it showing where there isn't any miss ?

Would be useful to have it since its an odd behavior to not have the column

jrouzierinverse:
It will only display it when there are missed counters

louismunro:
That is not very intuitive to the user.
Could we make that show 0 if there aren't any?

julsemaan:
Is it hard to add it because a user would expect the column to be there with a count of 0

jrouzierinverse:
Since this is "dynamic" information prepopulating would be difficult/pain.
I will change the display to split the display of the task counters and miss counters.

jrouzierinverse:
This has been done.

julsemaan:
When there are no miss count counters to show it should say it (following the same principles as the portal profiles) instead of showing an empty table.

jrouzierinverse:
Done

PR #1038, opened 2015-11-30
Feature/fingerbank cache rework

julsemaan:
# Description

Use the Fingerbank cache instead of caching its result
# Impacts

Fingerbank calls
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Fingerbank caching is now done through the Fingerbank library

### DIFF
--- a/lib/pf/fingerbank.pm
+++ b/lib/pf/fingerbank.pm
@@ -90,15 +90,6 @@ sub _query {
     my ( $args ) = @_;
     my $logger = pf::log::get_logger;
 
-
-    # Doing a shallow copy of the args hashref to modify the 'MAC' value and make it OUI rather than full MAC
-    my $cached_args = { %$args };
-    my $mac = $cached_args->{'mac'};
-    $mac =~ s/[:|\s|-]//g;      # Removing separators
-    $mac = lc($mac);            # Lowercasing
-    $mac = substr($mac, 0, 6);  # Only keep first 6 characters (OUI)
-    $cached_args->{'mac'} = $mac;
-
     my $cache = pf::CHI->new( namespace => 'fingerbank', expires_in => FINGERBANK_CACHE_EXPIRE );
     my $fingerbank = fingerbank::Query->new(cache => $cache);
     return $fingerbank->match($args);
--- a/lib/pf/fingerbank.pm
+++ b/lib/pf/fingerbank.pm
@@ -90,7 +90,6 @@ sub _query {
     my ( $args ) = @_;
     my $logger = pf::log::get_logger;
 
-    my $cache = pf::CHI->new( namespace => 'fingerbank' );
 
     # Doing a shallow copy of the args hashref to modify the 'MAC' value and make it OUI rather than full MAC
     my $cached_args = { %$args };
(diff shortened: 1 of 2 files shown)


lib/pf/fingerbank.pm
```diff
@@ -100,14 +99,9 @@ sub _query {
...
-        sub {
-            $logger->debug("Fingerbank result not in cache (either not present or expired). Querying Fingerbank for result");
-            my $fingerbank = fingerbank::Query->new;
-            return $fingerbank->match($args);
-        },
-        {expires_in => FINGERBANK_CACHE_EXPIRE}
-    );
+    my $cache = pf::CHI->new( namespace => 'fingerbank', expires_in => FINGERBANK_CACHE_EXPIRE );
+    my $fingerbank = fingerbank::Query->new(cache => $cache);
+    return $fingerbank->match($args);
```
fdurand, line 95: isn't it $cached_args ?

julsemaan, line 95: No and I actually need to remove cached_args from what I see

lib/pf/fingerbank.pm
```diff
@@ -100,14 +99,9 @@ sub _query {
...
 
-    return $cache->compute_with_undef($cached_args, 
-        sub {
-            $logger->debug("Fingerbank result not in cache (either not present or expired). Querying Fingerbank for result");
-            my $fingerbank = fingerbank::Query->new;
-            return $fingerbank->match($args);
-        },
-        {expires_in => FINGERBANK_CACHE_EXPIRE}
-    );
+    my $cache = pf::CHI->new( namespace => 'fingerbank', expires_in => FINGERBANK_CACHE_EXPIRE );
```
jrouzierinverse, line 93: put expires_in in the conf/chi.conf.example

julsemaan, line 93: We can't now as this will hit the maintenance branch and chi.conf will not be upgraded through it.

If we don't put it here and it's not in chi.conf then it will never expire.

We should create a chi.conf.defaults to handle these cases, as I agree with you, its not clean to have it here now

jrouzierinverse, line 93: Done #1041

julsemaan, line 93: Perfect.

I added a note to #1041 to remove this once its merged (as I doubt it will be in maintenance)

lib/pf/fingerbank.pm
julsemaan, line 95: Fixed, cached_args removed

lib/pf/fingerbank.pm
rouzier, line 93: #1041 could also be added to maintenance.

julsemaan, line 93: maintenance is usually only for bug fixes or to resolve issue that cannot wait for the next release.

Introducing a more complex change like reworking the CHI config seems out of scope for the maintenance

dwlfrth, line 93: I agree with @julsemaan

julsemaan:
Ready to merge so it can be part of 5.5.2 today ?

PR #1041, opened 2015-12-02
Default configuration file for chi.conf

jrouzierinverse:
## Description

Default configuration file for chi.conf
## Impacts

pf::CHI
## NEWS file entries

Enhancements
++++++++++++
- Default configuration file for cache layer
## Delete branch after merge

NO

### DIFF
--- a/conf/chi.conf.example
+++ b/conf/chi.conf.example
@@ -1,6 +1,10 @@
-#An example configuration for a new storage type
+# Look in /usr/local/pf/conf/chi.conf.defaults
+# for allowed parameters 
+# Also look in the CHI and the CHI::Driver::<Class>
+# documentation for allow parameters for cache storage
+#
 
+#An example configuration for a new storage type
 #[storage file]
 #driver=File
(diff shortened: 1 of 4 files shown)


conf/chi.conf.example
```diff
@@ -1,24 +1,10 @@
...
-driver = Redis
-redis_class = Redis::Fast
-server = 127.0.0.1:6379
-prefix = pf
-expires_on_backend = 1
-reconnect=60
+# Look in /usr/local/pf/conf/chi.conf.defaults
+# for allowed parameters 
+# Also look in the CHI and the CHI::Driver::<Class>
+# documentation for allow parameters for cache storage
```
julsemaan: *for allowed parameters

conf/chi.conf.example
```diff
@@ -1,24 +1,10 @@
...
+#
```
julsemaan, line 26: Can you add an example of modifying the cache expiration of a namspace since that would be the most common case

jrouzierinverse:
Done

PR #1042, opened 2015-12-03
Fix/ssidbroadcast agent

whitx:
# Description

Allow the option "Broadcast ssid" (is the ssid visible) to be usable
# Impacts

provisionner workflow
# Dependencies

from https://github.com/inverse-inc/packetfence-windows-agent/pull/4
# Issue

MailingList issue raised by Hack, Daniel, November the 30
# Delete branch after merge

YES
## Bug Fixes

Resolve an issue raised in the mailing list, where no matter if the checkbox was tick or not, it stayed the same (provisionner configuration)

### DIFF
--- a/lib/pf/provisioner/mobileconfig.pm
+++ b/lib/pf/provisioner/mobileconfig.pm
@@ -31,7 +31,7 @@ The set the default OS to IOS
 # Will always ignore the oses parameter provided and use ['Apple iPod, iPhone or iPad']
 has 'oses' => (is => 'ro', default => sub { ['Apple iPod, iPhone or iPad', 'Mac OS X'] }, coerce => sub { ['Apple iPod, iPhone or iPad', 'Mac OS X'] });
 
-=head2 hidden_ssid
+=head2 broadcast
 
 Is the ssid broadcasting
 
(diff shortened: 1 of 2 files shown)


lib/pf/provisioner/mobileconfig.pm
```diff
@@ -31,6 +31,14 @@ The set the default OS to IOS
 # Will always ignore the oses parameter provided and use ['Apple iPod, iPhone or iPad']
 has 'oses' => (is => 'ro', default => sub { ['Apple iPod, iPhone or iPad', 'Mac OS X'] }, coerce => sub { ['Apple iPod, iPhone or iPad', 'Mac OS X'] });
 
+=head2 hidden_ssid
```
julsemaan: its not the right variable name...

whitx:
Would be nice to put in 5.5.2, I tested it only for Windows. (affect all provisionner who use mobileconfig as a base)

julsemaan:
It can't because it changes the agent executable which is non-patchable using the maintenance branch script.

But from what I understand the fix and the updated agent are independant. Is that right ?

In that case we could add only the fix to the maintenance

whitx:
Yes, the agent did not handle the case before, which mean we can push this fix even if the .exe is not  updated yet.

julsemaan:
Perfect,

Has this been tested with Apple + Android to make sure there is no regression.

If yes, we'll merge and put 522b89565de3ba655e4122a50355a3660dbd8cb4 and 6b696d84f6ddda7bc12e8254f129b37143cd2d29 in maint

whitx:
Tested on AppleDevice, working, for Android I can't see the setting if the ssid is hidden or not, but I am able to connect in both case tried only with a ssid which is visible.

julsemaan:
Good enough.

As discussed, the agent will need to be reworked to support other config parameters (like cert check) but this at least makes it work.

PR #1048, opened 2015-12-07
Add a timeout for send/receive on the OMAPI socket

### LINKED ISSUE #1024: No timeout set when reading from OMAPI socket
There is not timeout set when reading from the OMAPI.

pfdhcplistener can wait indefinitly

julsemaan:
# Description

Added configurable timeout for send/receive operations on the OMAPI socket
# Impacts

OMAPI
# Issue

fixes #1024 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added configurable timeout for send/receive operations on the OMAPI socket

### DIFF
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -960,7 +960,7 @@ host=localhost
 # omapi.timeout
 #
 # The timeout on the OMAPI socket
-timeout=1
+timeout=0.5
 
 [provisioning]
 #
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -956,6 +956,11 @@ port=7911
 #
 # The OMAPI host
 host=localhost
+#
+# omapi.timeout
+#
+# The timeout on the OMAPI socket
+timeout=1
 
 [provisioning]
 #
--- a/lib/pf/OMAPI.pm
+++ b/lib/pf/OMAPI.pm
@@ -49,6 +49,14 @@ port of the dhcp server
 
 has port => (is => 'rw', default => sub { 7911 } );
 
+=head2 timeout
+
+Timeout to read from the OMAPI
+
+=cut
+
+has timeout => (is => 'rw');
+
 =head2 buffer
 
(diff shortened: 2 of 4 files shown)


fdurand:
Can it be simplified by https://metacpan.org/pod/IO::Socket::Timeout ?

julsemaan:
Could be, I will test it

julsemaan:
Reworked, please re-review

jrouzierinverse:
Update dependencies

julsemaan:
What dependencies ?

jrouzierinverse:
IO::Socket::Timeout

conf/pf.conf.defaults
```diff
@@ -956,6 +956,11 @@ port=7911
 #
 # The OMAPI host
 host=localhost
+#
+# omapi.timeout
+#
+# The timeout on the OMAPI socket
+timeout=1
```
jrouzierinverse: I think this is too long maybe 0.5 seconds would be better

julsemaan:
No new package necessary for IO::Socket::Timeout

It is already provided by the packages we already install

conf/pf.conf.defaults
julsemaan: Agreed

lib/pf/OMAPI.pm
```diff
@@ -241,6 +251,62 @@ sub connect {
...
+Validate the reply coming from the socket
+
+Return undef if invalid, return the reply if valid
+
+=cut
+
+sub validate_reply {
+    my ($self, $reply) = @_;
+    if (! $reply && ( 0+$! == ETIMEDOUT || 0+$! == EWOULDBLOCK )) {
+        pf::log::get_logger->error("Timeout while reading from the OMAPI socket");
```
jrouzierinverse: use pf::log; On top

julsemaan:
Fixed, re-review

fdurand:
Tested and works as expected, merge

PR #1057, opened 2015-12-08
New feature adds ability to reorder service start and stop 

louismunro:
Not ready.

### DIFF
--- a/lib/pf/cmd/pf/service.pm
+++ b/lib/pf/cmd/pf/service.pm
@@ -89,7 +89,7 @@ our %ACTION_MAP = (
 our $ignore_checkup = $FALSE;
 
 sub _byIndexOrder { 
-    $a->new->orderIndex <=> $b->new->orderIndex;
+    $a->orderIndex <=> $b->orderIndex;
 }
 
 sub parseArgs {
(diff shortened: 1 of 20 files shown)


lib/pf/cmd/pf/service.pm
```diff
@@ -88,15 +88,24 @@ our %ACTION_MAP = (
 
 our $ignore_checkup = $FALSE;
 
+sub _byIndexOrder { 
+    $a->new->orderIndex <=> $b->new->orderIndex;
+}
+
 sub parseArgs {
```
jrouzierinverse, line 95: To ensure that the sorting is "stable" add the following pragma.
use sort 'stable';

lib/pf/services/manager/pfsetvlan.pm
```diff
@@ -14,6 +14,7 @@ pf::services::manager::pfsetvlan
 use strict;
 use warnings;
 use Moo;
+use pf::config;
```
julsemaan: why ?

You don't seem to be doing anything with it

louismunro: I thought it looked better this way.

(I forgot to remove it when I refactored this code no to need it anymore...)

louismunro:
Review meeeeeeee.....

fdurand:
The code looks good to me. Tested and works. Ready to merge

PR #1058, opened 2015-12-09
change for support Chinese in configure setting

echotor:
I change something for version 5.4
It is supported for Chinese in configure setting and mysql.

### DIFF
--- a/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Authentication.pm
+++ b/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Authentication.pm
@@ -17,8 +17,10 @@ use HTTP::Status qw(:constants is_error is_success);
 use Moose;
 use namespace::autoclean;
 use POSIX;
+use utf8;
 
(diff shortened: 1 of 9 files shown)


julsemaan:
Can you confirm that your switches work properly (assigning a VLAN with a role that has non-ASCII characters)

Because what you did looks good on the administration interface side but I'm not sure what will happen in all the models.

echotor:
Yes, we confirm it in our company and it work well.
We change the code as this and change the default character of mysql. And we find it work well on the administration interface side and the authenticate model. (we use the Active Directory source)

echotor:
We do another thing that we change the default character of databases, tables and field of mysql from latin to utf8.

julsemaan:
That's a good thing then.

Only thing is looking at your code, this looks like a huge hack since there are a lot of places that may require to support the UTF-8 correctly as I stated in the issue you opened previously.

This requires a much cleaner job.

If it works for you now, then its all good but I doubt the team will want to integrate this PR as it is missing a lot of components and will require a lot of testing.

PR #1066, opened 2015-12-14
Feature/switch groups

julsemaan:
# Description

Adds switch groups
Adds Hierarchy ConfigStore to define inheritance in configuration sections
# Impacts

Interfaces ConfigStore (because of the change to the Group ConfigStore)
Switches configuration
SwitchFactory
radius_nas table
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added the ability to define switch groups in the switches configuration
# Upgrade notes
### Default type for the switches.

The default type for the switches now needs to be set explicitly. Add the following line in the default section of `/usr/local/pf/conf/switches.conf`

`type=Generic`


julsemaan:
Ready for review

jrouzierinverse:
ERROR: Caught exception in pfappserver::Controller::Configuration->section "Use of uninitialized value in exists at /usr/local/pf/lib/pf/ConfigStore.pm line 195." (pfappserver::PacketFence::Controller::Root::end)

html/pfappserver/lib/pfappserver/Model/Config/SwitchGroup.pm
```diff
@@ -0,0 +1,85 @@
...
+
+Override the parent method to validate we don't remove a group that has childs
+
+=cut
+
+sub remove {
+    my ($self, $id) = @_;
+    pf::log::get_logger->info("Deleting $id");
+    my @childs = $self->configStore->members($id, $self->idKey);
+    use Data::Dumper ;pf::log::get_logger->info(Dumper(\@childs));
```
jrouzierinverse: Remove Data::Dumper

t/unittest/switch_groups.t
```diff
@@ -0,0 +1,90 @@
...
+
+use pf::node;
+use Test::More tests => 8;
+
+#This test will running last
+use Test::NoWarnings;
+
+use pf::SwitchFactory;
+
+#use Data::Dumper ; print Dumper(\%pf::SwitchFactory::SwitchConfig);
```
jrouzierinverse: remove Data::Dumper

conf/switches.conf.example
```diff
@@ -6,6 +6,7 @@
 # http://www.fsf.org/licensing/licenses/gpl.html
 
 [default]
+type=Cisco::Catalyst_2960
```
jrouzierinverse: There should be no default.type

html/pfappserver/lib/pfappserver/Model/Config/Switch.pm
```diff
@@ -17,6 +17,7 @@ use namespace::autoclean;
 use pf::config::cached;
 use pf::config;
 use pf::ConfigStore::Switch;
+use List::MoreUtils qw(any);
```
jrouzierinverse: Remove

lib/pf/ConfigStore.pm
```diff
@@ -342,13 +362,13 @@ sub sortItems {
 
 =cut
 
-sub cleanupAfterRead { }
+sub cleanupAfterRead { return $_[1] }
```
jrouzierinverse: why are you returning this?

lib/pf/ConfigStore.pm
```diff
@@ -342,13 +362,13 @@ sub sortItems {
...
 
 =head2 cleanupBeforeCommit
 
 =cut
 
-sub cleanupBeforeCommit { }
+sub cleanupBeforeCommit { return $_[1] }
```
jrouzierinverse: why are you returning this?

conf/switches.conf.example
julsemaan: There should be...

Switches will always extend 'default'

There should even have been one already in my opinion.

Users are already setting one anyway, I'm just defining one so there is no need to handle the case where there is none and make the code heavier just for a config parameter.

I'm open to change it but that will require more arguments than a simple 'should not'

lib/pf/ConfigStore.pm
julsemaan: my bad, bad checkin

lib/pf/ConfigStore.pm
julsemaan: my bad, bad checkin

julsemaan:
For the exception you reported, you must not be running the latest version because line 195 is the declaration of a subroutine....

jrouzierinverse:
You are right I thought I had the latest version for testing.

However a new error as appeared.
I tried to create a new switch when saving I got the following error.

ERROR: Caught exception in pfappserver::Controller::Config::Switch->create "Use of uninitialized value in string eq at /usr/local/pf/lib/pf/ConfigStore/Switch.pm line 125."

conf/switches.conf.example
jrouzierinverse: My main issue here is you are changing previous behavior with no real added benefit.
In my opinion The default switch and by extension switch group is a "power user" feature.

By making the decision for the user that default type is "Cisco::Catalyst_2960"
I believe it is easier for people to make a mistake when configuring their switches.

Also by forcing user to add the switch type for a switch by default I feel it will reduce mistakes when configuring.

If a user decides to configure the default type then they should know the danger.

julsemaan: Yes but then we have to handle the 'mandatory' case for when there is and there isn't a default type set

It makes the code a bit harder to follow.

What we could do, is allow to use the 'pf::Switch' in the type and make it default so we at least have a base.

For me the default configuration should define all elements of the childs it has and they change them if needed.

I agree choosing the switch is not the best way but using pf::Switch could just be the best way.

Or create a 'Generic Switch' module which we currently lack and assign it as the default type

julsemaan:
For your error, are you using the switches conf with the default type being set :) ?

julsemaan:
Handled the case of your error.

Should be fine now

julsemaan:
Fixed, re-review

jrouzierinverse:
I am getting an error when I save a switch.
The switch is stored in the config.

image

julsemaan:
Logs ?

jrouzierinverse:
Have you tested saving a switch?

Dec 16 09:13:32 httpd.admin(25621) ERROR: Use of uninitialized value $error in concatenation (.) or string at /usr/local/pf/html/pfappserver/script/../lib/pfappserver/Base/Model/Config.pm line 310.

julsemaan:
Yes obviously, I even made a demo in front of you....

jrouzierinverse:
I meant afterwards.

html/pfappserver/root/config/switch/list.tt
```diff
@@ -73,23 +78,26 @@
         <tbody>
           [% FOREACH switch IN items %]
           <tr>
-            <td><a href="[% c.uri_for(c.controller('Config::Switch').action_for('view'), [ switch.id ]) %]">[% switch.id | html %]</a>
+            <td><a href="[% c.uri_for(c.controller(controller_namespace).action_for('view'), [ switch.id ]) %]">[% switch.id | html %]</a>
```
jrouzierinverse: c.controller(controller_namespace).action_for 

can be rewritten as

c.controller.action_for

Since c.controller() will return the current controller when there are no arguments passed to it.

html/pfappserver/root/config/switch/view.tt
```diff
@@ -1,16 +1,19 @@
 [% form.block('triggers').render | none %]
 
-<form autocomplete="off" name="modalSwitch" class="form-horizontal form-condensed" action="[% action_uri %]" method="post">
+<form autocomplete="off" name="modalSwitch" id="modalSwitch" class="form-horizontal form-condensed" action="[% action_uri %]" method="post">
```
jrouzierinverse: The modal has the same id as the form this could cause problems

html/pfappserver/root/static/admin/config/switches.css
```diff
@@ -0,0 +1,8 @@
+
+#modalSwitch{
```
jrouzierinverse, line 2: use the modal-large class on the modal in the index page instead.

lib/pf/ConfigStore.pm
```diff
@@ -165,6 +165,14 @@ format the id
 
 sub _formatId { return $_[1]; }
```
jrouzierinverse: Maybe this should be renamed to _formatSectionName or _cleanupSectionName to avoid confusion with _cleanupId

lib/pf/ConfigStore/Switch.pm
```diff
@@ -18,108 +18,37 @@ use pf::log;
...
+use List::MoreUtils qw(part any);
 use pfconfig::manager;
+use pf::freeradius;
 
 extends qw(pf::ConfigStore Exporter);
+with 'pf::ConfigStore::Hierarchy';
+
 
-sub configFile {$switches_config_file}
+sub configFile { $pf::file_paths::switches_config_file };
```
jrouzierinverse: Remove ; at the end

lib/pf/ConfigStore/Switch.pm
```diff
@@ -18,108 +18,37 @@ use pf::log;
...
 
 sub pfconfigNamespace {'config::Switch'}
 
-sub default_section { 'default' }
+sub default_section { undef }
```
jrouzierinverse: Remove this and let it inherit from the base class

conf/switches.conf.example
```diff
@@ -21,7 +22,7 @@ isolationRole = isolation
 macDetectionRole = macDetection
 voiceRole = voice
 inlineRole = inline
-VoIPEnabled = no
+VoIPEnabled = N
```
jrouzierinverse, line 25: Maybe it is time to move this to conf/switches.conf.defaults

julsemaan, line 25: Would switches.conf.defaults merge individual attributes in two sections with the same id in the config ?

If it doesn't then it would be next to useless because the default switch is always modified

jrouzierinverse, line 25: Since merging can mean a lot of things.

You can read how it works here
http://search.cpan.org/~shlomif/Config-IniFiles-2.88/lib/Config/IniFiles.pm#IMPORT_/_DELTA_FEATURES

julsemaan, line 25: This is what I mean :

In switches.conf :

```
[default]
param1=value1
```

In switches.conf.defaults :

```
[default]
param1=value69
param2=value2
```

We need to have the following result : 

```
[default]
param1=value1
param2=value2
```

I am asking you because the ConfigStore and pf::config::cached are doing a lot of magic above the Config::IniFiles library and you are the only one that knows how it works. So I would expect you to shed some light on this

jrouzierinverse, line 25: That is exactly how it works.

jrouzierinverse:
Should we allow people to clone 'default'?
Since it is type of super group I do not know if it make sense anymore.

image

julsemaan:
Why not ?

That gives more than 1 way of doing things and keeps it there for people that were used to using it.

jrouzierinverse:
OK I have no strong opinion on this I was wondering if it still makes sense.

html/pfappserver/root/static/admin/config/switches.css
julsemaan, line 2: modal large is too big

julsemaan:
Adjustements made except for the switches.conf.defaults since I think the default section already accomplishes the goal and the default section would be the only one in switches.conf.defaults

Can be discussed in a meeting but I'd say to leave it out for now

html/pfappserver/root/config/switch/index.tt
```diff
@@ -1,25 +1,48 @@
...
       [%- IF can_access("SWITCHES_CREATE") %]
       <div class="form-actions">
-        <a id="createSwitch" class="btn" href="[% c.uri_for(c.controller('Config::Switch').action_for('create')) %]">[% l('Add switch') %]</a>
+
+        [% IF controller_namespace == 'Config::Switch' %]
+          <div class="btn-group">
+            <a class="btn dropdown-toggle" data-toggle="dropdown" href="#">[% l('Add switch to group') %] <span class="caret"></span></a>
+            <ul class="dropdown-menu">
+              [% FOREACH group IN switch_groups %]
+              <li><a class="createSwitch" href="[% c.uri_for(c.controller(controller_namespace).action_for('create_in_group'), group) %]">[% group %]</a></li>
```
jrouzierinverse: c.controller(controller_namespace) to c.controller

html/pfappserver/root/config/switch/index.tt
```diff
@@ -1,25 +1,48 @@
...
+              [% #Add a separator after the default group %]
+              [% IF group == "default" %]<li role="separator" class="divider"></li>[% END %]
+              [% END %]
+            </ul>
+          </div>
+        [% ELSE %]
+          <a class="btn createSwitch" href="[% c.uri_for(c.controller(controller_namespace).action_for('create')) %]">[% l('Add ' _ model_name) %]</a>
```
jrouzierinverse: c.controller(controller_namespace) to c.controller

julsemaan:
Done

lib/pf/ConfigStore/Switch.pm
```diff
@@ -18,108 +18,34 @@ use pf::log;
...
 
-=head2 Methods
+=cut
 
-=over
+sub commit {
+    my ($self) = @_;
+    my ($result, $error) = $self->SUPER::commit();
+    pf::log::get_logger->info("commiting via Switch configstore");
+    freeradius_populate_nas_config( \%pf::SwitchFactory::SwitchConfig );
```
jrouzierinverse, line 46: I am asking just to make sure.
Would freeradius_populate_nas_config always get the latest version of pf::SwitchFactory::SwitchConfig?
Or is it possible it will get the old pf::SwitchFactory::SwitchConfig?

lib/pf/Switch/Generic.pm
```diff
@@ -0,0 +1,67 @@
...
+=over
+
+=cut
+
+# CAPABILITIES
+# access technology supported
+sub supportsWirelessDot1x { return $TRUE; }
+sub supportsWirelessMacAuth { return $TRUE; }
+sub supportsWiredDot1x { return $TRUE; }
+sub supportsWiredMacAuth { return $TRUE; }
```
jrouzierinverse, line 35: This has no description that means no user can select it from the GUI.

jrouzierinverse, line 35: After that it should be good

lib/pf/ConfigStore/Switch.pm
julsemaan, line 46: Yes because the ConfigStore commits synchronously so when returning from the super method the new SwitchConfig has been built

lib/pf/Switch/Generic.pm
julsemaan, line 35: you're right

fixed it

PR #1070, opened 2015-12-16
There shall be only one logger

jrouzierinverse:
## Description

Remove the use of two seperate logging subsystem
## Impacts

Logging in pfconfig namespace
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++

Bug Fixes
+++++++++
## Delete branch after merge

NO


jrouzierinverse:
This also slightly improves the performance of pfconfig.

lib/pfconfig/manager.pm
```diff
@@ -197,26 +200,18 @@ That sends the signal that the raw memory is expired
...
     $filename = untaint_chain($filename);
 
-    if ( !-e $filename ) {
-        my $fh;
-        unless ( open( $fh, ">$filename" ) ) {
-            $logger->error("Can't create $filename\nPlease run 'pfcmd fixpermissions'");
-            return 0;
-        }
+    if (sysopen(my $fh,$filename,O_RDWR | O_CREAT)) {
+        POSIX::2008::futimens(fileno $fh);
```
julsemaan: I had an issue with this, this is why it was changed to the ugly touch command.

There were a few microseconds of difference between this and the stat in Time::HiRes which made unit tests fail

julsemaan:
Other than my comment on the replacement of the file touching, this is all good

lib/pfconfig/manager.pm
rouzier: Ok I normalized POSIX::2008::futimens to use the same resolution of Time::HiRes.
My changes have been pushed please re-review

jrouzierinverse: All the test are passing now

PR #1072, opened 2015-12-17
The Null Source as a signup source

### LINKED ISSUE #786: Do not make a Null source exclusive
There are some scenarios that could call for a Null source to allowed to be used with other sources.
For instance Null and Facebook where you can allow users to choose to login without signing through facebook but have a lower time limit.

jrouzierinverse:
## Description

Make the null source as guest source instead of an exclusive source
## Impacts

The portal workflow
## Issues

Fixes #786
## NEWS file entries

Enhancements
++++++++++++
- Null Source is now considered an external source
## Delete branch after merge

NO


louismunro:
No objections your honor.
Merging.

PR #1074, opened 2015-12-18
Fix/pfconfig overlay

### LINKED ISSUE #1067: pfconfig : race condition when adding a namespace to the _namespace_overlay value
When adding a namespace to the namespace overlay there is a race condition which should be handled by locking writes on the backend on the keyed table or listing the overlayed namespaces using the list key capability of the backend

julsemaan:
# Description

Rework pfconfig overlay so it is based on the backend key list instead of a self maintained list of the overlayed namespaces
Prevents a race condition in an active/active cluster when both servers are reloading at the same time which makes the _namespace_overlay miss a few overlays
# Impacts

pfconfig overlayed namespaces (active/active clustering config)
pfconfig backends will now always need to be able to list their keys
# Issue

fixes #1067 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Reworked the configuration backends to prevent a race condition of the configuration namespaces in active/active cluster

### DIFF
--- a/lib/pf/cmd/pf/pfconfig.pm
+++ b/lib/pf/cmd/pf/pfconfig.pm
@@ -113,6 +113,45 @@ sub action_list {
     return $EXIT_SUCCESS;
 }
 
+=head2 action_list_overlayed
+
+List all pfconfig overlayed namespaces
+
+=cut
+
+sub action_list_overlayed {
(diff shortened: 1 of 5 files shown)


lzammit:
Looks good to me. :8ball:

PR #1077, opened 2015-12-21
Feature/radius rbac

fdurand:
## Description

Use attributes returned by radius (Radius authentication source) to define rules
## Impacts

No
## NEWS file entries

New Features
++++++++++++
- Radius authentication can use returned attributes to define rules
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -293,6 +293,11 @@ sub match {
 
     if (ref($source_id) eq 'ARRAY') {
         @sources = @{$source_id};
+    } elsif (!(defined(ref($source_id)) && $source_id ne '') {
+        my $source = getAuthenticationSource($source_id);
+        if (defined $source) {
+            @sources = ($source);
+        }
     } else {
         @sources = ($source_id);
     }
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -292,7 +292,7 @@ sub match {
     }
     if (ref($source_id) eq 'ARRAY') {
         @sources = @{$source_id};
-    } elsif (defined($source_id->id)) {
+    } elsif (defined(ref($source_id))) {
         @sources = ($source_id);
     } else {
         my $source = getAuthenticationSource($source_id);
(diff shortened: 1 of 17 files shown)


lib/pf/authentication.pm
```diff
@@ -243,19 +243,20 @@ sub authenticate {
...
         if ($result) {
             $logger->info("Authentication successful for $username in source ".$current_source->id." (".$current_source->type.")");
-            return ($result, $message, $current_source->id);
+            return ($result, $message, $current_source->id, $connection);
         }
     }
 
     $logger->trace("Authentication failed for '$username' for all ".scalar(@sources)." sources");
-    return ($FALSE, $message ? $message : $AUTH_FAIL_MSG);
+    return ($FALSE, $message ? $message : $AUTH_FAIL_MSG, undef);
```
julsemaan: Not sure this does what you want, the 3rd return is the source id not the connection but anyway it ends up as undef so just remove the additional undef

lib/pf/authentication.pm
```diff
@@ -243,19 +243,20 @@ sub authenticate {
...
         $logger->trace("Trying to authenticate '$username' with source '".$current_source->id."'");
         eval {
-            ($result, $message) = $current_source->authenticate($username, $password);
+            ($result, $message, $connection) = $current_source->authenticate($username, $password);
         };
         # First match wins!
         if ($result) {
             $logger->info("Authentication successful for $username in source ".$current_source->id." (".$current_source->type.")");
-            return ($result, $message, $current_source->id);
+            return ($result, $message, $current_source->id, $connection);
```
julsemaan: I am not sure that I like adding another param to the return.

I would suggest the following : 
Instead of returning the source ID, we return the source and then we populate an attribute that we make the same across all sources for the connection (like _connection)

This will require rewriting the code around authentication calls but would make a cleaner job IMO.

Or maybe something else than what is above but its just that returning 4 elements is not really in our common practice

extrafu:
Why close it?

fdurand:
I have to rework it, so i will reopen it after.

PR #1080, opened 2015-12-21
Feature/cache for get if index by nas port

fdurand:
## Description

Added a cache for getIfIndexByNasPortId to avoid snmpread on each radius request.
## Impacts

radius
## NEWS file entries

Enhancements
+++++++++++
- Cache the return of getIfIndexByNasPortId for one day to avoid snmpread on each radius request
## Delete branch after merge

YES

### DIFF
--- a/conf/chi.conf.defaults
+++ b/conf/chi.conf.defaults
@@ -64,3 +64,10 @@ expires_on_backend = 1
 #
 # How long to try to reconnect before giving up
 reconnect=60
+
+[storage switch]
+#
+# storage switch.expires_in
+#
+# Amount of time from the current time to expire an entry
+expires_in=1d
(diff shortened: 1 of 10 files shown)


julsemaan:
One word, 

NICE!

julsemaan:
Nothing to say on my side this looks pretty neat.

Did you test one of the switches to make sure it works ?

If yes, I would say to merge it for PF 5.6

Its been too long we don't have this

louismunro:
Two words: "very" .  "nice".

fdurand:
Tested with the 2960C,, on the first radius request the switch object contain the snmp answer, on the second one the snmp answer is empty but i can calculate the ifindex from snmp result in the cache.

jrouzierinverse:
It would be easier to just wrap get the response in pf::Switch instead of creating a pf::Net::SNMP module.

PR #1088, opened 2016-01-04
Have pfqueue read it's config from pfconfig

jrouzierinverse:
## Description

Have pfqueue read it's config from pfconfig
## Impacts

pfqueue
## NEWS file entries

Enhancements
++++++++++++
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/config/pfqueue.pm
+++ b/lib/pf/config/pfqueue.pm
@@ -1,0 +1,55 @@
+package pf::config::pfqueue;
+
+=head1 NAME
+
+pf::config::pfqueue -
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::config::pfqueue
+
+=cut
+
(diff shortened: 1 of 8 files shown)


lib/pf/config/pfqueue.pm
```diff
@@ -0,0 +1,55 @@
+package pf::config::pfqueue;
```
julsemaan, line 1: I like that new way of separating the configs

louismunro:
Redundant.
See #1089.

lib/pf/config/pfqueue.pm
```diff
@@ -0,0 +1,55 @@
...
+
+=head1 NAME
+
+pf::config::pfqueue -
```
julsemaan, line 5: Remove the `-` at the end

PR #1089, opened 2016-01-04
pfcmd pfqueue

jrouzierinverse:
## Description

New pfcmd "pfqueue" command for mananging the queue
## Impacts

pfcmd
## Dependencies

Depends on #1088
## NEWS file entries

New Features
+++++++++++
- New pfcmd "pfqueue" command for mananging the queue

Bug Fixes
++++++++
## Delete branch after merge

NO


lib/pf/config/pfqueue.pm
```diff
@@ -0,0 +1,55 @@
...
+use pfconfig::cached_hash;
+
+BEGIN {
+    use Exporter ();
+    our ( @ISA, @EXPORT, @EXPORT_OK );
+    @ISA = qw(Exporter);
+    @EXPORT = qw(%ConfigPfQueue);
+}
+
+tie our %ConfigPfQueue, 'pfconfig::cached_hash', 'config::Pfqueue';
```
julsemaan: Either the hash should be named ConfigPfqueue or the namespace be named config::PfQueue

From what I see, it seems the first one would be correct

lib/pf/config/pfqueue.pm
```diff
@@ -0,0 +1,55 @@
+package pf::config::pfqueue;
+
+=head1 NAME
+
+pf::config::pfqueue -
```
julsemaan: Remove the `-` at the end

lib/pf/pfqueue/stats.pm
```diff
@@ -0,0 +1,128 @@
+package pf::pfqueue::stats;
+
+=head1 NAME
+
+pf::pfqueue::stats - Object for collecting pfqueue stats
+
+=head1 DESCRIPTION
+
+
```
julsemaan: Please add a short description

lib/pf/cmd/pf/pfqueue.pm
```diff
@@ -0,0 +1,140 @@
...
+=head1 SYNOPSIS
+
+ pfcmd pfqueue <command> [options]
+
+  Commands:
+
+   clear <queue>    | clear the queue
+   list             | list all the current queues
+   stats            | show the stats
+   count <queue>    | show the count of the a queue
```
julsemaan: *of a queue

lib/pf/cmd/pf/pfqueue.pm
```diff
@@ -0,0 +1,140 @@
...
+sub action_count {
+    my ($self) = @_;
+    my ($queue) = $self->action_args;
+    my $redis = $self->redis;
+    my $real_queue = "Queue:$queue";
+    print $redis->llen($real_queue),"\n";
+    return $EXIT_SUCCESS;
+}
+
+sub redis {
```
julsemaan: missing zi doc

lib/pf/config/pfqueue.pm
```diff
@@ -0,0 +1,55 @@
+package pf::config::pfqueue;
+
+=head1 NAME
+
+pf::config::pfqueue - Configuration from conf/pfqueue.conf
```
julsemaan: Not sure your module template is correct,

It seems this should only be the name and the description should be below.

Now it's reversed

lib/pf/pfqueue/stats.pm
```diff
@@ -0,0 +1,128 @@
...
+use namespace::autoclean;
+use pf::config::pfqueue;
+use pf::constants::pfqueue qw($PFQUEUE_COUNTER $PFQUEUE_QUEUE_PREFIX $PFQUEUE_EXPIRED_COUNTER);
+use Redis::Fast;
+
+has redis => (is => 'ro', builder => 1, lazy => 1);
+
+=head1 METHODS
+
+=head2 my $counters = $self->counters()
```
julsemaan, line 24: Doesn't seem right.

Same across the file

lib/pf/pfqueue/stats.pm
```diff
@@ -0,0 +1,128 @@
...
+
+Get all the counters for a namespace
+
+=cut
+
+sub _get_counters_for {
+    my ($self, $counter_name) = @_;
+    my $redis = $self->redis;
+    my %counters = $redis->hgetall($counter_name);
+    my @counters = map { &_counter_map(\%counters, $_) } sort keys %counters;
```
julsemaan: Minor detail, is it necessary to put the `&` before the sub call ?

lib/pf/pfqueue/stats.pm
jrouzierinverse, line 24: https://www.youtube.com/watch?v=29M_VElHoFI

lib/pf/pfqueue/stats.pm
```diff
@@ -0,0 +1,128 @@
...
+}
+
+
+=head2 _build_redis
+
+=cut
+
+sub _build_redis {
+    my ($self) = @_;
+    return Redis::Fast->new( %{$ConfigPfQueue{consumer}{redis_args}});
```
julsemaan: There are multiple place where this is done, this should be in a central point so its not duplicated

lib/pf/pfqueue/stats.pm
julsemaan, line 24: Our standard is 

`=head2 counters`

Not to put the usage in the head name

lib/pf/cmd/pf/pfqueue.pm
```diff
@@ -0,0 +1,140 @@
...
+=head2 action_count
+
+=cut
+
+sub action_count {
+    my ($self) = @_;
+    my ($queue) = $self->action_args;
+    my $redis = $self->redis;
+    my $real_queue = "Queue:$queue";
+    print $redis->llen($real_queue),"\n";
```
julsemaan: I think it would be cleaner if this was in the producer so it is abstracted and would centralize the access.

lib/pfconfig/namespaces/config/Pfqueue.pm
```diff
@@ -0,0 +1,92 @@
...
+
+sub build_child {
+    my ($self) = @_;
+    my %tmp_cfg = %{ $self->{cfg} };
+    foreach my $queue_section ( $self->GroupMembers('queue') ) {
+        my $queue = $queue_section;
+        $queue =~ s/^queue //;
+        my $data = delete $tmp_cfg{$queue_section};
+        # Set defaults
+        $data->{workers} //= 10;
```
julsemaan: The 10 should be a constant

lib/pfconfig/namespaces/config/Pfqueue.pm
```diff
@@ -0,0 +1,92 @@
...
+        $data->{has_delayed_queue} = isenabled($data->{has_delayed_queue});
+        if($data->{has_delayed_queue}) {
+            $data->{delayed_queue_batch} //= 100;
```
julsemaan: The 100 should be a constant

lib/pfconfig/namespaces/config/Pfqueue.pm
```diff
@@ -0,0 +1,92 @@
...
+            $data->{delayed_queue_workers} //= 1;
```
julsemaan: The 1 should be a constant

lib/pfconfig/namespaces/config/Pfqueue.pm
```diff
@@ -0,0 +1,92 @@
...
+            # Normalize to milliseconds
+            $data->{delayed_queue_sleep} = ($data->{delayed_queue_sleep} // 100 ) * 1000;
```
julsemaan: The 100 should be a constant

lib/pf/cmd/pf/pfqueue.pm
jrouzierinverse: I think it will be better in the stats class.
The producer should not care about the length of a queue.

julsemaan: Just makes it harder to use the stats module with 2 different backends

The queuing backend should be swapable via a config parameter from what we discussed during the implementation meetings.

Having it directly in the module like this makes it harder to have two backends share the same code

jrouzierinverse: Just the opposite.

Maybe the name pf::pfqueue::stats is not the best name.

But this If there is a new backend 
We would just have to make a new pf::pfqueue::stats module.

pf::pfqueue::stats
pf::pfqueue::stats::redis
pf::pfqueue::stats::semaan

If we ever do change the backend queue I will do the work to refactor it.

PR #1129, opened 2016-01-12
Windows agent - How we provide it

whitx:
# Description

Change the way we provide the packetfence-windows-agent (provisioning)
# Impacts
- Provisioning flow
- Installation 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- packetfence-windows-agent is not in packetfence source code anymore. It will be download during the installation via inverse website.

This has been made because every time we update it, this add 15mb more the GitHub repo, where this could be managed differently. Also this will allow user to update their agent without update packetfence.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,4 +138,9 @@ fingerbank:
 	rm -f /usr/local/pf/lib/fingerbank
 	ln -s /usr/local/fingerbank/lib/fingerbank /usr/local/pf/lib/fingerbank \
 
+.PHONY: windows-agent
+
+windows-agent:
+	wget https://inverse.ca/downloads/packetfence-windows-agent.exe -P /usr/local/pf/html/captive-portal/content/
+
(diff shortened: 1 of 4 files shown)


julsemaan:
That's great and needed to be done !

Makefile
```diff
@@ -138,4 +138,9 @@ fingerbank:
 	rm -f /usr/local/pf/lib/fingerbank
 	ln -s /usr/local/fingerbank/lib/fingerbank /usr/local/pf/lib/fingerbank \
 
+.PHONY: windows-agent
+
+windows-agent:
+	wget https://inverse.ca/downloads/packetfence-windows-agent.exe -P /usr/local/pf/html/captive-portal/content/
```
fdurand: https://inverse.ca/downloads/PacketFence/addons/packetfence-windows-agent.exe

debian/control
```diff
@@ -21,7 +21,7 @@ Breaks: libdata-alias-perl
 Depends: ${misc:Depends}, vlan, make,
  openssl, openssl-blacklist, openssl-blacklist-extra,
  mysql-server|mariadb-server|mariadb-galera-server,
- snmp, snmptrapfmt, snmp-mibs-downloader, conntrack,
+ snmp, snmptrapfmt, snmp-mibs-downloader, conntrack, wget,
```
louismunro: This is a build dependency, not a runtime dependency.
This will break packaging.

Fix it for both debian and redhat please.

Issue #935, opened 2015-10-01
incorrect Cisco AVpair handling?


stefanmdor:
Hi,

I'm using the feature 'Role mapping by access list'. The access-list contains multiple lines, but only the first one is applied by the switch.
Packetfence Version: 5.3.1
Switch: WS-C2960X-24PS-L
Switch IOS Image: c2960x-universalk9-mz.152-3.E.bin

Debugging turns out that packetfence is sending the whole access-list as follows:
Oct  1 15:58:51.603: RADIUS:  Vendor, Cisco       [26]  39
Oct  1 15:58:51.603: RADIUS:   Cisco AVpair       [1]   33  "ip:inacl#101=permit tcp any any"
Oct  1 15:58:51.603: RADIUS:  Vendor, Cisco       [26]  39
Oct  1 15:58:51.603: RADIUS:   Cisco AVpair       [1]   33  "ip:inacl#101=permit udp any any"

The first entry with line number 101 is applied. The second entry has also line number 101 and is ignored by my switch. Other switches or other firmware might handle this differently. My resulting ACL is:
Extended IP access list GigabitEthernet1/0/1#v4#8A99BF4 (per-user)
    101 permit tcp any any

Is it possible to increment the line number for every ACL entry? I think, it has to look like this:
ip:inacl#101=permit tcp any any
ip:inacl#102=permit udp any any

kind regards
Stefan

fdurand:
Hi Stefan,
try that:

```
diff --git a/lib/pf/Switch/Cisco/Catalyst_2960.pm b/lib/pf/Switch/Cisco/Catalyst_2960.pm
index 1344f84..b803780 100644
--- a/lib/pf/Switch/Cisco/Catalyst_2960.pm
+++ b/lib/pf/Switch/Cisco/Catalyst_2960.pm
@@ -506,8 +506,10 @@ sub returnRadiusAccessAccept {
         if( defined($user_role) && $user_role ne ""){
             my $access_list = $self->getAccessListByName($user_role);
             my @av_pairs;
+            my $acl_num = 101;
             while($access_list =~ /([^\n]+)\n?/g){
-                push(@av_pairs, $self->returnAccessListAttribute."=".$1);
+                push(@av_pairs, $self->returnAccessListAttribute($acl_num)."=".$1);
+                $acl_num ++;
                 $logger->info("[$mac] (".$self->{'_id'}.") Adding access list : $1 to the RA
             } 
[...]

PR #1151, opened 2016-01-14
Increment the acl number because it doesn't work on some Cisco switch

fdurand:
## Description

Increment acl number because using only 101 as acl number does not works on cisco switches 
## Impacts

NO
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
Increment acl number when we do acl per port (#935)

Bug Fixes
+++++++++
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/Switch/Cisco/Catalyst_2960.pm
+++ b/lib/pf/Switch/Cisco/Catalyst_2960.pm
@@ -505,8 +505,10 @@ sub returnRadiusAccessAccept {
         if( defined($args->{'user_role'}) && $args->{'user_role'} ne ""){
             my $access_list = $self->getAccessListByName($args->{'user_role'});
             if ($access_list) {
+                my $acl_num;
                 while($access_list =~ /([^\n]+)\n?/g){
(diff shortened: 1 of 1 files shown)


lib/pf/Switch/Cisco/Catalyst_2960.pm
```diff
@@ -505,8 +505,10 @@ sub returnRadiusAccessAccept {
         if( defined($args->{'user_role'}) && $args->{'user_role'} ne ""){
             my $access_list = $self->getAccessListByName($args->{'user_role'});
             if ($access_list) {
+                my $acl_num;
```
julsemaan: make it `my $acl_num = 0` since otherwise the first one will be empty

fdurand: yes you are right, a copy and paste issue ...

julsemaan: That damned copy paste. Always messing up our code.

PR #1162, opened 2016-01-20
don't strip the username in the portal unless all the sources require it

julsemaan:
# Description

don't strip the username in the portal unless all the sources require it
# Impacts

Portal authentication
# Delete branch after merge

(REQUIRED)
YES | NO
# NEWS file entries
## Enhancements
- The username on the portal will not be stripped unless the authentication sources used all need it to be stripped

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -13,7 +13,7 @@ use pf::locationlog;
 use pf::authentication;
 use pf::Authentication::constants;
 use HTML::Entities;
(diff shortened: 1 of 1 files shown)


jrouzierinverse:
Good idea.
Maybe we should push the logic of stripping the username to the source itself.
That will allow use a mix of sources to have stripped_user_name enabled or disabled.

julsemaan:
We can't really make it as you say as the 802.1x connections already provide the stripped user name as provided by FreeRADIUS.

It could be confusing to not honor what FreeRADIUS does in 802.1x but then it might be confusing to have it done in two different places with 2 different logics (portal + FreeRADIUS)

louismunro:
Can't complain.
Merging.

PR #1166, opened 2016-01-25
Redefine provisioning template

### LINKED ISSUE #1119: Provisonning (Apple/Android)
Issue with the profile we provide in provisionning, 

On an apple devices, (peap configuration) you have to provide your username/password twice, once when you install the profile and once when you want to connect to the network.

On android devices, (peap) you need to renconfigure your network and adding identity/password when trying to connect to a network that was autoconfigured.

whitx:
# Description

Rework templates for the profile we are provisioning. Design 1 template per EAP type, (PEAP, EAP-TLS, No EAP).
# Impacts

Provisioning workflow
# Issue

fixes #1119 
# Delete branch after merge

NO
# NEWS file entries
## Enhancements
- Rework provisioning template
## Bug Fixes
- Issue while provisioning on IOS9 and Android, where PEAP need some manual tuning after auto-configuration. (#1119)


whitx:
[outdated] Tested with all profile on Windows + IOS , Android: EAP-TLS OK, PEAP issue because the agent is looking for password?

whitx:
PEAP, EAP-TLS, and standard WAP2-PSK tested on all platform, everything happens as expected

lib/pf/provisioner/mobileconfig.pm
```diff
@@ -157,6 +216,25 @@ sub sign_profile {
...
+Creates a template from the eap type
+
+=cut
+
+sub _build_profile_template {
+    my ($self) = @_;
+    my $eap_type = $self->eap_type;
+    if (defined($eap_type)) {
+        if ($eap_type == 13) {
+            return "wireless-profile-eap.xml";
```
julsemaan: This looks more like the TLS profile, not a generic EAP profile..

PR #1167, opened 2016-01-25
Feature/saml2

julsemaan:
# Description

Adds SAML authentication in the captive portal (does not do it in the admin)
# Impacts

Captive portal authentication
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added SAML authentication to the captive portal


addons/saml-sp-metadata.xml
```diff
@@ -0,0 +1,27 @@
...
+    <!--- Not implemented and probably will never be... Leaving it there should it prove to be mandatory for a SAML IDP
+    <md:SingleLogoutService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="[% protocol %]://[% hostname %]/saml/assertion"/>
+    -->
+    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="[% protocol %]://[% hostname %]/saml/assertion" index="0"/>
+    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:1.0:profiles:browser-post" Location="[% protocol %]://[% hostname %]/saml/assertion" index="1"/>
+    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Artifact" Location="[% protocol %]://[% hostname %]/saml/assertion" index="2"/>
+    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:1.0:profiles:artifact-01" Location="[% protocol %]://[% hostname %]/saml/assertion" index="3"/>
+  </md:SPSSODescriptor>
+</md:EntityDescriptor>
+
```
jrouzierinverse: Use the xml filter for all output
[% sp_entity_id | xml %]
Or use Template::AutoFilter 
my $tt = Template::AutoFilter->new( AUTO_FILTER => 'xml' );

http://search.cpan.org/~mithaldu/Template-AutoFilter/lib/Template/AutoFilter.pm

jrouzierinverse: Why is the in addons?
It should be under the html/captive-portal/ path

lib/pf/Authentication/Source/SAMLSource.pm
```diff
@@ -0,0 +1,244 @@
...
+
+Generate the metadata for the Service Provider (this server)
+
+=cut
+
+sub generate_sp_metadata {
+    my ($self) = @_;
+    require pf::config;
+    my $cert = read_file($self->sp_cert_path);
+    $cert = join("\n", map { ($_ !~ /^-----BEGIN CERTIFICATE-----/ && $_ !~ /^-----END CERTIFICATE-----/) ? $_ : () } split(/\n/, $cert));
```
jrouzierinverse: wrap getting the cert into a function.

lib/pf/Authentication/Source/SAMLSource.pm
```diff
@@ -0,0 +1,244 @@
...
+    my $vars = {
+        hostname => $pf::config::fqdn,
+        sp_entity_id => $self->sp_entity_id,
+        sp_cert => $cert,
+        entity_id => $self->sp_entity_id,
+        protocol => isenabled($pf::config::Config{captive_portal}{secure_redirect}) ? "https" : "http",
+    };
```
jrouzierinverse, line 196: You could just pass the object to the template like we do in the provisioners.
$vars = {
    saml_source => $self,
    hostname =>  $pf::config::fqdn,
    protocol => isenabled($pf::config::Config{captive_portal}{secure_redirect}) ? "https" : "http",
};

lib/pf/Authentication/Source/SAMLSource.pm
```diff
@@ -0,0 +1,244 @@
...
+
+    our $TT_OPTIONS = {ABSOLUTE => 1};
+    our $template = Template->new($TT_OPTIONS);
```
jrouzierinverse: To avoid issues with badly formed xml do the following.
our $TT_OPTIONS = {ABSOLUTE => 1, AUTO_FILTER => 'xml' };
our $template = Template::AutoFilter->new($TT_OPTIONS );

addons/saml-sp-metadata.xml
julsemaan: This is rendered by the admin and used by the portal at the same time so it doesn't really belong in any of them.

I agree with you addons is not the best place though.

Another suggestion other than html/captive-portal ?

rouzier: I would suggest either in
conf/
or
lib/pf/saml/

If we never expected it to be modified by the user then we should treat like code then lib/pf/saml2/.
However if the user can make modification to it then put in conf/.

lib/pf/Authentication/Source/SAMLSource.pm
jrouzierinverse: Or always use the xml filter in the template.

julsemaan:
Adjustments made

html/pfappserver/root/authentication/source/type/SAML.tt
```diff
@@ -0,0 +1,18 @@
+
+[% form.field('sp_entity_id').render | none %]
+[% form.field('sp_key_path').render | none %]
+[% form.field('sp_cert_path').render | none %]
+[% form.field('idp_metadata_path').render | none %]
+[% form.field('idp_entity_id').render | none %]
+[% form.field('idp_cert_path').render | none %]
+[% form.field('idp_ca_cert_path').render | none %]
+[% form.field('username_attribute').render | none %]
+[% form.field('authorization_source_id').render | none %]
```
jrouzierinverse, line 10: All the fields could be moved to a block.

julsemaan, line 10: They could but that would be the only authentication source to work this way...

I think that it is better to follow the standard of the other auth sources so it is easier to follow and search when we have to dig through the code

jrouzierinverse:
Also packaging needs to be taken care of.
Let me know if you need help with creating a package in the obs.

julsemaan:
I am blessed : 
http://inverse.ca/downloads/SOGo/RHEL6/x86_64/RPMS/

julsemaan:
Just need to validate no additionnal dependencies are needed other than lasso and lasso-perl

extrafu:
Ahh those SOGo guys, they are the best!

lzammit:
https://www.sogou.com ?

julsemaan:
Packages in lasso_rpms on the build system ready to be put in the devel repo
Packages are already provided in the usual debian repos

julsemaan:
All adjustments made

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -636,6 +636,54 @@ Here is a brief description of the fields :
...
+NOTE: PacketFence does not support logoff on the SAML Identity Provider. You can still define the URL in the metadata but it will not be used.
+
+Passthroughs
+^^^^^^^^^^^^
+
+In order for your users to be able to access the Identity Provider login page, you will need to activate passthroughs and add the Indentity Provider domain to the allowed passthroughs.
+
+To do so, go in 'Configuration -> Trapping', then check 'Passthrough' and add the Identity Provider domain name to the 'Passhtroughs' list.
+
+Restart PacketFence to apply the new passthroughs configuration.
```
extrafu: Can't we just restart a sub-service?

julsemaan: Well if the passthroughs have just been enabled, it requires to restart iptables and pfdns

If they were already enabled only pfdns.

I put the full restart so there is no doubt the proper services will be restarted.

The section on passthroughs still states to restart PacketFence.

The above needs to be addressed anyway since we shouldn't even have to restart pfdns after we add a passthrough as it should use pfconfig to auto refresh its values.

Should I still change it to the conditional one ?

extrafu: I think it would be good - having to restart the entire thing is always a bit amateurish.

louismunro:
_Nihil obstat_
Merging.

PR #1173, opened 2016-01-28
Install udp_reflector as a service instead of a task

fdurand:
## Description

Update documentation to configure udp_reflector as a service in windows server instead of a Task
## Impacts

No
## NEWS file entries

Enhancements
++++++++++++
- Configure udp_reflector as a service on windows server
## Delete branch after merge

YES


julsemaan:
Much cleaner, me very like

image

PR #1175, opened 2016-01-29
Merge vsa attribute in radius filter

fdurand:
# Description

Merge vsa attribute in radius filters if merge_answer = yes
# Impacts

radius filters
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix merge vsa attributes in radius filters if merge_answer is set to yes.

### DIFF
--- a/lib/pf/access_filter/radius.pm
+++ b/lib/pf/access_filter/radius.pm
@@ -79,6 +79,17 @@ sub handleAnswerInRule {
         if (defined($rule->{'merge_answer'}) && !(isenabled($rule->{'merge_answer'}))) {
             return ($radius_reply,$status);
         } else {
+            foreach my $key (keys %$radius_reply_ref) {
+                if (exists($radius_reply->{$key})) {
+                    if (defined(reftype($radius_reply->{$key})) && reftype($radius_reply->{$key}) eq 'ARRAY') {
+                        my @attribute;
+                        push(@attribute,@{$radius_reply_ref->{$key}});
+                        push(@attribute,@{$radius_reply->{$key}});
+                        $radius_reply_ref->{$key} = \@attribute;
+                        delete $radius_reply->{$key};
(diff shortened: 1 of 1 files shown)


lib/pf/access_filter/radius.pm
```diff
@@ -79,6 +79,17 @@ sub handleAnswerInRule {
         if (defined($rule->{'merge_answer'}) && !(isenabled($rule->{'merge_answer'}))) {
             return ($radius_reply,$status);
         } else {
+            foreach my $key (keys %$radius_reply_ref) {
+                if (exists($radius_reply->{$key})) {
+                    if (defined(reftype($radius_reply->{$key})) && reftype($radius_reply->{$key}) eq 'ARRAY') {
+                        my @attribute;
+                        push(@attribute,@{$radius_reply_ref->{$key}});
```
jrouzierinverse, line 87: What happens if radius_reply_ref->{$key} is not an array ref?

lib/pf/access_filter/radius.pm
```diff
@@ -79,6 +79,17 @@ sub handleAnswerInRule {
         if (defined($rule->{'merge_answer'}) && !(isenabled($rule->{'merge_answer'}))) {
             return ($radius_reply,$status);
         } else {
+            foreach my $key (keys %$radius_reply_ref) {
+                if (exists($radius_reply->{$key})) {
+                    if (defined(reftype($radius_reply->{$key})) && reftype($radius_reply->{$key}) eq 'ARRAY') {
```
jrouzierinverse: only call reftype once

lib/pf/access_filter/radius.pm
fdurand, line 87: If it's not an array ref then it mean that the guy who wrote the switch module did something wrong.

PR #1176, opened 2016-02-02
Fix/accounting uniqueid

### LINKED ISSUE #1169: Accounting procedure is broken
If there is more than one entry in radacct that matches the WHERE clause for : 

```
WHERE acctsessionid = p_acctsessionid
    AND username = p_username
    AND nasipaddress = p_nasipaddress
```

louismunro:
# Description

Fixes possible errors in accounting update and accounting stop when accounting IDs are reused.
# Impacts

The select has been made more specific and should only match closed accounting sessions.
# Issue

Fixes #1169 
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fixes broken accounting when the controller reuses session ids (#1169)

### DIFF
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -563,7 +563,6 @@ CREATE TABLE radacct_log (
 ) ENGINE=InnoDB;
 
 -- Adding RADIUS Updates Stored Procedure
-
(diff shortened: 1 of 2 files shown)


louismunro:
Review me...
Tested according to the mailing list.

louismunro:
@fdurand says it looks good.
Merging.

PR #1181, opened 2016-02-05
stripped_username in LDAPSource without RADIUS attribute

dwlfrth:
# Description

Will strip the username / realm part if configured to do so and that no 'stripped_username' attribute is present in RADIUS request.
# Impacts

LDAP authentication source matching
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Ability to strip a username in LDAP source even if not present in RADIUS request


louismunro:
Can't see anything wrong.
Merging.

PR #1183, opened 2016-02-08
dhcplistener ipv6 fix for filter out non-ip packets

dwlfrth:
# Description

Fixes bug introduced by myself that makes dhcplistener non-longer processing ipv6 packets
# Impacts

pfdhcplistener
# Delete branch after merge

YES

### DIFF
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -197,7 +197,6 @@ sub process_pkt {
             # a tagged frame and return the 'inner' ether_type of the frame.
             #
             # For this reason we chec to see if $l2->{tpid} is set
-            #
             if ( defined $l2->{tpid} ) {
                 $logger->trace("Skipping VLAN packets since it is probably addressed to another interface (offload decapsulated tagged packet).");
                 return;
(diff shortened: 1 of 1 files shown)


sbin/pfdhcplistener
```diff
@@ -213,8 +213,7 @@ sub process_pkt {
             # For these reasons, we must peek directly at the data.
             # skip the first 12 bytes of the frame (src + dst eth addr)
             # grab the ether_type field and discard the rest:
-            my ($ether_type, undef) = unpack("x[12]nC*", $pkt);
-            if ($ether_type != NetPacket::Ethernet::ETH_TYPE_IP) {
+            if ( ($l2->{type} ne ETH_TYPE_IP) && ($l2->{type} ne ETH_TYPE_IPv6) ) {
```
jrouzierinverse, line 206: This check should be done earlier in the code.
Right after 

```
my $l2 = NetPacket::Ethernet->decode($pkt);
```

So we can exit as early as possible

jrouzierinverse, line 206: As soon as this is done I will merge.

dwlfrth:
Done

PR #1191, opened 2016-02-11
Return random int in a range in radius filter

fdurand:
## Description

Added a way to return a random int value in the radius filter. Can be usefull if you want to return a timeout value from a specific range.
## Impacts

No
## NEWS file entries

Enhancements Features
++++++++++++
- Added a new computed return in radius filter (random number in a range).
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/access_filter/radius.pm
+++ b/lib/pf/access_filter/radius.pm
@@ -143,7 +143,7 @@ evaluate all the variables
 
 sub evalParam {
     my ($answer, $args) = @_;
-    $answer = random($answer) if rangeValidator($answer);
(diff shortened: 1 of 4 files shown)


lib/pf/access_filter/radius.pm
```diff
@@ -129,11 +130,52 @@ evaluate all the variables
 
 sub evalParam {
     my ($answer, $args) = @_;
+    $answer = random($answer) if rangeValidator($answer);
```
julsemaan: _random ?

fdurand:
Re review please

PR #1192, opened 2016-02-11
Pruned useless code from radius.pm.

louismunro:
# Description

Spring cleanup.
Deleting useless code we never used for years.
# Impacts

Should be none.
This is code that always returns true and has not been used for years.
# Delete branch after merge

YES 
# NEWS file entries
## Enhancements
- Spring cleaning (not worth mentioning in NEWS file).

### DIFF
--- a/lib/pf/radius.pm
+++ b/lib/pf/radius.pm
@@ -139,15 +139,6 @@ sub authorize {
     # let's check if an old port sec entry needs to be removed in another switch
     $self->_handleStaticPortSecurityMovement($args);
 
-    # TODO maybe it's in there that we should do all the magic that happened in the FreeRADIUS module
-    # meaning: the return should be decided by _doWeActOnThisCall, not always $RADIUS::RLM_MODULE_NOOP
-    my $weActOnThisCall = $self->_doWeActOnThisCall($args);
-    if ($weActOnThisCall == 0) {
-        $logger->info("We decided not to act on this radius call. Stop handling request from $switch_ip.");
-        $RAD_REPLY_REF = [ $RADIUS::RLM_MODULE_NOOP, ('Reply-Message' => "Not acting on this request") ];
-        goto CLEANUP;
-    }
-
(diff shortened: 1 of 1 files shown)


louismunro:
Give me a good reason why I should not do this...

extrafu:
Because Zaymmitte said so?

jrouzierinverse:
Just to play devil's advocate one reason to keep it is to make easier to write custom behavior in pf::radius::custom.

louismunro:
I thought about that, but since we haven't do that in years, how likely are we to start doing it now?

fdurand:
Hi devil's advocate, we can do it now in the radius filter.

jrouzierinverse:
You are correct but not as is.
We would have to add the a new scope to do so.

fdurand:
[wiremacauth]
filter = connection_type
operator = is
value = WIRED_MAC_AUTH

[wireeap]
filter = connection_type
operator = is
value = Ethernet-EAP

[wirelessmacauth]
filter = connection_type
operator = is
value = Wireless-802.11-NoEAP

[wirelesseap]
filter = connection_type
operator = is
value = Wireless-802.11-EAP

[1:!wiremacauth|!wireeap|!wirelessmacauth|!wirelesseap]
merge_answer = no
scope = returnRadiusAccessAccept
status = RLM_MODULE_NOOP
answer1 = Reply-Message => Not acting on this request

jrouzierinverse:
That would work.
However I meant adding a scope right after _handleStaticPortSecurityMovement to the extra work.
Maybe $switch->skipRadiusAccessAccept();

louismunro:
Let's do this right.
Moving to v6.

julsemaan:
Not sure where the debate is on this, I would personally go with the removal of these as they are not used at all, and as @fdurand said, this can be accomplished via a RADIUS filter, should need be.

If we are majorly happy with this, I'd say to merge that

PR #1205, opened 2016-02-18
performance improvment in voip detection

fdurand:
## Description

Cache the maximum of snmp static result.
Select the way to detect VoIP phone with CDP and/or LLDP.
Enable or disable the VoIP detection with DHCP Fingerprint 
## Impacts

NO
## NEWS file entries

New Features
++++++++++++

Enhancements
++++++++++++
Performance improvment for  VoIP detection

Bug Fixes
+++++++++
## Delete branch after merge

YES

### DIFF
--- a/lib/pf/Switch/Cisco/Catalyst_2950.pm
+++ b/lib/pf/Switch/Cisco/Catalyst_2950.pm
@@ -1144,7 +1144,8 @@ sub ifIndexToLldpLocalPort {
     my $oid_lldpLocPortDesc = '1.0.8802.1.1.2.1.3.7.1.4'; # from LLDP-MIB
 
     $logger->trace("SNMP get_table for lldpLocPortDesc: $oid_lldpLocPortDesc");
-    my $result = $self->{_sessionRead}->get_table( -baseoid => $oid_lldpLocPortDesc);
+    my $cache = $self->cache;
+    my $result = $cache->compute([$self->{'_id'},$oid_lldpLocPortDesc], sub { $self->{_sessionRead}->get_table( -baseoid => $oid_lldpLocPortDesc) } );
     # here's what we are getting here. Looking for the last element of the OID: lldpRemLocalPortNum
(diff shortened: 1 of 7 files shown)


lib/pf/Switch/Cisco/Catalyst_2950.pm
```diff
@@ -1144,7 +1144,8 @@ sub ifIndexToLldpLocalPort {
     my $oid_lldpLocPortDesc = '1.0.8802.1.1.2.1.3.7.1.4'; # from LLDP-MIB
 
     $logger->trace("SNMP get_table for lldpLocPortDesc: $oid_lldpLocPortDesc");
-    my $result = $self->{_sessionRead}->get_table( -baseoid => $oid_lldpLocPortDesc);
+    my $cache = $self->cache;
+    my $result = $cache->compute([$self->{'_id'},$oid_lldpLocPortDesc], sub { $self->{_sessionRead}->get_table( -baseoid => $oid_lldpLocPortDesc) } );
```
julsemaan, line 1148: use cachedSNMPRequest ?

julsemaan:
I think the caching should be deactivated by default.

What if I plug a computer in a port and we fetch the LLDP table, then I plug a phone in.
The cached table won't contain the phone.

Same case if the phone gets replaced.

Also, there are cases where the RADIUS request for the device comes in before the one for the phone.
In that case too we would use the cached data which wouldn't contain the phone.

We should have the caching as a toggle and deactivated by default, since this seems like a case of the alligator in the swamp

lib/pf/Switch/Cisco/Catalyst_2950.pm
fdurand, line 1148: Yes this is static info from the switch it's like ifindex -> Fastethernet1/0/1

fdurand:
You are wrong, we just cache the static table but not the LLDP dynamic info that tell us if the voip flag is on.

julsemaan:
But I wanted to make an allusion to the alligators :(

Fine then, I'm wrong...

PR #1207, opened 2016-02-22
don't allow trace mode in Apache

julsemaan:
# Description

Don't allow the TRACE HTTP method to be used in the portal process
# Impacts

Portal HTTP requests
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- The HTTP TRACE method is not allowed anymore on the captive portal


jrouzierinverse:
Great idea.
We should do this for all our web services not just the portal

julsemaan:
disabled from all processes

PR #1208, opened 2016-02-23
Feature Catalyst Layered Templates

### LINKED ISSUE #824: layered "profile-templates"
To ease modification and management of "profile-templates" files, the following changes should be considered:
- /usr/local/pf/html/captive-portal/templates now becomes the 'original' files where no modifications are made. This is in fact packaged by the PacketFence packages / sources.
- /usr/local/pf/html/captive-portal/profile-templates contains the different portal-profiles templates directories, but instead of each holding a copy of the original files, they will only contains copy of a modified file (a folder can be empty, which will lead to use default files)
- Installation / upgrade / source installation process would create a 'default' portal profile under /usr/local/pf/html/captive-portal/profile-templates which will stays empty. Modified file that will now become default ones can be copied over from the origin folder and modified.
- Creating a new portal profile would create an empty folder which may allow customization of files if needed.
- Rendering the templates from a portal perspective will behave the following: portal-profile folder, default folder, origin folder. If a template is found in the first folder, it uses it, otherwise it continues in the tree.
- Managing the files from the web admin GUI will behave as for the rendering. Listing will list all the files depending on the tree and will allow modification and create a copy in the appropriate folder

jrouzierinverse:
## Description

Have the portal templates layered to ease
## Impacts

Template rendering in the portal and the display of the portal in the profile
## Issue

Fixes #824
## Delete branch after merge

NO

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
@@ -380,7 +380,7 @@ sub _render {
         return $self->i18n_format(@_);
     };
 
-    get_logger->info("previous : ". $self->previous_module_id // "undef" .", current : ". $self->current_module_id // "undef");
+    get_logger->info("previous : ". ($self->previous_module_id // "undef") . ", current : " . ($self->current_module_id // "undef"));
     $self->detect_first_action();
     $args->{ show_restart } //= $self->session->{action_made};
 
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
@@ -230,9 +230,9 @@ sub detect_first_action {
     my ($self) = @_;
     if(defined($self->previous_module_id) && defined($self->current_module_id)
         && $self->previous_module_id ne $self->current_module_id){
-        $self->session->{action_made} = $TRUE;     
+        $self->session->{action_made} = $TRUE;
     }
-    $self->session->{action_made} //= $FALSE; 
+    $self->session->{action_made} //= $FALSE;
     return $self->session->{action_made};
 }
 
@@ -243,7 +243,7 @@ Processing that needs to occur before we execute the application
 =cut
 
 sub preprocessing {
-    my ($self) = @_; 
+    my ($self) = @_;
     my $timer = pf::StatsD::Timer->new({sample_rate => 1});
     $self->process_user_agent();
     $self->process_destination_url();
@@ -294,7 +294,7 @@ sub current_ip {
 
 Destination URL handling
 
-Will compute it using the following logic : 
+Will compute it using the following logic :
 - Use the Destination URL specified as a param or the one stored in the session
 - Check if the profile requires a forced destination URL, in this case use the one of the profile
 - If there is no destination URL at this point, use the one from the portal
@@ -351,7 +351,7 @@ sub render {
     my $content = $args->{layout} ? $self->_render('layout.html', $layout_args) : $inner_content;
 
     $self->template_output($content);
-   
+
     $self->empty_flash();
 }
 
@@ -363,9 +363,7 @@ Render a template using Template Toolkit.
 
 sub _render {
     my ($self, $template, $args) = @_;
-    
-    # this won't be needed once #1208 is merged
-    $self->profile->{_template_paths} = ["/usr/local/pf/html/captive-portal/templates"];
+
     our $TT_OPTIONS = {
         AUTO_FILTER => 'html',
         RELATIVE => 1,
@@ -379,7 +377,7 @@ sub _render {
         return $self->i18n(@_);
     };
     $args->{ i18n_format } = sub {
-        return $self->i18n_format(@_);  
+        return $self->i18n_format(@_);
     };
 
     get_logger->info("previous : ".$self->previous_module_id.", current : ".$self->current_module_id);
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
@@ -380,7 +380,7 @@ sub _render {
         return $self->i18n_format(@_);
     };
 
-    get_logger->info("previous : ".$self->previous_module_id.", current : ".$self->current_module_id);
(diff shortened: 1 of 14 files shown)


julsemaan:
Any ETA on when the admin part will be done ?

rouzier:
I would say monday if not the end of today

jrouzierinverse:
The admin part is done ready for review

html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm
```diff
@@ -477,6 +493,74 @@ sub _sync_file {
...
+        size => format_bytes(-s $full_path),
+        editable => $self->isEditable($full_path),
+        previewable => $self->isPreviewable($full_path),
+        delete_or_revert_disabled => $self->isDeleteOrRevertDisabled($short_path),
+        delete_or_revert => $self->revertableOfDeletable($short_path),
+    );
+    return \%data;
+}
+
+sub revertableOfDeletable {
```
julsemaan: think you meant or not of

html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm
```diff
@@ -16,16 +16,19 @@ use strict;
 use warnings;
```
julsemaan, line 16: Add quick descriptions of the subroutines on what they do in this file

julsemaan:
The directory tree appears completely expanded (all directories opened)

The current behavior is to have them collapsed and expand them on click

julsemaan:
Trying to save a new custom templates gives the following error : 

```
Mar 17 10:35:26 httpd.admin(22416) ERROR: Caught exception in pfappserver::Controller::Config::Profile::Default->save "write_file '/usr/local/pf/html/captive-portal/profile-templates/default/billing/billing.html' - sysopen: No such file or directory at /usr/local/pf/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm line 235" (pfappserver::PacketFence::Controller::Root::end)
```

You should mkdir -p the destination directory before writing I guess

julsemaan:
Reverting the file still has the old behavior where it copies the pristine template over to the portal profile instead of deleting it.

If the pristine is copied over, we get the same packaging upgrade problem as we have right now.

julsemaan:
Previewing a file that has includes in it fails since the lookup path doesn't include the templates dir I think.

jrouzierinverse:
You cannot modify the pristine from the gui (unless you found a way how).
Modifying the pristine templates is now a no-no in packetfence.

We can handle the pristine templates in the following way.
- They are no longer considered configurations so any modifications will be deleted on each upgrade.
- Package them as readonly to enforce the fact they should not be modified.
- If all else fails we send @julsemaan to scream at the user

jrouzierinverse:
Please re-review

julsemaan:
It would be good to have something identifying the files that are not prestine in the admin and only allow to revert those files.

Right now there is no way to tell which files are custom

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
```diff
@@ -379,10 +377,10 @@ sub _render {
         return $self->i18n(@_);
     };
     $args->{ i18n_format } = sub {
-        return $self->i18n_format(@_);  
+        return $self->i18n_format(@_);
     };
 
-    get_logger->info("previous : ".$self->previous_module_id.", current : ".$self->current_module_id);
+    get_logger->info("previous : ". $self->previous_module_id // "undef" .", current : ". $self->current_module_id // "undef");
```
julsemaan: This change makes it that only `previous :` display when previous_module_id is undef and it doesn't display the rest of the line

julsemaan:
As discussed together this should handle the revert of a file in a cluster. Otherwise all the operations sync properly except that one

html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm
julsemaan, line 16: There are still subroutines that need naming cleanup 
_makeDefaultFilePath
copyDefaultFiles

julsemaan:
Can you also add the same directory creation logic to pf::api::distant_download_configfile so that when a template syncs, it will handle the directory creation properly.

jrouzierinverse:
Please re review

lib/pf/api.pm
```diff
@@ -674,6 +674,38 @@ sub distant_download_configfile : Public {
     return 1;
 }
 
+our $ALLOWED_PATHS_TO_BE_EMPTIED = qr/^\Q$captiveportal_profile_templates_path\E\/[^\.]/;
+
+our $ALLOWED_DELETED_PATHS = qr/^\Q$captiveportal_profile_templates_path\E\/[^\.]/;
+
+=head2 directory_empty
+
+Empty a directory of all it's files
```
julsemaan: *its

julsemaan:
You need to allow the default profile to be fully revertable : 
html/pfappserver/root/config/profile/files.tt +208

jrouzierinverse:
done

James Rouzier

html/pfappserver/root/config/profile/files.tt
```diff
@@ -205,9 +205,7 @@ td a.btn { margin-left: 5px; }
         </tbody>
       </table>
 
-      [% IF id != 'default' %]<div class="form-actions">
         <a class="btn btn-danger call-modal" data-modal="revertAll" href="[% c.uri_for(c.controller('Config::Profile').action_for('revert_all'),[id]) %]" ><i class="icon-refresh"></i> [% l('Revert All Files')  %]</a>
-      </div>[% END %]
```
julsemaan, line 210: A bit too aggressive with your deletion :)

Leave the div around the button

html/pfappserver/root/config/profile/files.tt
```diff
@@ -206,6 +206,7 @@ td a.btn { margin-left: 5px; }
       </table>
 
         <a class="btn btn-danger call-modal" data-modal="revertAll" href="[% c.uri_for(c.controller('Config::Profile').action_for('revert_all'),[id]) %]" ><i class="icon-refresh"></i> [% l('Revert All Files')  %]</a>
+    </div>
```
julsemaan, line 210: You deleted the opening div as well in your previous commit....

jrouzierinverse, line 210: Fixed

James Rouzier

PR #1209, opened 2016-02-23
pfdetect trigger src and dst ip

fdurand:
# Description

Allow PacketFence to trigger a violation on the src ip or dst ip address based on trapping range configuration parameter from snort/suricata/security_onion.
Use pfqueue to trigger violation instead of httpd.webservices
# Impacts

pfdetect
# Delete branch after merge

YES
## Enhancements
- Trigger violation on source ip or destination ip addess if they are in the trapping range networks

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -23,6 +23,7 @@ use pf::Authentication::constants;
 use pf::config();
 use pf::config::util();
 use pf::config::cached;
+use pf::config::trapping_range;
 use pf::ConfigStore::Interface();
 use pf::ConfigStore::Pf();
 use pf::iplog();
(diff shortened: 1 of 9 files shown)


lib/pf/api.pm
```diff
@@ -61,22 +61,53 @@ use pf::dhcp::processor();
...
-    if ($srcmac) {
-
-        # trigger a violation
-        pf::violation::violation_trigger( { 'mac' => $srcmac, 'tid' => $id, 'type' => $type } );
 
+    my ($type, $id) = each %{$postdata{'events'}};
+    $logger->info("violation: $id - IP $postdata{'srcip'}") if (defined($postdata{'srcip'}));
+    if (defined ($pf::config::Config{'trapping'}{'range'}) && $pf::config::Config{'trapping'}{'range'} ne '') {
+        foreach my $network (split(',',$pf::config::Config{'trapping'}{'range'})) {
+            my $net_addr = NetAddr::IP->new($network);
```
julsemaan: These objects should be built in a pfconfig resource

Otherwise we have the cost of creating the object of each network every time we parse an alert

dwlfrth:
Is it ready to be reviewed ?

lib/pfconfig/namespaces/resource/trapping_range.pm
```diff
@@ -0,0 +1,70 @@
+package pfconfig::namespaces::resource::trapping_range;
```
julsemaan, line 1: add this to the child resources of config::Pf

PR #1212, opened 2016-02-25
LDAP Object Caching

jrouzierinverse:
## Description

Do the filtering of the LDAP object attributes in packetfence instead of LDAP
## Impacts

LDAP Auth Source
## NEWS file entries

Enhancements
++++++++++++
- Cache the LDAP object when doing matching
## Delete branch after merge

NO


lib/pf/Authentication/Condition.pm
```diff
@@ -33,76 +33,58 @@ sub description {
...
+        return $_[0] > $_[1];
+    }
+);
+
 sub matches {
     my ($self, $attr, $v) = @_;
-
     if (defined $v) {
-
         my ($time, $time_v);
```
cgx: Remove those variables as they are not used anymore.

cgx:
Remove the sub `match_in_subclass` as it's no longer called.

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -226,169 +228,194 @@ sub match {
...
+=head2 match_with_object
 
 =cut
 
-sub cache {
-    return pf::CHI->new( namespace => 'ldap_auth');
+sub match_with_object {
+    my ($self, $params) = @_;
+    $self->preMatchProcessing();
+    my $object = $self->get_user_object($params);
```
jrouzierinverse, line 246: Exit early if there was no object found

julsemaan, line 246: A reminder on your reminder :)

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -226,169 +228,194 @@ sub match {
...
+    my $rule = $self->match_first_rule($object, $params);
+    $self->postMatchProcessing();
+    return defined $rule ? $rule->{actions} : undef;
 }
 
-=head2 is_match_cacheable
-
-Checks to see if the match can be cached
+=head2 get_user_object
```
julsemaan, line 253: Document briefly

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -226,169 +228,194 @@ sub match {
...
+        next if ( (defined($params->{'rule_class'})) && ($params->{'rule_class'} ne $rule->{'class'}) );
+        if ($self->match_rule($rule, $object, $params, \%common_attributes_lookup)) {
+            return $rule;
         }
     }
-    return $self->cache_match;
+    return undef;
 }
 
+sub match_rule {
```
julsemaan, line 306: missing perldoc

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -226,169 +228,194 @@ sub match {
...
+               || any { $self->match_ldap_condition($_, $object) } @$ldap_conditions
+               || $self->match_member_conditions($object, $member_conditions, $rule->match);
+    }
+    return (@$common_conditions == 0 || all { $self->match_condition($_, $params) } @$common_conditions)
+           && (@$ldap_conditions == 0 || all { $self->match_ldap_condition($_, $object) } @$ldap_conditions)
+           && (@$member_conditions == 0 || $self->match_member_conditions($object, $member_conditions, $rule->match));
+}
 
-=head2 match_in_subclass
+sub match_member_conditions {
```
julsemaan, line 336: missing perldoc

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -226,169 +228,194 @@ sub match {
...
+      return 0;
+    }
+    if ($match eq $Rules::ANY ) {
+        return any { $self->match_group_filter($connection, $object, $dn_search, $_) } @$member_conditions;
+    }
+    return all { $self->match_group_filter($connection, $object, $dn_search, $_) } @$member_conditions;
+}
 
-C<$params> are the parameters gathered at authentication (username, SSID, connection type, etc).
+sub match_group_filter {
```
julsemaan, line 357: missing perldoc

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -226,169 +228,194 @@ sub match {
...
+    my ($self, $connection, $entry, $dn_search, $condition) = @_;
+    my $value = escape_filter_value($condition->{'value'});
+    my $attribute = escape_filter_value($entry->get_value($condition->{'attribute'}));
+    # Search for any type of group definition:
+    # - groupOfNames       => member (dn)
+    # - groupOfUniqueNames => uniqueMember (dn)
+    # - posixGroup         => memberUid (uid)
+    my $filter = "(|(member=${dn_search})(uniqueMember=${dn_search})(memberUid=${attribute}))";
```
julsemaan, line 365: There is another special member of attribute in the AD class that needs to be done directly on LDAP.

Is that working with this code ?

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -25,6 +25,8 @@ use pf::config;
 use List::MoreUtils qw(uniq);
```
julsemaan, line 25: missing perldoc on a few subroutines in that module

julsemaan:
Since this is pretty big and instrusive in a core module, that needs a lot of testing love, I will suggest we do the intense testing prior to merging

lib/pf/Authentication/Source/LDAPSource.pm
jrouzierinverse, line 365: This code was based off the original code so as long original was work it should work

julsemaan, line 365: Ok, so for example, attribute memberOf:1.2.840.113556.1.4.1941: cannot be looked up directly in the object you fetch and needs to be looked up directly on the AD.

Will that be the case ?

louismunro:
I am testing it.
Will only merge if satisfactory.

lib/pf/Authentication/Source/LDAPSource.pm
jrouzierinverse, line 365: Since memberOf:1.2.840.113556.1.4.1941 is a magical LDAP attribute it will always fail

julsemaan, line 365: Exactly, and I'm affraid of other surprises like that.

As we discussed together, I expect the perf gain from this PR will be minimal since the vast majority of the auth rules are group based.

We'll see what Mr. Munro thinks about it but from our discussions, me and James believe that this is more trouble than anything else, especially when taking in consideration the small amount of cases where the object can actually be used for matching.

jrouzierinverse:
I will close this pull request and cherry pick some of the improvements into devel.

PR #1215, opened 2016-02-29
Feature/portal rework

### LINKED ISSUE #1225: Remove email as a mandatory field from the SMS auth source

julsemaan:
# Description

Redesign the captive portal so that every action is modular and can be chained together and placed in any order.
Also rewrites all the templates.
# Impacts

All the captive portal
# NEW Package(s) required

CentOS : perl-Graph (available from OF, already on pfbuilder)
Debian : libgraph-perl (available from stock packages)
# Fixes

fixes #1225 
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Fully redesigned modular captive portal
# Upgrade notes
- Deprecates guests_self_registration.preregistration - Feature has not been ported
- Deprecates guests_self_registration.guest_pid - Now part of the portal module


bin/pfcmd.pl
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -T
+#!/usr/bin/perl
```
julsemaan, line 1: This needs to be discussed, but it is causing issues with Module::Plugable

julsemaan:
This is ready for testing, and for a first shot at reviewing
First more than the latest

bin/pfcmd.pl
jrouzierinverse, line 1: Which command is causing the problem?

julsemaan, line 1: pfcmd checkup when using the captiveportal::DynamicRouting::Factory

html/captive-portal/lib/captiveportal/DynamicRouting/Factory.pm
```diff
@@ -0,0 +1,242 @@
+package captiveportal::DynamicRouting::Factory;
+
+=head1 NAME
+
+pf::factory::firewallsso
```
jrouzierinverse: Clean up copy and paste

jrouzierinverse:
portal.psgi should be put in the t/ directory

julsemaan:
It should even be removed

jrouzierinverse:
Can you update addons/packages/packetfence.spec with any new packages used.

julsemaan:
Packages added to list.

Please re-review

bin/pfcmd.pl
jrouzierinverse, line 1: I was trying to figure out what is causing the issue.
It is not an issue Module::Pluggable since I am able to replicate this without it.

However it may not be worth trying to fix even though I would prefer to at this point and time.
Since it seems to be a very random problem.
We have a straightforward solution at the moment.

julsemaan, line 1: Indeed, and IMO running pfcmd with that flag doesn't represent the environment that the app runs in since (correct me if I'm wrong) but the other processes run without the tainting check.

As you said, we will revisit this, we can also brainstorm on the validity of these checks

jrouzierinverse, line 1: The reason we use the taint check for pfcmd is because it is running as root.
This is done to ensure we clean up our data before using it.

julsemaan:
Any other comments on this PR ?

debian/control
```diff
@@ -125,6 +125,8 @@ Depends: ${misc:Depends}, vlan, make,
  python-twisted-bin (<=13.0), python-twisted-core (<=13.0), graphite-web,
 # EAP TLS
  libcrypt-smime-perl
```
fdurand: packaging breaker !

addons/packages/packetfence.spec
```diff
@@ -270,6 +270,8 @@ Requires: iproute, vconfig
 Requires: wmi, perl(Net::WMIClient)
 # SAML
 Requires: lasso-perl 
+# Captive Portal Dynamic Routing
+Requires: perl(Graph)
```
fdurand, line 276: You probably forgot to add portal_modules.conf.defaults, portal_modules.conf.example and portal_modules.conf as configuration files. (rhel and debian)

julsemaan, line 276: You probably know me very well !

html/captive-portal/lib/captiveportal/DynamicRouting/Application.pm
```diff
@@ -0,0 +1,475 @@
...
+    # caching useragents, if it's the same don't bother triggering violations
+    my $cached_useragent = $self->user_agent_cache->get($mac);
+
+    # Cache hit
+    return
+      if ( defined($cached_useragent) && $user_agent eq $cached_useragent );
+
+    # Caching and updating node's info
+    $logger->debug("adding user-agent to cache");
+    $self->user_agent_cache->set( $mac, $user_agent, "5 minutes" );
```
fdurand: '5 minutes' in a constant

html/captive-portal/lib/captiveportal/DynamicRouting/Application.pm
```diff
@@ -0,0 +1,475 @@
...
+
+Render a template using Template Toolkit.
+
+=cut
+
+sub _render {
+    my ($self, $template, $args) = @_;
+    
+    # this won't be needed once #1208 is merged
+    $self->profile->{_template_paths} = ["/usr/local/pf/html/captive-portal/templates"];
```
fdurand: $install_dir/html/captive-portal/templates

jrouzierinverse:
Move html/captive-portal/lib/captiveportal/DynamicRouting to html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting

html/captive-portal/lib/captiveportal/DynamicRouting/Application.pm
jrouzierinverse: Do you want to merge #1208 first?

html/captive-portal/lib/captiveportal/DynamicRouting/Factory.pm
jrouzierinverse: Clean this

addons/packages/packetfence.spec
julsemaan, line 276: fixed

julsemaan:
Extension point files have been created in html/captive-portal/lib/PacketFence and modules/roles that shouldn't be extended moved to the appropriate place.

Good for another review

jrouzierinverse:
Are you going to rework the layered the template into this pull request?
Or are you going to merge it into devel then rebase this branch against devel?

julsemaan:
I would say to merge these and add the layered templates above it since they are two distinct features IMO

julsemaan:
Ready to merge ?

debian/control
fdurand: It miss a comma after libcrypt-smime-perl

addons/packages/packetfence.spec
```diff
@@ -813,6 +815,8 @@ fi
 %config(noreplace)      /usr/local/pf/conf/authentication.conf
 %config(noreplace)      /usr/local/pf/conf/chi.conf
 %config                 /usr/local/pf/conf/chi.conf.defaults
+%config(noreplace)      /usr/local/pf/conf/portal_modules.conf
```
fdurand, line 819: and portal_modules.conf.example

cgx:
You probably need to add some instructions to `UPGRADE.asciidoc`, don't you?

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
```diff
@@ -0,0 +1,478 @@
...
+
+Render a template using Template Toolkit.
+
+=cut
+
+sub _render {
+    my ($self, $template, $args) = @_;
+    
+    # this won't be needed once #1208 is merged
+    $self->profile->{_template_paths} = ["/usr/local/pf/html/captive-portal/templates"];
```
fdurand, line 368: $install_dir/html/captive-portal/templates

jrouzierinverse:
Maybe we should have a Test-a-Thon to workout all the bugs before merging this huge change.

louismunro:
Let's not start down that road.

jrouzierinverse:
This a huge change set that is very difficult to review on github.
I think we should at least see if it is working before merging it into devel.

addons/packages/packetfence.spec
julsemaan, line 819: From what I saw there is a rule for *.example above that.

I think that's why chi.conf.example isn't there but works.

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
julsemaan, line 368: This line will be replaced and is only temporary

julsemaan:
@cgx 
We do need upgrade instructions + a good section in the admin guide but I will do a separate PR for these two as this one is already pretty massive.

@jrouzierinverse 
Yes, agreed, but I would suggest we nominate testers quickly and have this done so we can start working on what is around this instead of having the PR just lying there

I will create a package based on this branch and post the link to it here so that it is easier to test

julsemaan:
Anybody taking a last shot at this ?

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
```diff
@@ -328,6 +328,7 @@ sub render {
         content => $inner_content,
         client_mac => $self->current_mac,
         client_ip => $self->root_module->current_ip,
+        current_module => $captiveportal::PacketFence::DynamicRouting::Factory::INSTANTIATED_MODULES{$self->current_module_id},
```
jrouzierinverse: Wrap getting the current_module in a function.

julsemaan:
Done, review

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module.pm
```diff
@@ -0,0 +1,326 @@
...
+Return this class name in a prettier way
+
+=cut
+
+sub pretty_class_name {
+    my ($self) = @_;
+    my $name = ref($self);
+    $name =~ s/::/-/g;
+    $name =~ s/^captiveportal-DynamicRouting-//g;
+    return $name;
```
cgx: Please return a lowercase string.

PR #1219, opened 2016-03-01
New source for EAP TLS certificate

jrouzierinverse:
## Description

New source for EAP TLS
## Impacts

The radius work flow
## NEWS file entries

New Features
++++++++++++
- New source for EAP TLs
## Delete branch after merge

NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/EAPTLS.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/EAPTLS.pm
@@ -2,7 +2,7 @@ package pfappserver::Form::Config::Authentication::Source::EAPTLS;
 
 =head1 NAME
 
-pfappserver::Form::Config::Authentication::Source::EAPTLS add documentation
+pfappserver::Form::Config::Authentication::Source::EAPTLS
 
 =cut
 
--- a/html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/EAPTLS.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/EAPTLS.pm
@@ -1,14 +1,14 @@
-package pfappserver::Form::Config::Authentication::Source::TLS;
+package pfappserver::Form::Config::Authentication::Source::EAPTLS;
 
 =head1 NAME
 
-pfappserver::Form::Config::Authentication::Source::TLS add documentation
+pfappserver::Form::Config::Authentication::Source::EAPTLS add documentation
 
 =cut
 
 =head1 DESCRIPTION
 
-pfappserver::Form::Config::Authentication::Source::TLS
+pfappserver::Form::Config::Authentication::Source::EAPTLS
 
 =cut
 
@@ -18,7 +18,7 @@ use HTML::FormHandler::Moose;
 extends 'pfappserver::Form::Config::Authentication::Source';
 with 'pfappserver::Base::Form::Role::Help';
 
-use pf::Authentication::Source::TLSSource;
+use pf::Authentication::Source::EAPTLSSource;
 
 =head1 AUTHOR
 
--- a/lib/pf/Authentication/Source/EAPTLSSource.pm
+++ b/lib/pf/Authentication/Source/EAPTLSSource.pm
@@ -22,7 +22,7 @@ use List::Util qw(first);
 extends 'pf::Authentication::Source';
 
 has '+class' => (default => 'external');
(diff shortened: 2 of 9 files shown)


html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/EAPTLS.pm
```diff
@@ -0,0 +1,50 @@
+package pfappserver::Form::Config::Authentication::Source::EAPTLS;
+
+=head1 NAME
+
+pfappserver::Form::Config::Authentication::Source::EAPTLS add documentation
```
julsemaan: You should fix your templates ;)

lib/pf/Authentication/Source/EAPTLSSource.pm
```diff
@@ -0,0 +1,104 @@
+package pf::Authentication::Source::EAPTLSSource;
+
+=head1 NAME
+
+pf::Authentication::Source::EAPTLSSource add documentation
```
julsemaan: zi tamplaite, zi is no goude

lib/pf/Authentication/Source/EAPTLSSource.pm
```diff
@@ -0,0 +1,104 @@
...
+use List::Util qw(first);
+
+extends 'pf::Authentication::Source';
+
+has '+class' => (default => 'external');
+has '+type' => (default => 'TLS');
+
+=head2 available_attributes
+
+Allow to make a condition on the user's email address.
```
julsemaan: I'm no expert but that doesn't look right

lib/pf/Authentication/Source/EAPTLSSource.pm
```diff
@@ -0,0 +1,104 @@
...
+use strict;
+use warnings;
+use Moose;
+use pf::constants;
+use pf::Authentication::constants;
+use List::Util qw(first);
+
+extends 'pf::Authentication::Source';
+
+has '+class' => (default => 'external');
```
julsemaan: That's more and internal source I think.
External is for systems that are outside of the corporate environment

lib/pf/Authentication/Source/EAPTLSSource.pm
```diff
@@ -0,0 +1,104 @@
...
+
+Allow to make a condition on the user's email address.
+
+=cut
+
+sub available_attributes {
+    my ($self) = @_;
+    my $super_attributes = $self->SUPER::available_attributes;
+    my @own_attributes = map { {value => $_, type => $Conditions::SUBSTRING}} qw(
+      TLS-Client-Cert-Serial
```
whitx, line 37: I don't think we will ever use this one, I might be wrong tho

louismunro, line 37: Yes, you might be wrong.
I can easily see a case where someone might want to reject a specific serial, or any serial number below X.

lib/pf/Authentication/Source/EAPTLSSource.pm
dwlfrth: I'm one and that doesn't look right.

lib/pf/Authentication/Source/EAPTLSSource.pm
whitx, line 37: Would it make more sense to revoke the cert in question and refuse it? Or refuse certs that were delivered before X date? From what I found, depending on the configuration of the CA, it can choose the serial randomly.

louismunro, line 37: Perhaps, but that is not for you to decide.
The user gets to make that call.

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/EAPTLS.pm
jrouzierinverse: Bad Copy and paste

jrouzierinverse:
Fixed please re-review

lib/pf/role.pm
```diff
@@ -449,6 +449,7 @@ sub getRegisteredRole {
                 SSID => $args->{'ssid'},
                 stripped_user_name => $stripped_user,
                 rule_class => 'authentication',
+                radius_request => $args->{radius_request},
```
julsemaan, line 452: This is missing from getNodeInfoForAutoreg

jrouzierinverse, line 452: Done

PR #1220, opened 2016-03-02
Feature/stop depends on services

dwlfrth:
# Description

Add the ability to instruct a service manager to stop other services before stopping it's own one
# Impacts

Services manager bro

### DIFF
--- a/bin/pfcmd-old.pl
+++ b/bin/pfcmd-old.pl
@@ -1261,7 +1261,7 @@ sub getManagers {
             my @managers;
             if ($includeDependsOn) {
                 push @managers, grep {defined $_}
-                  map {pf::services::get_service_manager($_)} @{$m->dependsOnServices};
+                  map {pf::services::get_service_manager($_)} @{$m->startDependsOnServices};
             }
             if($m->isa("pf::services::manager::submanager")) {
                 push @managers,$m->managers;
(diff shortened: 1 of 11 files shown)


dwlfrth:
Review me

dwlfrth:
Please review as soon as possible since it is required by something else :dancer:

bin/pfcmd-old.pl
```diff
@@ -75,7 +75,8 @@ =head1 SYNOPSIS
 use constant {
     INSTALL_DIR        => '/usr/local/pf',
     JUST_MANAGED       => 1,
-    INCLUDE_DEPENDS_ON => 2,
+    INCLUDE_START_DEPENDS_ON => 2,
+    INCLUDE_STOP_DEPENDS_ON => 3,
```
jrouzierinverse: This should be 4 not 3.

jrouzierinverse: These are flags not enumerations.
It should be changes to use bit notation.

0b0000001
0b0000010
0b0000100

To make it more clear

lib/pf/cmd/pf/service.pm
```diff
@@ -74,7 +74,8 @@ use pf::services;
 use List::MoreUtils qw(part any true all);
 use constant {
     JUST_MANAGED       => 1,
-    INCLUDE_DEPENDS_ON => 2,
+    INCLUDE_START_DEPENDS_ON => 2,
+    INCLUDE_STOP_DEPENDS_ON => 3,
```
jrouzierinverse: Same as above

dwlfrth:
Re-review mew

PR #1227, opened 2016-03-04
Feature/show role instead of vlan

fdurand:
# Description

In node location let see the role instead of the vlan id (not revelant when we use acl or role by role)
# Impacts

View of the location history of the node show the role instead of the vlan id
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Location history of a node show the role instead of the vlan id


julsemaan:
No need for jasing. GTM

PR #1229, opened 2016-03-04
Deregister a node that coming from a secure connection (managed by vlan filters)

fdurand:
# Description

Deregister with the vlan filters a device if the previous connection was secure and the new one is unsecure
# Impacts

Default deregister code in role.pm is remove
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Deregister a node from a secure connection to a unsecure one is now managed by the vlan filters.


julsemaan:
We should have a vlan_filters.conf.defaults that contains these, so this way, it will be upgrade proof and easier to determine which are PacketFence filters and which are the customers one

conf/vlan_filters.conf.example
```diff
@@ -253,3 +253,42 @@
...
+operator = is
+value = Wireless-802.11-EAP
+
+[nodereg]
+filter = node_info.status
+operator = is
+value = reg
+
+[nodeautoreg]
+filter = node_info.status
```
julsemaan: not sure this is really what you want

conf/vlan_filters.conf.example
```diff
@@ -253,3 +253,42 @@
 # For example, if you want to allow access from Monday to Friday, 9am to 11am and from 2pm to 4pm:
 # wd {Mon Tue Wed Thu Fri} hr {9am-11am}, wd {Mon Tue Wed Thu Fri} hr {2pm-4am}
 
+
+
+# This rules are predefined by packetfence and to disconnect device that comming from a secure connection to an open one (remove these rules mean that you know what you are doing).
+
+[wiremacauth]
```
julsemaan: some underscores or dashes would be welcome in all of those

lib/pfconfig/namespaces/config/VlanFiltersDefault.pm
```diff
@@ -0,0 +1,73 @@
+package pfconfig::namespaces::config::VlanFiltersDefault;
```
julsemaan, line 1: Just adjust the documentation naming in this file and its good to merge

PR #1232, opened 2016-03-07
Feature/campers parking

julsemaan:
# Description

Fear not, this is not a place where you bring your mobile home and where low hygiene standards are mixed with alcohol and drugs abuse.
This here, allows to make devices that 'camp' the registration VLAN use a different portal and have a longer lease length in order to reduce the load on the server.
# Impacts

Registration network
# Code / PR Dependencies

(OPTIONAL. REMOVE IF NOT NEEDED)
List the PRs or other factors on which this PR depends
# NEW Package(s) required

NONE ! YAY !
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added ability to 'park' a device that stays in the registration network for too long without registering

### DIFF
--- a/conf/dhcpd.conf
+++ b/conf/dhcpd.conf
@@ -10,3 +10,9 @@ log-facility local6;
 %%active%%
 
 %%networks%%
+
+group parking {
+  default-lease-time 3600;
+  max-lease-time 3600;
+}
+
--- a/conf/dhcpd.conf
+++ b/conf/dhcpd.conf
@@ -12,7 +12,7 @@ log-facility local6;
 %%networks%%
 
 group parking {
-  default-lease-time 3600;
-  max-lease-time 3600;
+  default-lease-time %%parking_lease_length%%;
+  max-lease-time %%parking_lease_length%%;
 }
 
--- a/conf/dhcpd.conf
+++ b/conf/dhcpd.conf
@@ -11,6 +11,7 @@ log-facility local6;
 
 %%networks%%
 
+# httpd.parking related
 group parking {
   default-lease-time %%parking_lease_length%%;
   max-lease-time %%parking_lease_length%%;
--- a/conf/dhcpd.conf
+++ b/conf/dhcpd.conf
@@ -11,9 +11,5 @@ log-facility local6;
 
 %%networks%%
 
-# httpd.parking related
-group parking {
-  default-lease-time %%parking_lease_length%%;
-  max-lease-time %%parking_lease_length%%;
-}
+%%parking%%
 
--- a/conf/httpd.conf.d/httpd.parking
+++ b/conf/httpd.conf.d/httpd.parking
@@ -481,7 +481,7 @@ HostnameLookups Off
 # logged here.  If you *do* define an error logfile for a <VirtualHost>
 # container, that host's errors will be logged there and not here.
 #
-ErrorLog logs/error_log
+ErrorLog /usr/local/pf/logs/httpd.parking.error
 
 #
 # LogLevel: Control the number of messages logged to the error_log.
@@ -523,7 +523,7 @@ LogFormat "%{User-agent}i" agent
 # For a single logfile with access, agent, and referer information
 # (Combined Logfile Format), use the following directive:
 #
-CustomLog logs/access_log combined
+CustomLog /usr/local/pf/logs/httpd.parking.access combined
 
 #
 # Optionally add a line containing the server version and virtual host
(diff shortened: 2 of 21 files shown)


whitx:
Best feature name

louismunro:
Now contains more sharks.

dwlfrth:
sharknado_1

julsemaan:
My sleeping bag when I go camping : 

image

lzammit:
Cute.

https://s-media-cache-ak0.pinimg.com/736x/0c/23/35/0c2335773550efcb7cd7e2c3c631f432.jpg

html/parking/index.html
```diff
@@ -0,0 +1,3 @@
+<h1>Parking</h1>
+
+<p>You're in parking bro...</p>
```
extrafu: You are in parking bro ... keep swimming forward and breathe so you don't get eaten by sharks.

julsemaan:
ready to review

conf/httpd.conf.d/httpd.parking
```diff
@@ -0,0 +1,1003 @@
...
+LoadModule proxy_balancer_module modules/mod_proxy_balancer.so
+LoadModule proxy_ftp_module modules/mod_proxy_ftp.so
+LoadModule proxy_http_module modules/mod_proxy_http.so
+LoadModule proxy_ajp_module modules/mod_proxy_ajp.so
+LoadModule proxy_connect_module modules/mod_proxy_connect.so
+LoadModule cache_module modules/mod_cache.so
+LoadModule suexec_module modules/mod_suexec.so
+LoadModule disk_cache_module modules/mod_disk_cache.so
+LoadModule cgi_module modules/mod_cgi.so
+LoadModule version_module modules/mod_version.so
```
fdurand, line 201: Do you need all these modules ?
Also does it work on debian ?

conf/iptables.conf.example
```diff
@@ -78,6 +78,8 @@
 -A input-internal-vlan-if --protocol tcp --match tcp --dport 80  --jump ACCEPT
 -A input-internal-vlan-if --protocol tcp --match tcp --dport 443 --jump ACCEPT
 -A input-internal-vlan-if --protocol tcp --match tcp --dport 647 --jump ACCEPT
+# HTTP (parking portal)
+-A input-internal-vlan-if --protocol tcp --match tcp --dport 5252 --jump ACCEPT
```
fdurand, line 82: can 5252 be a constant ?

lib/pf/parking.pm
```diff
@@ -0,0 +1,124 @@
...
+
+=cut
+
+sub remove_parking_actions {
+    my ($mac, $ip) = @_;
+    get_logger->info("Removing parking actions for $mac - $ip");
+    my $omapi = pf::OMAPI->get_client();
+    $omapi->delete_host($mac);
+
+    pf_run("LANG=C sudo ipset del $PARKING_IPSET_NAME $ip -exist 2>&1");
```
fdurand: don't we remove LANG=C ?

conf/iptables.conf.example
julsemaan, line 82: It can be a constant but it won't be usable in this file unless we inject it in the iptables manager.

That's why I put it this way.

Do you think it would be better to inject it in iptables.pm when we generate ?

conf/httpd.conf.d/httpd.parking
julsemaan, line 201: This is the default Apache config so we likely don't need all of them.

I'll try to remove some of them.

As for debian, as usual, no idea...
I assume there is something specific to do if you say that.

conf/iptables.conf.example
dwlfrth, line 82: I agree with mister @fdurand. We should try to use constants / config parameters for all ports that are non-standards.
It is not THAT big of a deal on the side of iptables generation :)

lib/pf/parking.pm
dwlfrth: Mister @fdurand is effectively right

conf/dhcpd.conf
```diff
@@ -10,3 +10,9 @@ log-facility local6;
 %%active%%
 
 %%networks%%
+
+group parking {
```
dwlfrth: Shouldn't we add a little small comment to say something like "httpd.portal parking related" or whatnot just so that we remember in 30 years that is what not a "Stationnement Montreal" feature related

lib/pf/OMAPI.pm
```diff
@@ -290,6 +303,54 @@ sub lookup {
...
+=head2 create_host
+
+Create a host entry using the OMAPI
+
+=cut
+
+sub create_host {
+    my ($self, $mac, $assignments) = @_;
+    $assignments //= {};
+    use Data::Dumper;
```
dwlfrth: Hello there

julsemaan:
@derekmaster you take care of the adjustments ?

Issue #1230, opened 2016-03-07
Excessive TCP socket usage for redis


julsemaan:
About 3500 ports being used for communication with redis.

And thats not the highest it can reach, thats just a stat I took at a random point during the day on the server : 

```
[root@packetfenceb ~]# netstat -ntu | awk '{print $5}' | grep 127.0.0.1:6379 | wc -l
2535
[root@packetfenceb ~]# netstat -ntu | awk '{print $5}' | grep 127.0.0.1:6380 | wc -l
947
```

Almost all of them are in TIME_WAIT

```
[root@packetfenceb ~]# netstat -ntu | awk '{print $5}' | grep 127.0.0.1:6380 | wc -l && netstat -ntu | grep TIME_WAIT | awk '{print $5}' | grep 127.0.0.1:6380 | wc -l
875
872
```

We need to figure out the high usage of sockets and see if we can lower it by establishing less connections and by making sure we close them when we are done with them as in the environment I analyzed, this accounts for half of the TCP connections on localhost. And this again, in non-peak time.

```
[root@packetfenceb ~]# netstat -ntu | grep TIME_WAIT | awk '{print $5}' | wc -l && netstat -ntu | grep TIME_WAIT | awk '{print $5}' | egrep '127.0.0.1:(6379|6380)' | wc -l
5131
2376
```

jrouzierinverse:
I think I have a good idea why that is happening.

jrouzierinverse:
The redis connection is not shared across each chi namespace.
I work on shared the connection

julsemaan:
Did you have a chance to look at the queuing service as well ?

It was using ~900 connections which were almost all in TIME_WAIT as well

jrouzierinverse:
Fixed by #1233

PR #1233, opened 2016-03-07
Share the redis connection across all the namespaces that use redis

jrouzierinverse:
## Description

Share the redis connection across all the namespaces that use redis
## Impacts

CHI
## Issue

Issue #1230

Enhancements
++++++++++++
Share redis connection between all chi namespaces
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/CHI.pm
+++ b/lib/pf/CHI.pm
@@ -1,11 +1,3 @@
-package pf::Redis;
-
-sub clone {
-    return $_[0];
-}
-
-sub CLONE_SKIP { 1 }
-
 package pf::CHI;
 
 =head1 NAME
@@ -24,8 +16,7 @@ use strict;
 use warnings;
 use base qw(CHI);
 use Module::Pluggable search_path => ['CHI::Driver', 'pf::Role::CHI'], sub_name => '_preload_chi_drivers', require => 1, except => qr/(^CHI::Driver::.*Test|FastMmap)/;
-use CHI::Driver;
-use Data::Clone();
+use Clone();
 use pf::file_paths;
 use pf::IniFiles;
 use Hash::Merge;
@@ -36,10 +27,10 @@ use Scalar::Util qw(tainted reftype);
 use pf::log;
 use Log::Any::Adapter;
 use pf::Redis;
-
 Log::Any::Adapter->set('Log4perl');
 
 my @PRELOADED_CHI_DRIVERS;
+
 
 Hash::Merge::specify_behavior(
     {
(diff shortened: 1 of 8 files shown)


lib/pf/CHI.pm
```diff
@@ -1,3 +1,11 @@
+package pf::Redis;
```
julsemaan: necessary ? 
If yes, document it briefly as to what it is used for

rouzier: pf::Redis is the layer for caching redis clients.

julsemaan: That I saw :)

I mean since you have pf::Redis in lib/pf/Redis.pm and you redefine it in another module which is not in our standards

rouzier: Oh sorry that was a search and replace mistake I will move that to pf::Redis

jrouzierinverse:
please review

lib/pf/dhcp/processor.pm
```diff
@@ -117,12 +120,28 @@ sub _get_redis_client {
...
 }
 
+
+=head2 _get_redis_config
+
+Get the redis config from pf::CHI
+
+=cut
+
+sub _get_redis_config {
```
julsemaan: that should be centralized in pf::Redis or another place for usage by other modules

jrouzierinverse: Between pf::Redis or pf::CHI.

julsemaan: Not sure what you mean ?
Do you mean where I would see it more ?
I would say in pf::CHI since thats where we get the config from but any generic place is good

jrouzierinverse:
Please review

lib/pf/Redis.pm
```diff
@@ -0,0 +1,76 @@
...
+
+use strict;
+use warnings;
+
+use Redis::Fast;
+use CHI;
+
+our $CHI_CACHE = CHI->new(driver => 'RawMemory', datastore => {});
+
+sub new {
```
julsemaan, line 29: Just add the doc headers to the method below and we're good to merge

julsemaan, line 29: bump !

jrouzierinverse, line 29: Re-review

julsemaan:
Will briefly test and merge afterwards

julsemaan:
Functionnality hasn't changed.

Wasn't able to test the full effect of it but it looks good.

PR #1236, opened 2016-03-09
Feature/expire persons

julsemaan:
# Description

Add expiration for the users.
# Impacts

The person table and password table
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added optional expiration of the users

### DIFF
--- a/lib/pf/constants.pm
+++ b/lib/pf/constants.pm
@@ -17,6 +17,7 @@ use Readonly;
 use base qw(Exporter);
 our @EXPORT = qw(
     $FALSE $TRUE $YES $NO $default_pid $admin_pid
+    %BUILT_IN_USERS
 );
 
 # some global constants
@@ -26,6 +27,11 @@ Readonly::Scalar our $YES => 'yes';
 Readonly::Scalar our $NO => 'no';
 Readonly::Scalar our $default_pid => 'default';
 Readonly::Scalar our $admin_pid => 'admin';
+
+Readonly::Hash our %BUILT_IN_USERS => (
+    $default_pid => 1, 
+    $admin_pid => 1,
+);
 
 =head1 AUTHOR
 
--- a/lib/pf/constants.pm
+++ b/lib/pf/constants.pm
@@ -17,7 +17,6 @@ use Readonly;
 use base qw(Exporter);
 our @EXPORT = qw(
     $FALSE $TRUE $YES $NO $default_pid $admin_pid
-    %BUILT_IN_USERS
 );
 
 # some global constants
@@ -28,7 +27,7 @@ Readonly::Scalar our $NO => 'no';
 Readonly::Scalar our $default_pid => 'default';
 Readonly::Scalar our $admin_pid => 'admin';
 
-Readonly::Hash our %BUILT_IN_USERS => (
+Readonly::Hash our %BUILTIN_USERS => (
     $default_pid => 1, 
     $admin_pid => 1,
 );
--- a/lib/pf/person.pm
+++ b/lib/pf/person.pm
@@ -416,6 +416,8 @@ sub person_cleanup {
                 get_logger->debug("Not deleting $pid because the local account is still valid.");
                 next;
             }
+            # We delete the password too
+            pf::password::_delete($pid);
         }
         # We're all good for deletion
         person_delete($pid);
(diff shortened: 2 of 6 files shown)


julsemaan:
Ready for review

lib/pf/person.pm
```diff
@@ -375,6 +388,42 @@ sub person_violations {
...
+                next;
+            }
+            # We delete the password too
+            pf::password::_delete($pid);
+        }
+        # We're all good for deletion
+        person_delete($pid);
+    }
+}
+
```
jrouzierinverse, line 431: This should be done in batches to avoid table locking.
Similar to your pattern you created in addons/database-cleaner.pl.

lib/pf/person.pm
```diff
@@ -164,6 +169,14 @@ sub person_db_prepare {
...
             WHERE pid = ? ]);
 
+    $person_statements->{'persons_without_nodes_sql'} = get_db_handle()->prepare(
+        qq[ 
+            SELECT person.pid as pid from person 
+            LEFT JOIN node on node.pid=person.pid  
+            GROUP BY pid 
+            HAVING count(node.mac)=0;
+            ]);
+
```
jrouzierinverse, line 180: Should join the password table to avoid the additional lookup later in person_cleanup.

julsemaan, line 180: I thought of it but the query about the persons without node could be useful by itself, that's why I made it into a separate sub routine and into a query that is independant of the password table

lib/pf/person.pm
julsemaan, line 431: I followed the same pattern as node_cleanup

I would suggest bringing the database cleaner into a generalized logic in lib/ at a later point and use it for all cleanups.

rouzier, line 431: The point is there is no LIMIT in the query

julsemaan, line 431: There is none either in the node cleanup and it works perfectly fine.

The node and person tables are less than a few tens of thousands of rows.

My point is there is no need to engineer a solution for this specific case.

We should instead address the problem in general using the database-cleaner logic and use it in all of our cleanups, not just do it for the case where it is almost useless

jrouzierinverse, line 431: As I thought about it since we are not doing bulk deletes in sql batching is not needed in these cases.
So this will not be problem for node_cleanup or person_cleanup.
However it will be slower since we are doing it in code instead of sql.
But like you said these are never really big tables not that big of a deal.

However there is no grace period for deletion from the password table.
This is a deviation from all the other expiration patterns we have.

lib/pf/person.pm
```diff
@@ -375,6 +388,42 @@ sub person_violations {
...
+=head2 person_cleanup
+
+Clean all persons that are not the owner of a node and that are not a local account that is still valid
+
+=cut
+
+sub person_cleanup {
+    my @to_delete = map { $_->{pid} } persons_without_nodes();
+    my $now = DateTime->now();
+    foreach my $pid (@to_delete) {
```
jrouzierinverse, line 411: We should add a check to exclude the default user and admin user from deletion.

julsemaan, line 411: good idea.

lib/pf/person.pm
julsemaan, line 431: You're right, but I didn't add it since the password table defines itself the expiration field versus the node table that defines when it was last active.

lib/pf/person.pm
julsemaan, line 411: done

julsemaan:
Re-review

lib/pf/constants.pm
```diff
@@ -17,6 +17,7 @@ use Readonly;
 use base qw(Exporter);
 our @EXPORT = qw(
     $FALSE $TRUE $YES $NO $default_pid $admin_pid
+    %BUILT_IN_USERS
```
dwlfrth: Me not like exporting for exporting :(

lib/pf/person.pm
```diff
@@ -375,6 +389,46 @@ sub person_violations {
...
+        if($BUILT_IN_USERS{$pid}){
```
dwlfrth: Me not like exporting only for exporting
pf::constants::BUILT_IN_USERS ?

lib/pf/constants.pm
dwlfrth: Also, picky but since we're there: -> BUILTIN_USERS :)

julsemaan:
re-review

PR #1237, opened 2016-03-10
Fix/less rpc in acct

julsemaan:
# Description

Reduce the amount of webservices calls for the handling of the accounting.
# Impacts

Accounting
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Reduced the number of webservices calls in accounting
- Enhancement 2

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -1153,9 +1153,7 @@ sub radius_rest_switch_authorize :Public :RestPath(/radius/rest/switch/authorize
 sub handle_accounting_metadata : Public {
     my ($class, %RAD_REQUEST) = @_;
     my $logger = pf::log::get_logger();
-    $logger->info("Entering handling of accounting metadata");
-    use Data::Dumper;
-    $logger->info(Dumper(\%RAD_REQUEST));
+    $logger->debug("Entering handling of accounting metadata");
     my $client = pf::client::getClient();
     
     my $mac = pf::util::clean_mac($RAD_REQUEST{'Calling-Station-Id'});
@@ -1173,7 +1171,7 @@ sub handle_accounting_metadata : Public {
         $client->notify("update_iplog", mac => $mac, ip => $RAD_REQUEST{'Framed-IP-Address'}) if ($RAD_REQUEST{'Framed-IP-Address'} );
     }
     else {
-        pf::log::get_logger->debug("Not handling packet because we're not configured to update the iplog on accounting packets.");
+        pf::log::get_logger->debug("Not handling iplog update because we're not configured to do so on accounting packets.");
     }
 
 }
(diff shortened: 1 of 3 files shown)


louismunro:
Merging!

raddb/mods-config/perl/packetfence.pm
```diff
@@ -339,16 +339,10 @@ sub accounting {
...
             $data = send_rpc_request($config, "radius_accounting", \%RAD_REQUEST);
-        } elsif ($RAD_REQUEST{'Acct-Status-Type'} eq 'Start') {
-            #
-            # Updating location log in on initial ('Start') accounting run.
-            #
-            $data = send_rpc_request($config, "radius_update_locationlog", \%RAD_REQUEST);
-        }
-        # Tracking IP address.
-        send_rpc_request($config, "update_iplog", {mac => $mac, ip => $RAD_REQUEST{'Framed-IP-Address'}, source => "accounting"}) if ($RAD_REQUEST{'Framed-IP-Address'} );
+        }
```
jrouzierinverse, line 345: We should also handle the Stop and Interim-Update to

julsemaan, line 345: I don't think so because we want to send it for every type that exists so we can extract out metadata out of every packet.

I initially made this piece of code for accounting fingerprinting and its present in every packet so we want the metadata to be handled on every packet

lib/pf/api.pm
```diff
@@ -1156,6 +1149,35 @@ sub radius_rest_switch_authorize :Public :RestPath(/radius/rest/switch/authorize
 
     return \%mapped_return;
 }
+
+sub handle_accounting_metadata : Public {
+    my ($class, %RAD_REQUEST) = @_;
+    my $logger = pf::log::get_logger();
+    $logger->debug("Entering handling of accounting metadata");
+    my $client = pf::client::getClient();
```
fdurand, line 1157: don't we have to use the queue ?

julsemaan, line 1157: Since this call is executed in the webservices, calling getClient will return the local one meaning it will be executed in the same process which is desirable.

Otherwise the webservices call will queue 2 tasks while its already being executed async

extrafu:
Get some traction on this PR so we can merge it and be ready for v6.

lib/pf/api.pm
```diff
@@ -1156,6 +1149,35 @@ sub radius_rest_switch_authorize :Public :RestPath(/radius/rest/switch/authorize
...
+    
+    my $mac = pf::util::clean_mac($RAD_REQUEST{'Calling-Station-Id'});
+    if ($RAD_REQUEST{'Acct-Status-Type'} eq 'Start') {
+        #
+        # Updating location log in on initial ('Start') accounting run.
+        #
+        $logger->info("Updating locationlog from accounting request");
+        $client->notify("radius_update_locationlog", %RAD_REQUEST);
```
fdurand, line 1165: We must return the return the value of radius_update_locationlog because the radius server will answer $RADIUS::RLM_MODULE_FAIL

julsemaan, line 1165: It shouldn't because I used a notification for this call instead of send_rpc_request
From my tests, it behaved correctly.

PR #1238, opened 2016-03-10
Update locationlog with the last connection_type and sub_type

fdurand:
# Description

Update locationlog entry based on the latest one.
Since online/offline close the locationlog on accounting stop i need to fetch the latest connection type info from the locationlog to insert a new one.
# Impacts

Switch that support supportsRoamingAccounting
# Issue

Fix the update location log with the incorrect connection type
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix the update of the locationlog from accounting start with the wrong connection type

### DIFF
--- a/lib/pf/radius.pm
+++ b/lib/pf/radius.pm
@@ -381,7 +381,7 @@ sub update_locationlog_accounting {
             return [ $RADIUS::RLM_MODULE_OK, ('Reply-Message' => "Update locationlog from accounting ok") ];
         }
     }
-    return [ $RADIUS::RLM_MODULE_OK, ('Reply-Message' => "Not updated locationlog from accounting ok") ];
+    return [ $RADIUS::RLM_MODULE_OK, ('Reply-Message' => "Did not update locationlog from the accounting") ];
 }
 
 =item * _parseRequest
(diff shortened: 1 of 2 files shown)


lib/pf/radius.pm
```diff
@@ -365,44 +365,46 @@ sub update_locationlog_accounting {
...
-            connection_sub_type => $connection_sub_type,
-            radius_request => $radius_request,
-        };
-        my $role_obj = new pf::role::custom();
-        my $role = $role_obj->fetchRoleForNode($args);
-        $switch->synchronize_locationlog($port, $vlan, $mac, undef, $connection_type, $connection_sub_type, $user_name, $ssid, $stripped_user_name, $realm, $role->{role});
-    }
-    return [ $RADIUS::RLM_MODULE_OK, ('Reply-Message' => "Update locationlog from accounting ok") ];
+    }
+    return [ $RADIUS::RLM_MODULE_OK, ('Reply-Message' => "Not updated locationlog from accounting ok") ];
```
julsemaan: "Did not update the locationlog from the accounting"

julsemaan:
I'm not sure I fully understand the value of this.

Doesn't the RADIUS accounting request already contains the connection type for the insertion in the locationlog ?

lib/pf/radius.pm
julsemaan: Just fix that detail and its good to merge

PR #1239, opened 2016-03-11
Feature/cache deauth acctsessionid

julsemaan:
# Description

Use CHI to cache the latest accounting request so we don't have to record accounting in the SQL tables for the deauthentication
# Impacts

That big beast that we call accounting
Deauthentication
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added option to deactivate the accounting recording in the SQL tables
- Added caching of the latest accounting request for use in access reevaluation.

### DIFF
--- a/lib/pf/accounting.pm
+++ b/lib/pf/accounting.pm
@@ -75,8 +75,6 @@ Readonly our $ACCOUNTING_TRIGGER_RE => qr/
 Readonly our $DIRECTION_IN => 'IN';
 Readonly our $DIRECTION_OUT => 'OUT';
 
-our $CACHE = pf::CHI->new(namespace => "accounting");
-
 =head1 SUBROUTINES
 
 =over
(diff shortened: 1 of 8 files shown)


julsemaan:
Ready for review

lib/pf/accounting.pm
```diff
@@ -74,6 +75,8 @@ Readonly our $ACCOUNTING_TRIGGER_RE => qr/
 Readonly our $DIRECTION_IN => 'IN';
 Readonly our $DIRECTION_OUT => 'OUT';
 
+our $CACHE = pf::CHI->new(namespace => "accounting");
```
rouzier: Wrap creating the cache in a function.
This will cause problems in a threaded environment

julsemaan:
As the frenchs would say : fixaide.

Re-review

lzammit:
Une baguette et un croissant au beurre s'il vous plait*

PR #1242, opened 2016-03-14
added proper RADIUS REST handling

julsemaan:
# Description

Add object to raise API error
Add REST encoder/decoder for RADIUS request/response
# Impacts

FR3 integration and REST API
# NEW Package(s) required

No, no, package no is needed
# Delete branch after merge

YES

### DIFF
--- a/lib/pf/radius/rest.pm
+++ b/lib/pf/radius/rest.pm
@@ -1,0 +1,89 @@
+package pf::radius::rest;
+
+=head1 NAME
+
+pf::radius::rest
+
+=cut
+
+=head1 DESCRIPTION
+
+Helper methods to use REST with FreeRADIUS
+
+=cut
+
+use strict;
+use warnings;
+
+use pf::log;
+use Apache2::Const -compile =>
+  qw(DONE OK DECLINED HTTP_UNAUTHORIZED HTTP_NOT_IMPLEMENTED HTTP_UNSUPPORTED_MEDIA_TYPE HTTP_PRECONDITION_FAILED HTTP_NO_CONTENT HTTP_NOT_FOUND SERVER_ERROR HTTP_OK HTTP_INTERNAL_SERVER_ERROR);
+use pf::api::error;
+
+=head2 format_response
+
+Format a PacketFence RADIUS response to the format expected by the FreeRADIUS REST module
+
+=cut
+
+sub format_response {
+    my ($response) = @_;
+
+    my $radius_return = shift @$response; 
+    my %mapped_object = @$response; 
+    %mapped_object = ( %{$mapped_object{"RADIUS_AUDIT"}}, %mapped_object);
+
+    delete $mapped_object{"RADIUS_AUDIT"};
+    use Data::Dumper; get_logger->debug(sub { use Data::Dumper ; "RADIUS REST object : ". Dumper(\%mapped_object) });
+    $response = \%mapped_object; 
+    
+    unless ($radius_return == 2) { 
+        die pf::api::error->new(status => Apache2::Const::HTTP_UNAUTHORIZED, response => $response);
+    }
+
+    return $response;
+}
+
+=head2 format_request
+
+Format a FreeRADIUS REST RADIUS request to the format PacketFence expects it
+
+=cut
+
+sub format_request {
+    my ($request) = @_;
+    # transform the request according to what radius_authorize expects
+    my %remapped_radius_request = map {
+        $_ => $request->{$_}->{value}->[0];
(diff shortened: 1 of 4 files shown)


julsemaan:
Review

lib/pf/radius/rest.pm
```diff
@@ -0,0 +1,89 @@
...
+
+Format a FreeRADIUS REST RADIUS request to the format PacketFence expects it
+
+=cut
+
+sub format_request {
+    my ($request) = @_;
+    # transform the request according to what radius_authorize expects
+    my %remapped_radius_request = map {
+        $_ => $request->{$_}->{value}->[0];
```
jrouzierinverse, line 62: This will lose multi value entries.
My question is do we current use them in our API?
If so then we need to handle it this case since it will break code.

If not then should we handle it now or take care in the future?

julsemaan, line 62: To my knowledge we currently do not use multi value entries which is why I did it this way.

Again, to my knowledge, having multi values as arrays in the radius request hash can even lead to error in the code.

This requires input from the whole team though to be sure this is right, but it should be...

We should only handle arrays in the answer.

louismunro, line 62: We have never handled multiple values correctly.
This has actually caused bugs in some cases (e.g. when receiving multiple Calling-Station-Ids).
I think it's better to flatten the list  if we can't or won't handle them.

louismunro, line 62: Furthermore, handling arrays correctly now would require that we rewrite significant parts of the code.
Role.pm, radius.pm, switch.pm etc.

rouzier, line 62: Louis when you say flatten do you mean do a join by some separator?
Or do you mean just use the first value?

louismunro, line 62: Just use the first.
Joining them would break as many things as using a list in the first place.

I know it's not ideal, but we have lived with it for a long time anyway.
There was always the risk that the NAS could send multiple values.
We never dealt with them properly before either.

lib/pf/radius/rest.pm
```diff
@@ -0,0 +1,89 @@
...
+
+sub format_response {
+    my ($response) = @_;
+
+    my $radius_return = shift @$response; 
+    my %mapped_object = @$response; 
+    %mapped_object = ( %{$mapped_object{"RADIUS_AUDIT"}}, %mapped_object);
+
+    delete $mapped_object{"RADIUS_AUDIT"};
+    use Data::Dumper; get_logger->debug(sub { use Data::Dumper ; "RADIUS REST object : ". Dumper(\%mapped_object) });
```
rouzier: Remove Data::Dumper

julsemaan:
Review again

lib/pf/radius/rest.pm
```diff
@@ -0,0 +1,89 @@
...
+    my %mapped_object = @$response; 
+    %mapped_object = ( %{$mapped_object{"RADIUS_AUDIT"}}, %mapped_object);
+
+    delete $mapped_object{"RADIUS_AUDIT"};
+    get_logger->trace(sub { use Data::Dumper ; "RADIUS REST object : ". Dumper(\%mapped_object) });
+    $response = \%mapped_object; 
+    
+    unless ($radius_return == 2) { 
+        die pf::api::error->new(status => Apache2::Const::HTTP_UNAUTHORIZED, response => $response);
+    }
```
louismunro, line 47: we might want to extend the code to handle the subtly different following cases (from conf/radiusd/rest.conf):

```
    #  Authorize/Authenticate
    #
    #  Code   Meaning       Process body  Module code
    #  404    not found     no            notfound
    #  410    gone          no            notfound
    #  403    forbidden     no            userlock
    #  401    unauthorized  yes           reject
    #  204    no content    no            ok
    #  2xx    successful    yes           ok/updated
    #  5xx    server error  no            fail
    #  xxx    -             no            invalid
```

julsemaan, line 47: Right but we will need to map the return code from Switch.pm to an HTTP code which I'm not sure about

louismunro, line 47: I think we could do with just supporting userlock (-1) for now.
That's the only one I know we sometimes use.

julsemaan, line 47: adjusted

louismunro, line 47: Did you push the change?

julsemaan, line 47: I sure did*

`* To the wrong repo`

PR #1252, opened 2016-03-21
Feature/portal titles

### LINKED ISSUE #1251: There is no way to customize the titles of the portal profile pages
Previously in the portal templates allowed the titles of the pages to customized.
It customized from within the templates themselves.
Now this has been completely removed.

julsemaan:
# Description

Add title setting in the new captive portal
Also cleans up outdated Catalyst controllers implementations
Also reworks the i18n calls to be in the templates instead of the controllers
# Impacts

Title of HTML pages
Outdated catalyst controllers
# Issue

fixes #1251 
# Delete branch after merge

YES


cgx:
Food for thought: I would try to adopt a standard guideline when localizing strings. Should we internationalize strings in the controllers or in the HTML templates? Current trend is to localize the strings in the templates.

julsemaan:
Right now the titles in the HTML templates don't do anything (i need to clean them)

I discussed with @jrouzierinverse last week and setting them in the controllers seems to make more sense especially with the new dynamic nature of templates.

So with this PR, the titles will always be set in the controllers

cgx:
All right. So all string variables exposed to the templates are assumed to be already localized.

julsemaan:
That is correct

jrouzierinverse:
In my opinion it would be better to internationalize in the template.

Since it is easier to notice if a template was not internationalize.
However if we do this in the code we do this in the controller it would be more difficult to find or we might even internationalize a string twice.

julsemaan:
I agree and disagree with you.

The problem with that is when strings need to be internationalized with parameters.

In that case, it becomes complex to pass the parameters as well as the string in a clean way (in our current codebase)

For example, in our current case, the title of a provisioning page contains the name of the provisioner.

And putting the titles in the templates wouldn't be a solution for that IMO as it led us to have half of the pages not have a title because it is not done consistently.

jrouzierinverse:
Julien so to be clear I was not talking about having titles in the templates.
That discussion is done and the decision is to place them in the configuration.

I would refering to having the internationalization in the templates vs doing it in the controller.
I see three choices (There may be more).

### We only do it in the templates

Advantages
- There is only one place to do internationalization
- It simplifies the code in the controllers
- It is easy to see if it was not done.

Disadvantages
- When using parameters in internationalization it can be awkward.

julsemaan:
Are we missing a comment ? I only see one choice.

jrouzierinverse:
### We only do it in the controller

Advantages
- There is only one place to do internationalization
- It simplifies the templates
- It can handle parameters easier

Disadvantages
- It can be more difficult to find out if internationalization was done
- Internationalization can be done twice on a string

### We do it in both

Advantages
- Where the controller can be simplified they are.
- The template code does not have to deal with parameters

Disadvantages
- We have to make this judgement call every time we need to something special

julsemaan:
I think doing it in both is a no-no as like you said, its a judgement call everytime.

My opinion would tend to do it in the controller just for the parameters issue.

So stashed variables would need to be pre-i18n and the hard-coded strings in the template would use i18n + i18n_format

Passing parameters to i18n_format would require to adapt it to accept an array ref that contains the parameters but then in the template we would have to know the nature of the variable so we choose the right sub (i18n vs i18n_format).

If we can find a solution that looks clean for the parameters issue, we can put it in the templates but again we need to handle cases like the errors (and their eventual parameters) so the only solution I see would be to have i18n (not i18n_format) handle an array ref of a string + its parameters.

jrouzierinverse:
So to be clear the two scenarios at the moment that would need multiple parameters are titles and error messages.

If we were to handle it in the templates here is how is one approach it.

Let me know if you think this approach would cover all situations that could arise.

The internalization function of the template would have the logic to handle the following 3 cases.
- One parameter i18n(msg) => i18n("String")
- Two parameters i18n(msg, array_ref)  => i18n("String %s String", ['String'])
- Two parameters i18n(msg, param)  => i18n("String %s String", 'String')

There is a fourth case for convenience only but we do not really need it.
Since it can be handled in the template or controller code.
- More than two parameters => i18n(msg, param1, param2, param3) 
  However it could be rewritten like this i18n(msg, [param1, param2, param3])

Here are how would use it the template.

### Examples

**Simple**

```
[% i18n("Text to show") %]
```

**Titles**

```
<title>[% i18n(title, title_params)  %]</title>
```

**Errors**

```
[% FOREACH error IN errors %]
<div class="[% error.class %]">[% i18n(error.msg, error.msg_params) %]</div>
[% END %]
```

julsemaan:
Having to put a variable and another one for its params seems a bit unnatural

In order to get rid of the params variable, I would have the following

```
title = ["An amazing %s", "page"]
i18n(title)
```

Then i18n would check the type of the first argument to know what to do with it.
This would IMO be easier to read in the code and adding params to existing strings would also be easier (only change it in the code, template will adjust accordingly)

julsemaan:
Also with the errors, they are currently not objects so we would need to rework all of the errors in the portal which will consume a bit of time

jrouzierinverse:
I agree the single parameter array/scalar approach is a much cleaner solution.
As for the error handling that is a different fish to fry.

julsemaan:
Well right now, the portal uses a flash and the flash type already contains the error class.

So we could have the following : 

```
app->flash->{notice} = ["%s is the best", "Zaymitte"]
```

Then in the template : 

```
[% FOREACH type IN flash %]
<div class="[% type %]">[% i18n(flash.$type) %]</div>
[% END %]
```

jrouzierinverse:
Do you want to take care of i18n changes in this pull request or should we open a new one?

julsemaan:
Might as well add it to this one.

I'll write in the PR when I'm done

jrouzierinverse:
Ok cool

julsemaan:
Ready for review with discussed i18n changes

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
```diff
@@ -87,6 +87,7 @@ sub code : Path : Args(1) {
         node_modify($c->portalSession->clientMac, unregdate => $unregdate);
         pf::activation::set_status_verified($code);
         $c->stash(
+            title => "Access granted",
```
jrouzierinverse, line 90: Should the title be allowed to be configured or should it be done by i18n?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
```diff
@@ -120,6 +121,7 @@ sub login : Private {
         $c->clear_errors;
     }
     $c->stash(
+        title => "Guest Sponsor Login",
```
jrouzierinverse, line 124: Should the title be allowed to be configured or should it be done by i18n?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
```diff
@@ -226,6 +228,7 @@ sub doSponsorRegistration : Private {
 
         # send to a success page
         $c->stash(
+            title => "Sponsor request accepted",
```
jrouzierinverse, line 231: Should the title be allowed to be configured or should it be done by i18n?

html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm
```diff
@@ -61,15 +61,15 @@ sub index : Path : Args(0) {
...
                 $c->detach('landing');
             }
         } else {
-            $c->stash(txt_auth_error => i18n_format("Please verify the provided MAC address."));
+            $c->stash(txt_auth_error => "Please verify the provided MAC address.");
         }
     }
     # User is authenticated so display registration page
-    $c->stash(template => 'device-registration/registration.html');
+    $c->stash(title => "Registration", template => 'device-registration/registration.html');
```
jrouzierinverse, line 72: Should the title be allowed to be configured or should it be done by i18n?

html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm
```diff
@@ -119,12 +119,12 @@ sub login : Local : Args(0) {
         $c->stash->{txt_auth_error} = join(' ', grep { ref ($_) eq '' } @{$c->error});
         $c->clear_errors;
     }
-    $c->stash( template => 'device-registration/login.html' );
+    $c->stash( title => "Login", template => 'device-registration/login.html' );
```
jrouzierinverse, line 122: Should the title be allowed to be configured or should it be done by i18n?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Remediation.pm
```diff
@@ -52,6 +52,7 @@ sub index : Path : Args(0) {
 
         my $node_info = node_view($mac);
         $c->stash(
+            'title'        => "violation: quarantine established",
```
jrouzierinverse, line 55: Should the title be allowed to be configured or should it be done by i18n?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
julsemaan, line 231: i18n or directly in the template IMO

The configuration will get heavy for something that is never changed

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
julsemaan, line 90: i18n or directly in the template IMO

The configuration will get heavy for something that is never changed

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
julsemaan, line 124: i18n or directly in the template IMO

The configuration will get heavy for something that is never changed

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
julsemaan, line 231: i18n or directly in the template IMO

The configuration will get heavy for something that is never changed

html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm
julsemaan, line 72: i18n or directly in the template IMO

The configuration will get heavy for something that is never changed

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
```diff
@@ -388,7 +397,7 @@ sub _render {
 
     our $processor = Template::AutoFilter->new($TT_OPTIONS);;
     my $output = '';
-    $processor->process($template, $args, \$output) || die("Can't generate template $template: ".$processor->error."Error : ".$@);
+    $processor->process($template, $args, \$output, {binmode => ':utf8'}) || die("Can't generate template $template: ".$processor->error."Error : ".$@);
```
jrouzierinverse: It did not see the code to handle array ref or a single scalar.

julsemaan: Not sure I understood that comment

julsemaan: But the binmode is garbage, I'll remove it

jrouzierinverse: This is not handled yet.

$stash->{cmd1} = ["Kneel before %s", "Zod"];
$stash->{cmd2} = "Kneel before Zaymitte";

[% i18n(cmd1)%] vs [% i18n(cmd2)%]

julsemaan: It's in lib/pf/web.pm

lib/pf/web.pm
```diff
@@ -71,7 +72,15 @@ Warning: The list of subroutine is incomplete
 sub i18n {
     my $msgid = shift;
 
-    return gettext($msgid);
+    if(ref($msgid) eq "ARRAY"){
+        my @infos = @$msgid;
+        my $id = shift @infos;
```
jrouzierinverse: Unnecessary work just pass the array to i18n_format.

return i18n_format(@$msgid);

html/captive-portal/lib/captiveportal/PacketFence/Controller/Violation.pm
```diff
@@ -92,6 +94,29 @@ sub index : Path : Args(0) {
...
+Get the subtemplate in the right portal profile
+
+=cut
+
+sub getSubTemplate {
+    my ( $self, $c, $template ) = @_;
+    my $portalSession = $c->portalSession;
+    return "violations/$template.html";
+#    my $langs         = $portalSession->getRequestLanguages();
+    my $langs         = [];
```
jrouzierinverse, line 108: Are we no longer supporting different languages for the violation pages?

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
```diff
@@ -363,10 +367,15 @@ Render a template using Template Toolkit.
...
 
+    if(defined($args->{title})){
+        $self->title($args->{title});
+    }
+
     our $TT_OPTIONS = {
         AUTO_FILTER => 'html',
         RELATIVE => 1,
         INCLUDE_PATH => $self->profile->{_template_paths},
+        ENCODING => 'utf8',
```
jrouzierinverse: To make it easier to extend the template processor move the creation of the TT_OPTIONS into a method.

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
```diff
@@ -363,10 +367,15 @@ Render a template using Template Toolkit.
...
     };
 
     use Template::Stash;
```
jrouzierinverse: Is this needed?

html/captive-portal/lib/captiveportal/PacketFence/Controller/Violation.pm
julsemaan, line 108: Not even sure what this does :P I copy/pasted it from another controller I deleted.

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
```diff
@@ -110,7 +110,7 @@ sub do_sponsor_registration {
     foreach my $key (qw(firstname lastname telephone company sponsor)) {
         $info{$key} = $self->request_fields->{$key};
     }
-    $info{'subject'} = $self->app->i18n_format( "%s: Guest access request", $Config{'general'}{'domain'} );
+    $info{'subject'} = [ "%s: Guest access request", $Config{'general'}{'domain'} ];
     utf8::decode($info{'subject'});
```
jrouzierinverse: This no longer makes sense

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
```diff
@@ -110,7 +110,7 @@ sub do_sponsor_registration {
     foreach my $key (qw(firstname lastname telephone company sponsor)) {
         $info{$key} = $self->request_fields->{$key};
     }
-    $info{'subject'} = $self->app->i18n_format( "%s: Guest access request", $Config{'general'}{'domain'} );
+    $info{'subject'} = [ "%s: Guest access request", $Config{'general'}{'domain'} ];
```
jrouzierinverse: This would need to be internationalized in pf::activation::create_and_send_activation_code since it is used for send email

html/captive-portal/templates/layout.html
```diff
@@ -1,12 +1,14 @@
 <html>
   <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <link href="/content/new-styles.css" media="screen" rel="stylesheet" type="text/css" />
     <script src="/common/jquery-1.11.3.min.js" type="text/javascript"></script>
     <script src="/common/pf.js" type="text/javascript"></script>
+    <title>[% IF title.defined ; i18n(title) ; ELSE ; i18n("PacketFence Registration system") ; END %]</title>
   </head>
   <h1>PacketFence captive portal</h1>
   [% FOREACH level IN flash.keys %]
```
jrouzierinverse, line 10: The order of the keys in flash can be random.
Which mean we have no control on how the the flash is displayed.

julsemaan, line 10: You are right.

Not sure if this is really an issue though.

Not sure either what it has to do with this PR

jrouzierinverse, line 10: Just pointing out a potential issue that is all.

html/captive-portal/lib/captiveportal/PacketFence/Controller/Violation.pm
julsemaan, line 108: So looking at it, seems we do support them (well we ship a few of the templates with locales in them)

julsemaan:
all fixed up except for the localized violation sub templates.

Not sure what we should do about those...

cgx:
@julsemaan Check how we do it in [captiveportal::PacketFence::Controller::Remediation](https://github.com/inverse-inc/packetfence/blob/devel/html/captive-portal/lib/captiveportal/PacketFence/Controller/Remediation.pm#L62). We must continue to support localized violation templates.

jrouzierinverse:
When researching the issue that seemed to be broken for a while.
Do you want deal with it in this pull request or do you want me to deal with it in another.

julsemaan:
Would prefer if you could do it in another since that one is already getting quite big and would need merging asap as it contains testable features

jrouzierinverse:
I agree I will create an issue and assign it to myself

html/captive-portal/lib/captiveportal/PacketFence/Controller/Remediation.pm
```diff
@@ -91,14 +92,15 @@ sub scan_status : Private {
...
+        title => "scan: scan in progress",
         template    => 'scan-in-progress.html',
         timer         => $Config{'trapping'}{'redirtimer'},
-        txt_message => i18n_format(
+        txt_message => [
             'scan in progress contact support if too long',
             $scan_start_time
-        ),
+        ],
         txt_auto_refresh =>
```
cgx: Where is this variable used? Can't find it nowhere.

julsemaan: You are right, this is dead code

I'll remove it

html/captive-portal/lib/captiveportal/PacketFence/Controller/Violation.pm
```diff
@@ -92,6 +94,29 @@ sub index : Path : Args(0) {
     }
 }
 
+=head2 getSubTemplate
+
+Get the subtemplate in the right portal profile
+
+=cut
+
+sub getSubTemplate {
```
cgx: You should explicitly mark this function has `Private`.

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Billing.pm
```diff
@@ -121,7 +121,7 @@ sub verify {
     };
     if ($@) {
         get_logger->error($@);
-        $self->app->flash->{billing_error} = $self->app->i18n("Unable to process payment");
+        $self->app->flash->{billing_error} = "Unable to process payment";
```
cgx: Please use `error` instead of `billing_error`.

julsemaan:
Adjustments made.
Can we go ahead and merge this ?

cgx:
Don't you want to add the titles to the .po file?

Having the strings far from the `i18n` function call restrains `extract_i18n_strings_portal.sh` from extracting the localized strings ..

julsemaan:
Do they need to be added to all the po files manually or just en works ?

Otherwise this seems way too manual and we'll need to find a way to extract them with the new way

cgx:
Just the english/source .po. The others will be translated and fetched from Transifex.

julsemaan:
i18n completed.

Good enough to merge ?

conf/locale/en/LC_MESSAGES/packetfence.po
```diff
@@ -796,7 +796,7 @@ msgstr "If you believe that you must be allowed to connect then contact your loc
 # html/captive-portal/templates/provisioner/android.html
 # html/captive-portal/templates/provisioner/mobileconfig.html
 # html/captive-portal/templates/provisioner/windows.html
-# html/captive-portal/templates/release.html
+# tml/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module.pm
```
jrouzierinverse: html

jrouzierinverse:
After that fix I will push the big green button

julsemaan:
Go ahead

PR #1254, opened 2016-03-21
added RADIUS clients for management IPs

julsemaan:
# Description

Add RADIUS clients for the management IPs of PacketFence so that radtest can be used to monitor the FreeRADIUS server
# Impacts

RADIUS clients + PacketFence Switches
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added the PacketFence management IP addresses as switches so that they can be used in monitoring tools

### DIFF
--- a/lib/pfconfig/namespaces/config/Switch.pm
+++ b/lib/pfconfig/namespaces/config/Switch.pm
@@ -29,7 +29,9 @@ use base 'pfconfig::namespaces::config';
 sub init {
     my ($self) = @_;
     $self->{file}            = $switches_config_file;
-    $self->{child_resources} = [ 'resource::default_switch', 'resource::switches_ranges' ];
+    $self->{child_resources} = [ 'resource::default_switch', 'resource::switches_ranges', 'interfaces::management_network' ];
+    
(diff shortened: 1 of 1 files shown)


lib/pfconfig/namespaces/config/Switch.pm
```diff
@@ -46,6 +48,17 @@ sub build_child {
...
+    push @management_ips, $self->{management_network}->tag('vip') if(defined($self->{management_network}->tag('vip')));
+    push @management_ips, $self->{management_network}->tag('ip') if(defined($self->{management_network}->tag('ip')));
+    foreach my $management_ip (@management_ips){
+        $tmp_cfg{$management_ip} = {
+            type => 'PacketFence', 
+            mode => 'production', 
+            radiusSecret => 'testing1234', 
+        };
+    }
+
```
jrouzierinverse, line 61: This could be done directly added in the function.
pf::freeradius::freeradius_populate_nas_config
Instead of creating fake switches here

julsemaan, line 61: It makes it a valid switch as well so that if we want to run an eapol_test and go inside the PacketFence webservices call, we will be able too

PR #1255, opened 2016-03-21
Rename Target Vlan to Target Role in violation

fdurand:
# Description

Replaced the Target Vlan to Target Role in violation admin gui
# Impacts

Admin gui
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Renamed Vlan to Role in admin gui violation

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Violation.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Violation.pm
@@ -176,12 +176,12 @@ has_field 'button_text' =>
 has_field 'vlan' =>
   (
    type => 'Select',
-   label => 'VLAN',
+   label => 'Role',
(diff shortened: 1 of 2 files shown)


html/pfappserver/lib/pfappserver/Form/Violation.pm
```diff
@@ -176,12 +176,12 @@ has_field 'button_text' =>
 has_field 'vlan' =>
```
jrouzierinverse, line 176: Why not also change the name of field to avoid confusion on what is for

fdurand, line 176: Changing this mean that we have to change it every where in the code (db, violations.conf ...)

PR #1256, opened 2016-03-21
Assign the voice category to voip devices

### LINKED ISSUE #1280: DB schema: Do not specify node category ID when creating schema
Column 'id' of the node_category table is auto-increment. There is no need to specify the ID when creating the schema.

fdurand:
# Description

Assign the voice role if PacketFence detect that the device is a VoIP
# Impacts

Update node and locationlog with the voice role
# Delete branch after merge

YES
# Fixes

Fixes #1280 
# NEWS file entries
## Enhancements
- Assign the voice role to VoIP devices when PacketFence detect it


lib/pf/radius.pm
```diff
@@ -454,7 +454,7 @@ sub _authorizeVoip {
             ('Reply-Message' => "Server reported: VoIP authorization over RADIUS not supported for this network device")
         ];
     }
-    $args->{'switch'}->synchronize_locationlog($args->{'ifIndex'}, $args->{'switch'}->getVlanByName('voice'), $args->{'mac'}, $VOIP, $args->{'connection_type'}, $args->{'connection_sub_type'}, $args->{'user_name'}, $args->{'ssid'});
+    $args->{'switch'}->synchronize_locationlog($args->{'ifIndex'}, $args->{'switch'}->getVlanByName('voice'), $args->{'mac'}, $VOIP, $args->{'connection_type'}, $args->{'connection_sub_type'}, $args->{'user_name'}, $args->{'ssid'}, undef, undef, 'voice');
```
jrouzierinverse: Create a constant like VOICE_ROLE

lib/pf/role.pm
```diff
@@ -567,6 +567,7 @@ sub getNodeInfoForAutoReg {
     # this might look circular but if a VoIP dhcp fingerprint was seen, we'll set node.voip to VOIP
     if ($args->{'isPhone'}) {
         $node_info{'voip'} = $VOIP;
+        $node_info{'category'} = 'voice';
```
jrouzierinverse: use a constant

jrouzierinverse:
Since we are using this as an internal role we should have a flag that would refuse to delete/modify internal roles.

I would keep that code out of this pull request since it out of scope for the pull request.

fdurand:
What i can do is to revert back the change in lib/pf/Switch/constants.pm and use uniq on the role array.

extrafu:
Bump - let's get it in there ASAP.

db/upgrade-X-X-X-X-Y-Z.sql
```diff
@@ -0,0 +1,28 @@
...
+INSERT INTO pf_version (id, version) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION, @SUBMINOR_VERSION));
+--
+-- PacketFence SQL schema upgrade from 5.7.0 to 5.8.0
+--
+
+--
+-- Insert 'VoIP' category
+--
+
+INSERT INTO `node_category` (name,notes) VALUES ("voice","VoIP devices");
```
julsemaan: Move that insert above the pf_version setting since the latest needs to be done at the very end

julsemaan: Also that script is wrong.

This needs to go in upgrade-X.X.X-X.Y.Z.sql not upgrade-X-X-X-X-Y-Z.sql

dwlfrth:
Rebase please

db/pf-schema-X.Y.Z.sql
```diff
@@ -122,6 +122,12 @@ INSERT INTO `node_category` (category_id,name,notes) VALUES ("2","guest","Guests
 INSERT INTO `node_category` (category_id,name,notes) VALUES ("3","gaming","Gaming devices");
 
 --
+-- Insert 'voice' category
+--
+
+INSERT INTO `node_category` (category_id,name,notes) VALUES ("4","voice","VoIP devices");
```
dwlfrth: Why are we specifying an ID ('4' in this case) if the column is auto-increment.
I know that the upgrade schema does not, but unsure why doing it when creating the schema

(That is also applicable to 3 other roles :P)

jrouzierinverse: The answer is simple. Because that is the way it was always done.
It was that way since node_category was introduced in PacketFence 1.9.0.

The bigger question is do we have any code that depends on those roles being set to those category_ids.

If not the I would recommend to remove it.
Else if we do then we create an issue and address in the next release.

jrouzierinverse:
Rebased as you asked.

db/pf-schema-X.Y.Z.sql
dwlfrth: dangerous

jrouzierinverse:
@dwlfrth Do you know if there is any code that depends on the hard coded category_id of the node_category table
My inspections says no.
Can you also check to make sure so we can have a direction on the topic.

dwlfrth:
Unfortunately I can't tell...
I'll have a look

lib/pf/constants/role.pm
```diff
@@ -0,0 +1,78 @@
...
+
+=cut
+
+Readonly::Scalar our $VOICE_ROLE => 'voice';
+Readonly::Scalar our $REGISTRATION_ROLE => 'registration';
+Readonly::Scalar our $ISOLATION_ROLE => 'isolation';
+Readonly::Scalar our $MAC_DETECTION_ROLE => 'macDetection';
+Readonly::Scalar our $INLINE_ROLE => 'inline';
+
+Readonly::Array our @ROLES => (
```
dwlfrth, line 49: Missing $VOICE_ROLE ?

rouzier, line 49: No the list is for roles not in the DB by default

dwlfrth, line 49: So, why : https://github.com/inverse-inc/packetfence/pull/1256/files#diff-7a8b09d4b1d078251d635d6dab5e9caeL39

rouzier, line 49: None of those roles other are in the node_category table

dwlfrth, line 49: It is missing from what I can see the @ROLES is used for...
https://github.com/inverse-inc/packetfence/blob/feature/assign_role_for_voip/html/pfappserver/lib/pfappserver/Form/Role.pm#L54

julsemaan, line 49: Seems like this prevents creating a role that is a reserved name (from my understanding)

Since the voice role will now exist, it doesn't need to be validated there since it will be a duplicate of voice which is already in the DB

One thing we need to make sure though is that the voice role _cannot_ be deleted from the admin. Otherwise calls to node_modify($ziBigMac, category => $VOICE_ROLE) will fail and they are there in role.pm

rouzier, line 49: That was something I thought about also maybe a flag in the database saying if it is an internal role or should we handle in via code.

julsemaan, line 49: I would say via code is enough.

If somebody goes pro style and deletes the voice role directly in the DB, he probably knows what he is doing or deserves to have issues.

Preventing the deletion in the DB may lead to other side effects which we may not want.

Both are good though, as long as pressing the small red button doesn't delete it, I'm happy

dwlfrth, line 49: As @julsemaan said, I'm happy :P

dwlfrth, line 49: Handle that case in code and merge away !

db/pf-schema-X.Y.Z.sql
dwlfrth: Related opened issue #1280

lib/pf/constants/role.pm
julsemaan, line 49: Yes that's what I said.
Had an extra drink last night.... You see what that does

jrouzierinverse:
Please review

html/pfappserver/lib/pfappserver/Model/Roles.pm
```diff
@@ -129,6 +130,9 @@ sub delete {
     my ($self, $role_ref) = @_;
 
     my $logger = get_logger();
+    if (exists $STANDARD_ROLES{$role_ref->{name}}) {
+        return ($STATUS::FORBIDDEN, "The role $role_ref->{name} a standard role cannot be remove");
```
dwlfrth: - I'd put the role name between '' ie: The role 'patate' ...
- is a standard role ?
- cannot be removed ?

dwlfrth: Other than that, I'd say ready to merge

PR #1261, opened 2016-03-22
Dynamic profile filter fields

jrouzierinverse:
## Description

When changing a profile filter type the match field will be changed to suit the filters type's needs.

The following filter types are now dynamic types
- Switch
- Roles
- Connection Types
- Connection Sub Types
## Impacts

The Admin GUI Profiles page
## Dependencies

List the PRs or other factors on which this PR depends
## NEWS file entries

New Features
++++++++++++
- Profile filters provides the appropriate values according its type
## Delete branch after merge

NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm
+++ b/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm
@@ -118,11 +118,11 @@ Append additional data after the view
 after view => sub {
     my ($self, $c) = @_;
     my ($status, $roles) = $c->model('Roles')->list;
-    $c->stash(
+    $c->stash({
         connection_types => [ keys %connection_type ],
         connection_sub_types => [ sort keys %RADIUS_EAP_TYPE_2_VALUES ],
         node_roles => $roles,
-    )
+    });
 };
 
 sub sort_profiles :Local :Args(0) :AdminRole('PORTAL_PROFILES_READ') {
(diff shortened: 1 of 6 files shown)


julsemaan:
me like this

jrouzierinverse:
@lzammit Brought up a point about the Realm profile filter not having a drop down list.
I just want to make sure if it make sense to have a drop down list for realms also.

julsemaan:
I'm not sure...

Is there a way to switch from dropdown to manual ?

jrouzierinverse:
We could create a typeahead for the realms allowing someone to see the configured realms but also allow them to put in their own user defined realm.

jrouzierinverse:
Spoke to King Louis he said we should not worry about this. 

image

html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Profile.pm
```diff
@@ -356,7 +373,7 @@ Returns all the merge paths
 
 sub mergedPaths {
     my ($self, $c) = @_;
-    return (catfile($captiveportal_profile_templates_path, $c->stash->{id}), $self->parentPaths($c));
+    return grep { -d } (catfile($captiveportal_profile_templates_path, $c->stash->{id}), $self->parentPaths($c));
```
julsemaan, line 376: Does this belong here ?
Just making sure

rouzier, line 376: I found a bug while working on the branch so I fixed it.

PR #1265, opened 2016-03-23
Adding support for RFC7710

dwlfrth:
# Description

Adding support for Captive-portal DHCP attribute (RFC7710)
Even if there's probably 1 or 2 devices around the globe that are handling it, we now support it... wow
# Impacts

DHCP
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added support for Captive-Portal DHCP attribute (RFC7710)

### DIFF
--- a/lib/pf/services/manager/dhcpd.pm
+++ b/lib/pf/services/manager/dhcpd.pm
@@ -47,7 +47,7 @@ sub generateConfig {
     $tags{'networks'} = '';
     $tags{'active'} = '';
 
-    my $captive_portal_uri = pf::util::get_captive_portal_uri();
+    my $captive_portal_uri = pf::config::util::get_captive_portal_uri();
 
     my $failover_activated = 0;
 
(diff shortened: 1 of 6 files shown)


lib/pf/services/manager/dhcpd.pm
```diff
@@ -47,6 +47,8 @@ sub generateConfig {
     $tags{'networks'} = '';
     $tags{'active'} = '';
 
+    my $captive_portal_url = $Config{'general'}{'hostname'} . "." . $Config{'general'}{'domain'};
```
extrafu: shouldn't we prefix that with https:// or http:// ? the option must be an URI.

dwlfrth: yup!

lib/pf/services/manager/dhcpd.pm
```diff
@@ -47,6 +47,9 @@ sub generateConfig {
     $tags{'networks'} = '';
     $tags{'active'} = '';
 
+    my $captive_portal_uri = isenabled($Config{'captive_portal'}{'secure_redirect'}) ? $HTTPS : $HTTP;
+    $captive_portal_uri .= "://" . $Config{'general'}{'hostname'} . "." . $Config{'general'}{'domain'};
```
julsemaan: I think it would be the time to create a helper for the portal URL in pf::config::util

image

PR #1269, opened 2016-03-24
Dedicated pfconfig process for testing 

julsemaan:
# Description

Dedicates a pfconfig process for testing
Allows to override the configuration and PID file path of pfconfig via the command line
Makes use of a BDB backend for testing
Allow to change the backend of pfconfig via configuration
# Impacts

pfconfig for the new options introduced
The unit tests
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- The pfconfig backend can now be configured.

### DIFF
--- a/t/PfFilePaths.pm
+++ b/t/PfFilePaths.pm
@@ -15,11 +15,11 @@ Overrides the the location of config files to help with testing
 use strict;
 use warnings;
 
-use File::Slurp qw(read_file);
 
 our $PFCONFIG_TEST_PID_FILE;
 
 BEGIN {
+    use File::Slurp qw(read_file);
     use File::Path qw(remove_tree);
     use File::Spec::Functions qw(catfile catdir rel2abs);
     use File::Basename qw(dirname);
--- a/t/PfFilePaths.pm
+++ b/t/PfFilePaths.pm
@@ -15,11 +15,17 @@ Overrides the the location of config files to help with testing
 use strict;
 use warnings;
 
+use File::Slurp qw(read_file);
+
+our $PFCONFIG_TEST_PID_FILE;
+
 BEGIN {
     use File::Path qw(remove_tree);
     use File::Spec::Functions qw(catfile catdir rel2abs);
     use File::Basename qw(dirname);
     use pf::file_paths;
+    use pfconfig::constants;
+    use pfconfig::manager;
     remove_tree('/tmp/chi');
     my $test_dir = rel2abs(dirname($INC{'PfFilePaths.pm'})) if exists $INC{'PfFilePaths.pm'};
     $test_dir ||= catdir($install_dir,'t');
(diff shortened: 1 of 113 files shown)


jrouzierinverse:
:+1:

t/PfFilePaths.pm
```diff
@@ -32,12 +38,21 @@ BEGIN {
...
-}
 
-# we need to load the proper data in pfconfig
-use pfconfig::manager;
-pfconfig::manager->new->expire_all;
+    $pfconfig::constants::CONFIG_FILE_PATH = catfile($test_dir, 'data/pfconfig.conf');
+    $pfconfig::constants::SOCKET_PATH = "/usr/local/pf/var/run/pfconfig-test.sock";
 
+    $PFCONFIG_TEST_PID_FILE = "/usr/local/pf/var/run/pfconfig-test.pid";
+    `/usr/local/pf/sbin/pfconfig -s $pfconfig::constants::SOCKET_PATH -p $PFCONFIG_TEST_PID_FILE -c $pfconfig::constants::CONFIG_FILE_PATH -d`;
```
jrouzierinverse: Change the name to -n pfconfig-test

jrouzierinverse:
When testing this if there is not a pfconfig service listening at var/run/pfconfig.sock then the tests hangs.

jrouzierinverse:
More specifically t/pf.t will hang

julsemaan:
Not able to replicate that issue. Executing pf.t and smoke.t work.

But, that is with the latest commits I made. Shouldn't affect it but can you try again with the latest code

julsemaan:
One thing that doesn't work without the main pfconfig though is compile tests of the modperl requires and other binaries but there isn't much we can do about it as these are spawned independently from the process.

I don't think there is an easy way out of it. We'll always need the main pfconfig to be there but this PR aims mainly at not conflicting with our dev environments

jrouzierinverse:
This is perl there is always a way ;)

I was thinking about how to solve this while driving home.
The new check would be something like this.

perl -I/usr/local/pf/t/ -Mtest_paths -c path

Where test paths is something like this.

package test_paths;
BEGIN {
use pf::file_paths;
use pfconfig::constants;
#Override the file paths below
}

James Rouzier

julsemaan:
That good old perl...

I'll give it a shot but looks like it will work correctly

julsemaan:
All ready for another review with all failing unit tests fixed (except the one for the interfaces that @jrouzierinverse is fixing)

t/setup_test_config.pm
```diff
@@ -0,0 +1,91 @@
...
+    use pfconfig::manager;
+    use pfconfig::constants;
+    use File::Spec::Functions qw(catfile);
+    use File::Slurp qw(read_file);
+
+    use File::Path qw(remove_tree);
+    remove_tree('/tmp/chi');
+
+    $PFCONFIG_TEST_PID_FILE = "/usr/local/pf/var/run/pfconfig-test.pid";
+    `/usr/local/pf/sbin/pfconfig -n pfconfig-test -s $pfconfig::constants::SOCKET_PATH -p $PFCONFIG_TEST_PID_FILE -c $pfconfig::constants::CONFIG_FILE_PATH -d`;
```
rouzier: It would also make sense for pfconfig process to use test_paths also.
perl -I/usr/local/pf/t -M test_paths /usr/local/pf/sbin/pfconfig -n pfconfig-test -p $PFCONFIG_TEST_PID_FILE -d

t/test_paths.pm
```diff
@@ -34,12 +39,13 @@ BEGIN {
     $pf::file_paths::firewall_sso_config_file = catfile($test_dir,'data/firewall_sso.conf');
     $pf::file_paths::config_file = catfile($test_dir,'data/pf.conf');
     $pf::file_paths::pf_config_file = catfile($test_dir,'data/pf.conf');
```
jrouzierinverse, line 41: It would be better to only override pf.conf if an environmental variable is set.
So it can be use our own special pf.conf for testing.

jrouzierinverse:
Outside those two comments I mentioned everything else looks good.

t/test_paths.pm
julsemaan, line 41: Only thing is that the content of pf.conf should be predictable.

For example, in mine I have proxy passthroughs that make the unit tests fail.

I won't have a choice to point it to mine since it contains my IP addresses.

Would it sound right to have the pf.conf overriden via ENV and have the IPs and DB options overriden via ENV also ?

This way you don't have to absolutly override that value and do it if you want to while keeping the other ones so setup dependant variables can be set.

jrouzierinverse, line 41: Sounds good

PR #1273, opened 2016-03-25
added server error template for fatal error instantiating the Dynamic…

julsemaan:
# Description

Added minimalistic server error page when the DynamicRouting appliaction can't instantiate
# Impacts

Portal errors
# Delete branch after merge

YES

### DIFF
--- a/html/captive-portal/templates/server_error.html
+++ b/html/captive-portal/templates/server_error.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Fatal error !</title>
+  </head>
+
+  <body>
+    <h2>Fatal error !</h2>
+    <p>A server error has occured, please try again or try to contact your local support staff</p>
+  </body>
+</html>
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
@@ -23,6 +23,9 @@ use List::MoreUtils qw(uniq);
 use captiveportal::DynamicRouting::Factory;
 use captiveportal::DynamicRouting::Application;
 use pf::StatsD::Timer;
+use pf::file_paths;
+use File::Slurp qw(read_file);
+use pf::error;
 
 BEGIN { extends 'captiveportal::Base::Controller'; }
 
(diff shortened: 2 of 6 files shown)


html/captive-portal/templates/server_error.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Fatal error !</title>
+  </head>
+
+  <body>
+    <h2>Fatal error !</h2>
+    <p>A server error has occured, please try again or try to contact your local support staff</p>
```
dwlfrth, line 9: Shouldn't we include a "standard" footer (maybe the one we already have but more abstracted) to provide support information that can be customized by clients (email, phone number, ip address, mac address) ?

julsemaan, line 9: This displays when s**t hits the fan and doesn't go through any templating at all
Since this display when there is an application error, the renderer won't be available.

Adding templating to that would add a possibility of having an error in the error handling

image

jrouzierinverse:
Could you add the ability to use a custom server_error.html from the profile directory

julsemaan:
Well you can have a per-profile server_error.html so I'm not really sure there is a plus-value to that

jrouzierinverse:
It seems incomplete without it.

James Rouzier

julsemaan:
Not sure why...

Can you point out a use case where that would be used

jrouzierinverse:
If a customer wants to completely control the look and feel of their PacketFence install down to the server error template without that they cannot.
Or in reality if they do they would have to have to worry about when they upgrade since it will be overridden on every install.

It is about giving the complete control of the look and feel.

If you do not want add this feature at minimum disable the modification of the server_error.html from the Admin GUI.
So it will not give the false impression that this file is allowed to be modified.

julsemaan:
Well it is allowed to be modified.

You don't have to think about the upgrade if you have a fully custom template because the upgrade doesn't touch the portal profile files so you do have full control on it

I really don't see where is your point with that.

Whether you change server_error.html or zammitbug_hit.html in the portal profile the result stays the exact same and behaves exactly the same way when you upgrade

jrouzierinverse:
It was agreed upon that all files in captive-portal/templates will be overwritten on a upgrade.
So captive-portal/templates/server_error.html will be also overwritten on an upgrade.

Unless this has been changed.

julsemaan:
But if I edit server_error.html in the portal profile it creates a server_error.html in 'html/captive-portal/profile-templates/Ziportal' which never gets overridden

jrouzierinverse:
But it your code at the moment does not do that.

jrouzierinverse:
The code only read from captive-portal/templates/server_error.html not from a profile

julsemaan:
Well then all along the issue was that it should look it up, not that it should be configurable.

I'll simply adjust it so it looks it up in the profile template paths and we're all good

julsemaan:
added lookup in portal, re-review

PR #1274, opened 2016-03-26
Added crontab file in /etc/cron.d at install (for db backup)

fdurand:
# Description

Added packetfence crontab at install
# Impacts

Be sure to remove previous database maintenance in crontab
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Install packetfence crontab at install
# UPGRADE file entries

Remove previous crontab job in global crontab (crontab -e)

### DIFF
--- a/packetfence.cron.d
+++ b/packetfence.cron.d
@@ -1,0 +1,7 @@
+# /etc/cron.d/packetfence: crontab entries for the packetfence package
+
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+30 00 * * * /usr/local/pf/addons/database-backup-and-maintenance.sh
+
--- a/packetfence.cron.d
+++ b/packetfence.cron.d
@@ -3,5 +3,5 @@
 SHELL=/bin/sh
 PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 
-30 00 * * * /usr/local/pf/addons/database-backup-and-maintenance.sh
+30 00 * * * root /usr/local/pf/addons/database-backup-and-maintenance.sh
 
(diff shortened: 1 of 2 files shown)


julsemaan:
This conflicts with the maintenance in the admin.

Not really sure this should be activated by default as it will clear out data from the DB with no way for the user to know it except from digging in his system.

He also won't get why he sets the iplog expiration to 2 weeks in the admin but its only keeps 1 week :)

We should either remove the cleanup from the script or make it use the same values as the admin interface.
In the last case, we need to make sure pfmon doesn't do the job and lets the script do it

dwlfrth:
I agree with @julsemaan, we should remove cleanup from the script.

julsemaan:
If that is the case, we need to make sure the defaults in pf.conf expire after 1 week

dwlfrth:
We should look at adjusting the followings:
- maintenance.locationlog_cleanup_window
- maintenance.iplog_rotation_window
- maintenance.iplog_cleanup_window
- maintenance.radius_audit_log_cleanup_window

Not sure there's anything related to radacct (radacct_log) and auth_log in pfmon jobs...

julsemaan:
Don't think they are there (neither is radacct I think)

We need to handle all these before merging that.

I personally suggest that we push that to 6.1

packetfence.cron.d
```diff
@@ -0,0 +1,7 @@
+# /etc/cron.d/packetfence: crontab entries for the packetfence package
+
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+30 00 * * * /usr/local/pf/addons/database-backup-and-maintenance.sh
```
julsemaan: Missing the user on that line

julsemaan:
We need to complete the migration of the cleanup inside pfmon before merging this. Otherwise, it will duplicate the job.

To my knowledge, only the accounting tables are missing.

Being what they are (our biggest pain), we could remove the cleanup of the other tables already handled by pfmon from it and leave the accounting for now.

fdurand:
Bump

PR #1275, opened 2016-03-26
Escape the attribute before trying to do a search

fdurand:
# Description

Escape the ldap value (sometimes it can contain unescape caracters)
# Impacts

NO
# Issue

Fix wrong filter if the value returned by ldap is not escaped.
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix unescape value in ldap search

### DIFF
--- a/lib/pf/Authentication/Source/LDAPSource.pm
+++ b/lib/pf/Authentication/Source/LDAPSource.pm
@@ -345,7 +345,7 @@ sub match_in_subclass {
         # Perform match on a static group condition since they require a second LDAP search
         foreach $condition (grep { $_->{'operator'} eq $Conditions::IS_MEMBER } @{$own_conditions}) {
             $value = escape_filter_value($condition->{'value'});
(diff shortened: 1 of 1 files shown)


lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -346,6 +346,7 @@ sub match_in_subclass {
         foreach $condition (grep { $_->{'operator'} eq $Conditions::IS_MEMBER } @{$own_conditions}) {
             $value = escape_filter_value($condition->{'value'});
             $attribute = $entry->get_value($condition->{'attribute'});
+            $attribute = escape_filter_value($attribute);
```
jrouzierinverse, line 349: Since you are already in the code can you add a check if the attribute is undef.
`$attribute = $entry->get_value($condition->{'attribute'}) // ''`;
Then I will merge it.

PR #1276, opened 2016-03-26
Compute the unregdate in registeredRole with secure connection

fdurand:
# Description

Unregdate was never computed in registeredRole for secure connection.
# Impacts

Compute the unregdate for a secure connection
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Compute the unregdate for a secure connection

### DIFF
--- a/lib/pf/role.pm
+++ b/lib/pf/role.pm
@@ -441,7 +441,7 @@ sub getRegisteredRole {
                 radius_request => $args->{radius_request},
             };
             $role = &pf::authentication::match([@sources], $params, $Actions::SET_ROLE, \$source);
-            my $value = &pf::authentication::match([@sources], $params, $Actions::SET_UNREG_DATE);
+            my $unregdate = &pf::authentication::match([@sources], $params, $Actions::SET_UNREG_DATE);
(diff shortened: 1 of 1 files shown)


lib/pf/role.pm
```diff
@@ -441,6 +441,7 @@ sub getRegisteredRole {
                 radius_request => $args->{radius_request},
             };
             $role = &pf::authentication::match([@sources], $params, $Actions::SET_ROLE, \$source);
+            my $value = &pf::authentication::match([@sources], $params, $Actions::SET_UNREG_DATE);
```
dwlfrth: my $unregdate would be better than "value" :dancer:

dwlfrth: Also, not sure performance wise of calling match two times.
I think that by not passing any action, match will return all of them from which you can then choose what you want.

jrouzierinverse: You are correct @dwlfrth but it does not guarantee that the SET_UNREG_DATE action will be in the list of actions.

When you pass the action it will match the first rule and action combination.

jrouzierinverse:
Looks good to me

PR #1288, opened 2016-03-30
Fix the online/offline search

jrouzierinverse:
## Description

Fix the location closing and online/offline search
## Impacts

Radius accounting and the admin gui
## NEWS file entries

Bug Fixes
+++++++++
- Remove locationlog closing on radius accounting stop

### DIFF
--- a/sbin/pfmon
+++ b/sbin/pfmon
@@ -252,13 +252,8 @@ sub registertasks  {
         }
     );
     register_task(
-        'update node online stats', 
-        'online_node_count_interval', 
-        \&update_node_online_stats
-    );
-    register_task(
-        'fingerbank data update', 
-        'fingerbank_data_update_interval', 
+        'fingerbank data update',
+        'fingerbank_data_update_interval',
         \&update_fingerbank_data
     );
 }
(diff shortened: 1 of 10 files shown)


jrouzierinverse:
Bump

sbin/pfmon
```diff
@@ -252,13 +252,8 @@ sub registertasks  {
...
     register_task(
-        'update node online stats', 
-        'online_node_count_interval', 
-        \&update_node_online_stats
-    );
-    register_task(
-        'fingerbank data update', 
-        'fingerbank_data_update_interval', 
+        'fingerbank data update',
+        'fingerbank_data_update_interval',
```
louismunro, line 256: You're sure this should be in this PR?

louismunro:
On it.
I want to test it before we merge so we don't screw up the maintenance branch.

sbin/pfmon
jrouzierinverse, line 256: That is just white space noise

louismunro, line 256: Ah, got it now.
Forget I said anything.

PR #1317, opened 2016-04-04
Move pf::file_paths to use @EXPORT_OK

### LINKED ISSUE #1315: Move pf::file_paths to use @EXPORT_OK
Since pf::file_paths exports all its variables to a module namespace (which is just a hash internally) it will increase the memory usage of any module that imports it.

So by moving to EXPORT_OK there will be a reduction of memory across all modules in packetfence ~4 megabytes

jrouzierinverse:
## Description

Since pf::file_paths exports all its variables to a module namespace (which is just a hash internally) it will increase the memory usage of any module that imports it.

So by moving to EXPORT_OK there will be a reduction of memory across all modules in packetfence ~4 megabytes
## Impacts

Nothing
## NEWS file entries

Enhancements
++++++++++++
- Reduction of memory usage

Bug Fixes
+++++++++
Fixes #1315
## Delete branch after merge

NO


jrouzierinverse:
Review me please

lib/pf/pfqueue/producer/redis.pm
```diff
@@ -16,8 +16,7 @@ use strict;
 use warnings;
 use Moo;
 extends qw(pf::pfqueue::producer);
-use pf::Redis;
-use pf::file_paths;
+use Redis::Fast;
```
julsemaan: Aren't we using pf::Redis ?

julsemaan:
I don't think anyone of us has the patience of going through 355 of copy/pasting stuff
Checked it quickly and didn't see anything abnormal except a small comment.

I would say the question is more, should we merge it for v6 or wait..
Risk is minimal since these are all variables so compilation errors arise quickly.

My opinion would be to merge it. IDK what the others think

lib/pf/pfqueue/producer/redis.pm
jrouzierinverse: My bad

jrouzierinverse:
All the smoke test pass very little risk and it bring some reduction of memory.

julsemaan:
I would agree with you.
Lets wait for someone else to say its OK so we can put the full blame on him if this PR causes issues

extrafu:
@lzammit told me it's all good!

jrouzierinverse:
If there are no objections I will merge this at the end of the day.

louismunro:
Waiting is for wusses.

PR #1318, opened 2016-04-04
Fix/portal session state

julsemaan:
# Description

Fix the MAC based portal session being lost when there is no session Cookie to initiate it.
This removes the MAC_based store and uses the MAC as the state of the session instead
# Impacts

Portal session
# Delete branch after merge

YES


julsemaan:
for quick review and merge please as all IOS sessions are bugged in the v6 portal without it.

lib/Catalyst/Plugin/Session/State/MAC.pm
```diff
@@ -0,0 +1,115 @@
...
+
+=cut
+
+sub get_session_id {
+    my $c = shift;
+
+    my $mac = $c->portalSession->clientMac;
+    if(valid_mac($mac)){
+        $mac =~ s/\://g;
+        return $mac;
```
jrouzierinverse, line 38: Why the extra step?
I do not see what we gain by removing the ':'

julsemaan, line 38: Catalyst does not consider it valid.
A catalyst session ID can only contain hexadecimal characters

jrouzierinverse, line 38: That would be a good reason

julsemaan:
Can we merge this ?

lib/Catalyst/Plugin/Session/State/MAC.pm
```diff
@@ -0,0 +1,115 @@
...
+=head2 browser_session_id
+
+Get the browser session ID (not tied to MAC address)
+
+=cut
+
+sub browser_session_id {
+    my ($c) = @_;
+    if($c->request->cookie('CGISESSION')){
+        $c->request->cookie('CGISESSION')->value();
```
jrouzierinverse: should this be return $c->request->cookie('CGISESSION')->value();

jrouzierinverse:
No we cannot merge this

PR #1335, opened 2016-04-06
Move pf::config to use @EXPORT_OK

### LINKED ISSUE #1315: Move pf::file_paths to use @EXPORT_OK

jrouzierinverse:
## Description 

Since pf::config exports all its variables to a module namespace (which is just a hash internally) it will increase the memory usage of any module that imports it.

So by moving to EXPORT_OK there will be a reduction of memory across all modules in packetfence ~5 megabytes

All test have passed
## Impacts

Nothing
## NEWS file entries

Enhancements
++++++++++++
- Reduction of memory usage
## Delete branch after merge

NO


jrouzierinverse:
Bump

julsemaan:
This is as boring to review as it was to make it.

Reviewed it quickly without seeing anything abnormal.

Like pf::file_paths export rework I shall raise the question of the merge-ability for v6 while agreeing with the merge

jrouzierinverse:
You are a brave man to even look at this

jrouzierinverse:
@louismunro do you want to do the honors?

louismunro:
I dub thee Sir Mergealot.

rouzier:
I like big merges I cannot lie

PR #1347, opened 2016-04-11
add memory cached sources to portal profile

julsemaan:
# Description

Optimize fetching of portal profile sources
Adds a new in-memory cache to pfconfig that can be tied to a namespace
# Impacts

Portal profile sources
# Delete branch after merge

YES

### DIFF
--- a/lib/pfconfig/memory_cached.pm
+++ b/lib/pfconfig/memory_cached.pm
@@ -1,0 +1,96 @@
+package pfconfig::memory_cached;
+
+=head1 NAME
+
+pfconfig::memory_cached
+
+=cut
+
+=head1 DESCRIPTION
+
+pfconfig::memory_cached
+
+This module is used to compute values and tie them to a pfconfig namespace so they are only created when the pfconfig namespace gets updated
+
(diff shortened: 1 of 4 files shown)


julsemaan:
For quick review and merge please, this is for v6

lib/pfconfig/memory_cached.pm
```diff
@@ -0,0 +1,96 @@
...
+sub compute {
+    my ( $self, $key, $sub ) = @_;
+    my $logger = $self->logger;
+
+    my $result = $self->compute_from_subcache($key, sub {
+        return $sub->();
+    });
+
+    return $result;
+}
```
jrouzierinverse: Why not just call compute_from_subcache directly in the code as this is the same thing as
$self-> compute_from_subcache($key,  $sub);

julsemaan: Just thought it would be cleaner this way if we add additional logic to the computing.

rouzier: This will be called often IMO it should be kept "light" until we need that extra layer of abstraction.

julsemaan:
re-review

PR #1380, opened 2016-04-14
Feature/statsd level

julsemaan:
# Description

Allow to set a level of importance to statsd timer to reduce the amount of whisper DBs while keeping fine grained timers, all of that without restarting a single service (looks too good to be true...)
# Impacts

Statsd timers and anything that uses them
# Delete branch after merge

YES

### DIFF
--- a/lib/pf/db.pm
+++ b/lib/pf/db.pm
@@ -177,7 +177,7 @@ sub get_db_handle {
 
 sub db_data {
     my ($from_module, $module_statements_ref, $query, @params) = @_;
-    my $timer = pf::StatsD::Timer->new({ 'stat' => called() . ".$query", sample_rate => 0.1, level => 9});
+    my $timer = pf::StatsD::Timer->new({ 'stat' => called() . ".$query",  level => 9});
 
     my $sth = db_query_execute($from_module, $module_statements_ref, $query, @params) || return (0);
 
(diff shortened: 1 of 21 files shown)


julsemaan:
@jrouzierinverse can you make sure I didn't go crazy with that one

lib/pf/db.pm
```diff
@@ -177,7 +177,7 @@ sub get_db_handle {
 
 sub db_data {
     my ($from_module, $module_statements_ref, $query, @params) = @_;
-    my $timer = pf::StatsD::Timer->new({ 'stat' => called() . ".$query", sample_rate => 0.1});
+    my $timer = pf::StatsD::Timer->new({ 'stat' => called() . ".$query", sample_rate => 0.1, level => 9});
```
louismunro: Maybe a bit high here? 
We may want some of that good stuff without going all the way to 11.

lib/pf/db.pm
```diff
@@ -214,7 +214,7 @@ sub db_data {
 # afterwards: remove export of $<module>_db_prepared and <module>_db_prepare()
 sub db_query_execute {
     my ($from_module, $module_statements_ref, $query, @params) = @_;
-    my $timer = pf::StatsD::Timer->new({ 'stat' => called() . ".$query", sample_rate => 0.1});
+    my $timer = pf::StatsD::Timer->new({ 'stat' => called() . ".$query", sample_rate => 0.1, level => 9});
```
louismunro: Same here. Can we say 7 or 8?
And how about that spoon?

jrouzierinverse:
You a few redundant  "sample_rate => 0.1" in the code.

Nothing a
 `perl -pi -e's/sample_rate => 0.1,//' $(grep -PRl 'sample_rate => 0.1,' lib/ html/)` 
could not fix.

Outside of that the code looks solid

lib/pf/db.pm
julsemaan: as discussed verbally, it isn't :)

lib/pf/db.pm
julsemaan: as discussed verbally, it isn't :)

julsemaan:
That code isn't going to reboot my machine isn't ?

jrouzierinverse:
Run it and find out

PR #1386, opened 2016-04-15
Fix #1384

### LINKED ISSUE #1384: Radius reply contain all internal PacketFence attributes
capture du 2016-04-15 12 46 39

fdurand:
# Description

Remove PacketFence's radius internal attributes in radius audit log
# Impacts

Radius audit log
# Issue

fixes #1384
# Delete branch after merge

YES


jrouzierinverse:
Oh I just fixed this by return the radius audit attributes in the control instead of the reply in the rest module.

PR #1387, opened 2016-04-17
Replaced call to packetfence.pm with an unlang policy.

louismunro:
# Description

Fixes broken eap-tls authentication in FR 3.
EAP-TLS sometimes sends very long "EAP-Message" attributes which it can't seem to pass to packetfence.pm.
This PR removes the need to call packetfence.pm in "authorize", thus obviating the risk of such an occurence.
Performance improvements are also a possible benefit.
# Impacts

All RADIUS authentication will hit this policy.
Nothing should actually trigger it unless it's EAP-MAC auth which is rare.
# Delete branch after merge

YES 
# NEWS file entries

This is purely internal.
I don't think it warrants an entry an in the NEWS file.
## Bug Fixes

None that were known before I hit one and fixed it here.

### DIFF
--- a/conf/radiusd/packetfence.example
+++ b/conf/radiusd/packetfence.example
@@ -94,7 +94,7 @@ authorize {
 			&control:Auth-Type := Accept
 		}
 	}
-	packetfence
+	packetfence-eap-mac-policy
 	#
 	#  Look in an SQL database.  The schema of the database
 	#  is meant to mirror the "users" file.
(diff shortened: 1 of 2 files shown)


julsemaan:
First, what are you doing working on a beautiful day like this.

Second, we should remove all calls to packetfence.pm except for accounting IMO

louismunro:
If I were to only work on bad days we would never release.

I agree about the removal, but time is short. It will probably have to wait for 6.1.

PR #1388, opened 2016-04-17
Fixed case where we may never receive acct-stop.

louismunro:
# Description

Fixes the case of the missing accounting stop packet.
This now closes all existing sessions if we receive a newer "start" packet before opening the new session.
# Impacts

More accurate accounting.
# Delete branch after merge

YES

### DIFF
--- a/db/upgrade-X.X.X-X.Y.Z.sql
+++ b/db/upgrade-X.X.X-X.Y.Z.sql
@@ -64,6 +64,24 @@ CREATE PROCEDURE acct_start (
     IN p_acctstatustype varchar(25)
 )
 BEGIN
+
+    
+# We make sure there are no left over sessions for which we never received a "stop"
+DECLARE Previous_Session_Time int(12);
+SELECT acctsessiontime
+INTO Previous_Session_Time
+FROM radacct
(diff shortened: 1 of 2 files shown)


db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -64,6 +64,24 @@ CREATE PROCEDURE acct_start (
...
+AND (acctstoptime IS NULL OR acctstoptime = 0);
+
+IF (Previous_Session_Time IS NOT NULL) THEN
+    UPDATE radacct SET
+      acctstoptime = p_timestamp,
+      acctterminatecause = 'UNKNOWN',
+      WHERE acctuniqueid = p_acctuniqueid
+      AND (acctstoptime IS NULL OR acctstoptime = 0);
+END IF;
+
```
jrouzierinverse, line 84: Wouldn't the insert fail since the acctuniqueid already exists?

louismunro, line 84: It's not UNIQUE. Contrary to what the name seems to imply, there is no garantee that it won't be reused by the NAS.

PR #1397, opened 2016-04-19
Support register node by IP address in webservice

meggarr:
# Description

(REQUIRED)
Allow PacketFence to provide Web Service API to register a node by an IP address, this is useful when a captive portal is from a 3rd party, and the client host is proxied or redirect to the 3rd party captive portal.
The captive portal will know the IP address of the client (via X-Fowarded-For or HTTP Request origin), with the IP address, the captive portal can decide on its own strategy to register (or not) the node by IP.
# Impacts

(REQUIRED)
Web Services
# Delete branch after merge

(REQUIRED)
NO
# NEWS file entries

## Enhancements
- Register node via IP address in Web Services

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -566,29 +566,9 @@ sub register_node : Public {
     my ($class, %postdata )  = @_;
     my @require = qw(mac pid);
     my @found = grep {exists $postdata{$_}} @require;
-    return unless pf::util::validate_argv(\@require,  \@found);
+    return (0) unless pf::util::validate_argv(\@require,  \@found);
 
-    pf::node::node_register($postdata{'mac'}, $postdata{'pid'}, %postdata);
-    return;
-}
-
-=head2 register_node_ip
-
-Register a node by ip address
-
-=cut
-
-sub register_node_ip : Public {
-    my ($class, %postdata )  = @_;
-    my @require = qw(ip pid);
-    my @found = grep {exists $postdata{$_}} @require;
-    return unless pf::util::validate_argv(\@require,  \@found);
-
-    my $mac = pf::iplog::ip2mac($postdata{'ip'});
-    die "Cannot find host with IP address $postdata{'ip'}" unless $mac;
-
-    pf::node::node_register($mac, $postdata{'pid'}, %postdata);
-    return;
+    return pf::node::node_register($postdata{'mac'}, $postdata{'pid'}, %postdata);
 }
 
 =head2 deregister_node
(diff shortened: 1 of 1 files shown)


lib/pf/api.pm
```diff
@@ -572,6 +572,25 @@ sub register_node : Public {
...
+    my ($class, %postdata )  = @_;
+    my @require = qw(ip pid);
+    my @found = grep {exists $postdata{$_}} @require;
+    return unless pf::util::validate_argv(\@require,  \@found);
+
+    my $mac = pf::iplog::ip2mac($postdata{'ip'});
+    die "Cannot find host with IP address $postdata{'ip'}" unless $mac;
+
+    pf::node::node_register($mac, $postdata{'pid'}, %postdata);
+    return;
```
julsemaan: return the result of the node_register instead, that will give better feedback : 
`return pf::node::node_register($mac, $postdata{'pid'}, %postdata);`

lib/pf/api.pm
```diff
@@ -572,6 +572,25 @@ sub register_node : Public {
     return;
 }
 
+=head2 register_node_ip
+
+Register a node by ip address
```
julsemaan: *IP address

meggarr: Ok, I can change this one.

lib/pf/api.pm
meggarr: That's ok, do you mean we can change the API `register_node`'s return value as well ?

BTW, do you know why the `pf::node::node_register` returns `(1|0)` instead of boolean directly ?

julsemaan: > That's ok, do you mean we can change the API register_node's return value as well ?

Might as well while you are at it.

> BTW, do you know why the pf::node::node_register returns (1|0) instead of boolean directly ?

Booleans don't exist in perl (at least the version we use) $TRUE and $FALSE are constants pointing to 0 and 1. We just haven't rewriten all of the 0s and 1s to those yet

meggarr: Thanks! I am pretty new to Perl. Frankly, I grabbed some Perl tutorial from Google to hack PF's code.

lib/pf/api.pm
```diff
@@ -566,10 +566,9 @@ sub register_node : Public {
     my ($class, %postdata )  = @_;
     my @require = qw(mac pid);
     my @found = grep {exists $postdata{$_}} @require;
-    return unless pf::util::validate_argv(\@require,  \@found);
+    return (0) unless pf::util::validate_argv(\@require,  \@found);
```
julsemaan: remove the `(0)` from the 4 places you added it since it creates a different behavior than all the other API methods.

Other than that, it will be good to merge

PR #1414, opened 2016-04-21
Fix for corruption of /etc/sysconfig/network & new feature: 'deauthsourceip' in pf.conf

ptrrkssn:
# Description

pfappserver sometimes corrupts /etc/sysconfig/network when it chomp():s on the old content (replaces newlines with space, causing HOSTNAME=xxx to be put on the same line as NETWORKING=yes, sometimes.

Also added a new feature - deauthsourceip=IP in pf.conf allowing you to override the automatic source IP for deauthentication requests when needed. (In our environment sending deauth requests from the MGMT interface IP is wrong).
# Impacts

Probably should put the deauthsourceip option under some other name than "liupf" (our local modifications goes under that name).
# Delete branch after merge

YES
# NEWS file entries
## New Features

deauthsourceip=IP pf.conf option
## Fixes

Corruption of /etc/sysconfig/network

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1824,3 +1824,10 @@ options=enabled|disabled
 description=<<EOT
 Show the parking portal to the device instead of the usual portal
 EOT
+
+[liupf.deauthsourceip]
+type=text
+description=<<EOT
+Source IP address for deauthentication requests.
+EOT
+
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -3130,7 +3130,7 @@ Takes into account the active/active clustering and centralized deauth
 
 =cut
 
-sub deauth_source_ip {
+sub orig_deauth_source_ip {
     my ($self) = @_;
     if($cluster_enabled){
         return isenabled($Config{active_active}{centralized_deauth}) ? pf::cluster::management_cluster_ip() : pf::cluster::current_server->{management_ip};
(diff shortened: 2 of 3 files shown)


conf/documentation.conf
```diff
@@ -1824,3 +1824,10 @@ options=enabled|disabled
 description=<<EOT
 Show the parking portal to the device instead of the usual portal
 EOT
+
+[liupf.deauthsourceip]
```
julsemaan, line 1828: Not sure where liupf comes from :)

It seems it would be better in `advanced`

Also, you need to add the parameter in pf.conf.defaults even if it will be empty

lib/pf/Switch.pm
```diff
@@ -3130,7 +3130,7 @@ Takes into account the active/active clustering and centralized deauth
 
 =cut
 
-sub deauth_source_ip {
+sub orig_deauth_source_ip {
```
julsemaan, line 3133: `orig_deauth_source_ip` doesn't sound exactly right to me.

I would personally merge the two subroutines together or call that one `_deauth_source_ip` or another more meaningful name than orig_...

extrafu:
Closing, no activity in almost one year and no longer relevant with https://github.com/inverse-inc/packetfence/pull/1887

julsemaan:
#1887 already does something similar to chose the deauth IP and this was left to rot

PR #1419, opened 2016-04-22
Feature/coa control

julsemaan:
# Description

Control whether a CoA or RADIUS disconnect should be used when changing a role
# Impacts

Aruba + Avaya network modules
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added control of CoA vs Disconnect when deauthenticating a device.

### DIFF
--- a/lib/pf/Switch/Aruba.pm
+++ b/lib/pf/Switch/Aruba.pm
@@ -497,8 +497,7 @@ sub radiusDisconnect {
         # merging additional attributes provided by caller to the standard attributes
         $attributes_ref = { %$attributes_ref, %$add_attributes_ref };
 
-        # Roles are configured and the user should have one
-        if ( defined($role) && isenabled($self->{_RoleMap}) && isenabled($self->{_useCoA}) ) {
+        if ( $self->shouldUseCoA({role => $role}) ) {
 
             $attributes_ref = {
                 %$attributes_ref,
--- a/lib/pf/Switch/Aruba.pm
+++ b/lib/pf/Switch/Aruba.pm
@@ -498,7 +498,7 @@ sub radiusDisconnect {
         $attributes_ref = { %$attributes_ref, %$add_attributes_ref };
 
         # Roles are configured and the user should have one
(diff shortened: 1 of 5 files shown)


julsemaan:
Please review

lib/pf/Switch/Aruba.pm
```diff
@@ -498,7 +498,7 @@ sub radiusDisconnect {
         $attributes_ref = { %$attributes_ref, %$add_attributes_ref };
 
         # Roles are configured and the user should have one
-        if ( defined($role) && (defined($node_info->{'status'}) && isenabled($self->{_RoleMap}) ) ) {
+        if ( defined($role) && isenabled($self->{_RoleMap}) && isenabled($self->{_useCoA}) ) {
```
jrouzierinverse: This would be best wrapped in a function.

julsemaan: Here we say vrapped.

But I can still agree with you

julsemaan:
Re-review

lib/pf/Switch/Aruba.pm
jrouzierinverse: vrapped?

julsemaan: Forget about it :) just another joke out of my bad jokes bucket (which happen to contain all my jokes)

lib/pf/Switch/Avaya.pm
```diff
@@ -589,8 +589,7 @@ sub radiusDisconnect {
         # merging additional attributes provided by caller to the standard attributes
         $attributes_ref = { %$attributes_ref, %$add_attributes_ref };
 
-        # Roles are configured and the user should have one
-        if (defined($role) && (defined($node_info->{'status'}) ) ) {
+        if ( $self->shouldUseCoA({role => $role}) ) {
```
jrouzierinverse, line 592: This did not previously check for _RoleMap.
Was this a bug in the old code?

julsemaan, line 592: That is correct. It was likely to have been forgotten when RoleMap was brought in.
It should use the same logic as the other modules

PR #1424, opened 2016-04-22
Fix/fr cluster clients

### LINKED ISSUE #1245: Generate raddb/clients.conf.inc based on the cluster members instead of a range
If a new switch is part of the management network range, it will not be used even if it is there in the radius_nas table.

This doesn't seem to appear for every switch.

Defining the whole range is also a small security threat as a client out of the pf cluster could start doing requests

julsemaan:
# Description

Declare cluster members individually in cluster clients
Change generic cluster secret to an on-install generated one.
# Impacts

FR clustering
Local test scripts
# Issue

fixes #1245 
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes

- Declare cluster members individually in cluster clients (#1245)
- Change generic cluster secret to an on-install generated one.
# UPGRADE file entries

The management IP(s) of PacketFence are now defined as switches with a forced RADIUS secret defined in `/usr/local/pf/conf/local_secret`. Make sure you reconfigure the secret in the file if necessary and that this file is synchronized on all your cluster members if that applies. Note that this doesn't affect the RADIUS secret you have configured for wireless controllers and switches. It only affects RADIUS requests that originate from the management IP(s)

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -649,6 +649,11 @@ if [ ! -f /usr/local/pf/conf/pf_omapi_key ]; then
     /usr/bin/openssl rand -base64 -out /usr/local/pf/conf/pf_omapi_key 32
 fi
 
+# Create server local RADIUS secret
+if [ ! -f /usr/local/pf/conf/local_secret ]; then
+    date +%s | sha256sum | base64 | head -c 32 > /usr/local/pf/conf/local_secret
+fi
+
 for service in snortd httpd snmptrapd portreserve redis
 do
 %if 0%{?el6}
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,9 @@ conf/ssl/server.crt:
 conf/pf_omapi_key:
 	/usr/bin/openssl rand -base64 -out /usr/local/pf/conf/pf_omapi_key 32
 
+conf/local_secret:
+	date +%s | sha256sum | base64 | head -c 32 > /usr/local/pf/conf/local_secret
+
 bin/pfcmd: src/pfcmd.c
 	$(CC) -O2 -g -std=c99  -Wall $< -o $@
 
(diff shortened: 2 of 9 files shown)


julsemaan:
Please review

addons/packages/packetfence.spec
```diff
@@ -649,6 +649,11 @@ if [ ! -f /usr/local/pf/conf/pf_omapi_key ]; then
     /usr/bin/openssl rand -base64 -out /usr/local/pf/conf/pf_omapi_key 32
 fi
 
+# Create server local RADIUS secret
+if [ ! -f /usr/local/pf/conf/local_secret ]; then
+    date +%s | sha256sum | base64 | head -c 32 > /usr/local/pf/conf/local_secret
```
louismunro, line 654: Can you make that file chmod 600?
Perhaps by setting the umask just prior to creating the file or else setting the mode explicitly.

julsemaan, line 654: Question is, should we really ?

pf_omapi_key doesn't do it this way so if I would change this one, pf_omapi_key needs to be changed as well to be consistent.

I personally don't see any problem having this using the default rights even though it is a key/password.

pf.conf + switches.conf contain that same type of info in clear text and they use the default rights.

Just playing the devils advocate so as even if you only have a semi-valid argument I'll change it :)

louismunro, line 654: Mmm, fair point.
We might have to review the whole permissions scheme.
I'll let this one slide for now.
We'll revisit the issue.

louismunro:
I especially like "enhancement 2" above.
Seems like a killer selling point.

julsemaan:
I just realized that I added it to the news entry but that the 'enhancement 2' is not yet fully ready.

Hope we don't lose too much market share because of the delays of that feature

PR #1428, opened 2016-04-25
Feature/services status cluster

### LINKED ISSUE #714: Show services status for all servers in active/active cluster
/admin/status#service is showing only 1 server.

Either we should make it so we can select the server or merge the servers into different columns

It may be worth it to have a 'Cluster status' page where you see the services and communication status of the servers

julsemaan:
# Description

ie: Allow PacketFence to trigger the coffee machine on wireless EAP 802.1X connection to brew a fresh espresso.
# Impacts

Show service status for all cluster members
# Issue

fixes #714 
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added page to visualize the status of the services on all cluster members

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -46,6 +46,7 @@ use pf::file_paths qw($captiveportal_profile_templates_path);
 use pf::CHI;
 use pf::access_filter::dhcp;
 use pf::metascan();
+use pf::services();
 
 use List::MoreUtils qw(uniq);
 use List::Util qw(pairmap);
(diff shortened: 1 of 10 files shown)


julsemaan:
Please review

lib/pf/api.pm
```diff
@@ -1313,6 +1314,21 @@ sub handle_accounting_metadata : Public {
 
 }
 
+sub services_status : Public {
```
jrouzierinverse, line 1323: documentation

julsemaan:
re-review

lib/pf/api.pm
```diff
@@ -1313,6 +1314,27 @@ sub handle_accounting_metadata : Public {
...
+
+=cut
+
+sub services_status : Public {
+    my ($class, $services) = @_;
+    my @managers = pf::services::getManagers($services);
+
+    my $statuses = {};
+    foreach my $manager (@managers){
+        my $isManaged = $manager->isManaged;
```
jrouzierinverse: Unused variable

lib/pf/api.pm
```diff
@@ -1313,6 +1314,27 @@ sub handle_accounting_metadata : Public {
...
+        my $status = $manager->status;
```
jrouzierinverse: This should be done within the if below.
Since we do not use any where else

julsemaan:
adjustments done, good for another review

PR #1432, opened 2016-04-26
change permenant error to temporary in radius audit log

### LINKED ISSUE #1141: Radius Log: Error message doesn't fade away.
If you don't put the start time and/or the end time, you will get an error message but they never seem to disappear, I need to remove them by clicking the cross.

### LINKED ISSUE #1210: RADIUS Audit Log error message
When you do not put a starting date to search some entries in the RADIUS audit log, the message 'Start Date field is required' appear.

If you set a starting date and press search again, the message does not disappear.

whitx:
# Description

Make error message temporary to not have them stay after refreshing a search in the radius audit log
# Impacts

JS, error message in the radius audit log
# Issue

fixes #1210 
fixes #1141
# Delete branch after merge

NO
# NEWS file entries
## Bug Fixes
- Issue with Error message in RadiusAuditLog(#1210)

### DIFF
--- a/html/pfappserver/root/radiuslog/search.tt
+++ b/html/pfappserver/root/radiuslog/search.tt
@@ -8,7 +8,7 @@
       <!--th[% " class=\"$class\"" IF class %]><a href="[% c.uri_for(search_action,page_num,per_page,params)  %]">[% title | html %][% IF by == column %] <i class="icon-sort-[% IF new_direction == 'desc' %]down[% ELSE %]up[% END %]"></i>[% END %]</a></th-->
       <th[% " class=\"$class\"" IF class %]>[% title | html %][% IF by == column %] <i class="icon-sort-[% IF new_direction == 'desc' %]down[% ELSE %]up[% END %]"></i>[% END %]</th>
 [% END -%]
-      <table id="RadiusAuditLog" class="table table-condensed table-striped items">
+      <table id="radiusAuditLog" class="table table-condensed table-striped items">
            <thead>
              <tr>
                <td colspan="[% display_columns.size + 1%]">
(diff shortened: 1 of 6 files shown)


whitx:
Is anyone against the fact that we merge this? 
All error messages affected by items.js will be fading out after few seconds (~10) and won't be permanent.

julsemaan:
It is a big behavior change IMO.

Usually, we clear the error messages when doing a new query (ex : https://github.com/inverse-inc/packetfence/blob/devel/html/pfappserver/root/static/admin/config/violations.js#L184)

That is IMO a better solution than making the error not permanent.
Let's say it takes forever to save my config due to a cluster member being down (without me knowing) and I decide to go to ZammitBucks (ZammitCorp's overpriced coffee company), then when I come back, I won't see any error because it faded away and may assume it worked in the case of saving a config section (which your code applies to as well)

So no, doesn't seem like the right thing to do. We should find a general way to reset the error when doing a query (whether it is saving a config or doing another RADIUS audit log search)

whitx:
Re-review please

html/pfappserver/root/radiuslog/search.tt
```diff
@@ -8,7 +8,7 @@
       <!--th[% " class=\"$class\"" IF class %]><a href="[% c.uri_for(search_action,page_num,per_page,params)  %]">[% title | html %][% IF by == column %] <i class="icon-sort-[% IF new_direction == 'desc' %]down[% ELSE %]up[% END %]"></i>[% END %]</a></th-->
       <th[% " class=\"$class\"" IF class %]>[% title | html %][% IF by == column %] <i class="icon-sort-[% IF new_direction == 'desc' %]down[% ELSE %]up[% END %]"></i>[% END %]</th>
 [% END -%]
-      <table id="items" class="table table-condensed table-striped items">
+      <table id="RadiusAuditLog" class="table table-condensed table-striped items">
```
julsemaan: HTML ids start lowercase in our code convention from what I know.

Change this to radiusAuditLog here and in radiusauditlog.js#14

julsemaan:
Other than the ID detail, this works so will merge when it gets fixed

whitx:
fixed

PR #1453, opened 2016-05-02
Allow the pcap packet size in pfdhcplistener to be configurable

jrouzierinverse:
## Description

Allow the pcap packet size in pfdhcplistener to be configurable
## Impacts

pfdhcplistener

Enhancements
++++++++++++
- Allow the pcap packet size in pfdhcplistener to be configurable
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/pfcmd/checkup.pm
+++ b/lib/pf/pfcmd/checkup.pm
@@ -173,6 +173,7 @@ sub sanity_check {
     db_check_version();
     valid_certs();
     portal_modules();
+    advanced_config();
 
     return @problems;
 }
@@ -1294,6 +1295,18 @@ sub cert_has_expired {
     my $cert = Crypt::OpenSSL::X509->new_from_file($path);
     my $expiration = str2time($cert->notAfter);
     return time > $expiration;
+}
+
+=item advanced_config
+
+Check the advanced config section
+
+=cut
+
+sub advanced_config {
+    if ($Config{advanced}{pfdhcplistener_packet_size} < 576) {
+        add_problem($FATAL, "advanced.pfdhcplistener_packet_size is set to a value less than 576");
+    }
 }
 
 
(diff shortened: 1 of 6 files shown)


sbin/pfdhcplistener
```diff
@@ -169,7 +169,7 @@ sub dhcp_detector {
     # updating process name so we know what interface we are listening on
     # WARNING: the format is expected by watchdog in pf::services. Don't change lightly.
     $PROGRAM_NAME = basename($PROGRAM_NAME) . ": listening on $interface";
-    $pcap = Net::Pcap::pcap_open_live( $interface, 576, 1, 0, \$err );
+    $pcap = Net::Pcap::pcap_open_live( $interface, $Config{advanced}{pfdhcplistener_packet_size} || 576, 1, 0, \$err );
```
julsemaan: No need for `||` since we already defined it in pf.conf.defaults

rouzier: The was more of a check to make sure that someone did not put 0.
But you are right we should just take it from pf.conf.defaults.
I will add a check in checkup.pm to not allow a size less than 576

julsemaan: Even better

jrouzierinverse:
Please review

lib/pf/pfcmd/checkup.pm
```diff
@@ -1296,6 +1297,18 @@ sub cert_has_expired {
     return time > $expiration;
 }
 
+=item advanced_config
+
+Check the advanced config section
+
+=cut
+
+sub advanced_config {
```
julsemaan: wouldn't it be better to check that all numeric fields defining a minimum are respecting it.

Would be less of a case by case and make it cleaner

jrouzierinverse: True

jrouzierinverse:
Please re-review

PR #1457, opened 2016-05-04
Support challenge response workflow in authentication

jrouzierinverse:
## Description

Support challenge response workflow in authentication
## Impacts
- RADIUS authorize workflow
- Admin authentication
- Portal authentication
## NEWS file entries

New Features
++++++++++++
- Support Challenge Response authentication 
## Delete branch after merge

NO

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
@@ -35,10 +35,22 @@ has 'challenge_template' => (is => 'ro', default => sub { "challenge.html" } );
 
 has 'challenge_data' => (is => 'rw', builder => '_build_challenge_data', lazy => 1, trigger => \&_trigger_challenge_data);
 
+=head2 _build_challenge_data
+
+Get the challenge data from the session cache
+
+=cut
+
 sub _build_challenge_data {
     my ($self) = @_;
     return $self->app->session->{challenge_data};
 }
+
+=head2 _trigger_challenge_data
+
+Set the challenge data in the session cache
+
+=cut
 
 sub _trigger_challenge_data {
     my ($self, $data, $old_data) = @_;
@@ -192,16 +204,22 @@ sub authenticate {
     $self->done();
 }
 
+=head2 challenge
+
+Handle the challenge request
+
+=cut
+
 sub challenge {
     my ($self) = @_;
     my $password = $self->request_fields->{password};
-    my ($results, $message) = $self->source->challenge($self->username, $password, $self->challenge_data);
-    if ($results == $LOGIN_FAILURE) {
+    my ($result, $message) = $self->source->challenge($self->username, $password, $self->challenge_data);
+    if ($result == $LOGIN_FAILURE) {
         $self->app->flash->{error} = $message;
         $self->display_challenge();
         return;
     }
-    if ($results == $LOGIN_CHALLENGE) {
+    if ($result == $LOGIN_CHALLENGE) {
         $self->challenge_data($message);
         $self->display_challenge();
         return;
@@ -210,6 +228,12 @@ sub challenge {
     $self->done();
 
 }
+
+=head2 display_challenge
+
+Display the challenge form
+
+=cut
 
 sub display_challenge {
     my ($self, $args) = @_;
@@ -222,6 +246,11 @@ sub display_challenge {
         %{$args},
     });
 }
+
+
+=head2 allowed_urls_auth_module
+
+=cut
 
 sub allowed_urls_auth_module { ['/challenge'] }
 
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
@@ -43,7 +43,7 @@ Get the challenge data from the session cache
 
 sub _build_challenge_data {
     my ($self) = @_;
-    return $self->app->session->{challenge_data};
+    return $self->session->{challenge_data};
 }
 
 =head2 _trigger_challenge_data
@@ -54,7 +54,7 @@ Set the challenge data in the session cache
 
 sub _trigger_challenge_data {
     my ($self, $data, $old_data) = @_;
-    $self->app->session->{challenge_data} = $data;
+    $self->session->{challenge_data} = $data;
 }
 
 sub sources_classes {
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
@@ -21,7 +21,7 @@ use pf::config::util;
 use List::MoreUtils qw(all);
 use pf::auth_log;
 use pf::person;
-use pf::Authentication::constants;
+use pf::Authentication::constants qw($LOGIN_SUCCESS $LOGIN_FAILURE $LOGIN_CHALLENGE);
 use pf::web::guest;
 use pf::node qw(node_view);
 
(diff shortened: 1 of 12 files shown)


whitx:
tested on CenOS7, works great, package have been added for centos6/7 need to be done for debian

whitx:
Package added for wheezy, already existing for jessie

jrouzierinverse:
Please review

whitx:
This has been tested at CUSM with their OTP service and is working properly.

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
```diff
@@ -30,8 +30,22 @@ has '+sources' => (isa => 'ArrayRef['.join('|', @{sources_classes()}).']');
 
 has '+multi_source_object_classes' => (default => sub{sources_classes()});
 
+has 'challenge_template' => (is => 'ro', default => sub { "challenge.html" } );
+
+has 'challenge_data' => (is => 'rw', builder => '_build_challenge_data', lazy => 1, trigger => \&_trigger_challenge_data);
+
+sub _build_challenge_data {
```
julsemaan, line 44: Missing documenation headers for the new subs in the file

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
```diff
@@ -30,8 +30,22 @@ has '+sources' => (isa => 'ArrayRef['.join('|', @{sources_classes()}).']');
...
+    my ($self) = @_;
+    return $self->app->session->{challenge_data};
```
julsemaan: Would be better to use the module session IMO. This way, if there is a chain of logins, it will not need to be cleared.

`$self->session->{challenge_data}`

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
```diff
@@ -127,19 +162,50 @@ sub authenticate {
...
+
     $self->update_person_from_fields();
-    $self->username($username);
     $self->done();
 }
 
+sub challenge {
+    my ($self) = @_;
+    my $password = $self->request_fields->{password};
+    my ($results, $message) = $self->source->challenge($self->username, $password, $self->challenge_data);
```
julsemaan: There is only one result so it would be `$result`

html/pfappserver/lib/pfappserver/PacketFence/Controller/Admin.pm
```diff
@@ -111,10 +133,48 @@ sub login :Local :Args(0) {
...
     } elsif ($c->user_in_realm( 'admin' )) {
-        $c->response->redirect($c->uri_for($c->controller->action_for('index')));
+        $c->response->redirect($c->uri_for($self->action_for('index')));
+        $c->detach();
+    } elsif ($req->params->{'redirect_action'}) {
+        $c->stash->{redirect_action} = $req->params->{'redirect_action'};
+    }
+}
+
+sub challenge :Local :Args(0) {
```
julsemaan, line 149: missing doc

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -46,33 +47,90 @@ sub available_attributes {
...
+    if (!defined $radius) {
+        $logger->error("Unable to perform RADIUS authentication on any server: " . Authen::Radius::get_error());
+        return ($FALSE, $COMMUNICATION_ERROR_MSG);
+    }
+    my $result = $self->check_radius_password($radius, $username, $password);
+    return $self->_handle_radius_request($radius, $result);
+}
 
-     if ($radius->get_error() eq 'ENONE') {
+sub challenge {
```
julsemaan, line 70: missing doc header in this sub and the other ones in the file

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -46,33 +47,90 @@ sub available_attributes {
...
+    my ($self, $username, $password, $challenge_data) = @_;
+    my $attribute = {
```
julsemaan, line 72: Seems this should be `$attributes` since there are multiple ones.

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -46,33 +47,90 @@ sub available_attributes {
...
+    }
+    my $result = $self->check_radius_password($radius, $username, $password, undef, $attribute);
+    return $self->_handle_radius_request($radius, $result);
+}
 
-       if ($result) {
+sub _handle_radius_request {
+    my ($self, $radius, $result) = @_;
+    my $logger = get_logger();
+    if ($radius->get_error() ne 'ENONE') {
```
julsemaan: ENONE should be a constant

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -46,33 +47,90 @@ sub available_attributes {
...
+    }
+    return ($FALSE, $AUTH_FAIL_MSG);
+}
 
-   return ($FALSE, $COMMUNICATION_ERROR_MSG);
- }
+sub _make_challenge_data {
+    my ($self, $result, $radius) = @_;
+    my @attributes = $radius->get_attributes;
+    my ($state_attribute) = grep { $_->{Name} eq 'State'} @attributes;
```
julsemaan: State should use a constant

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -46,33 +47,90 @@ sub available_attributes {
...
+    my ($message_attribute) = grep { $_->{Name} eq 'Reply-Message'} @attributes;
```
julsemaan: Reply-Message should use a constant

lib/pf/Authentication/Source/RADIUSSource.pm
rouzier, line 72: This is only one attribute

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
jrouzierinverse: What is the difference?

lib/pf/Authentication/Source/RADIUSSource.pm
julsemaan, line 72: My bad you are right

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
jrouzierinverse: Agreed

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
julsemaan: So example:

I have a chained login:
- First is ZammitCorp RADIUS login - uses module zammitcorp_login - a RADIUS login with challenge
- Second is JamaisseInc LDAP login - uses module bug_killer_3000

If you put the challenge in the global session then when you login on ZammitCorp it will put the challenge data in the global session meaning the JamaisseInc module will try to use it and go through the challenge code path.

If you put it in the module session, then the challenge data of ZammitCorp stays encapsulated into the module and the JamaisseInc authentication won't see it and won't go through the challenge codepath.

So the session of the module was made for that reason, encapsulating the state of the module into itself so data doesn't clash between modules.

jrouzierinverse:
Please Review

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm
```diff
@@ -62,10 +88,24 @@ Execute this module
 sub execute_child {
     my ($self) = @_;
     if($self->app->request->method eq "POST"){
-        $self->authenticate();
+        if($self->app->request->path eq "challenge") {
```
julsemaan, line 91: One detail.. I don't see that path defined in lib/pf/web/constants.pm
To my knowledge, it won't work without it in the Apache captive portal

jrouzierinverse, line 91: You are correct let me fix that

rouzier:
Done

PR #1473, opened 2016-05-12
Move to use the mod_proxy_html for the portal preview

jrouzierinverse:
## Description

Move to use the mod_proxy_html for the portal preview
## Impacts

The Admin GUI preview portal functionality
## NEWS file entries

Bug Fixes
+++++++++
- Fix some edge cases when proxying the portal through the admin

Dependencies
+++++++++
mod_proxy_html
## Delete branch after merge

NO


fdurand:
ProxyHTMLLogVerbose and RewriteLog doesn't exist with apache 2.4

jrouzierinverse:
Done

dwlfrth:
Still seems a bit buggy here and there.
- Return button (top-right corner) is not working properly
- Sometimes getting the 'default' flow rather than the one we want to preview

jrouzierinverse:
Can you provide instructions on how to replicate where the default flow 
is being used instead of the  one we want to preview.

James Rouzier

lzammit:
Portal profile with custom CSS in layout.html doesn't display it on the portal preview but works on the registration VLAN.

dwlfrth:
Now proxying using the MAC address of the device we are previewing with...
Which means, if the device I am accessing the admin interface to preview the portal profiles is registered, I will hit the "Sorry your network is being enabled ..."
We should use a "proxied mac address / ip address" which will not be part of the 'nodes' and for which we will not keep a session.

addons/packages/packetfence.spec
```diff
@@ -88,7 +88,7 @@ Requires: chkconfig, coreutils, grep, openssl, sed, tar, wget, gettext, conntrac
 Requires: procps
 Requires: libpcap, libxml2, zlib, zlib-devel, glibc-common,
 Requires: httpd, mod_ssl
-Requires: mod_perl, mod_qos, mod_evasive
+Requires: mod_perl, mod_qos, mod_evasive, mod_proxy_html
```
fdurand, line 91: What about debian ?

julsemaan, line 91: You know what they say about Debian...

fdurand:
Still missing libapache2-mod-proxy-html in debian packaging

jrouzierinverse:
It is already there.

dwlfrth:
mod_proxy_html is missing from PacketFence repositories

PR #1482, opened 2016-05-18
Fix/graphite hostname and SSL URLs

### LINKED ISSUE #658: Graphite provided images should be requested over https.
Currently they are http while the dashboard is https.

louismunro:
# Description

Enforce SSL access to graphite URLs in the admin and sets the URL to point to hostname instead of IP.
# Impacts

The httpd.graphite config will now require SSL.
# Issue

Fixes #658 
# Delete branch after merge

YES 
# NEWS file entries
## Bug Fixes
- Graphite URLs are now requested over HTTPS (#658)


html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
```diff
@@ -296,7 +296,7 @@ sub _buildGraphiteURL :Private {
 
     my $options =
       {
-       graphite_host => $management_ip,
+       graphite_host => $Config{'general'}{'hostname'} . '.' .  $Config{'general'}{'domain'},
```
fdurand: If the dns entry is not set correctly then it will not work. There is probably an env variable in catalyst with the host value.

louismunro:
I see a bad merge rising.
I see trouble on the way.
Will reopen.

PR #1488, opened 2016-05-24
add condition for mariadb

whitx:
# Description

Add a check to verify if we run MariaDB or MySQL
# Impacts

database-backup-and-maintenance script
# Delete branch after merge

NO
# NEWS file entries
## Enhancements
- Add SQL engine detection for database maintenance script

### DIFF
--- a/addons/database-backup-and-maintenance.sh
+++ b/addons/database-backup-and-maintenance.sh
@@ -57,9 +57,16 @@ else
     echo -e $BACKUP_DIRECTORY$BACKUP_PF_FILENAME ", file already created. \n"
 fi
 
+# check if MySQL or MariaDB is installed
+$SQL_ENGINE=''
+if mysql -V | grep "MariaDB"; then
+    $SQL_ENGINE='mariadb'
+else
+    $SQL_ENGINE="mysql"
+fi
 
 # is MySQL running? meaning we are the live packetfence
-if [ -f /var/run/mysqld/mysqld.pid ]; then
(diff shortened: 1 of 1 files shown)


addons/database-backup-and-maintenance.sh
```diff
@@ -57,9 +58,15 @@ else
     echo -e $BACKUP_DIRECTORY$BACKUP_PF_FILENAME ", file already created. \n"
 fi
 
+# check if MySQL or MariaDB is installed
+if mysql -V | grep "MariaDB"; then
```
louismunro: use grep -q if possible.

addons/database-backup-and-maintenance.sh
```diff
@@ -57,9 +58,15 @@ else
...
+    $SQL_ENGINE='mariadb'
```
louismunro: This is a bug. 
It does not do what you think it does.
Shell variable assignment must not be prefixed with $.

addons/database-backup-and-maintenance.sh
```diff
@@ -29,6 +29,7 @@ BACKUP_PF_FILENAME='packetfence-files-dump'
 ARCHIVE_DIRECTORY=$BACKUP_DIRECTORY
 ARCHIVE_DB_FILENAME='packetfence-archive'
 PERCONA_XTRABACKUP_INSTALLED=0
+SQL_ENGINE=''
```
louismunro: This is not really necessary.

PR #1496, opened 2016-05-28
Delete environment variable `LANGUAGE` when initializing i18n

tribut:
# Description

From `man 3 gettext`:

```
If the LANGUAGE environment variable is set to a nonempty  value,
and the  locale  is not the "C" locale, the value of LANGUAGE is
assumed to contain a colon separated list of  locale  names.  The
functions will attempt  to  look  up  a translation of msgid in
each of the locales in turn. This is a GNU extension.
```

Therefore, we have to make sure `$LANGUAGE` is empty, otherwise it will override our setting for `LC_MESSAGES`. On Debian, the Installer sets this in `/etc/default/locale` when a different language and locale are selected (such as `english` and `de_DE.UTF-8`).
# NEWS file entries
## Bug Fixes
- Localization now works even if the LANGUAGE environment variable is set


louismunro:
Seems like a simple fix.
I will schedule it for testing to be included in 6.1.

PR #1504, opened 2016-05-31
Fix/destination url computing

julsemaan:
# Description

Adds a javascript snippet that ensures the destination URL is computed from a javascript capable browser (thus one used by a human) and not by an application doing an API call
# Impacts

Destination URL in the portal
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix the bad handling of the destination URL that would sometimes be the one of a background application


julsemaan:
confirmed working, merging and putting in maintenance

PR #1506, opened 2016-05-31
Use correct constant in CURLOPT_HTTPAUTH

tribut:
# Description

Packetfence never actually sent username/password with the jsonrpcclient class, because a wrong flag was used. This patch fixes that problem.
# NEWS file entries
## Bug Fixes
- Send username/password in RPC requests when https is used

### DIFF
--- a/lib/pf/api/jsonrpcclient.pm
+++ b/lib/pf/api/jsonrpcclient.pm
@@ -183,7 +183,7 @@ sub curl {
             $curl->setopt(CURLOPT_PASSWORD, $self->password);
         }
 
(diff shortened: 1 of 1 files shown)


julsemaan:
Its weird how it always worked :/ with that option with our server implementation 

I'll test this against our API and if it works, I'll check the code and merge

julsemaan:
Works fine.

Looked at the curl doc, and most likely the reason why it always worked is that the default value of `CURLOPT_HTTPAUTH` is `CURLAUTH_BASIC`. I think you run debian, thus a different libcurl than most of us on CentOS so that might be the reason why there may be a different behavior when setting the option to an invalid value.

Still, better setting it to what we want than to an invalid value :)

Merging.

PR #1507, opened 2016-05-31
Add Firewall/SSO type "JSON-RPC"

tribut:
# Description

This adds a rather generic json-rpc client for firewall/SSO.

We were looking to integrate packetfence with our linux firewall. Unfortunately there is no "generic" firewall/SSO type yet and I wanted something as simple as possible.

The class works well with the equally trivial server from https://github.com/tribut/ipset-rpcd

I would be happy to write some documentation/integration guide. Just let me know if you generally consider merging something like this.
# Code / PR Dependencies

Needs https://github.com/inverse-inc/packetfence/pull/1506 for http auth to work.
# NEWS file entries
## New Features
- New Firewall/SSO type: JSON-RPC

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JSONRPC.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JSONRPC.pm
@@ -1,0 +1,119 @@
+package pfappserver::Form::Config::Firewall_SSO::JSONRPC;
+
+=head1 NAME
+
+pfappserver::Form::Config::Firewall_SSO::JSONRPC - Web form for a JSON-RPC device
+
+=head1 DESCRIPTION
+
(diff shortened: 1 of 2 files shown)


julsemaan:
Looks like a good idea although I doubt it will be used widely.

Still, IMO, that would be a merge-able feature, unless someone in the team would have an opposite opinion

louismunro:
@julsemaan can you review it and decide whether this should be in 6.1 or 6.2?

julsemaan:
I'd say 6.1 is already fairly packed up.

Also, this would need a bit of doc and examples

louismunro:
Moved it to 6.2.

julsemaan:
@tribut 

We'll need a bit of documentation around this so its integrated inside PacketFence

julsemaan:
Closing this as it hasn't got any updates

Please update according to the comments and re-open

PR #1508, opened 2016-05-31
Admin allowed node roles

jrouzierinverse:
## Description

Provide the ability to restrict the modification of nodes based off roles
## Impacts

The admin GUI
## NEWS file entries

New Features
++++++++++++
- Provide the ability to Restrict the modification of node roles
## Delete branch after merge

NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Node.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Node.pm
@@ -16,6 +16,7 @@ with 'pfappserver::Base::Form::Role::AllowedOptions';
 
 use HTTP::Status qw(is_error);
 use pf::config qw(%Config);
+use pf::log;
 
 # Form select options
 has 'roles' => ( is => 'ro' );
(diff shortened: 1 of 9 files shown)


rouzier:
Please Review

html/pfappserver/lib/pfappserver/Form/Node.pm
```diff
@@ -190,6 +195,67 @@ sub validate {
...
+    my ($self) = @_;
+    my $init_object = $self->init_object;
+    return undef unless defined $init_object;
+    my $role = $self->init_object->{category};
+    return undef unless defined $role;
+    my %allowed_node_roles = map {$_ => undef} $self->_get_allowed_options('allowed_node_roles');
+    return
+        keys %allowed_node_roles == 0     ? undef
+      : exists $allowed_node_roles{$role} ? undef
+      :                                     1;
```
julsemaan, line 256: Seems to be a bit too much logic for the ternary operator

julsemaan:
Seems quite big for a last minute addition.

Will require a good review + some good testing love

Would be in favor of moving it to 6.2 since 6.1 is getting quite fat

louismunro:
Agreed. Moved to 6.2.

html/pfappserver/lib/pfappserver/Form/Node.pm
rouzier, line 256: If it is not clear what is going then it make sense to change it

PR #1509, opened 2016-05-31
Implement configuration of autoregistration from the profiles.

louismunro:
# Description

Allows autoregistration to be enabled from the GUI, per profile.
# Impacts

No longer need a vlan filter or custom code to autoregister.
Even Mamadou can do it.
This seems too easy. I must be missing something.
# Issue

Fixes my annoyance at not having done this sooner
# Would like to thank

@fdurand for the radius.pm refactor which allowed this.
My mother, the academy.
# Delete branch after merge

YES 
# NEWS file entries

## New Features
- Automatic registration of devices is now configurable from the GUI on a per profile basis.

### DIFF
--- a/lib/pf/role.pm
+++ b/lib/pf/role.pm
@@ -378,17 +378,8 @@ sub getRegisteredRole {
     my ($self, $args) = @_;
     my $logger = $self->logger;
 
-    my $options = {};
-    $options->{'last_connection_type'} = connection_type_to_str($args->{'connection_type'}) if (defined( $args->{'connection_type'}));
-    $options->{'last_switch'}          = $args->{'switch'}->{_id} if (defined($args->{'switch'}->{_id}));
-    $options->{'last_port'}            = $args->{'ifIndex'} if (defined($args->{'ifIndex'}));
-    $options->{'last_ssid'}            = $args->{'ssid'} if (defined($args->{'ssid'}));
-    $options->{'last_dot1x_username'}  = $args->{'user_name'} if (defined($args->{'user_name'}));
-    $options->{'realm'}                = $args->{'realm'} if (defined($args->{'realm'}));
-
-    my $profile = pf::Portal::ProfileFactory->instantiate($args->{'mac'},$options);
-
     my ($vlan, $role, $result, $person, $source, $portal);
+    my $profile = $args->{'profile'};
 
     if (defined($args->{'node_info'}->{'pid'})) {
         $person = pf::person::person_view_simple($args->{'node_info'}->{'pid'});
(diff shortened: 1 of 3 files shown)


lib/pf/role.pm
```diff
@@ -678,6 +659,10 @@ sub shouldAutoRegister {
             return $role;
         }
     }
+    if ($args->{'profile'}->{'_autoregister'} eq "enabled") {
```
julsemaan: image

lib/pf/role.pm
```diff
@@ -678,6 +659,10 @@ sub shouldAutoRegister {
             return $role;
         }
     }
+    if ($args->{'profile'}->{'_autoregister'} eq "enabled") { 
+        $logger->debug("Autoregistration set on profile " . $args->{'profile'}->getName() );
+        return 1;
```
julsemaan: image

julsemaan:
Other than the two details, this looks good to merge

lib/pf/role.pm
louismunro: Well, the rest of the sub doesn't use it either...
Consistency or rewrite?

lib/pf/role.pm
louismunro: I find it butt ugly to be honest.

julsemaan: Still better than a hardcoded string :P

lib/pf/role.pm
julsemaan: Well I'd say to use $TRUE even if it is inconsistent with the rest of the sub and move to them as we rewrite code.

Or if you feel like it rewrite the 0s and 1s to the constant

lib/pf/role.pm
louismunro: I would rather be able to just write 
if ($args->{'profile'}->{'_autoregister'}) { do stuff; } 
but can't right now because it's either set to 'disabled' or 'enabled' which are both true.

lib/pf/role.pm
louismunro: I'll revrite ze bittes.

lib/pf/role.pm
julsemaan: Fully agree with that wish, but I'd say for now it has to be `isenabled($args->{'profile'}->{'_autoregister'})` so we stay consistent in our code

louismunro: :sick: allright, I'll do it.
Still uglier than a pigs ass though...

louismunro:
Changes done. 
Review, season to taste and merge.

PR #1518, opened 2016-06-03
Fix/provisioning pending release

julsemaan:
# Description

Fix the provisioning release logic when being in pending (it should reuse the role + unregdate of the device)
Also adds detaching capability to the dynamic routing
# Impacts

Portal release
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes

- Fixed release process when re-insuring compliance with a provisioner

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
@@ -252,7 +252,7 @@ sub apply_new_node_info {
     }
     else {
         $self->app->error("Couldn't register your device. Please contact your local support staff.");
-        $self->detach()
+        $self->detach();
     }
 }
 
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
@@ -28,7 +28,7 @@ has '+route_map' => (default => sub {
 
 use pf::log;
 use pf::node;
-use pf::config;
+use pf::config qw($default_pid);
 use pf::constants qw($TRUE $FALSE);
 use pf::util;
 use pf::violation;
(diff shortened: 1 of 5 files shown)


julsemaan:
Please review   0d6fdeb and c04b8f8

For some reason github refuse to see that the other 3 commits are already in devel...

sbin/pfmon
```diff
@@ -408,18 +409,24 @@ sub _runtask {
...
         }
+        my $final_interval = $time_taken > 0 ? $interval - int($time_taken+1) : $interval;
+        if($final_interval < 0) {
+            $final_interval = 0;
+        }
         # run these functions every $maintenance_interval
-        $logger->trace("$id is sleeping for $interval");
-        alarm $interval;
+        $logger->trace("$id is sleeping for $final_interval ($interval from configuration - $time_taken)");
+        alarm $final_interval;
```
jrouzierinverse: only call the alarm pause combo if the final_interval is greater than zero.
Since this will cause the process to pause until it gets a signal.

jrouzierinverse: More details.
If the alarm value is 0
Then no alarm is scheduled.
This will cause pause to stay paused until the process gets a signal.

sbin/pfmon
```diff
@@ -408,18 +409,24 @@ sub _runtask {
...
         pause;
         last unless $running;
         $logger->trace("$id is running");
         eval {
-            #reload all cached configs before running the task
+            my $start = Time::HiRes::gettimeofday;
             $task->();
+            $time_taken = Time::HiRes::gettimeofday - $start;
```
jrouzierinverse: Move the calculation of the time taken outside of the eval to take in consideration partial failures.

sbin/pfmon
```diff
@@ -408,18 +409,24 @@ sub _runtask {
             pause;
```
jrouzierinverse, line 410: reset time_taken in this section.

julsemaan, line 410: There is no need to reset it since it is not used until it is set to the new value, and it will be set unconditionally

sbin/pfmon
```diff
@@ -410,24 +410,22 @@ sub _runtask {
...
         my $final_interval = $time_taken > 0 ? $interval - int($time_taken+1) : $interval;
-        if($final_interval < 0) {
-            $final_interval = 0;
-        }
-        # run these functions every $maintenance_interval
         $logger->trace("$id is sleeping for $final_interval ($interval from configuration - $time_taken)");
-        alarm $final_interval;
-        pause;
+        # Only wait if the interval is positive
+        unless($final_interval < 0) {
```
jrouzierinverse:
```
unless($final_interval <= 0) {
...
}
```

Or more clearly

```
if ($final_interval > 0) {
...
}
```

julsemaan:
fixed

sbin/pfmon
```diff
@@ -410,24 +410,22 @@ sub _runtask {
             next;
```
jrouzierinverse, line 411: I suggested to reset it to avoid a leftover time_taken value to be used in a calculation after someone disables and then enables a maintenance task.
Which could happen if the interval is changed to zero during the running of a task.

sbin/pfmon
```diff
@@ -410,24 +410,22 @@ sub _runtask {
...
-        # run these functions every $maintenance_interval
         $logger->trace("$id is sleeping for $final_interval ($interval from configuration - $time_taken)");
-        alarm $final_interval;
-        pause;
+        # Only wait if the interval is positive
+        if($final_interval > 0) {
+            alarm $final_interval;
+            pause;
+        }
         last unless $running;
```
jrouzierinverse: Move this check inside the if above since that check only matters after doing the pause.

julsemaan:
fixed

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
```diff
@@ -235,13 +235,24 @@ Apply the new node info in the session to the node
...
+
     if(node_register($self->current_mac, $self->username, %{$self->new_node_info()})){
         $self->app->flash->{notice} = [ "Role %s has been assigned to your device with unregistration date : %s", $self->new_node_info->{category}, $self->new_node_info->{unregdate} ];
         return $TRUE;
     }
     else {
-        $self->app->flash->{error} = "Couldn't register your device. Please contact your local support staff.";
-        return $FALSE;
+        $self->app->error("Couldn't register your device. Please contact your local support staff.");
+        $self->detach()
```
jrouzierinverse: semi colon ;

julsemaan:
fixed

PR #1520, opened 2016-06-03
Remove router_id from config template for keepalived

tribut:
# Description

Keepalived will use the hostname as `router_id` if not explicitly set making notifications mails much more meaningful. This is true at least for the version in Debian jessie.


julsemaan:
Seems like a good idea. Will just need to test if the same logic applies to the keepalive package on CentOS

julsemaan:
Unfortunately, this is not true for CentOS 7.

It puts `(null)` even though the hostname is properly configured.

With that issue there, we can't merge it. We'll have to see if this is something that can be fixed.

extrafu:
Closing, will do proper fix.

PR #1526, opened 2016-06-06
Fix/graphite configurable tls

### LINKED ISSUE #1521: Admin dashboard: Graphite in HTTPS breaks the status in Firefox
Firefox requires to add the cert exception for every host:port couple.
This means that in order to see the dashboard, we need to access https://host:9000 then access https://host:1443 so the dashboard displays properly.

Following items would fix it:
- Check for the SSL error in the AJAX call and display a meaningful error in the admin interface
- Make graphite use http vs https configurable in pf.conf - monitoring

louismunro:
# Description

Adds an option to toggle TLS usage for the graphite URLs in the dashboard.
# Impacts

Allows Firefox to use the dashboard.
# Issue

fixes #1521 
# Delete branch after merge

YES
# UPGRADE file entries

Graphite URLs are now queried over HTTPS by default in the dashboard.
An option is provided to query them over HTTP.

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1777,6 +1777,13 @@ Useful if your load balancer cannot handle both tasks.
 Changing this requires to restart radiusd.
 EOT
 
+[monitoring.graphite_tls]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Query the graphite URLs in the dashboard over HTTPS.
+EOT
+
 [monitoring.graphite_host]
 type=text
 description=<<EOT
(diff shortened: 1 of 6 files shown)


julsemaan:
One comment on this, I think we need to catch the connection error somehow and present a meaningful error message to the user. Otherwise, a user will just assume that pf is b0rken.

louismunro:
Do you feel up to it? 
I concede to be a very bad js developer.

julsemaan:
I'll give it a shot today or tomorrow.

From my (extremelly fast) research, doesn't seem we'll get a specific code from an HTTPS error. Still we can show a help message when the connection fails (like check that graphite is running and that you have an HTTPS exception for https://IP:9000 by clicking here)

louismunro:
Don't bother.
This needs thinking and rework.
We should probably do things the right way and proxy the requests to the graphite httpd instead of sending them to a different port.

louismunro:
I thought better of this.
We have to support both HTTP(s) at all times.

Issue #1503, opened 2016-05-31
HASH(0x7fcfb7cf2040) value on Portail profil Filter after saving


piellick:
Hello,
when i try to add filter on a portail profil, after saving the option. My value is remplaced by SSID and HASH(0x7fcfb7cf2040) .

I'm on Packetfilter 6.0.2 and after somes checks with "oeufdur" on IRC, i have the good "HTML-FormHandler" packet : 

apt-cache search HTML-FormHandler libhtml-formhandler-model-dbic-perl - base class that holds DBIC model role libhtml-formhandler-perl - HTML forms using Moose root@packetfence:~# apt-cache show libhtml-formhandler-model-dbic-perl Package: libhtml-formhandler-model-dbic-perl Version: 0.29-1

![filters](https://cloud.githubusercontent.com/assets/18485789/15676796/faff1b02-2747-11e6-8dd5-b0d38c1274b3.PNG)
apt-cache show libhtml-formhandler-perl Package: libhtml-formhandler-perl Version: 1:0.40019-1 Architecture: all Maintainer: root root@jessiedeb.inverse.local Installed-Size: 1201

piellick:
another exemple:

on profiles.conf : 

[interne_DR7]
filter=ssid:HASH(0x7fcfb7ffe530),ssid:HASH(0x7fcfb7fdf0f8),network:192.168.15.0/24
block_interval=10m
sources=ad_dr7_cnrs_fr
logo=/common/cnrs_interne.png
description=portail_interne_DR7
filter_match_style=all
reuse_dot1x_credentials=enabled

ON WEB UI : 
![filters2](https://cloud.githubusercontent.com/assets/18485789/15679373/f3ce085c-2751-11e6-930e-7e209b291f37.PNG)

On configuration file, last filter is OK not on the WEB UI after reloading page.

![filters3](https://cloud.githubusercontent.com/assets/18485789/15679457/363d47e8-2752-11e6-80a1-383eb829f314.PNG)

jrouzierinverse:
Can you verify if you do not have another HTML::Formhandler installed?

piellick:
from my aptitude cache : 
``
root@packetfence:/usr/local/pf/conf# apt-cache search Formhandler
libhtml-formhandler-model-dbic-perl - base class that holds DBIC model role
libhtml-formhandler-perl - HTML forms using Moose

``

`
apt-cache show libhtml-formhandler-model-dbic-perl
[...]

PR #1541, opened 2016-06-09
Fix profile filter hash stringification

jrouzierinverse:
## Description

Fix profile filter hash stringification  
## Impacts

The admin profile filters
## Issue

Issue #1503
## NEWS file entries

Bug Fixes
+++++++++
- Fix profile filter saving incorrectly on jessie 
## Delete branch after merge

NO


fdurand:
Tested on jessie and works.
Also centos packages are available too (CentOS 6 and CentOS 7)

PR #1575, opened 2016-06-22
Fix/1465 local accounts override

### LINKED ISSUE #1465: local accounts: Create a local account for an existing user overwrite the existing user
Reported by @lzammit (AKA Zaymitte)

Using the web admin to create a local account may overwrite existing information if the username already exists.

We should not use "person_modify" in this specific use case, or should prompt the admin to either overwrite or cancel.

https://github.com/inverse-inc/packetfence/blob/devel/html/pfappserver/lib/pfappserver/Model/User.pm#L355

dwlfrth:
# Description

Make sure a user does not exists before creating it from the webadmin interface
# Impacts

User creation using webadmin
# Issue

fixes #1465 
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- No longer override an existing user when creating a new one from web admin (#1465)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/User/Create/Import.pm
+++ b/html/pfappserver/lib/pfappserver/Form/User/Create/Import.pm
@@ -35,6 +35,15 @@ has_field 'delimiter' =>
    ],
   );
 
+has_field 'pid_overwrite' => (
+    type    => 'Checkbox',
+    label   => 'Username (PID) overwrite',
+    tags    => {
+        after_element   => \&help,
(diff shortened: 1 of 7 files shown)


dwlfrth:
Review me

julsemaan:
We need to have the 'override if exists' checkbox, at least for the create multiple.

Some users use it to recreate vouchers with the same usernames but different password once their vouchers expire.

dwlfrth:
Review me

PR #1576, opened 2016-06-23
NetworkManager workarounds to make sure it doesn't mess with DNS configuration

tlaurion:
# Description

windbindd_child.pm: ifcg file (/etc/sysconfig/network-scripts/ifcfg) is generated if it doesn't exist yet so NetworkManager will not manage AD's virtual interface.

PacketFence_Administration_Guide.asciidoc: Big fat warning and commands to run to deactivate DNS management of NetworkManager if still active.

PacketFence_Clustering_Guide.asciidoc: Warning before running hostname related commands, since doing so with an active NetworkManager who manages dns (default)  will replace /etc/resolv.conf file with an empty file. Link to bug still currently open at RedHat.
# Impacts

winbindd_child.pm creates AD associated ifcfg file.
Cluster documentation refers to the NetworkManager bug, and to the administration guide where corrective actions need to be taken at system initial installation.

NOTE: if NetworkManager is not deactivated/instructed to not mess with DNS config, it will since it listens on hostname file changes!
# Delete branch after merge

YES
## Enhancements
- All PF' created interfaces are excluded from NetworkManager

### DIFF
--- a/docs/PacketFence_Administration_Guide.asciidoc
+++ b/docs/PacketFence_Administration_Guide.asciidoc
@@ -210,6 +210,12 @@ RedHat-based systems
 
 NOTE: Applies to CentOS and Scientific Linux but only the x86_64 architecture is supported.
 
+Explicitly instruct NetworkManager to never interact with your DNS configuration (https://bugzilla.redhat.com/show_bug.cgi?id=1344303[Source]):
+
+  echo "[main]
+  dns=none" > /etc/NetworkManager/conf.d/99-no-dns.conf
+  service NetworkManager restart
+
 RHEL 6.x
 ^^^^^^^^
 
--- a/lib/pf/services/manager/winbindd_child.pm
+++ b/lib/pf/services/manager/winbindd_child.pm
@@ -74,6 +74,12 @@ sub build_namespaces(){
         my $OUTERLOGDIRECTORY="$CHROOT_PATH/$LOGDIRECTORY";
         my $OUTERRUNDIRECTORY="$CHROOT_PATH/var/run/samba$domain";
         my $PIDDIRECTORY="$var_dir/run/$domain";
+        my $DOMAIN_IF="$domain-b";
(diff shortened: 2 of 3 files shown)


lib/pf/services/manager/winbindd_child.pm
```diff
@@ -74,6 +74,12 @@ sub build_namespaces(){
         my $OUTERLOGDIRECTORY="$CHROOT_PATH/$LOGDIRECTORY";
         my $OUTERRUNDIRECTORY="$CHROOT_PATH/var/run/samba$domain";
         my $PIDDIRECTORY="$var_dir/run/$domain";
+        my $DOMAIN_IF="$domain-b";
+        my $DOMAIN_IFCFG_VIRT_INT="/etc/sysconfig/network-scripts/ifcfg-$DOMAIN_IF";
+        unless (-e $DOMAIN_IFCFG_VIRT_INT) {
+           my $cmd='echo -e "DEVICE='.$DOMAIN_IF.'\nNM_CONTROLLED=no" |sudo tee '.$DOMAIN_IFCFG_VIRT_INT;
```
tlaurion, line 80: This is more for coherence then anything else. (all packetfence linked interfaces are told to not be managed by nm).

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -210,6 +210,12 @@ RedHat-based systems
 
 NOTE: Applies to CentOS and Scientific Linux but only the x86_64 architecture is supported.
 
+Explicitly instruct NetworkManager to never interact with your DNS configuration (https://bugzilla.redhat.com/show_bug.cgi?id=1344303[Source]):
+
+  echo "[main]
+  dns=none" > /etc/NetworkManager/conf.d/99-no-dns.conf
```
tlaurion, line 216: This is what really stops nm from messing with dns configuration. NM won't be fixed.

PR #1577, opened 2016-06-23
Extendable Profile Filters

jrouzierinverse:
## Description

Allow third parties to extend what can be match in profile filters
## Impacts

Profile filtering
## NEWS file entries

New Features
++++++++++++
- Allow third parties to extend profile filter matching
## Delete branch after merge

NO


lib/pf/node.pm
```diff
@@ -1327,6 +1331,32 @@ sub _cleanup_attributes {
...
+
+sub node_extended_data {
+    my ($mac) = @_;
+    my @namespaces = qw(mse);
+    if (@{$Config{advanced}{extended_profile_filter_namespaces}}) {
+        my %hash;
+        my $redis = pf::Redis->new(server => 'localhost:6379');
+        my $json = JSON::MaybeXS->new;
+        foreach my $namespace (@namespaces) {
+            my ($data) = $redis->get("extended:${namespace}:${mac}");
```
julsemaan: Wouldn't it be better to use the hashes in Redis instead of this to prevent race conditions while updating the JSON data.

Or maybe support both JSON in a key and hashes ?

jrouzierinverse: I am not sure how using hashes will prevent race conditions.
Can you please provide a bit more detail on how that will work?

The main reason not use hashes is that we can support different namespaces that each have a  different TTLs.

julsemaan: Well example, I have a JSON blob in a key that contains the status of the mac in different MDM providers:
{
'MobileIron' : 0,
'OPSWAT' : 1,
'ZammitDM' : 0
}

Now if I want to change OPSWAT to 0, I have to fetch the redis key, decode the JSON blob, modity the hash, re-encode it and re-set it in the redis key. If in the meanwhile another operation is being done on that key to change 'ZammitDM' to 1, then it will be lost when I write the operation for OPSWAT.

If it would be a redis hash, I'd simply do `hset extended:whatever:00:11:22:33:44:55 OPSWAT 1` without risking to impact other keys in the hash.

Then in this piece of code doing HGETALL instead of decoding the JSON piece will give the whole hash.

The keys in the hash don't support complex structures so this is why I suggesting having both of them.

rouzier: IMO it should be the job of the 3rd party integration to know the state of the object it is storing.

So there two ways this could be done.

Different namespaces this way we can have separate expiration for each namespace.

set extended:MobileIron:00:11:22:33:44:55 '{ "allowed": 0 }' EX 1 # Expires in one second
set extended:OPSWAT:00:11:22:33:44:55 '{ "allowed": 1 }' EX 10 # Expires in 10 seconds
set extended:ZammitDM:00:11:22:33:44:55 '{ "allowed": 0 }'         # ZammitDM never expires

Or one namespace

set extended:sso:00:11:22:33:44:55 '{"MobileIron":0,"OPSWAT":1,"ZammitDM":0}'

Also my concern about using hashes it introduces more ways to get it wrong.

It introduces questions like the following
- How do we decide when to serialize values using json vs just using plain text?
- Which hash keys do we update and when?

julsemaan: Yeah I guess different namespaces would do it, just less user friendly but then, this is such an advanced feature that it can live with advanced ways to configured it to obtain keys that are safe to race conditions

fdurand:
Do we really need to review it since it's included in pr #1578

jrouzierinverse:
No need close this is you want.

PR #1590, opened 2016-06-29
Fix/portal timeout

### LINKED ISSUE #1570: Add timeout to mod_perl
Right now, Apache requests cannot timeout when they are handed to mod_perl.

This means that if there is an infinite loop, the process can hang forever and eat as much RAM as he wants until OOM gets him.

We should either install a timeout in the mod_perl handlers or at least in the handle_request of Catalyst.

Other suggestions are also welcome.

julsemaan:
# Description

Make captive portal timeout after X amount of seconds.
# Impacts

Captive portal
# Issue

fixes #1570
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Added timeout to captive portal to prevent long running requests (#1570).


julsemaan:
Please review

lib/pf/web/dispatcher.pm
```diff
@@ -43,6 +43,9 @@ use pf::Portal::Session;
 use pf::web::externalportal;
 use pf::inline;
 
+use pfconfig::cached_scalar;
+tie our $portal_request_timeout, 'pfconfig::cached_scalar', 'resource::portal_request_timeout';
```
jrouzierinverse: Since this is already in $Config{captive_portal}{request_timeout}
Why the need to add another namespace?

lib/pf/web/dispatcher.pm
```diff
@@ -59,6 +62,13 @@ Reference: http://perl.apache.org/docs/2.0/user/handlers/http.html#PerlTransHand
 =cut
 
 sub handler {
+    alarm $portal_request_timeout;
+    my $res = _handler(@_);
+    alarm 0;
+    return $res;
+}
```
jrouzierinverse, line 72: This does not follow the usual alarm pattern in perl.
See the example here for setting an alarm in mod_perl.
https://perl.apache.org/docs/2.0/user/coding/coding.html#Using_Signal_Handlers

lib/pf/web/dispatcher.pm
julsemaan: This is so captiveportal in Catalyst doesn't have to use pf::config so its lighter

rouzier: pf::config will be loaded by all the controller and model in catalyst so it really does not change that much.

julsemaan:
Adjustments done, re-review please

PR #1613, opened 2016-07-12
Fix/1574 external portal switch instantiation

### LINKED ISSUE #1574: switche ranges: Not working with external portal
https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/web/externalportal.pm#L62

dwlfrth:
# Description

Redesigned the way external portal works... switch module / object is responsible for parsing and externalportal.pm is abstracted
# Impacts
- External portal workflow...
- Switches modules.... (WLC, Catalyst_2960, Xirrus, AeroHIVE, Meraki)
# Issue

fixes #1574 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Easier external portal integration flow
## Bug Fixes

- switch ranges: Not working with external portal (#1574)
# UPGRADE file entries

Rework of the external captive portal capabilities involves some significant changes in the switch modules configuration.
Some switch modules have been moved to other ones and some others have been removed. Please adjust configuration (type) accordingly within switches.conf.
AeroHIVE::AP_http -> AeroHIVE::AP
Meraki::AP_http -> Meraki::MR
Meraki::AP_http_V2 -> Meraki::MR_v2
Xirrus:AP_http -> Xirrus

To instruct a switch module to perform external captive portal enforcement, a new switch configuration parameter have been added. Make sure to adjust to your needs
ExternalPortalEnforcement = Y

External captive portal URLs have also changed. 

AeroHIVE: http://portal_IP/AeroHIVE::AP
Aruba: http://portal_IP/Aruba
Cisco Catalyst 2960: http://portal_IP/Cisco::Catalyst_2960
Cisco WLC: http://portal_IP/Cisco::WLC
CoovaChilli: http://portal_IP/CoovaChilli
Meraki: http://portal_IP/Meraki::MR
Ruckus: http://portal_IP/Ruckus
Xirrus: http://portal_IP/Xirrus


fdurand:
Still have _http switch module.
There is no "Enable external portal" switch configuration parameter .
Doc must be updated to explain how to define the redirect url.

dwlfrth:
Review me

lib/pf/Switch/Cisco/WLC.pm
```diff
@@ -429,10 +411,10 @@ sub returnRadiusAccessAccept {
     my @av_pairs = defined($radius_reply_ref->{'Cisco-AVPair'}) ? @{$radius_reply_ref->{'Cisco-AVPair'}} : ();
 
     my $role = $self->getRoleByName($args->{'user_role'});
-    if ( isenabled($self->{_UrlMap}) && $self->supportsUrlBasedEnforcement ) {
+    if ( isenabled($self->{_UrlMap}) && $self->externalPortalEnforcement ) {
         if ( defined($args->{'user_role'}) && $args->{'user_role'} ne "" && defined($self->getUrlByName($args->{'user_role'}) ) ) {
             my $redirect_url = $self->getUrlByName($args->{'user_role'});
-            $args->{'session_id'} = "cep".$self->setSession($args) if ($redirect_url =~ /\$session_id/);
+            $args->{'session_id'} = "sid".$self->setSession($args) if ($redirect_url =~ /\$session_id/);
```
jrouzierinverse: Why the change from cep to sid?

lib/pf/Switch/MockedSwitch.pm
```diff
@@ -2951,26 +2950,21 @@ sub parseRequest {
...
+
+Parse external portal request using URI and it's parameters then return an hash reference with the appropriate parameters
+
+See L<pf::web::externalportal::handle>
+
+=cut
+
+sub parseExternalPortalRequest {
+    my ( $self, $r, $req ) = @_;
     return "";
```
jrouzierinverse: This should return an empty hash

jrouzierinverse (changes requested):

lib/pf/Switch/Cisco/WLC.pm
dwlfrth: "cep" originally stands for "cisco external portal". "sid" stands for "session id" which is more generic if it comes that we support other vendors that uses session based external portal flow.

lib/pf/Switch/MockedSwitch.pm
dwlfrth: You are right !

fdurand:
I like what you did but there is limitation since if you enable ExternalPortalEnforcement then you can only do web auth.
What i would like to have is a way to enable or disabled it based on criterias like the ssid, the connection type ...

extrafu:
Bump. What's the status on this?

lib/pf/web/constants.pm
```diff
@@ -93,18 +93,15 @@ Readonly::Scalar our $URL_GAMING_REGISTRATION   => '/gaming-registration';
...
-Readonly::Scalar our $REQ_ARUBA                 => 'apname';
-Readonly::Scalar our $REQ_CISCO_PORTAL          => '/cep(.*)';
-Readonly::Scalar our $REQ_RUCKUS                => 'sip';
-Readonly::Scalar our $REQ_AEROHIVE              => 'RADIUS-NAS-IP';
-
 # External Captive Portal URL detection constant
-Readonly::Scalar our $EXT_URL_XIRRUS            => '^/Xirrus::AP_http';
-Readonly::Scalar our $EXT_URL_MERAKI            => '^/Meraki::AP_http';
-Readonly::Scalar our $EXT_URL_COOVACHILLI       => '^/CoovaChilli';
+Readonly::Scalar our $EXT_URL_AEROHIVE              => '^/AeroHIVE::AP';
```
fdurand, line 97: Just /AeroHIVE should be enough

lib/pf/web/constants.pm
```diff
@@ -93,18 +93,15 @@ Readonly::Scalar our $URL_GAMING_REGISTRATION   => '/gaming-registration';
...
+Readonly::Scalar our $EXT_URL_ARUBA                 => '^/Aruba';
+Readonly::Scalar our $EXT_URL_CISCO_CATALYST_2960   => '^/Cisco::Catalyst_2960';
```
fdurand, line 99: Too long, i prefer something like /Cisco::Switch

lib/pf/web/constants.pm
```diff
@@ -93,18 +93,15 @@ Readonly::Scalar our $URL_GAMING_REGISTRATION   => '/gaming-registration';
 Readonly::Scalar our $URL_DEVICE_REGISTRATION   => '/device-registration';
 Readonly::Scalar our $URL_DEVICE_REG_LOGOUT     => '/device-registration/logout';
 
-# External Captive Portal detection constant
-Readonly::Scalar our $REQ_CISCO                 => 'ap_mac';
-Readonly::Scalar our $REQ_MAC                   => 'mac';
-Readonly::Scalar our $REQ_ARUBA                 => 'apname';
-Readonly::Scalar our $REQ_CISCO_PORTAL          => '/cep(.*)';
```
fdurand, line 100: Can we try to be backward compatible or maybe add a check to test if /cep\* is still there ? (switch.conf)

dwlfrth:
@fdurand request (ability to use multiple enforcement methods) is a good one but a bit out of this scope.
It can actually be achieved using RADIUS filters.
The scope of this PR was really just to achieve the same behavior as before, but easier and more versatile to then add new feature like the multiple enforcement methods.

lib/pf/web/constants.pm
dwlfrth: What do you mean? In case a configuration have not been modified ?

dwlfrth: You want to handle requests coming in with the /cep\* address ?
If yes, I'm not quite sure we should be backward compatible since it will involve adding other part of code for that kind of processing.
(Processing have been changed and is now using switch module used in the URI).
I can sure add an 'if' in the processing part to say if we match cep URI then use the Cisco module but we would have to mark it as deprecated and actually remove it before long...

fdurand: In fact i don't want backward compatibility because we will keep the code for long but just a warn in checkup that told the guy to read the upgrade guide.

dwlfrth: I like that ;)

dwlfrth:
@fdurand you might want to check last commit "https://github.com/inverse-inc/packetfence/pull/1613/commits/d842faa8fdc029e815f66337d55b5831a52940d8"

lib/pf/web/constants.pm
dwlfrth: After verification, the "cep" part was never handled by the user / configuration parameter itself but was added in the code based of if $sessionid was part of the URL.
There is then, no check to do.
I also reworked the logic of the session id no longer needed to be in the configuration since it is a de facto flow with Cisco (https://github.com/inverse-inc/packetfence/pull/1613/commits/d842faa8fdc029e815f66337d55b5831a52940d8)

lib/pf/web/constants.pm
dwlfrth: We now use the module name in the URI. It need to be configured as the module name in the configuration so I made the constants for matching the same.

lib/pf/web/constants.pm
dwlfrth: As for AeroHive, we now use the module name in the URI. It need to be configured as the module name in the configuration so I made the constants for matching the same.

fdurand: I understand but it mean that it only works for Catalyst_2960, a more generic name should be better.

lib/pf/Switch/Cisco/Catalyst_2960.pm
```diff
@@ -513,9 +513,10 @@ sub returnRadiusAccessAccept {
     if ( isenabled($self->{_UrlMap}) && $self->externalPortalEnforcement ) {
         if( defined($args->{'user_role'}) && $args->{'user_role'} ne "" && defined($self->getUrlByName($args->{'user_role'}))){
             my $mac = $args->{'mac'};
+            $args->{'session_id'} = "sid".$self->setSession($args);
             my $redirect_url = $self->getUrlByName($args->{'user_role'});
-            $args->{'session_id'} = "sid".$self->setSession($args) if ($redirect_url =~ /\$session_id/);
-            $redirect_url =~ s/\$([a-zA-Z_0-9]+)/$args->{$1} \/\/ ''/ge;
+            $redirect_url .= '/' unless $redirect_url =~ m(\/$);
+            $redirect_url .= $args->{'session_id'};
```
fdurand, line 519: Yes but no, imagine a violation role like antivirus what we want is to forward the user to http://www.antivirus.ca and not  http://www.antivirus.ca/sid9028493048093248

lib/pf/web/constants.pm
dwlfrth: Not quite sure how you want to implement it.
If we are 100% sure that all the Cisco equipment works the same way, we can move the parsing stuff into to parent (Cisco.pm) class and then use that as the URI. Otherwise, we need to refer to the module responsible for the handling

lib/pf/Switch/Cisco/Catalyst_2960.pm
julsemaan: You're right but I don't see the point that much. A normal flow would want the portal to be displayed with some instructions with a link to http://www.antivirus.ca

Redirecting the user straight to that website wouldn't make much sense.

IMO general configuration shouldn't be built to handle the 0.01% corner cases unless it has absolutely no impact on the interaction.

Also, if someone would want to handle that extreme corner case, he could do so using the RADIUS filters.

fdurand: Ok, then don't forget to change the doc (PacketFence_Network_Devices_Configuration_Guide.asciidoc)

lib/pf/web/constants.pm
fdurand: Arf ok, this is because you instantiate the switch module with that ... so forget and let's use a long name.

extrafu:
@julsemaan and @fdurand  - is it ready to be merged?

fdurand:
@extrafu it miss documentation change and upgrade guide

dwlfrth:
@fdurand @extrafu 
- UPGRADE info are in the PR description
- PacketFence Network Devices guide is updated

@inverse-inc/packetfence 
Need to carefully test at least WLC to make sure everything is still OK
Also, if possible, Meraki::MR_v2...

extrafu:
@fdurand Review and merge. Testing will be done during the release test process.

dwlfrth:
Just keep in mind we need to test Meraki MR_v2 since it was "BETA" at the time of development...

docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -3770,9 +3770,9 @@ Portal configuration
 ++++++++++++++++++++
 * Go in *Configuration -> Authentication -> Captive Web Portals* and create a new portal
 * Select Select *Registration Type* = *External Authentication*
-* Go in the section *Captive Web Portal Login Page Settings* set the *Login URL* to http://pf_ip/ and *Password Encryption* to *No Encryption*
+* Go in the section *Captive Web Portal Login Page Settings* set the *Login URL* to http://pf_ip/AeroHIVE::AP and *Password Encryption* to *No Encryption*
```
fdurand: <your_captive_portal_ip> instead of pf_ip

fdurand: or 192.168.1.5

docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -1634,7 +1634,7 @@ PacketFence switch configuration:
 
 * Select the type to 'Cisco Catalyst 2960'
 * Set the 'Registration' role to 'registration' (If left empty then it will use the role name)
-* Set Role by Web Auth URL for registration to 'http://<your_captive_portal_ip>/$session_id'
+* Set Role by Web Auth URL for registration to 'http://<your_captive_portal_ip>/Cisco::Catalyst_2960'
```
fdurand, line 1637: or 192.168.1.5

dwlfrth:
Please re-review

fdurand:
Ok we are good to go.

PR #1631, opened 2016-07-19
Feature/cluster conflict handling

julsemaan:
# Description

Adds configuration conflict handling for active/active clusters
# Impacts

Configuration synchronization in active/active clustering
# Delete branch after merge

YES
# NEWS file entries

## New Features
- Active/active cluster nodes will now handle configuration conflicts automatically

### DIFF
--- a/docs/PacketFence_Clustering_Guide.asciidoc
+++ b/docs/PacketFence_Clustering_Guide.asciidoc
@@ -999,6 +999,38 @@ Now restart haproxy and iptables in order to complete the configuration
 
 You should now be able to connect to the dashboard on the following URL : http://pf.local:1025/stats
 
+Configuration conflict handling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+NOTE: It is not recommended to perform onfiguration while one or more node of the cluster is experiencing issues. Still, should that be the case, this section will explain the conflict resolution that will occur when the nodes will reattach together.
+
+When modifying the configuration through the administration interface, the configuration will be automatically synchronized to all the nodes that are online. In the event that one or more nodes cannot be updated, an error message will be displayed with affected nodes.
+
+A scheduled check runs on the management server (controlled through maintenance.cluster_check_interval) in order to validate if all servers are running the same configuration version.
+When the failed node(s) will come back online, that scheduled check will ensure that the new configuration is pushed on the new node(s).
+You can disable this check by setting maintenance.cluster_check_interval to 0 and restarting pfmon. In that case, you will need to manually resolve the conflict when the node(s) come back online by running `# /usr/local/pf/bin/cluster/sync --as-master` on the node you want to keep the configuration of.
+
+General facts about conflict resolution:
+
+ * If the configuration is not pushed to at least half of the servers of your cluster, when the failed nodes will come back online, they will have quorum on the previous configuration and the one they are running will be pushed to all the servers.
+ * In a two node cluster, the most recent configuration is always selected when resolving a conflict.
+ * In a two node cluster, no decision is taken unless the peer server has its webservices available.
+
+Going deeper in the conflict handling
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The section below will explain with more details, the steps that are taken in order to take the decision of which server should be declared as the master when one or more servers have conflicting configuration version.
+
+The first step is to get the configuration version from each server through a webservice call.
+
+The results are then organized by version identifier. Should all alive servers run the same version, the state is considered as healthy and nothing happens.
+
+Then, should there be more than one version identifier across the alive servers, the algorithm validates that there are at least 2 servers **configured** in the cluster. If there aren't, then the most recent version is pushed on the peer node.
+
+After that, the algorithm looks at which version is on the most servers. In the event that the dead servers are in higher number than the alive ones, the most recent version is taken. Otherwise, the version that is present on the most servers will be selected.
+
+When pushing a version to the other servers, if the current version has the most recent version or is part of the majority (depending on which push strategy was defined above), then he will be the one pushing the new configuration to the other servers. Otherwise, a webservice call is made to one of the servers running the selected version so that he pushes its configuration to its peers.
+
 Unsupported features in active/active
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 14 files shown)


lib/pf/cluster.pm
```diff
@@ -396,6 +408,205 @@ sub call_server {
...
+
+    # We make sure we have the right version for this node (in case webservices is currently dead)
+    $servers_map->{$host_id} = $version;
+
+
+    if(keys(%$versions_map) == 2 && (defined($versions_map->{0}) && @{$versions_map->{0}} > 0)) {
+        get_logger->warn("Not all servers were checked for the configuration version but all alive ones are running the same version.");
+        return;
+    }
+    elsif(keys(%$versions_map) > 1) {
```
jrouzierinverse, line 500: No need for the elsif since you return in the previous if.

lib/pf/cluster.pm
```diff
@@ -396,6 +408,205 @@ sub call_server {
...
+                get_logger->warn("Failed to find quorum in the cluster. Most recent configuration will be selected.");
+                goto SYNC_MOST_RECENT;
+            }
+        }
+        else {
+            get_logger->info("Quorum not possible in 2 servers clusters. Most recent configuration will be selected.");
+            goto SYNC_MOST_RECENT;
+        }
+    }
+    else {
```
jrouzierinverse: No need for the else since since you are just logging and returning.

lib/pf/cluster.pm
```diff
@@ -396,6 +408,205 @@ sub call_server {
...
+        my $server = $versions_map->{$quorum_version}->[0];
+        get_logger->info("Using $server from the servers holding quorum to sync as the master.");
+        call_server($server, 'sync_config_as_master');
+    }
+
+    return;
+
+}
```
jrouzierinverse, line 572: The equivalent behavior could achieved by doing `return sync_most_recent();` or `return sync_quorum();` instead of using gotos.

Since you are just returning after each goto section it does not make sense to use them in this case.

However if the workflow required something like below then I would not have any issues with the use of the gotos.

```
SYNC_MOST_RECENT:
  sync_most_recent();
SYNC_QUORUM:
  sync_quorum();
  return
```

docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -999,6 +999,38 @@ Now restart haproxy and iptables in order to complete the configuration
 
 You should now be able to connect to the dashboard on the following URL : http://pf.local:1025/stats
 
+Configuration conflict handling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+NOTE: It is not recommended to perform onfiguration while one or more node of the cluster is experiencing issues. Still, should that be the case, this section will explain the conflict resolution that will occur when the nodes will reattach together.
```
jrouzierinverse: onfiguration => configuration

sbin/pfmon
```diff
@@ -543,6 +548,46 @@ sub update_fingerbank_data {
...
+            $cache->set('last_config_healthy_timestamp', time);
+        }
+        elsif($last_conflict_at > $conflict_resolution_threshold) {
+            get_logger->info("Configuration is unhealthy since $conflict_resolution_threshold seconds. Will attempt to resolve the conflict");
+            pf::cluster::handle_config_conflict();
+        }
+    }
+    else {
+        get_logger->info("All cluster members are running the same configuration version");
+        $cache->set('last_config_healthy_timestamp', time);
```
jrouzierinverse: last_config_healthy_timestamp is set twice.

sbin/pfmon
```diff
@@ -543,6 +548,46 @@ sub update_fingerbank_data {
...
+
+    my $conflict_resolution_threshold = $Config{active_active}{conflict_resolution_threshold};
+
+    get_logger->info("Using $conflict_resolution_threshold resolution threshold");
+    
+    my $interval = $Config{'maintenance'}{'cluster_check_interval'};
+    my ($servers_map, $version_map) = pf::cluster::get_all_config_version();
+
+    # Making sure we have all available data for the decision and that there are multiple versions detected
+    if(defined($last_conflict_at) && defined($last_config_checked) && keys(%$version_map) > 1) {
```
jrouzierinverse, line 572: Would it be better to call time only once to be consistent?

sbin/pfmon
```diff
@@ -543,6 +548,46 @@ sub update_fingerbank_data {
...
+            get_logger->info("Configuration is unhealthy since $conflict_resolution_threshold seconds. Will attempt to resolve the conflict");
+            pf::cluster::handle_config_conflict();
+        }
+    }
+    else {
+        get_logger->info("All cluster members are running the same configuration version");
+        $cache->set('last_config_healthy_timestamp', time);
+    }
+    $cache->set('last_config_checked_timestamp', time);
+}
```
jrouzierinverse, line 591: Time::HiRes::time should be used.

docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -999,6 +999,38 @@ Now restart haproxy and iptables in order to complete the configuration
...
+
+When modifying the configuration through the administration interface, the configuration will be automatically synchronized to all the nodes that are online. In the event that one or more nodes cannot be updated, an error message will be displayed with affected nodes.
+
+A scheduled check runs on the management server (controlled through maintenance.cluster_check_interval) in order to validate if all servers are running the same configuration version.
+When the failed node(s) will come back online, that scheduled check will ensure that the new configuration is pushed on the new node(s).
+You can disable this check by setting maintenance.cluster_check_interval to 0 and restarting pfmon. In that case, you will need to manually resolve the conflict when the node(s) come back online by running `# /usr/local/pf/bin/cluster/sync --as-master` on the node you want to keep the configuration of.
```
jrouzierinverse: No '#' in front of the command.

lib/pf/cluster.pm
julsemaan, line 572: As we discussed, the gotos are there so that the logic has access to the scope of the subroutine in order to dry up the code in case additional variables would be needed

sbin/pfmon
julsemaan: I don't really see where

jrouzierinverse: You are right my bad.

sbin/pfmon
julsemaan, line 572: done

julsemaan:
Adjustments made, please re-review

PR #1642, opened 2016-08-04
Feature/eduroam as a source

dwlfrth:
# Description

Create a new authentication source type (Eduroam) that will allow integration with FreeRADIUS
# Impacts
- Authentication sources
- RADIUS (without / with clustering)
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Added new "Eduroam" authentication source type to ease eduroam integration with FreeRADIUS
# UPGRADE file entries

*\* May be needed**

### DIFF
--- a/conf/radiusd/eduroam.conf.example
+++ b/conf/radiusd/eduroam.conf.example
@@ -1,0 +1,139 @@
+
+pidfile = %%pid_file%%
+
+$INCLUDE radiusd.conf
+
+listen {
+    ipaddr = %%management_ip%%
+    port = %%eduroam_auth_listening_port%%
+    type = auth
+    virtual_server = eduroam
+}
+
+listen {
+    ipaddr = %%management_ip%%
+    port = %%eduroam_acct_listening_port%%
+    type = acct
+    virtual_server = eduroam
+}
+
+listen {
+    type = control
+    socket = %%socket_file%%
+    mode = rw
+}
+
+log {
+	#
+	#  Destination for log messages.  This can be one of:
+	#
+	#	files - log to "file", as defined below.
+	#	syslog - to syslog (see also the "syslog_facility", below.
+	#	stdout - standard output
+	#	stderr - standard error.
+	#
+	#  The command-line option "-X" over-rides this option, and forces
+	#  logging to go to stdout.
+	#
+	destination = files
+
+	#
+	#  Highlight important messages sent to stderr and stdout.
+	#
+	#  Option will be ignored (disabled) if output if TERM is not
+	#  an xterm or output is not to a TTY.
+	#
+	colourise = yes
+
+	#
+	#  The logging messages for the server are appended to the
+	#  tail of this file if destination == "files"
+	#
+	#  If the server is running in debugging mode, this file is
+	#  NOT used.
+	#
+	file = ${logdir}/radius-eduroam.log
+
+	#
+	#  If this configuration parameter is set, then log messages for
+	#  a *request* go to this file, rather than to radius.log.
+	#
+	#  i.e. This is a log file per request, once the server has accepted
+	#  the request as being from a valid client.  Messages that are
+	#  not associated with a request still go to radius.log.
+	#
+	#  Not all log messages in the server core have been updated to use
+	#  this new internal API.  As a result, some messages will still
+	#  go to radius.log.  Please submit patches to fix this behavior.
+	#
+	#  The file name is expanded dynamically.  You should ONLY user
+	#  server-side attributes for the filename (e.g. things you control).
+	#  Using this feature MAY also slow down the server substantially,
+	#  especially if you do thinks like SQL calls as part of the
+	#  expansion of the filename.
+	#
+	#  The name of the log file should use attributes that don't change
+	#  over the lifetime of a request, such as User-Name,
+	#  Virtual-Server or Packet-Src-IP-Address.  Otherwise, the log
+	#  messages will be distributed over multiple files.
+	#
+	#  Logging can be enabled for an individual request by a special
+	#  dynamic expansion macro:  %{debug: 1}, where the debug level
+	#  for this request is set to '1' (or 2, 3, etc.).  e.g.
+	#
+	#	...
+	#	update control {
+	#	       Tmp-String-0 = "%{debug:1}"
+	#	}
+	#	...
+	#
+	#  The attribute that the value is assigned to is unimportant,
+	#  and should be a "throw-away" attribute with no side effects.
+	#
+	#requests = ${logdir}/radiusd-%{%{Virtual-Server}:-DEFAULT}-%Y%m%d.log
+
+	#
+	#  Which syslog facility to use, if ${destination} == "syslog"
+	#
+	#  The exact values permitted here are OS-dependent.  You probably
+	#  don't want to change this.
+	#
+	syslog_facility = daemon
+
+	#  Log the full User-Name attribute, as it was found in the request.
+	#
+	# allowed values: {no, yes}
+	#
+	stripped_names = no
+
+	#  Log authentication requests to the log file.
+	#
+	#  allowed values: {no, yes}
+	#
+	auth = yes
+
+	#  Log passwords with the authentication requests.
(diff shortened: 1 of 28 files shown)


dwlfrth:
Would be ready for first round of comments

louismunro:
TOP POST!

extrafu:
Rebase ASAP so we can finish reviewing it and we merge it.

dwlfrth:
Rebased.
Please pay attention to the "use" parts of the files due to 3 rebases occurred since the opening of the pull-request and just want to make sure nothing have been removed

conf/radiusd/eduroam.conf.example
```diff
@@ -0,0 +1,132 @@
...
+	#
+	destination = files
+
+	#
+	#  Highlight important messages sent to stderr and stdout.
+	#
+	#  Option will be ignored (disabled) if output if TERM is not
+	#  an xterm or output is not to a TTY.
+	#
+	colourise = yes
```
louismunro, line 39: colourise? 
not colourize?

dwlfrth: Yeah, I tend to conserve what others do and that I don't need to change ;) :P
https://github.com/inverse-inc/packetfence/blame/devel/conf/radiusd/auth.conf.example#L46
https://github.com/inverse-inc/packetfence/blame/devel/conf/radiusd/acct.conf.example#L39

dwlfrth: Let's involve Alan !!
https://github.com/FreeRADIUS/freeradius-server/blob/v4.0.x/raddb/radiusd.conf.in#L256

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Eduroam.pm
```diff
@@ -0,0 +1,95 @@
...
+    type        => 'Text',
+    label       => 'TLRS2 server address',
+    required    => 1,
+    tags        => {
+        after_element   => \&help,
+        help            => 'Eduroam TLRS2 server address',
+    },
+);
+
+has_field 'tlrs_radius_secret' => (
```
louismunro: can we be sure the secret will be the same for all eduroam servers?

conf/radiusd/eduroam.conf.example
louismunro: Err....
I say the dictionary is wrong... 
colourise it is then!

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Eduroam.pm
dwlfrth: It can be different but the provided doc will ask to set the same (easier) since we need to create new tlrs entries on eduroam for PacketFence

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Eduroam.pm
```diff
@@ -0,0 +1,95 @@
...
+
+use strict;
+use warnings;
+
+use HTML::FormHandler::Moose;
+
+extends 'pfappserver::Form::Config::Authentication::Source';
+with 'pfappserver::Base::Form::Role::Help';
+
+has_field 'tlrs1_server_address' => (
```
louismunro: You know, I am not so sure calling them tlrs1 and tlrs2 is a great idea.
Those are the current North-American servers names, but they will not make sense elsewhere.
Could we not just have server1 and server2?
Or something that would make sense globally?

louismunro:
Is there any documentation?

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Eduroam.pm
dwlfrth: Hmm TLRS stands for Top-Level RADIUS Server.
Don't think it's a "North-America" thing.

dwlfrth:
Documentation will come once we agreed on the configuration steps and flow since I don't want to write documentation that will be changed 2-3 times :) Lazy me :P

louismunro:
okay then.

dwlfrth:
But, no merge before doc ;)

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Eduroam.pm
louismunro: I don't remember seeing it elsewhere though.
Nevertheless, it's not worth waiting for this.

louismunro:
looks good enough.
Ready for merging.

lib/pf/services/manager/radiusd_child.pm
```diff
@@ -246,11 +283,49 @@ $options
...
     }
+
+    # Eduroam configuration
+    if ( @{pf::authentication::getAuthenticationSourcesByType('Eduroam')} ) {
+        my @eduroam_authentication_source = @{pf::authentication::getAuthenticationSourcesByType('Eduroam')};
+        my $tlrs1_server_address = $eduroam_authentication_source[0]{'tlrs1_server_address'};   # using array index 0 since there can only be one 'eduroam' authentication source ('unique' attribute)
+        my $tlrs2_server_address = $eduroam_authentication_source[0]{'tlrs2_server_address'};   # using array index 0 since there can only be one 'eduroam' authentication source ('unique' attribute)
+        my $tlrs_radius_secret = $eduroam_authentication_source[0]{'tlrs_radius_secret'};   # using array index 0 since there can only be one 'eduroam' authentication source ('unique' attribute)
+
+        $tags{'eduroam'} = <<"EOT";
```
fdurand, line 294: I think it should be in site-enabled/eduroam

fdurand:
I suppose that the assumption is that there is a 'eduroam ssid' but if the wifi admin want to have both on one ssid, can we imagine to play with the realm instead too ?

dwlfrth:
@fdurand eduroam guidelines mentions that eduroam should be operated on it's own SSID called 'eduroam'.
Sure thing, we have clients having different needs but I think that the purpose of this PR is really to be able to integrate with eduroam in a "standard" way based on the eduroam guidelines.

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/Eduroam.pm
dwlfrth: You are right.
Here it is called TLRS by eduroam (as said, for Top-level RADIUS server) but it looks like it may also be called FLR (Federetion level RADIUS) elsewhere.
I'll have the name changed

julsemaan:
I would opt for having the possibility of using an SSID named something else than eduroam if that isn't already the case for testing purposes. Not all admins will want to have the SSID named eduroam while testing

dwlfrth:
If I remember correctly, there is no reference to the SSID anywhere in the code (correct me if I'm wrong).
The only "association" is the RADIUS port.
You can then have your famous "ZammitIsTheBestBecauseHeIsTheNumeroUnoWorldwideLikePitbullMister305" SSID, as long as it is configured to interact with PacketFence RADIUS server on port XXXX (as configured), you will be golden.

julsemaan:
@dwlfrth 

We're golden from my point of view then.

lib/pf/services/manager/radiusd_child.pm
dwlfrth: Not sure why you say the proxy part should be in the eduroam virtual server configuration file ?

fdurand: Just for consistency , we do that in packetfence-cluster

dwlfrth: "proxy" in packetfence-cluster are the PacketFence servers themselves... i think that proxy.conf should refer to "outside" proxy servers... some that are not "PacketFence managed"
But I must say that I agree with you on the point that eduroam configuration should be all together in the same file (that would also include the clustering configuration)...
Let me rework that and I'll let you know

PR #1656, opened 2016-08-12
feature/pfsetvlan-refactor

jrouzierinverse:
## Description

Move snmp processing to pfqueue
## Impacts

Switches using snmp port security
## NEWS file entries

Enhancements
++++++++++++
- SNMP traps are now handled in pfqueue
## Delete branch after merge

NO


louismunro:
Unit tests would make me feel all warm inside...

jrouzierinverse:
So would some whiskey

lib/pf/task/pfsnmp.pm
```diff
@@ -0,0 +1,972 @@
...
+
+Handle the setvlan task
+
+=cut
+
+my $logger = get_logger();
+
+sub doTask {
+    my ($self, $args) = @_;
+    use Data::Dumper;get_logger->info(Dumper($args));
```
louismunro, line 53: Should this still be here?

lib/pf/task/pfsnmp.pm
```diff
@@ -0,0 +1,972 @@
...
+
+=head2 doTask
+
+Handle the setvlan task
+
+=cut
+
+my $logger = get_logger();
+
+sub doTask {
```
louismunro, line 51: Please add statsd metrics for timing.

lib/pf/task/pfsnmp.pm
jrouzierinverse: No

lib/pf/task/pfsnmp.pm
jrouzierinverse: Ok

lib/pf/snmptrapd.pm
```diff
@@ -0,0 +1,96 @@
...
+    foreach my $variable (@$variables) {
+        $variable->[0] = $variable->[0]->quote_oid;
+    }
+}
+
+NetSNMP::TrapReceiver::register("all", \&receiver) || warn "failed to register perl trap handler\n";
+
+sub _redis_client {
+    return pf::Redis->new(
+        server    => '127.0.0.1:6380',
```
louismunro, line 62: Fix Me hardcoded.

lib/pf/file_paths.pm
```diff
@@ -166,6 +168,7 @@ $log_conf_dir  = catdir( $conf_dir,"log.conf.d" );
 
 $generated_conf_dir   = catdir( $var_dir,"conf");
 $tt_compile_cache_dir = catdir( $var_dir,"tt_compile_cache");
+$control_dir  = catdir( $var_dir,"control");
```
julsemaan, line 175: This looks like it maps to /usr/local/pf/var/control which is the directory for the pfconfig control files.

pfconfig will wipe the directory contents when restarting which will certainly cause a problem at some point.

I would suggest using another directory for that

lib/pf/services/manager/snmptrapd.pm
```diff
@@ -43,7 +43,7 @@ if (ref($management_network)) {
     $management_ip .= ':162';
 }
 
-has '+launcher' => (default => sub { "%1\$s -n -c $generated_conf_dir/snmptrapd.conf -C -A -Lf $install_dir/logs/snmptrapd.log -p $install_dir/var/run/snmptrapd.pid -On $management_ip" } );
+has '+launcher' => (default => sub { "%1\$s -n -c $generated_conf_dir/snmptrapd.conf -C -A -Lf $install_dir/logs/snmptrapd.log -p $install_dir/var/run/snmptrapd.pid -On 0.0.0.0:162" } );
```
fdurand: Why not 0.0.0.0 but what about the mgmt vip or the mgmt ip ?

fdurand:
Question, what about active/active setup ? 
Can we imagine to have a udp load balancer in front of snmptrapd in order to load balance ? (Maybe it's out of scope and need to devellop another feature for that).
Btw can we add something like that in iptables (-A POSTROUTING --protocol udp --match udp --dport 161 -o mgmt_int -j SNAT --to mgmt_vip) in order to just define one snmp server in the switch config ?

lib/pf/services/manager/snmptrapd.pm
jrouzierinverse: Added this when testing and you are right we should bind to the proper interface.

Should it be the vip or ip at the moment.

julsemaan:
Seems like this was merged with pending requested changes and some look quite important.

Is it an error that this was merged ?

PR #1671, opened 2016-08-24
Invalidate email registration URL upon their timeout

### LINKED ISSUE #1670: Email activation URL timeout is hard-coded to 31 days.
Email activation URLs expire after 31 days, regardless of the timeout allowed by the email authentication source.

It's thus possible to reuse a URL even though the timeout is passed.

louismunro:
# Description

Email activations URLs were hard-coded to expire after 31 days.
This makes them expire when the email source times out.
# Impacts

Try to use a registration URL after the email source has timed out (e.g. 10 minutes).
It should not be allowed.
# Issue

fixes #1670
# Delete branch after merge

YES 
## Bug Fixes
- Fixes email registration URL not expiring at same time as the source times out (#1670)

### DIFF
--- a/lib/pf/activation.pm
+++ b/lib/pf/activation.pm
@@ -127,7 +127,7 @@ sub activation_db_prepare {
     $activation_statements->{'activation_find_unverified_code_sql'} = get_db_handle()->prepare(qq[
         SELECT code_id, pid, mac, contact_info, activation_code, expiration, status, type, portal, email_pattern as carrier_email_pattern, unregdate
         FROM activation LEFT JOIN sms_carrier ON carrier_id=sms_carrier.id
-        WHERE activation_code = ? AND status = ?
+        WHERE activation_code = ? AND status = ? AND expiration >= NOW()
     ]);
 
     $activation_statements->{'activation_find_code_sql'} = get_db_handle()->prepare(qq[
@@ -342,7 +342,7 @@ Returns the activation code
 =cut
 
 sub create {
-    my ($mac, $pid, $pending_addr, $type, $portal, $provider_id) = @_;
+    my ($mac, $pid, $pending_addr, $type, $portal, $provider_id, $timeout) = @_;
     my $logger = get_logger();
 
     unless($mac){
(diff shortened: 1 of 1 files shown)


lib/pf/activation.pm
```diff
@@ -342,9 +342,19 @@ Returns the activation code
...
+        $pid 		= $args->{'pid'}, 
+	$pending_addr 	= $args->{'pending'}, 
+	$type 		= $args->{'type'}, 
+	$portal 	= $args->{'portal'}, 
+	$provider_id 	= $args->{'provider_id'}, 
+	$timeout 	= $args->{'timeout'} );
+	
     my $logger = get_logger();
 
+    $mac = $args->{'mac'};
```
cgx: Duplicated code.

lib/pf/activation.pm
```diff
@@ -501,7 +514,10 @@ sub create_and_send_activation_code {
     my ($mac, $pid, $pending_addr, $template, $type, $portal, %info) = @_;
 
     my ($success, $err) = ($TRUE, 0);
-    my $activation_code = create($mac, $pid, $pending_addr, $type, $portal);
+    my %args = ( 
+	mac 	=> $mac,  pid 	 => $pid, 	pending => $pending_addr,
```
dwlfrth: One per line is better (coding style)

lib/pf/activation.pm
```diff
@@ -574,7 +590,10 @@ sub sms_activation_create_send {
     $phone_number =~ s/\D//g;
 
     my ($success, $err) = ($TRUE, 0);
-    my $activation_code = create($mac, $pid, $phone_number, 'sms', $portal, $provider_id);
+    my %args = ( 
+	mac  => $mac,  pid    => $pid,    pending     => $phone_number,
```
dwlfrth: Again, one per line is better (coding style)

dwlfrth (changes requested):

dwlfrth:
Fix duplicate code mentionned by @cgx and requested changes by me and we are golden to merge.

lib/pf/activation.pm
louismunro: I disagree in this case.
It makes the subroutine too long.
A tabular format is better if there are a lot of arguments.
Conciseness is also a virtue.

louismunro: Come to think of it.
You are right.
I am changing it.

dwlfrth: Looks great when printed but PITA when it comes to editing and other.

lib/pf/activation.pm
louismunro: Done.

dwlfrth (approved):

PR #1708, opened 2016-09-19
remove outdated scan configuration

### LINKED ISSUE #1635: Section [scan] of pf.conf should be removed
Its outdated and was replaced (but never removed) by scan.conf

julsemaan:
# Description

Cleanup outdated [scan] section of pf.conf
# Impacts

Should be none
# Issue

fixes #1635 
# Delete branch after merge

(REQUIRED)
YES | NO


fdurand:
Rebase please

lib/pf/web.pm
```diff
@@ -201,10 +201,11 @@ sub generate_scan_start_page {
     my $logger = get_logger();
 
     $portalSession->stash({
-        timer           => $Config{'scan'}{'duration'},
+        # Hardcoded here since the scan section is gone.
+        # In case this codepath is still called (pf::web::release is still called by scan in some cases)
+        timer           => 60,
```
rouzier, line 206: Could this be moved into the scanner configuration?

julsemaan: Not even sure that codepath is still being used so I don't think we should really put efforts in it.

Would like to remove it completely but I'm not confident enough about it

extrafu:
Rebase

jrouzierinverse:
Bump

fdurand:
I have rebased

PR #1715, opened 2016-09-23
Feature/distributed cache

julsemaan:
# Description

Create a distributed cache for CHI through MySQL to distribute the accounting cache
# Impacts

CHI accounting namespace cache (so deauth)
# NEW Package(s) required

None, new driver is included in lib/ since its custom to us
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added a distributed cache for the accounting information to safely disable the SQL accounting records in active/active clusters

### DIFF
--- a/db/chi_cache.sql
+++ b/db/chi_cache.sql
@@ -1,0 +1,5 @@
+CREATE TABLE `chi_cache` (
+  `key` VARCHAR(767),
+  `value` LONGTEXT,
+  PRIMARY KEY (`key`)
+);
--- a/db/chi_cache.sql
+++ b/db/chi_cache.sql
@@ -1,5 +1,0 @@
-CREATE TABLE `chi_cache` (
-  `key` VARCHAR(767),
-  `value` LONGTEXT,
-  PRIMARY KEY (`key`)
-);
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -1239,12 +1239,6 @@ END /
 DELIMITER ;
 
 --
--- Updating to current version
---
-
-INSERT INTO pf_version (id, version) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION, @SUBMINOR_VERSION));
-
---
 -- Creating auth_log table
 --
 
@@ -1261,3 +1255,20 @@ CREATE TABLE auth_log (
   KEY pid (pid),
   KEY  attempted_at (attempted_at)
 ) ENGINE=InnoDB;
+
+--
+-- Creating chi_cache table
+--
+
+CREATE TABLE `chi_cache` (
+  `key` VARCHAR(767),
+  `value` LONGTEXT,
+  PRIMARY KEY (`key`)
+);
+
+--
+-- Updating to current version
+--
+
+INSERT INTO pf_version (id, version) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION, @SUBMINOR_VERSION));
+
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -1263,6 +1263,7 @@ CREATE TABLE auth_log (
 CREATE TABLE `chi_cache` (
   `key` VARCHAR(767),
   `value` LONGTEXT,
+  `expires_at` REAL,
   PRIMARY KEY (`key`)
 );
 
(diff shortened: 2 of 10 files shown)


db/chi_cache.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE `chi_cache` (
+  `key` VARCHAR(767),
+  `value` LONGTEXT,
+  PRIMARY KEY (`key`)
```
extrafu: Shouldn't we also add a `created` column that would be set to the creation date of the entry? It could be useful to remove old entries automatically, in case the cache gets dirty over time.

julsemaan: The CHI driver itself manages to expire properly these entries and to ensure it doesn't get dirty.
So on on that side there is no gain except for the event we would need to deep debug CHI (which I  doubt we would need to do)

jrouzierinverse: move db/chi_cache.sql to db/upgrade-X.X.X-X.Y.Z.sql and db/pf-schema-X.Y.Z.sql

extrafu:
Do we want to get this in v7? It's now or never.

julsemaan:
without this the accounting cache doesn't work correctly and its a big help in environment where accounting is needed for the deauth but not for anything else

extrafu:
Ok, let's move the create table kungfu to the upgrade SQL file and we'll merge.

julsemaan:
done

julsemaan:
I did some testing around the expiration and we need a job that will go through the expired keys in the storage to remove them from there as they aren't removed by default

Would be a matter of calling this for each namespace that uses it:
```
# /usr/local/pf/bin/pfcmd cache accounting expire
```

Or to bring back the pfmon cache cleanup job

In the case of the accounting namespace it cannot grow over the amount of MAC addresses that will connect to the setup so the size can be limited in some cases, but in a hotspot environment that sees different devices every day it would grow out of control without calling the pfcmd cache expire

IMO, we can't merge without having some cleanup in place, so lets see if:
1. We want to keep this technology knowing that
2. We want to create a pfmon job or call the pfcmd cache expire in database-backup-and-maintenance.sh

extrafu:
I suggest a pfmon job to do it.

julsemaan:
@jrouzierinverse 

Can you review 95908da974359638b25af00fca1f3bb5fd249151..1f198538760fb018ed3647d3fc44554280a2eed5 which adds the expires_at field in the DB and handles the expiration in pfmon

db/pf-schema-X.Y.Z.sql
```diff
@@ -1261,3 +1255,21 @@ CREATE TABLE auth_log (
...
   KEY  attempted_at (attempted_at)
 ) ENGINE=InnoDB;
+
+--
+-- Creating chi_cache table
+--
+
+CREATE TABLE `chi_cache` (
+  `key` VARCHAR(767),
+  `value` LONGTEXT,
```
jrouzierinverse: Should be LONGBLOB not LONGTEXT.
Since we are storing binary data attempting to store it as text could cause issues.

db/pf-schema-X.Y.Z.sql
```diff
@@ -1261,3 +1255,21 @@ CREATE TABLE auth_log (
...
+  `expires_at` REAL,
```
jrouzierinverse: I would suggest to use DATETIME.
So we can use a single source of when to expire of when to expire an entry instead of the individual clients.

lib/CHI/Driver/DBI.pm
```diff
@@ -0,0 +1,313 @@
...
+
+    my $strings = {
+        fetch    => "SELECT $value FROM $table WHERE $key = ? and ? < $expires_at",
+        store    => "INSERT INTO $table ( $key, $value, $expires_at ) VALUES ( ?, ?, ? )",
+        store2   => "UPDATE $table SET $value = ?, $expires_at = ? WHERE $key = ?",
+        remove   => "DELETE FROM $table WHERE $key = ?",
+        clear    => "DELETE FROM $table where $key like $key_prefix_match",
+        get_keys => "SELECT DISTINCT $key FROM $table where $key like $key_prefix_match and ? < $expires_at",
+        create   => "CREATE TABLE IF NOT EXISTS $table ("
+          . " $key VARCHAR( 300 ), $value TEXT, $expires_at REAL,"
```
jrouzierinverse: This create should match what is in db/pf-schema-X.Y.Z.sql.

db/pf-schema-X.Y.Z.sql
julsemaan: Downside is that the now() function doesn't exist in all storages which the driver supports

Using the UNIX timestamp was the best way I found to support the different storages and keep the code simpler

I'm open to another implementation but this one works well

lib/CHI/Driver/DBI.pm
julsemaan: I fixed it without the KEY since SQLite3 doesn't support it and the driver tests are based on it

jrouzierinverse:
Just update the translation file for the admin then I will merge

julsemaan:
updaytaide

PR #1717, opened 2016-09-27
Fix 1681 iptables parking redirect

### LINKED ISSUE #1681: iptables rules for parking
While using the iptables rule for parking :
-A input-internal-vlan-if --protocol tcp --match tcp --dport 5252 --jump ACCEPT
it is not taken into account (doesn't seems to be everywhere (centos6 only for what I experienced))
in the meantime replace with
-I INPUT --protocol tcp --match tcp --dport 5252 --jump ACCEPT

dwlfrth:
# Description

Moving the INPUT RELATED,ESTABLISHED iptables at the end of the FILTER table since it is kind of act as a "default" rule.
# Impacts
- VLAN enforcement
- Inline enforcement
- Web auth enforcement
- Firewalling
# Issue

fixes #1681 
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes

- iptables rules for parking (#1681)
# UPGRADE file entries
- iptables.conf template file have been modified. Make sure to use latest template and migrate custom if required.

### DIFF
--- a/conf/iptables.conf.example
+++ b/conf/iptables.conf.example
@@ -6,8 +6,6 @@
 :INPUT DROP [0:0]
 # accept loopback stuff
 -A INPUT --in-interface lo --jump ACCEPT
-# accept anything related
--A INPUT --match state --state ESTABLISHED,RELATED --jump ACCEPT
 # Accept Ping (easier troubleshooting)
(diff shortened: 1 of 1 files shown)


dwlfrth:
Not quite sure of the "side effects" that may create... :/

dwlfrth:
Closed since #1720 now addresses it.

PR #1720, opened 2016-09-29
Fix 1681 iptables parking redirect

### LINKED ISSUE #1681: iptables rules for parking

dwlfrth:
# Description

Added a targeting rule for vlan enforcement for the interface IP even when clusterized
# Impacts

Should not have any
# Issue

fixes #1681 
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- iptables rules for parking (#1681)


dwlfrth:
Review me

extrafu:
What's the status on this?

PR #1731, opened 2016-10-03
feature/interactive-reports

jrouzierinverse:
## Description

Make the reports more interactive
## Impacts

Admin GUI reports

Enhancements
++++++++++++
- Make admin reports more interactive
## Delete branch after merge

NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
+++ b/html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
@@ -728,12 +728,7 @@ sub connectiontype :Local :AdminRole('REPORTS') {
                                    'count' => 'connections' },
                      }
                     );
-    my $items = $c->stash->{items};
-    if ($items) {
-        for my $item (@$items) {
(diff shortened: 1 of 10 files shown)


html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
```diff
@@ -725,6 +728,12 @@ sub connectiontype :Local :AdminRole('REPORTS') {
                                    'count' => 'connections' },
                      }
                     );
+    my $items = $c->stash->{items};
+    if ($items) {
+        for my $item (@$items) {
+            $item->{link} = '/admin/nodes#/node/advanced_search?searches.0.name=connection_type&searches.0.op=equal&searches.0.value=' . $pf::config::connection_type_explained_to_str{$item->{label}};
```
cgx: To reuse `_add_links`, could you define a new attribute to `$item` and use it as the `$value_key` argument?

fdurand:
Tested and works, i have just one detail, when i go on report to see SSIDs i have 9 devices on the ssid but when i click on the ssid name it return more than 50 devices. Probably not related to this PR but we have to check why.

fdurand:
My bad, wrong time period. Looks ok to me

PR #1738, opened 2016-10-05
limit httpd.portal's mod_perl to 1024MB of RAM

julsemaan:
# Description

limit httpd.portal's mod_perl to 1024MB of RAM
Prevents the process from eating up all the RAM of a server because of an infinite loop or a memory leak that occurs before the 1000 requests per child
# Impacts

httpd.portal
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Added memory limitation for httpd.portal processes


fdurand:
We already deploy it ... so merge

PR #1756, opened 2016-10-12
Set Calling-Station-Id in accounting by User-Name attribute if it is undefine

fdurand:
# Description

Set Calling-Station-Id if it doesn't exist by the User-Name attribute if it represent a mac address.
# Impacts

RADIUS Accounting
# Issue

Fix empty Calling-Station-Id in radius accounting
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix empty Calling-Station-Id in radius accounting

### DIFF
--- a/conf/radiusd/packetfence.example
+++ b/conf/radiusd/packetfence.example
@@ -249,6 +249,9 @@ preacct {
 #  Accounting.  Log the accounting data.
 #
 accounting {
+	rewrite_user_name
+	rewrite_calling_station_id
+	rewrite_called_station_id
 	# Add in PacketFence specific configuration
 	update {
 		&request:FreeRADIUS-Client-IP-Address := "%{Packet-Src-IP-Address}"
@@ -257,9 +260,8 @@ accounting {
 		&control:PacketFence-RPC-User = ${rpc_user}
 		&control:PacketFence-RPC-Pass = ${rpc_pass}
 		&control:PacketFence-RPC-Proto = ${rpc_proto}
+		&request:Calling-Station-Id = "%{User-Name}"
 	}
-	rewrite_calling_station_id
-	rewrite_called_station_id
 	#
 	#  Log traffic to an SQL database.
 	#
--- a/conf/radiusd/packetfence.example
+++ b/conf/radiusd/packetfence.example
@@ -251,7 +251,7 @@ preacct {
 accounting {
 	rewrite_user_name
 	rewrite_calling_station_id
-	rewrite_called_station_id
+	set_calling_station_id
 	# Add in PacketFence specific configuration
 	update {
 		&request:FreeRADIUS-Client-IP-Address := "%{Packet-Src-IP-Address}"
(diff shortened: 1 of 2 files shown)


conf/radiusd/packetfence.example
```diff
@@ -257,9 +260,8 @@ accounting {
 		&control:PacketFence-RPC-User = ${rpc_user}
 		&control:PacketFence-RPC-Pass = ${rpc_pass}
 		&control:PacketFence-RPC-Proto = ${rpc_proto}
+		&request:Calling-Station-Id = "%{User-Name}"
```
julsemaan: Unless I don't understand something from this logic but it seems like its going to set the Calling-Station-Id to the User-Name in the request whether it looks like a MAC or not, meaning if its a 802.1x connection with username 'lzammit', its going to put 'lzammit' in the Calling-Station-Id.

Only god knows what will happen if 'lzammit' becomes the value of the Calling-Station-Id

julsemaan: I think I found a picture of the best case:

image

fdurand: In fact it will set the Calling-Station-Id attribute to the user-name if only Calling-Station-Id is undef.
Btw you are right, i will change the policy to something like that:
rewrite_user_name {
         if (&User-Name && (&User-Name =~ /^${policy.mac-addr-regexp}$/i)) {
                 update request {
                         &User-Name := "%{tolower:%{1}:%{2}:%{3}:%{4}:%{5}:%{6}}"
                         &Calling-Station-Id =  "%{tolower:%{1}:%{2}:%{3}:%{4}:%{5}:%{6}}"
                 }
                 updated
         }
         else {
                 noop
         }
 }
And remove &request:Calling-Station-Id = "%{User-Name}"

raddb/policy.d/canonicalization
```diff
@@ -111,3 +111,16 @@ rewrite_calling_station_id {
                 noop
         }
 }
+
+set_calling_station_id {
+        if (&User-Name && (&User-Name =~ /^${policy.mac-addr-regexp}$/i)) {
```
louismunro, line 116: I would be careful here.
policy.mac-addr-regexp would also match "aafacceebb01"  which could be a user.
You could overwrite a valid username.

fdurand: Yes sure, btw this is only use in accounting.
OK so let's start the name challenge, we have to find a username with a or b or c or d or e or f with 12 characters long !

extrafu:
Status?

raddb/policy.d/canonicalization
julsemaan: I think one of Zayme's cousin is named Abdea Befafa
Which could translate to username abdeabefafa1

Jokes aside, I tend to agree with @fdurand on that point where it is extremely unlikely that will happen and since a few controllers tend to send «vidange requests» then we will fix 99.999999% of the cases and break the remaining. 

That seems like a good trade-off and likely the only one we can make as I don't see another way to address this.

louismunro: All right then. 
Don't come to me when Mr. Befafa complains though :wink:

PR #1768, opened 2016-10-19
Changing default dhcp types to extract and interpret to DHCPACK and D…

### LINKED ISSUE #1747: PacketFence should only process DHCPREQUESTS and DHCPACKS?
UDP Reflector should be launched to neglect discovers/offers (server port 67 <-> dhcp client port 68) and only receive client's confirmation (ACKS) of the accepted lease sent to server port 68  from dhcp client port 67](https://www.linux.com/blog/troubleshooting-connection-problems-tcpdump-dhcp).

If I understand well, the behavior of UDP reflector and ip helpers would then be the same and the choice between the two tools would not be linked to the difference of behavior.

Actual [recommended](https://packetfence.org/doc/PacketFence_Administration_Guide.html#_dhcp_remote_sensor) line:
`C:\udp-reflector\udp_reflector.exe -s pcap0:67 -d 192.168.1.5:767 -b 25000`

What if:
`C:\udp-reflector\udp_reflector.exe -s pcap0:68 -d 192.168.1.5:767 -b 25000`

What do you think?

tlaurion:
# Description

Changing the defaults for pfdhcplistener to only extract DHCPACK and DHCPREQUEST so that no processing and no logging is done when other types are captured.
# Impacts

Only DHCPACK and DHCPREQUEST will be processed and logged.
# Issue

fixes #1747 
# Delete branch after merge

YES
## Enhancements
- phdhcplistenr consumes less ressources since it doesn't look for dhcp subtypes it is not interested in. 
# UPGRADE file entries

If you want pfdhcplistener to parse and log about all dhcp packets received, change your pf.conf configuration back to the old value:
dhcp_filter_by_message_types=DHCPOFFER,DHCPREQUEST,DHCPDECLINE,DHCPACK,DHCPRELEASE

Else, only DHCPREQUEST and DHCPACK packets will be processed and corresponding logs generated for those packet types.

### DIFF
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -63,7 +63,7 @@ dhcpoption82logger=disabled
 #
 # Which dhcp message types are being listened by the pfdhcplistener.
 # Do not change unless you know what you are doing
-#Possitibilities are: DHCPOFFER,DHCPREQUEST,DHCPDECLINE,DHCPACK,DHCPRELEASE
+#Possibilities are: DHCPOFFER,DHCPREQUEST,DHCPDECLINE,DHCPACK,DHCPRELEASE
 dhcp_filter_by_message_types=DHCPREQUEST,DHCPACK
 #
 # network.force_listener_update_on_ack
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -63,7 +63,8 @@ dhcpoption82logger=disabled
 #
 # Which dhcp message types are being listened by the pfdhcplistener.
 # Do not change unless you know what you are doing
-dhcp_filter_by_message_types=DHCPOFFER,DHCPREQUEST,DHCPDECLINE,DHCPACK,DHCPRELEASE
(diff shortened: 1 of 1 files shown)


conf/pf.conf.defaults
```diff
@@ -63,7 +63,8 @@ dhcpoption82logger=disabled
 #
 # Which dhcp message types are being listened by the pfdhcplistener.
 # Do not change unless you know what you are doing
-dhcp_filter_by_message_types=DHCPOFFER,DHCPREQUEST,DHCPDECLINE,DHCPACK,DHCPRELEASE
+#Possitibilities are: DHCPOFFER,DHCPREQUEST,DHCPDECLINE,DHCPACK,DHCPRELEASE
```
julsemaan: *Possibilities

extrafu: Possitibilities are infinite possibilities! I love it!

tlaurion: Voilà!

PR #1770, opened 2016-10-19
Fixes for WMI

julsemaan:
# Needs to be carefully tested
# Description

Ensure a post-registration scan will occur even if device hasn't gone through portal registration
Fix scan engine instantiation that created invalid scan objects
# Impacts

Scanning + DHCP processing
# Delete branch after merge

YES
# NEWS file entries
## Bug Fixes
- Fix scan engine instantiation resulting in post-registration scans breaking
- Fix post-registration scan not triggering when using 802.1x

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -947,6 +947,8 @@ sub trigger_scan : Public : Fork {
     # post_registration (production vlan)
     # We sleep until (we hope) the device has had time issue an ACK.
     if (pf::util::is_prod_interface($postdata{'net_type'})) {
+        # We add the violation to ensure there is one opened
+        pf::violation::violation_add( $postdata{'mac'}, $pf::constants::scan::POST_SCAN_VID );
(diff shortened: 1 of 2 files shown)


lib/pf/api.pm
```diff
@@ -947,6 +947,11 @@ sub trigger_scan : Public : Fork {
     # post_registration (production vlan)
     # We sleep until (we hope) the device has had time issue an ACK.
     if (pf::util::is_prod_interface($postdata{'net_type'})) {
+        my $profile = pf::Portal::ProfileFactory->instantiate($postdata{'mac'});
+        my $scanner = $profile->findScan($postdata{'mac'});
+        if (defined($scanner) && pf::util::isenabled($scanner->{'post_registration'})) {
+            pf::violation::violation_add( $postdata{'mac'}, $pf::constants::scan::POST_SCAN_VID );
```
fdurand, line 953: Make sense since the violation can't be add twice.

fdurand:
Agree to merge

PR #1776, opened 2016-10-19
Fix/wmi tab

### LINKED ISSUE #1771: WMI node tab: When WMI call fails, it shows template not found
In my case, the username/password is not valid in the AD

See screenshot below:

image

whitx:
# Description

Make sure WMI tab is only available for Windows and fixes some issues with the tab WMI
# Impacts

WMI tab
# Issue

fixes #1771 
# Delete branch after merge

NO

Probably should go to maintenance since at the moment there is a template toolkit error if there is an "Access denied"

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
@@ -104,7 +104,9 @@ sub parseWmi {
     }
     $scan_config->{_scanIp} = $host;
     my $scan_result = $scan->runWmi($scan_config, $rule_config);
-    if ($scan_result =~ /0x80041010/ || !@$scan_result) {
+    if ($scan_result =~ /ACCESS_DENIED/) {
+        $rule_config->{item_exist} = "Access denied";
+    }elsif ($scan_result =~ /0x80041010/ || !@$scan_result) {
         $rule_config->{item_exist} = 'No';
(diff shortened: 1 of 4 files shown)


html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
```diff
@@ -103,7 +104,9 @@ sub parseWmi {
     }
     $scan_config->{_scanIp} = $host;
     my $scan_result = $scan->runWmi($scan_config, $rule_config);
-    if ($scan_result =~ /0x80041010/ || !@$scan_result) {
+    if ($scan_result =~ /ACCESS_DENIED/) {
+        $rule_config->{item_exist} = "Access denied";
+    }elsif ($scan_result =~ /0x80041010/ || !@$scan_result) {
```
fdurand: Constant

html/pfappserver/root/node/tab_WMI_process.tt
```diff
@@ -1,5 +1,5 @@
 [% IF rules.id == "Process_Running" %]
-  [% IF rules.item_exist == "No" %]
+  [% IF rules.item_exist == "No" || rules.item_exist == "Access denied" %]
     <span> Error while getting running process. [% rules.item_exist | html %]</span><br/>
```
fdurand: Can you localize string ? [% l(' Error while getting running process. ') %]
Item scanned, Is Present, Name, Version, too. (in this file)

PR #1782, opened 2016-10-20
Removed login for dns filter (can be done in the rule)

fdurand:
# Description

Removed log when a dns filter match ( should be enabled in the rule itself)
# Impacts

pfdns log
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- Remove useless log in pfdns.log


extrafu:
@tlaurion  and @whitx - bump

PR #1786, opened 2016-10-20
feature/ha_int_for_reals

jrouzierinverse:
## Description

Store the interface object instead of the interface name for @pf::config::ha_ints
## Impacts

iptables generation and pf admin interface editing
## Delete branch after merge

NO


tlaurion:
Can't use string ("eth0") as a HASH ref while "strict refs" in use at /usr/local/pf/lib/pf/iptables.pm line 245.

jrouzierinverse:
Do a configreload

tlaurion:
Reason why it was not working was because of a typo in high_availability (high_availability). a config reload makes it work between code changes.

tlaurion:
@fdurand? I'll do a pull request that depends upon this to be merged in devel.

PR #1787, opened 2016-10-21
feature/double-auth-match

### LINKED ISSUE #1279: Combine the calls for matching rules role and regdate
Since the rules are now split between auth and admin it is now possible to combine the calls to pf::authentication::match for the role and regdate actions.

jrouzierinverse:
## Description

Avoid the calling pf::authentication::match twice
## Impacts

Everything
## Issue

Fixes #1279
## Delete branch after merge

NO

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -1054,10 +1054,10 @@ sub dynamic_register_node : Public {
     }
     my $values = $matched->{values};
     my $role = $values->{$Actions::SET_ROLE};
-    my $value = $values->{$Actions::SET_UNREG_DATE};
-    if (defined $value) {
+    my $unregdate = $values->{$Actions::SET_UNREG_DATE};
+    if (defined $unregdate) {
         my %info = (
-            'unregdate' => $value,
+            'unregdate' => $unregdate,
             'category' => $role,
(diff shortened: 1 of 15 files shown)


lib/pf/api.pm
```diff
@@ -1040,9 +1040,14 @@ sub dynamic_register_node : Public {
...
-    #Compute autoreg if we use autoreg
-    my $value = &pf::authentication::match([@sources], $params, $Actions::SET_UNREG_DATE);
+    my $matched = pf::authentication::match2([@sources], $params);
+    unless ($matched) {
+        $logger->warn("Did not find any actions to match");
+        return;
+    }
+    my $values = $matched->{values};
+    my $role = $values->{$Actions::SET_ROLE};
+    my $value = $values->{$Actions::SET_UNREG_DATE};
```
julsemaan: This would be clearer if the variable was called $unregdate

julsemaan:
This code looks pretty good given the code paths it plays into.

Would require testing before merge but other than the minor detail above, it's good.

jrouzierinverse:
Rebase please review

extrafu:
@julsemaan Review please!

PR #1788, opened 2016-10-21
Feature/p0f with drbd bpf

### LINKED ISSUE #1745: P0f should be launched with a bpf discarding DRBD traffic
P0f shuld be launched with a bpf discarding DRBD traffic (ip address linked to ha_interface)

tlaurion:
# Description
- Requires https://github.com/jrouzierinverse/packetfence/tree/feature/ha_int_for_reals
- If "high-availability" is defined on interfaces in cluster.conf, extract the ip adresses of those interfaces and exclude DRBD port in a bpf at p0f launch.
- Modify PacketFence_Cluster docs part saying that high-availability is required: high-avail in pf.conf is there only for consistency in regard of cluster.conf
# Impacts
- DRBD traffic will not be inspected by p0f.
# Issue

fixes #1745 
# Delete branch after merge

YES
# NEWS file entries
## Enhancements
- more lightweight P0f processing, excluding DRDB traffic for high-availibility interfaces.

### DIFF
--- a/lib/pf/services/manager/p0f.pm
+++ b/lib/pf/services/manager/p0f.pm
@@ -47,7 +47,7 @@ has '+launcher' => (
         {
             $p0f_cmdline="sudo %1\$s -d -i any -p -f $p0f_map -s $p0f_sock > /dev/null && pidof $name > $pid_file";    
         }
-        print $p0f_cmdline;
+        return $p0f_cmdline;
     }
 );
 
--- a/lib/pf/services/manager/p0f.pm
+++ b/lib/pf/services/manager/p0f.pm
@@ -31,6 +31,7 @@ has '+launcher' => (
         my $p0f_sock = $FingerbankConfig->{tcp_fingerprinting}{p0f_socket_path};
         my $pid_file = $self->pidFile;
         my $name = $self->name;
+        my $p0f_cmdline;
         if (@ha_ints)
         {
             my @ha_ips;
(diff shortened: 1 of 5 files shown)


lib/pf/services/manager/p0f.pm
```diff
@@ -30,7 +31,23 @@ has '+launcher' => (
...
+            }
+            my @tmp_bpf_filter = map { "( host $_ and port 7788)"} @ha_ips;
+            my $p0f_bpf_filter = join(" or not", @tmp_bpf_filter);
+            $p0f_cmdline="sudo %1\$s -d -i any -p -f $p0f_map -s $p0f_sock" . " 'not $p0f_bpf_filter' " . " > /dev/null && pidof $name > $pid_file";
+        }
+        else
+        {
+            $p0f_cmdline="sudo %1\$s -d -i any -p -f $p0f_map -s $p0f_sock > /dev/null && pidof $name > $pid_file";    
+        }
+        print $p0f_cmdline;
```
julsemaan: Your issue is because you're printing the command instead of returning the launcher string

So change this by : 
`return $p0f_cmdline;`

And let the magic happen

jrouzierinverse: After you fix the what @julsemaan recommended I will review.

extrafu:
@jrouzierinverse Have you completed your review? We need this in ASAP.

lib/pf/services/manager/p0f.pm
```diff
@@ -30,7 +31,23 @@ has '+launcher' => (
...
+        my $p0f_cmdline;
+        if (@ha_ints)
+        {
+            my @ha_ips;
+            foreach my $ha_int (@ha_ints)
+            {
+                push (@ha_ips, $ha_int->{Tip});
+            }
+            my @tmp_bpf_filter = map { "( host $_ and port 7788)"} @ha_ips;
+            my $p0f_bpf_filter = join(" or not", @tmp_bpf_filter);
```
jrouzierinverse: I think that should be ' and not '.

lib/pf/services/manager/p0f.pm
```diff
@@ -30,7 +31,23 @@ has '+launcher' => (
...
+        if (@ha_ints)
+        {
+            my @ha_ips;
+            foreach my $ha_int (@ha_ints)
+            {
+                push (@ha_ips, $ha_int->{Tip});
+            }
+            my @tmp_bpf_filter = map { "( host $_ and port 7788)"} @ha_ips;
+            my $p0f_bpf_filter = join(" and not", @tmp_bpf_filter);
+            $p0f_cmdline="sudo %1\$s -d -i any -p -f $p0f_map -s $p0f_sock" . " 'not $p0f_bpf_filter' " . " > /dev/null && pidof $name > $pid_file";
```
jrouzierinverse: One more thing.
I would prefer the filter to built this way since to avoid confusion if we have to refactor it later.

```
my @tmp_bpf_filter = map { "not ( host $_ and port 7788)"} @ha_ips;
my $p0f_bpf_filter = join(" and ", @tmp_bpf_filter);
$p0f_cmdline="sudo %1\$s -d -i any -p -f $p0f_map -s $p0f_sock" . " '$p0f_bpf_filter' " . " > /dev/null && pidof $name > $pid_file";

```

PR #1792, opened 2016-10-25
fix/service-status

### LINKED ISSUE #1040: pfdhcplistener status is bogus in the web admin
In the web admin, pfdhcplistener is only listed once, even if multiple dhcplistener are configured. This is fine but if one is going down, the pfdhcplistener entry still state as "Started". The only way to see if all the pfdhcplistener are started would be to use the command line 'service packetfence status'.

Solutions:
- List all of the pfdhcplistener in the status page with their own status
- Keep only one entry but if one is down, it should not report as "Started"
- Fancy listing with a + sign beside pfdhcplistener to list all of the configured ones along with their status

jrouzierinverse:
## Description

Display the sub managers also
## Impacts

Admin GUI
## Issue

Fixes #1040

Enhancements
++++++++++++
- Display the status for sub services also
## Delete branch after merge

NO


jrouzierinverse:
Please Review

dwlfrth:
LGTM

PR #1806, opened 2016-11-07
Disabled .htaccess for all apache services

fdurand:
# Description
Removed the access to .htaccess for each http request (all apache process)

# Impacts
apache process

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Remove useless htaccess search for each http request


extrafu:
@jrouzierinverse Please review quickly and merge if all right.

PR #1824, opened 2016-11-11
Protecting VRRP against attacks

fdurand:
# Description
Protecting VRRP against attacks
Changed from plain text to IPSEC-AH in keepalived in order to protect vrrp traffic.

# Impacts
keepalived

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Protecting VRRP against attacks


louismunro:
Good enough.

PR #1826, opened 2016-11-14
Person Lookup Cache

fdurand:
# Description
Cache the result of person lookup for 1h, it will reduce a lot the ldap search.
Optimize call to person_add in order to reduce the stress on the db.

# Impacts
pfqueue

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Cache person_lookup to reduce the call to ldap_search

### DIFF
--- a/conf/chi.conf.defaults
+++ b/conf/chi.conf.defaults
@@ -78,3 +78,10 @@ reconnect=60
 #
 # Amount of time from the current time to expire an entry
 expires_in=1d
+
+[storage person_lookup]
+#
+# storage person_lookup.expires_in
+#
+# Amount of time from the current time to expire an entry
+expires_in=10m
--- a/conf/chi.conf.defaults
+++ b/conf/chi.conf.defaults
@@ -84,4 +84,4 @@ expires_in=1d
 # storage person_lookup.expires_in
 #
 # Amount of time from the current time to expire an entry
-expires_in=1h
+expires_in=24h
--- a/conf/chi.conf.defaults
+++ b/conf/chi.conf.defaults
@@ -84,4 +84,4 @@ expires_in=1d
 # storage person_lookup.expires_in
 #
 # Amount of time from the current time to expire an entry
-expires_in=10m
+expires_in=1h
--- a/lib/pf/lookup/person.pm
+++ b/lib/pf/lookup/person.pm
@@ -51,7 +51,7 @@ sub lookup_person {
         return undef;
     }
     my $person = $CHI_CACHE->get("$source_id.$pid");
-    unless($reply){
+    unless($person){
         $person = $source->search_attributes($pid);
         if (!$person) {
            $logger->debug("Cannot search attributes for user '$pid'");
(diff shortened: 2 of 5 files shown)


conf/chi.conf.defaults
```diff
@@ -78,3 +78,10 @@ reconnect=60
 #
 # Amount of time from the current time to expire an entry
 expires_in=1d
+
+[storage person_lookup]
+#
+# storage person_lookup.expires_in
+#
+# Amount of time from the current time to expire an entry
+expires_in=1h
```
julsemaan: I'd even go with 24 hours and maybe even more, its not like your lastname will change every hour.

Unless you're called Zammit/Zayme/Zaymitte/ZiCEO

extrafu:
@jrouzierinverse Review please.

lib/pf/lookup/person.pm
```diff
@@ -36,26 +39,31 @@ sub lookup_person {
...
 
     unless (person_exist($pid)) {
-        return "Person $pid is not a registered user!\n";
+        return undef;
     }
-    my ($person, $msg) = $source->search_attributes($pid);
-    if (!$person) {
-       $logger->info("Cannot search attributes for user '$pid'");
-       return "Cannot search attributes for user '$pid'!\n";
+    my $person = $CHI_CACHE->get("$source_id.$pid");
```
jrouzierinverse, line 54: use compute

lib/pf/lookup/person.pm
```diff
@@ -36,26 +39,31 @@ sub lookup_person {
...
     my $source = pf::authentication::getAuthenticationSource($source_id);
     if (!$source) {
        $logger->info("Unable to locate the source $source_id");
-       return "Unable to locate the source $source_id!\n";
+       return undef;
     }
 
     unless (person_exist($pid)) {
-        return "Person $pid is not a registered user!\n";
+        return undef;
```
jrouzierinverse: Log the previous message

lib/pf/lookup/person.pm
```diff
@@ -36,26 +39,31 @@ sub lookup_person {
...
+            $logger->info("Successfully did a person lookup for $pid");
+            person_modify($pid, %$person);
+            return undef;
+        }
     }
-
-    $logger->info("Successfully did a person lookup for $pid");
-
-    person_modify($pid, %$person);
+    $logger->info("Already did a person lookup for $pid");
```
jrouzierinverse, line 67: return a true value

lib/pf/lookup/person.pm
fdurand: If i use compute then each time the person has been found in the cache then i will do a person_modify and i don't want to do it.

lib/pf/lookup/person.pm
jrouzierinverse: Why are you returning a value when the value is not used?

lib/pf/lookup/person.pm
```diff
@@ -36,26 +39,33 @@ sub lookup_person {
     my $logger = get_logger();
     unless (defined $source_id) {
         $logger->info("undefined source id provided");
-        return "Unknown source";
+        return undef;
```
jrouzierinverse: missed this one

PR #1830, opened 2016-11-14
Improvements to DB backups: last_backup file, no double compression.

### LINKED ISSUE #1827: Database back and maintenance script not cleaning the db backup files
The db back done with persona xtrabackup are not removed after the $NB_DAYS_TO_KEEP_DB.

louismunro:
# Description
Minor issues fixed with the backup script.
This removes double compression on the xbstream if using innobackupex.
Also add a touchfile in /usr/local/pf/var/run/last_backup that indicates when the last sucessful backup happened.
Useful for monitoring.

# Impacts
See above.

# Issue
If the disk fills up there is no warning that the backups cannot be taken normally.
This should allow monitoring of the last backup timestamp.
fixes #1827 

# Delete branch after merge
YES

# NEWS file entries

# Enhancement
* Database backups can now be monitored for success through /usr/local/pf/var/run/last_backup

### DIFF
--- a/addons/database-backup-and-maintenance.sh
+++ b/addons/database-backup-and-maintenance.sh
@@ -90,21 +90,25 @@ if [ -f /var/run/$SQL_ENGINE/$SQL_ENGINE.pid ]; then
     MYSQL_USED_SPACE=`du -s /var/lib/mysql | awk '{ print $1 }'`
     if (( $BACKUPS_AVAILABLE_SPACE > (( $MYSQL_USED_SPACE /2 )) )); then 
         if [ $PERCONA_XTRABACKUP_INSTALLED -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


julsemaan:
Does that mean the script we installed during the rollout was b0rken ?

louismunro:
Not really, but it makes recovery more complicated than it should be.

julsemaan:
As long as it works, we're good, just don't want to have the backups stopped by a full disk

extrafu:
@julsemaan That has never happened in the past. 🤕 🔫

julsemaan:
@extrafu, obviously that never happened !!
I was just saying that for prevention

Please don't stop paying me like that other french guy!!

louismunro:
@julsemaan You don't want the backups blocked by a full disk?
What should they do then?
Overwrite the previous backups?
Or perhaps overwrite those useless files in /chroot

julsemaan:
Our current strategy is good (stopping).

From #1827 it sounded like the backups were never cleaned which would have filled the disk. Just wanted to make sure that this cleanup bug wasn't part of the script we rolled out. Otherwise I would have gone back on the setups and updated them.

PR #1831, opened 2016-11-14
Adding fault tolerance on RADIUS connection check

dwlfrth:
# Description
Adding fault tolerance to avoid false-positive when there is a certain load on the server

# Impacts
Monitoring

# Delete branch after merge
YES

### DIFF
--- a/addons/monit/monit_configurations/monit_general.tt
+++ b/addons/monit/monit_configurations/monit_general.tt
@@ -22,11 +22,13 @@ A copy of this alert have been sent to the PacketFence support team.
 # - Do not alert on 'nonexist' because we alert on 'restart'
 # - Do not alert on 'pid' because we alert on 'restart
 # - Do not alert on 'content' because there is a separate alert configuration with reminders
+# - Do not alert on 'connection' because there is a separate alert configuration with reminders
 [% FOREACH email IN EMAILS %]
-    set alert [% email %] but not on { resource nonexist pid content }
-    set alert [% email %] { resource } with reminder on 450 cycle   # reminder every 15 minutes (450 cycles) if 1 cycle is 2 seconds
+    set alert [% email %] but not on { resource nonexist pid content connection }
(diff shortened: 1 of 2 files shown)


dwlfrth:
Review me

extrafu:
Looks good to me, @julsemaan review and merge.

addons/monit/monit_configurations/monit_general.tt
```diff
@@ -22,11 +22,13 @@ A copy of this alert have been sent to the PacketFence support team.
 # - Do not alert on 'nonexist' because we alert on 'restart'
 # - Do not alert on 'pid' because we alert on 'restart
 # - Do not alert on 'content' because there is a separate alert configuration with reminders
+# - Do not alert on 'connection' because there is a separate alert configuration with reminders
 [% FOREACH email IN EMAILS %]
-    set alert [% email %] but not on { resource nonexist pid content }
-    set alert [% email %] { resource } with reminder on 450 cycle   # reminder every 15 minutes (450 cycles) if 1 cycle is 2 seconds
+    set alert [% email %] but not on { resource nonexist pid content connection }
+    set alert [% email %] { resource } with reminder on 450 cycle       # reminder every 15 minutes (450 cycles) if 1 cycle is 2 seconds
```
julsemaan, line 28: Just to make sure, is it normal you added 'connection' there ?
Is it a monit keyword ?
And that looks like it will flush the alerts on the connection alerts. If so, I don't see the point of that test at all

julsemaan:
Other than the comment of me probably just not understanding stuff (business as usual), we're clear

addons/monit/monit_configurations/monit_general.tt
dwlfrth: Yep... As mentioned in the comment a couple line over, removing the standard alert configuration for "connection" since I redefined it couple lines below (31) to add a reminder.
That means, rather than alerting only when the connection fails and when it comes back, it will alert every minute if it stays "failed"

PR #1841, opened 2016-11-16
Honor "controllerPort" configuration parameter in WLC radiusDisconnect

### LINKED ISSUE #1840: Cisco WLC do not honor controllerPort for CoA/Disconnect
If a "controllerPort" is configured in a WLC switch configuration, the radius "radiusDisconnect" method will not honor it.

dwlfrth:
# Description
See issue (#1840)

# Impacts
WLC radiusDisconnect

# Issue
fixes #1840 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Cisco WLC do not honor controllerPort for CoA/Disconnect (#1840)

### DIFF
--- a/lib/pf/Switch/Cisco/WLC.pm
+++ b/lib/pf/Switch/Cisco/WLC.pm
@@ -490,6 +490,10 @@ sub radiusDisconnect {
             nas_port => $nas_port,
         };
 
+        if (defined($self->{'_controllerPort'}) && $self->{'_controllerPort'} ne '') {
+            $connection_info->{'nas_port'} = $self->{'_controllerPort'};
+        }
+
         $logger->debug("network device (".$self->{'_id'}.") supports roles. Evaluating role to be returned");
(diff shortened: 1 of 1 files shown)


dwlfrth:
Review me

julsemaan:
The code seems useless :P 

Its already setting the nas_port using the controllerPort attribute

I might be reading it wrong though but it really seems you're setting an attribute that is already set

Are you sure #1840 applies to devel ?

lib/pf/Switch/Cisco/WLC.pm
```diff
@@ -542,6 +546,11 @@ sub radiusDisconnect {
                 LocalAddr => $self->deauth_source_ip(),
                 nas_port => $nas_port,
```
julsemaan, line 547: Its being set here...

dwlfrth:
As @julsemaan said, do not apply to 'devel'.
@dwlfrth... you're drunk... go home

PR #1845, opened 2016-11-17
Feature/rate limit dhcp

### LINKED ISSUE #1722: pfdhcplistener rate-limiting
pfdhcplistener is currently "vulnerable" to "mass-trafic" attacks (it is actually more pfqueue but POE is pfdhcplistener). The daemon will receive DHCP packets then forwards them to the appropriate queue which may create a big processing queue in the case a device is sending a lot of DHCP discover / request.

pfdhcplistener should be able to "rate-limit" itself based on number of request from a same device per X time.

julsemaan:
# Description
Rate limit the pfdhcplistener
Move the decoding logic back in the pfdhcplistener (since it has to take decision on the L4 payload)

# Impacts
pfdhcplistener

# Issue
fixes #1722 

# Delete branch after merge
NO (to backport on setup)

# NEWS file entries
## Enhancements
* Rate limited the DHCP listener processes to prevent specific device(s) from performing a denial of service on the DHCP listening processes (#1722)

### DIFF
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -35,6 +35,14 @@ timezone=America/Montreal
 # meaning a span port is not necessary.  This feature is highly recommended if the internal network is DHCP-based.
 dhcpdetector=enabled
 #
+# network.dhcp_rate_limiting
+#
+# Will rate-limit DHCP packets that contain the same information.
+# For example, a DHCPREQUEST for the same MAC/IP will only be processed once in the timeframe configured below.
+# This is independant of the DHCP server/relay handling the packet and is only based on the IP, MAC Address and DHCP type inside the packet.
+# A value of 0 will disable the rate limitation.
+dhcp_rate_limiting=5s
+#
 # network.rogue_dhcp_detection
 #
 # Tries to identify Rogue DHCP Servers and triggers the 1100010 violation if one is found.
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -107,6 +107,9 @@ my $pcap;
 my $net_type;
 my $process;
 
+my $rate_limit_hash = {};
+my $rate_limit_cache = CHI->new( driver => 'Memory', datastore => $rate_limit_hash );
+
 sub reload_config {
     if ( $net_type eq "internal" ) { 
         $process = $TRUE;
@@ -258,6 +261,41 @@ sub process_pkt {
                 };
                 return if (!$success);
                 $args{dhcp} = $dhcp;
+                
+                # adding to dhcp hashref some frame information we care about
+                $dhcp->{'src_mac'} = $args{'src_mac'};
+                $dhcp->{'dest_mac'} = $args{'dest_mac'};
+                $dhcp->{'src_ip'} = $args{'src_ip'};
+                $dhcp->{'dest_ip'} = $args{'dest_ip'};
+
+
+                # grab DHCP information
+                if ( !defined($dhcp->{'chaddr'}) ) {
+                    $logger->debug("chaddr is undefined in DHCP packet");
+                    return;
+                }
+
+                $dhcp->{'chaddr'} = clean_mac( substr( $dhcp->{'chaddr'}, 0, 12 ) );
+
+                if (!valid_mac($dhcp->{'src_mac'})) {
+                    $logger->debug("Source MAC is invalid. skipping");
+                    return;
+                }
+
+                if ( $dhcp->{'chaddr'} ne "00:00:00:00:00:00" && !valid_mac($dhcp->{'chaddr'}) ) {
+                    $logger->debug( sub {
+                        "invalid CHADDR value ($dhcp->{'chaddr'}) in DHCP packet from $dhcp->{src_mac} ($dhcp->{src_ip})"
+                    });
+                    return;
+                }
+
+
+                my $dhcp_mac = $dhcp->{'chaddr'};
+                # If its a DHCPREQUEST, we take the IP address from the options
+                my $dhcp_ip = ( $dhcp->{'options'}{'53'} == 3 ) ? $dhcp->{'options'}{'50'} : $dhcp->{yiaddr};
+                # Key is with BOOT type, DHCP type, IP and MAC
+                my $rate_limit_key = $dhcp->{'op'} . "-" . $dhcp->{'options'}{'53'} . "-$dhcp_ip-" . $dhcp_mac;
+                $logger->trace("Rate limit key : $rate_limit_key");
 
                 if ($is_inline_vlan) {
                     my $ip = new NetAddr::IP::Lite clean_ip($l3->{'src_ip'});
@@ -267,8 +305,14 @@ sub process_pkt {
                         $args{inline_sub_connection_type} = $NET_TYPE_INLINE_L3;
                    }
                 }
-                my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
-                $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
+                if(!$rate_limit_cache->get($rate_limit_key)) {
+                    my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
+                    $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
+                    $rate_limit_cache->set($rate_limit_key, 1, "5 second");
+                }
+                else {
+                    $logger->info("Ignoring packet due to rate-limiting");
+                }
             }
             else {
                 my $apiclient = pf::api::queue->new;
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -262,31 +262,29 @@ sub process_pkt {
                 return if (!$success);
                 $args{dhcp} = $dhcp;
                 
+                my $dhcp_mac = clean_mac( substr( $dhcp->{'chaddr'}, 0, 12 ) );
+                if ( $dhcp_mac ne "00:00:00:00:00:00" && !valid_mac($dhcp_mac) ) {
+                    $logger->debug( sub {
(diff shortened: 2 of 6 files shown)


julsemaan:
This is ready for review.

Opinions wanted on whether or not this should be on by default (right now it rate limits the packets for 5 seconds by default)

sbin/pfdhcplistener
```diff
@@ -105,6 +107,9 @@ my $pcap;
 my $net_type;
 my $process;
 
+my $rate_limit_hash = {};
+my $rate_limit_cache = CHI->new( driver => 'Memory', datastore => $rate_limit_hash );
```
fdurand, line 111: Is there a way to see what is in the cache with pfcmd cache ...

conf/pf.conf.defaults
```diff
@@ -35,6 +35,14 @@ timezone=America/Montreal
 # meaning a span port is not necessary.  This feature is highly recommended if the internal network is DHCP-based.
 dhcpdetector=enabled
 #
+# network.dhcp_rate_limiting
+#
+# Will rate-limit DHCP packets that contain the same information.
+# For example, a DHCPREQUEST for the same MAC/IP will only be processed once in the timeframe configured below.
+# This is independant of the DHCP server/relay handling the packet and is only based on the IP, MAC Address and DHCP type inside the packet.
+# A value of 0 will disable the rate limitation.
+dhcp_rate_limiting=5s
```
fdurand, line 44: maybe something lower than the maximum we set for dhcp lease time (29s)

julsemaan: The thing is we could miss some renewals and if someone sets the DHCP lease to 15 seconds, it won't adjust this automatically, so for the default value I wanted to keep something that would work for all the cases and wouldn't create a hard to debug issue.

sbin/pfdhcplistener
julsemaan: I'll change it to use a pf::CHI namespace

sbin/pfdhcplistener
```diff
@@ -252,8 +305,17 @@ sub process_pkt {
...
+                my $rate_limiting = $Config{network}->{dhcp_rate_limiting};
+                if($rate_limiting == 0 || !$rate_limit_cache->get($rate_limit_key)) {
+                    my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
+                    $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
+                    if($rate_limiting != 0) {
+                        $rate_limit_cache->set($rate_limit_key, 1, $rate_limiting);
+                    }
+                }
+                else {
+                    $logger->info("Ignoring packet due to rate-limiting");
```
fdurand: $logger->info("[mac:$dhcp_mac] Ignoring packet due to rate-limiting");

sbin/pfdhcplistener
julsemaan: Just looked into it and if we want to have this available through pfcmd then we can't use the memory storage for obvious reasons..

Now the question is whether we prefer using Redis or the memory storage. I put it in memory for faster access and reduced load on Redis so we'll need to discard this advantage if we want access through pfcmd

fdurand: IMO keep it faster as possible

sbin/pfdhcplistener
```diff
@@ -244,6 +249,54 @@ sub process_pkt {
...
+                    $logger->debug( sub {
+                        "invalid CHADDR value ($dhcp->{'chaddr'}) in DHCP packet from $dhcp->{src_mac} ($dhcp->{src_ip})"
+                    });
+                    return;
+                }
+
+
+                my $dhcp_mac = $dhcp->{'chaddr'};
+                # If its a DHCPREQUEST, we take the IP address from the options
+                my $dhcp_ip = ( $dhcp->{'options'}{'53'} == 3 ) ? $dhcp->{'options'}{'50'} : $dhcp->{yiaddr};
```
fdurand: It seems that sometimes $dhcp_ip is empty, we probably have to add a condition if define($dhcp_ip) ...after in the code

julsemaan: Correct, I will need to find why its empty sometimes though as it shouldn't be

sbin/pfdhcplistener
```diff
@@ -244,6 +249,54 @@ sub process_pkt {
...
+                $dhcp->{'dest_ip'} = $args{'dest_ip'};
+
+
+                # grab DHCP information
+                if ( !defined($dhcp->{'chaddr'}) ) {
+                    $logger->debug("chaddr is undefined in DHCP packet");
+                    return;
+                }
+
+                $dhcp->{'chaddr'} = clean_mac( substr( $dhcp->{'chaddr'}, 0, 12 ) );
```
jrouzierinverse: Create a temporary variable to hold $dhcp->{'chaddr'} to avoid accessing the hash multiple times for the same variable.

sbin/pfdhcplistener
```diff
@@ -244,6 +249,54 @@ sub process_pkt {
...
+                    $logger->debug("Source MAC is invalid. skipping");
+                    return;
+                }
+
+                if ( $dhcp->{'chaddr'} ne "00:00:00:00:00:00" && !valid_mac($dhcp->{'chaddr'}) ) {
+                    $logger->debug( sub {
+                        "invalid CHADDR value ($dhcp->{'chaddr'}) in DHCP packet from $dhcp->{src_mac} ($dhcp->{src_ip})"
+                    });
+                    return;
+                }
```
jrouzierinverse: Move any logic that would exit early before creating the %args hash and doing the base64 of the packet to avoid extra work.

sbin/pfdhcplistener
jrouzierinverse: Just a question is the rate limiting key too specific?

sbin/pfdhcplistener
jrouzierinverse: This could create a lot of noise in the log files can you change to debug or trace.

jrouzierinverse (changes requested):
Should move any logic that is used for deciding to send or not send early in the flow to avoid unnecessary work.

sbin/pfdhcplistener
julsemaan: Why would it be too specific ?

jrouzierinverse: I was curious why use the ip in calculating the key and not just $op . $options_53 . $mac

julsemaan: Well since the iplog is mac<->ip based if a device is assigned a different address for whatever reasons, then it won't be ignored by rate limiting.

julsemaan:
Adjustments completed for @fdurand and @jrouzierinverse reviews

sbin/pfdhcplistener
```diff
@@ -244,6 +249,55 @@ sub process_pkt {
...
+
+                # adding to dhcp hashref some frame information we care about
+                $dhcp->{'src_mac'} = $args{'src_mac'};
+                $dhcp->{'dest_mac'} = $args{'dest_mac'};
+                $dhcp->{'src_ip'} = $args{'src_ip'};
+                $dhcp->{'dest_ip'} = $args{'dest_ip'};
+                $dhcp->{'chaddr'} = $dhcp_mac;
+
+                # grab DHCP information
+                if ( !defined($dhcp_mac) ) {
```
jrouzierinverse: This check would be better after the clean_mac of $dhcp->{'chaddr'}

sbin/pfdhcplistener
```diff
@@ -252,8 +306,21 @@ sub process_pkt {
...
-                my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
-                $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
+                my $rate_limiting = $Config{network}->{dhcp_rate_limiting};
+                # We send the packet to be processed if:
+                #  - there is no rate limit key (packet doesn't contain info to be rate-limited)
+                #  - there is no rate limiting configured
+                #  - there is no cache entry (packet wasn't seen in the last $rate_limiting seconds)
+                if(!defined($rate_limit_key) || $rate_limiting == 0 || !$rate_limit_cache->get($rate_limit_key)) {
+                    my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
+                    $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
```
jrouzierinverse: We do not need to send the udp_payload_b64 anymore for dhcpv4

So to avoid additional processing we should only base64 it when sending it process_dhcpv6.

Also remove the decoding of the udp_payload_b64 in pf::api::process_dhcp

jrouzierinverse (changes requested):

julsemaan:
changes pushaide

sbin/pfdhcplistener
```diff
@@ -244,6 +248,56 @@ sub process_pkt {
...
+                my $dhcp_mac = clean_mac( substr( $dhcp->{'chaddr'}, 0, 12 ) );
+                if ( $dhcp_mac ne "00:00:00:00:00:00" && !valid_mac($dhcp_mac) ) {
+                    $logger->debug( sub {
+                        "invalid CHADDR value ($dhcp_mac) in DHCP packet from $dhcp->{src_mac} ($dhcp->{src_ip})"
+                    });
+                    return;
+                }
+                
+                # grab DHCP information
+                if ($dhcp_mac) {
```
jrouzierinverse: if (!defined $dhcp_mac) {

julsemaan: clean_mac returns "0" when it can't clean the mac

Since you made me move the clean_mac at the top, now the defined($clean_mac) doesn't apply anymore which I've fixed at the same time

jrouzierinverse: Then
if (!$dhcp_mac) {

julsemaan: good catch, let me fit it

jrouzierinverse (approved):

PR #1853, opened 2016-11-18
Fix/1852 pfmon activeactive

### LINKED ISSUE #1852: pfmon: Should run on every server in active-active
pfmon should be running on all servers in an active-active setup but should only "runtask" on the "management" server.

dwlfrth:
# Description
Allow pfmon to run on all servers in an active-active cluster and runtasks only on the "management" one

# Impacts
pfmon

# Issue
fixes #1852 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* pfmon: Should run on every server in active-active (#1852)

### DIFF
--- a/sbin/pfmon
+++ b/sbin/pfmon
@@ -87,7 +87,7 @@ POSIX::sigaction(
 POSIX::sigaction(
     &POSIX::SIGHUP,
     POSIX::SigAction->new(
-        'normal_sighandler' , POSIX::SigSet->new(), &POSIX::SA_NODEFER
+        'reload_config' , POSIX::SigSet->new(), &POSIX::SA_NODEFER
     )
 ) or die("pfmon could not set SIGHUP handler: $!");
 
@@ -137,6 +137,7 @@ flock($fh, LOCK_EX | LOCK_NB) or die "cannot lock $pidfile another pfmon is runn
 $HAS_LOCK = 1;
 
 our $running = 1;
+our $process = 0;
 
 # standard signals and daemonize
 daemonize($PROGRAM_NAME) if ($daemonize);
@@ -144,6 +145,7 @@ our $PARENT_PID = $$;
 
 
 sub start {
+    reload_config();
     registertasks();
     runtasks();
     waitforit();
@@ -381,6 +383,24 @@ sub normal_sighandler {
     $running = 0;
 }
 
+=head2 reload_config
+
+=cut
+
+sub reload_config {
+    if ( pf::cluster::is_management ) {
+        $process = $TRUE;
+    }
+    elsif ( !$pf::cluster::cluster_enabled ) {
+        $process = $TRUE;
+    }
+    else {
+        $process = $FALSE;
+    }
+
+    $logger->debug("Reload configuration with status $process");
+}
+
 =head2 runtasks
 
 run all runtasks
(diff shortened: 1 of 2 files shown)


dwlfrth:
Review me

sbin/pfmon
```diff
@@ -391,7 +411,7 @@ run all runtasks
 sub runtasks {
     my $mask = POSIX::SigSet->new(POSIX::SIGCHLD());
     sigprocmask(SIG_BLOCK,$mask);
-    while(@REGISTERED_TASKS) {
+    while(@REGISTERED_TASKS && $process) {
```
julsemaan, line 413: Shouldn't we spawn the worker processes even if they are going to sleep.

That will make it more consistent when looking at the processes list as if it goes from 'slave -> master -> slave', it seems like its going to have them there anyway

sbin/pfmon
```diff
@@ -382,6 +384,24 @@ sub normal_sighandler {
     $running = 0;
 }
 
+=head2 reload_config
+
+=cut
+
+sub reload_config {
+    if ( pf::cluster::is_management ) {
+        $process = $TRUE;
```
julsemaan, line 392: I'm not really sure this will change it in the child processes, meaning we will have to send the signal to all pfmon processes. @jrouzierinverse can probably assert this much better than me.

sbin/pfmon
dwlfrth: From my POV having sleepy workers is really just having useless process... It is really more easy to tell that pfmon is running here or there by looking at the workers.
And spawning is pretty instantanious on "failover"

julsemaan:
You're missing a fairly important piece :) sending the SIGHUP to the pfmon processes on cluster state change so that it refreshes its state.

Guess we've been focussing a bit too much on the ZammitCorp vs Inverse war during the afternoon

sbin/pfmon
julsemaan: In that case, we should kill the child workers when we go back to slave state. 

Otherwise we'll have 2 possible slave states, one with child processes and one without.

julsemaan: oupsy, trash my last comment :)

dwlfrth: They are being killed on runtime since no longer required to run on that server "slave"

sbin/pfmon
dwlfrth: No need to signal the childs, we only tell the parent to run (or not to run) the childs

dwlfrth:
It is handled in a different way using the "waitforit" since, as asked to @jrouzierinverse , there is no SIGHUP signal sent on cluster state change

sbin/pfmon
julsemaan: Just to make sure, you seem to be shifting @REGISTERED_TASKS, will that re-spawn the processes if it goes in slave mode more than once ? From my reading it doesn't, but I might be reading it wrong (that tends to happen it seems :P)

julsemaan:
I've been getting everything wrong then.

Screw it, I'm going for a beer.

That's what happens when you start reviewing code a Friday afternoon

dwlfrth:
@julsemaan is like @dwlfrth yesterday... "go home, you're drunk"... but he's already home

sbin/pfmon
dwlfrth: Not touching @REGISTERED_TASKS :P
Yep everything is spawning back as supposed !

sbin/pfmon
```diff
@@ -391,7 +411,7 @@ run all runtasks
 sub runtasks {
```
jrouzierinverse, line 410: I would be better to check if it is the master pfmon process early.
To avoid blocking the child signal.

dwlfrth:
Bump

julsemaan:
Bump your rebase

dwlfrth:
Review me

dwlfrth:
It's been rebased and no review :(

dwlfrth:
Reeeevvviiiiieeeewwww meeee

PR #1862, opened 2016-11-24
Fix/no log die

### LINKED ISSUE #1734: pfconfig: improve FS full recovery
Currently, when the root filesystem is full, pfconfig doesn't recover gracefully after that and just jams and requires a restart.

This should be addressed to make the service more robust

julsemaan:
# Description
Don't die when logger fails to write line

# Impacts
Logging

# Issue
fixes #1734 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Prevent logging failure from making a process die (fixes #1734)

# UPGRADE file entries

In order to be sure all your logging facilities use the new logging backend which ensures the processes won't die in case of a logging failure, you must execute the following command:
```
# find conf/log.conf.d/ -type f -exec sed -i.bak "s/Log::Log4perl::Appender::File/pf::log::FileAppender/g" {} \; ; find conf/log.conf.d/ -name '*.bak' -delete
```

### DIFF
--- a/conf/log.conf.d/httpd.aaa.conf.example
+++ b/conf/log.conf.d/httpd.aaa.conf.example
@@ -6,7 +6,7 @@ log4perl.rootLogger = INFO, HTTPD_AAA
 #log4perl.category.pf.SNMP = WARN
 
 ### Logging for httpd.aaa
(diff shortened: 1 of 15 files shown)


julsemaan:
Ready for review

PR #1863, opened 2016-11-24
Collectd Upgrade

fdurand:
# Description
Updates collect version in order to have more information from the system

# Impacts
Graph

# NEW Package(s) required
Collectd >= 5.5

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Collectd updated in order to have new graph


lib/pf/services/manager/collectd.pm
```diff
@@ -58,7 +61,10 @@ sub generateCollectd {
     $tags{'graphite_host'} = "$Config{'monitoring'}{'graphite_host'}";
     $tags{'graphite_port'} = "$Config{'monitoring'}{'graphite_port'}";
     $tags{'hostname'}      = hostname;
-    $tags{'db_host'}       = "$Config{'database'}{'host'}";
+    $tags{'db_host'}
+        = $cluster_enabled
+        ? $management_network->tag('ip')
```
julsemaan: Not sure we can expect a MySQL database on this in cluster.

Unless its used for something else than to connect to MySQL

fdurand: Are you sure ? https://github.com/inverse-inc/packetfence/blob/devel/docs/PacketFence_Clustering_Guide.asciidoc#mysql-network-configuration

julsemaan: But we should be going through haproxy to find the server that is currently alive.

If we're on the server that isn't the one offering DB service, that won't connect to a valid MySQL DB

So this is already set to 127.0.0.1 (database.host) which should be the right configuration

fdurand: Yes but the goal of that is just to reduce the db query since it's useless to fetch the same data from each servers.
Like you have a cluster of an hundred servers and each server will query the db for stats.

html/pfappserver/root/admin/status.tt
```diff
@@ -24,6 +24,12 @@
               [%- IF can_access("REPORTS") %]
               <li><a href="#graph/dashboard">[% l('Dashboard') %]</a></li>
               [%- END %]
+              [%- IF can_access("REPORTS") %]
```
louismunro: Why is this not inside a single IF for the can_access?

PR #1865, opened 2016-11-28
Yes lo

fdurand:
# Description
When the vip is local to the server then the radius traffic between the load balancer and the radius server use the lo interface

# Impacts
Graph

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* radius graph in dashboard


julsemaan:
I just love the PR title

julsemaan:
Should this go in the maintenance as well ?

fdurand:
Maybe, why not

PR #1867, opened 2016-11-28
Reduce p0f cpu usage

fdurand:
# Description
P0f doen't allow to listen on multiples interfaces (-i eth0 -i eth1 ...), so in order to reduce the number of packets it has to process this code generate a bpf filter to remove traffic between each packetfence's server (cluster) and remove all the 'lo' traffic (net 127.0.0.0/8 and 0:0:0:0:0:0:0:1 ipv6 address).

# Impacts
Reduce cpu usage

# NEW Package(s) required
Algorithm::Combinatorics (http://search.cpan.org/~fxn/Algorithm-Combinatorics-0.27/Combinatorics.pm)

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Reduce p0f cpu usage

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -274,6 +274,7 @@ Requires: perl(IO::Interface)
 Requires: perl(Time::Period)
 Requires: perl(Time::Piece)
 Requires: perl(Number::Range)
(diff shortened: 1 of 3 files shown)


tlaurion:
Awesome!

PR #1868, opened 2016-11-28
Feature/infoblox syslog

julsemaan:
# Description
Adds DHCP syslog parser to update iplog from Infoblox and ISC DHCP syslog traffic

# Impacts
pfdetect

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Added syslog receiver to update the iplog from Infoblox and ISC DHCP syslog lines


fdurand:
Looks good to me.
Btw it looks that iscdhcp log should match with that parser so why not add iscdhcp support too.

julsemaan:
Updated code so its not Infoblox related but general DHCP parsing

PR #1875, opened 2016-11-29
Adjusted documentation to have systemd manage MariaDB in RHEL/CentOS7…

dwlfrth:
# Description
Adjusted the Clustering guide to instruct configuration of pcs resource using systemd to manage MariaDB rather than manually starting it

# Impacts
None

# Delete branch after merge
YES

### DIFF
--- a/docs/PacketFence_Clustering_Guide.asciidoc
+++ b/docs/PacketFence_Clustering_Guide.asciidoc
@@ -701,12 +701,7 @@ Corosync configuration
 
   # pcs -f clust_cfg constraint order promote MySQLClone then start DRBD_fs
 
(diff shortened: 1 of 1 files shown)


dwlfrth:
Review me...
Only documentation changes but I used the PR way to "discuss" possible side effects of using systemd in pcs rather than manually starting it.

julsemaan:
This seems like a no-brainer to me

We already use systemd to manage mariadb in active/active so there was no point in managing it manually using pcs

Unless somebody puts a veto on that, I'll merge it today

dwlfrth:
Applied at a client site last night, still alive.

fdurand:
Looks good to me too. Merge !

PR #1876, opened 2016-11-30
Do not log weak authentication methods passwords (ex. PAP) in CLI audit logs.

tlaurion:
To generalize?

# Description
Modify RADIUS Packetfence-CLI so that even weak password authentication methods do not result in password leaks in PacketFence audit logs.

# Impacts
RADIUS audit logs.

# Delete branch after merge
YES

## Enhancements

# UPGRADE file entries
How should we proceed?

### DIFF
--- a/conf/radiusd/load_balancer.conf.example
+++ b/conf/radiusd/load_balancer.conf.example
@@ -127,8 +127,8 @@ log {
 	#
 	#  allowed values: {no, yes}
 	#
-	auth_badpass = yes
-	auth_goodpass = yes
+	auth_badpass = no
+	auth_goodpass = no
 
 	#  Log additional text at the end of the "Login OK" messages.
 	#  for these to work, the "auth" and "auth_goodpass" or "auth_badpass"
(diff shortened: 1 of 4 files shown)


extrafu:
Please update the PR text to reflect the actual PR - otherwise it's rather a mess when we go over for the NEWS file entry.

whitx:
You should also remove the password from the logs, in conf/radiusd/load_balancer.conf

janegilv:
Would you mind looking at something similar packetfence-tunnel? I did your fix there, which fixed the same issue for EAP-TTLS GTC, but don't know whether this has any other side effects.

tlaurion:
@louismunro : 
- EAP-MSCHAPV2: Success.  
- EAP-GTC, EAP-TTLS GTC: Login incorrect (eap: Failed continuing EAP GTC (6) session.  EAP sub-module failed)

In the logs:
```
WARNING: pap: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING: pap: !!! Ignoring control:User-Password.  Update your        !!!
WARNING: pap: !!! configuration so that the "known good" clear text !!!
WARNING: pap: !!! password is in Cleartext-Password and NOT in        !!!
WARNING: pap: !!! User-Password.                                      !!!
WARNING: pap: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

Is that even supposed to work? There is no reference to Cleartext-Password in our confs.

tlaurion:
Tested locally with PAP. Password is obfuscated.

PR #1879, opened 2016-12-01
Fix/reg after status

### LINKED ISSUE #1877: Portal: Registration can be available after accessing the status page
If you access the status page, and then access /captive-portal you can register your device again because the status page bypasses the reg check so that billing can be accessed from the status page.

We need to have this enabled/disabled from the portal profile

julsemaan:
# Description
This will prevent access to the registration portal after reaching the status page unless explicitly allowed by the administrator on a portal profile basis

# Impacts
Status page flow

# Issue
fixes #1877 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fix devices gaining access to registration section of the portal even though they were already registered (#1877)

# UPGRADE notes

PacketFence will not allow access to the registration section of the portal (signup/billing/login) when a device is registered unless access_registration_when_registered is enabled in the portal profile. That means if you have users that need to be able to register on the portal even when they are already registered (like when renewing paid access), make sure you activate this parameter on your portal profile.

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
@@ -209,7 +209,10 @@ sub execute_child {
     # HACK alert : E-mail registration has the user registered but still going in the portal
     # release_bypass is there for that. If it is set, it will keep the user in the portal
     my $node = node_view($self->current_mac);
-    if(defined($node->{status}) && $node->{status} eq "reg" && !$self->app->session->{release_bypass}){
+    if($self->app->profile->canAccessRegistrationWhenRegistered() && $self->app->session->{release_bypass}) {
+        get_logger->info("Allowing user through portal even though he is registered as the release bypass is set and the portal profile is configured to let registered users use the registration module of the portal.");
+    }
+    elsif(defined($node->{status}) && $node->{status} eq "reg"){
         return $self->unknown_state();
     }
     $self->SUPER::execute_child();
--- a/conf/profiles.conf.defaults
+++ b/conf/profiles.conf.defaults
@@ -26,3 +26,4 @@ scans=
 reuse_dot1x_credentials=0
 sources=
 provisioners=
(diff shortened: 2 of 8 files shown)


julsemaan:
Would someone be available to review this.

That bug causes weird behavior for already registered devices

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
```diff
@@ -209,7 +209,10 @@ sub execute_child {
     # HACK alert : E-mail registration has the user registered but still going in the portal
     # release_bypass is there for that. If it is set, it will keep the user in the portal
     my $node = node_view($self->current_mac);
-    if(defined($node->{status}) && $node->{status} eq "reg" && !$self->app->session->{release_bypass}){
+    if($self->app->profile->canAccessRegistrationWhenRegistered() && $self->app->session->{release_bypass}) {
+        get_logger->info("Allowing user through portal even though he is registered as the release bypass is set and the portal profile is configured to let registered users use the registration module of the portal.");
+    }
+    elsif(defined($node->{status}) && $node->{status} eq "reg"){
```
jrouzierinverse, line 215: should you continue to check for ```!$self->app->session->{release_bypass}``` since that is still can be a valid state

jrouzierinverse:
Add something to the upgrade guide about this

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
julsemaan: No, that isn't necessary because unless $self->app->profile->canAccessRegistrationWhenRegistered() is true, we don't want to consider release_bypass

julsemaan:
Added upgrade notes to description

jrouzierinverse:
Put a default value for access_registration_when_registered in conf/profiles.conf.defaults

julsemaan:
Just added the value

PR #1884, opened 2016-12-02
Fixed "autoreg" field not set in db.

### LINKED ISSUE #1883: autoreg is never set to 'yes'
Autoregistered nodes do not have the autoreg field set to 'yes' in the database.

This breaks any code that relies on the field (e.g. vlan filters).

louismunro:
# Description
fetchRoleForNode overwrites the autoreg field in the database.
Therefore the field is never set to "yes" and any filter that relies upon it will never match.
This fixes that.

# Impacts
autoreg should equal 'yes' in the node table when a device has been autoregistered
Conversely, when the device has been unregistered, or not autoregistered the field should be equal to 'no'.

# Issue
fixes #1883

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes
* autoreg flag is now set correctly (#1883)

### DIFF
--- a/lib/pf/node.pm
+++ b/lib/pf/node.pm
@@ -867,13 +867,6 @@ sub node_modify {
         return (0);
     }
 
-    # hack to support an additional autoreg param to the sub without changing the hash to a reference everywhere
-    my $auto_registered = 0;
-    if (defined($data{'auto_registered'})) {
-        $auto_registered = 1;
-        delete($data{'auto_registered'});
-    }
-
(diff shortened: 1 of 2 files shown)


julsemaan:
I got a complaint about this issue in one of my deployments.

Just to make sure, was this tested in the classic use case (802.1x SSID + Guest MAB SSID) ?

louismunro:
@julsemaan Affirmative.

PR #1886, opened 2016-12-06
feature/ipset-cache

jrouzierinverse:
Description
-----------
Reduce the number of ipset calls made when adding ports for Active Directory SRV records

Impacts
-----------

NEWS file entries
-------

Enhancements
++++++++++++
* Reduce the number of ipset calls when adding ports for Active Directory

Bug Fixes
++++++++++++
* Fix adding a udp SRV record port as a tcp port

Delete branch after merge
-------------------------
NO

### DIFF
--- a/lib/pf/ipset_cache.pm
+++ b/lib/pf/ipset_cache.pm
@@ -150,9 +150,11 @@ sub populate_cache {
     }
     my @pairs;
     while (my $line = shift @lines) {
-        if ($line =~ /(\d+(?:\.\d+){3}),((?:tcp|udp):(?:\d+))/) {
-            push @pairs, { ip => $1, port => $2}
+        unless ($line =~ /(\d+(?:\.\d+){3}),((?:tcp|udp):(?:\d+))/) {
+            $logger->error("Could not match line : '$line'");
+            next;
         }
+        push @pairs, { ip => $1, port => $2}
     }
     $self->_add_pairs_to_cache(\@pairs);
 }
--- a/lib/pf/ipset_cache.pm
+++ b/lib/pf/ipset_cache.pm
@@ -1,0 +1,136 @@
+package pf::ipset_cache;
+
+=head1 NAME
+
+pf::ipset_cache -
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::ipset_cache
+
+=cut
+
+use strict;
+use warnings;
+use Moo;
+use CHI;
+use pf::log;
+use pf::util qw(pf_run);
+
+our $logger = get_logger();
+
+has ipset_binary => (is => 'ro', default => sub { 'ipset' });
+
+has cache => (is => 'ro', builder => 1, lazy => 1);
+
+has setname => (is => 'ro', required => 1);
+
+sub _build_cache {
+    my ($self) = @_;
+    return CHI->new(driver => 'RawMemory', datastore => {}, namespace => $self->setname);
+}
+
+sub add_pairs {
+    my ($self, $ip_port_pairs) = @_;
+    my @pairs = grep {!$self->is_in_cache($_) } @$ip_port_pairs;
+    if (@pairs) {
+        unless ($self->_add_pairs_to_ipset(\@pairs)) {
+            return undef;
+        }
+        $self->_add_pairs_to_cache(\@pairs);
+    }
+    return scalar @pairs;
+}
+
+sub is_in_cache {
+    my ($self, $pair) = @_;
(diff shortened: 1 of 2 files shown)


lib/pf/ipset_cache.pm
```diff
@@ -0,0 +1,135 @@
+package pf::ipset_cache;
```
julsemaan, line 1: Missing perldoc for all subs in this file

sbin/pfdns
```diff
@@ -379,6 +388,17 @@ sub regzone {
     return @answer;
 }
 
+sub add_answers_to_ip_port_pairs {
```
julsemaan: perldoc

julsemaan: I would suggest returning the pairs instead of doing it by reference.

Most of our code uses returns instead of passing things by reference

That would make the code clearer IMO

lib/pf/ipset_cache.pm
```diff
@@ -0,0 +1,135 @@
...
+
+=head1 NAME
+
+pf::ipset_cache -
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::ipset_cache
```
julsemaan, line 11: Cool description

lib/pf/ipset_cache.pm
```diff
@@ -0,0 +1,135 @@
+package pf::ipset_cache;
+
+=head1 NAME
+
+pf::ipset_cache -
```
julsemaan: `-` at the end doesn't seem useful

lib/pf/ipset_cache.pm
```diff
@@ -0,0 +1,135 @@
...
+=cut
+
+use strict;
+use warnings;
+use Moo;
+use CHI;
+use pf::log;
+use pf::util qw(pf_run);
+
+our $logger = get_logger();
```
julsemaan, line 24: If the logger is declared here, does that mean we won't be able to see the subroutine name in the log lines ?

lib/pf/ipset_cache.pm
```diff
@@ -0,0 +1,135 @@
...
+    my ($self) = @_;
+    my $setname = $self->setname;
+    my @lines = pf_run("sudo ipset list $setname 2>/dev/null", accepted_exit_status => [1]);
+
+    while (my $line = shift @lines) {
+        last if $line =~ /Members:/;
+    }
+    my @pairs;
+    while (my $line = shift @lines) {
+        if ($line =~ /(\d+(?:\.\d+){3}),((?:tcp|udp):(?:\d+))/) {
```
julsemaan: Could there be a trace or debug if that doesn't match

This way we could troubleshoot any malformed pairs

julsemaan:
Also needs labels and milestone

lib/pf/ipset_cache.pm
jrouzierinverse: Doing this does not affect subroutine name in the log lines.

julsemaan: So if I understand correctly, there will be no notable difference between this and instantiating one at the beginning of every sub ?

If so, going forward should we consider it good practice to always declare it on the top of the module to save some processing time ?

julsemaan:
Code looks good, I'd @fdurand to look at it as well as I'm not as familiar as him with the passthrough/ipset

lib/pf/ipset_cache.pm
```diff
@@ -0,0 +1,189 @@
...
+=cut
+
+sub _add_pairs_to_ipset {
+    my ($self, $pairs) = @_;
+    my $setname = $self->setname;
+    my $binary = $self->ipset_binary;
+    my $data = join("\n", (map { "add $setname " . _format_pair($_) } @$pairs), "");
+    my $pid = open(my $ipset, "| LANG=C sudo $binary -! restore 2>&1");
+    unless (defined $pid) {
+        $logger->error("Cannot start ipset ");
```
fdurand, line 89: ipsetd ?

lib/pf/ipset_cache.pm
jrouzierinverse: Yes, However for modules that are sub classes the better approach is to wrap getting the logger in a method.
To ensure we are getting the correct logger for that class.

jrouzierinverse:
Please review

extrafu:
@jrouzierinverse Please document what does it fix so we can test it and merge it.

dwlfrth:
@jrouzierinverse Please document what does it fix so we can test it and merge it.

jrouzierinverse:
Please review

PR #1887, opened 2016-12-06
Select radius interface

fdurand:
# Description
Define for an interface if you want to have radius available

# Impacts
Radius Workflow

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Define on interface if radius must listen on.

### DIFF
--- a/lib/pf/ConfigStore/Network.pm
+++ b/lib/pf/ConfigStore/Network.pm
@@ -142,10 +142,11 @@ sub cleanupBeforeCommit {
     my $config = $self->cachedConfig;
     unless ( $config->SectionExists($id) ) {
         # Set default values when creating a new network
+        my @types = split(',',$network->{type});
         $network->{named} = 'enabled' unless ($network->{named});
         $network->{dhcpd} = 'enabled' unless ($network->{dhcpd});
         $network->{fake_mac_enabled} = 'disabled' if ($network->{type} ne $pf::config::NET_TYPE_INLINE_L3);
-        $network->{'domain-name'} = $network->{type} . "." . $Config{general}{domain}
+        $network->{'domain-name'} = $types[0] . "." . $Config{general}{domain}
             unless $network->{'domain-name'};
     } else {
         if ($network->{type} && $network->{type} eq $pf::config::NET_TYPE_INLINE_L3) {
--- a/lib/pf/ConfigStore/Network.pm
+++ b/lib/pf/ConfigStore/Network.pm
@@ -142,6 +142,7 @@ sub cleanupBeforeCommit {
     my $config = $self->cachedConfig;
     unless ( $config->SectionExists($id) ) {
         # Set default values when creating a new network
+        $network->{type} =~ s/\s+//;
         my @types = split(',',$network->{type});
         $network->{named} = 'enabled' unless ($network->{named});
         $network->{dhcpd} = 'enabled' unless ($network->{dhcpd});
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -3120,7 +3120,7 @@ sub deauth_source_ip {
                                          if ($interface_src[1] eq 'via') {
                                              return $interface_src[4];
                                          } else {
-                                             return $interface_scr[2];
+                                             return $interface_src[2];
                                          }
                                       }
                            );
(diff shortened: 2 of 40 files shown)


lib/pf/Switch.pm
```diff
@@ -3112,12 +3112,22 @@ Takes into account the active/active clustering and centralized deauth
 =cut
 
 sub deauth_source_ip {
-    my ($self) = @_;
+    my ($self,$dst_ip) = @_;
+    my $chi = pf::CHI->new(namespace => 'route_int');
+    my $int = $chi->compute($dst_ip, sub {
```
julsemaan: We need to validate that the result of this is a valid interface configured in PacketFence

lib/pf/services/manager/keepalived.pm
```diff
@@ -59,7 +60,7 @@ sub generateConfig {
 
     $tags{'vrrp'} = '';
     $tags{'mysql_backend'} = '';
-    my @ints = uniq(@listen_ints,@dhcplistener_ints, map { $_->{'Tint'} } @portal_ints);
+    my @ints = uniq(@listen_ints,@dhcplistener_ints, map { $_->{'Tint'} } @portal_ints, @radius_ints);
```
julsemaan: wrap the map in parentheses so its clearer what you are mapping and to make sure it only applies to what you want:
`uniq(@listen_ints,@dhcplistener_ints, (map { $_->{'Tint'} } @portal_ints, @radius_ints));`

julsemaan:
I'll find some time to test it a bit

lib/pf/Switch.pm
```diff
@@ -3112,12 +3112,32 @@ Takes into account the active/active clustering and centralized deauth
...
+                           );
+    if (defined($Config{ 'interface ' . $int })) {
+        if($cluster_enabled){
+            return isenabled($Config{active_active}{centralized_deauth}) ? pf::cluster::cluster_ip($int) : pf::cluster::current_server->{"interface $int"}->{ip};
+        }
+        else {
+            return $Config{ 'interface ' . $int }{'vip'} || $Config{ 'interface ' . $int }{'ip'}
+        }
+    } else {
+        $logger->warn("Interface $int has not been found in the configuration, using the management interface");
```
julsemaan, line 3136: that logger is not defined

julsemaan:
If I add radius as an additional listening daemon on a registration network, it doesn't add it in pf.conf and makes the interface become of the 'Other' type in the admin

julsemaan:
Also, there is no way to put an interface in type 'radius' from the admin, its only possible to put it as an additionnal listening daemon

julsemaan:
Other than my 2 comments above it seems to do what it says

lib/pf/ConfigStore/Network.pm
```diff
@@ -142,10 +142,11 @@ sub cleanupBeforeCommit {
     my $config = $self->cachedConfig;
     unless ( $config->SectionExists($id) ) {
         # Set default values when creating a new network
+        my @types = split(',',$network->{type});
```
julsemaan: split while cleaning up spaces

lib/pf/Switch.pm
```diff
@@ -3113,12 +3113,33 @@ Takes into account the active/active clustering and centralized deauth
...
-    if($cluster_enabled){
-        return isenabled($Config{active_active}{centralized_deauth}) ? pf::cluster::management_cluster_ip() : pf::cluster::current_server->{management_ip};
-    }
-    else {
-        return $management_network->tag('vip') || $management_network->tag('ip');
+    my ($self,$dst_ip) = @_;
+    my $logger = $self->logger();
+    my $chi = pf::CHI->new(namespace => 'route_int');
+    my $int = $chi->compute($dst_ip, sub {
+                                         my @interface_src = split(" ", `LANG=C sudo ip route get $dst_ip`);
```
julsemaan: Use pf_run ?

lib/pf/services/manager/radsniff.pm
```diff
@@ -40,4 +44,18 @@ has '+launcher' => (
 
 has startDependsOnServices => ( is => 'ro', default => sub { [qw(collectd)] } );
 
+sub make_filter {
```
julsemaan: Missing perldoc

lib/pfconfig/namespaces/interfaces/radius_ints.pm
```diff
@@ -0,0 +1,64 @@
+package pfconfig::namespaces::interfaces::radius_ints;
```
julsemaan, line 1: perldoc for the two subs in this module

julsemaan (changes requested):
A few changes here and there and we're good to merge

extrafu:
Bump

PR #1891, opened 2016-12-08
Added route service in order to manage it in a separate service

fdurand:
# Description
Moved static routes management from dhcpd service to routes service

# Impacts
Static routes

# Issue
In case of a webauth layer 3 network when the dhcp service has been deactivated there is no way to manage static routes.

# Delete branch after merge
YES

# NEWS file entries
## New Features
* New routes service to manage statics routes

### DIFF
--- a/lib/pf/services/manager/routes.pm
+++ b/lib/pf/services/manager/routes.pm
@@ -1,0 +1,151 @@
+package pf::services::manager::routes;
+
+=head1 NAME
+
+pf::services::manager::routes add documentation
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::services::manager::routes
+
+=cut
+
+use strict;
+use warnings;
+use Moo;
+use pf::file_paths qw($install_dir);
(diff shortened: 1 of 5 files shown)


julsemaan:
Looked at the code, and it seems all right

Will test it a bit and merge

fdurand:
Change has been done, ready for a review

fdurand:
Bump

lib/pf/services/manager/routes.pm
```diff
@@ -88,8 +89,8 @@ sub isAlive {
             $route_exist = $network;
         }
     }
-
-    my $routes_applied = defined(pf_run("route | grep ".$route_exist));
+    my $routes_applied = 0;
```
julsemaan: $FALSE

julsemaan:
Will attempt to test today

julsemaan:
Did a few tests and it seems to be behaving correctly.

PR #1893, opened 2016-12-09
Let's haproxy listen on portal interfaces

fdurand:
# Description
Enable haproxy on portal interfaces

# Impacts
Portal

# Issue
If we define portal interfaces in cluster mode in another place than on the management interface then haproxy will not listen on the vip

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Enable haproxy on portal interface

### DIFF
--- a/lib/pf/services/manager/haproxy.pm
+++ b/lib/pf/services/manager/haproxy.pm
@@ -23,6 +23,7 @@ use pf::config qw(
     @listen_ints
     @dhcplistener_ints
     $management_network
+    @portal_ints
 );
 use pf::file_paths qw(
     $generated_conf_dir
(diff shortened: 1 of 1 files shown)


fdurand:
Bump

lib/pf/services/manager/haproxy.pm
```diff
@@ -113,7 +114,7 @@ $backend_ip_config
 
 EOT
         }
-        if ($cfg->{'type'} eq 'internal') {
+        if ($cfg->{'type'} =~ /internal/ | $cfg->{'type'} =~ /portal/) {
```
julsemaan: Not sure that binary OR is really doing what you want

Should be `||`

julsemaan (changes requested):

fdurand:
Change done

PR #1894, opened 2016-12-09
Because sometimes we need a different port for the coa and disconnect

fdurand:
# Description
Some wlc use the 2 port , one for disconect and the other one for coa.
Added an option to define both.

# Impacts
Cisco WLC

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Added a way to define 2 differents port for dicsonnect and coa

# UPGRADE file entries
Update controllerPort to disconnectPort in switches.conf (find . -name "switches.conf" -exec sed -i "s/controllerPort/disconnectPort/g" '{}' \;)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
@@ -300,7 +300,7 @@ has_field macSearchesSleepInterval  =>
 
 has_block definition =>
   (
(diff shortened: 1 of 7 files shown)


dwlfrth:
I know it is more work, but I'd rename the switch object (and everything that refers to it) from 'controllerPort' to 'disconnectPort'

julsemaan:
I'd agree with @dwlfrth although its a bit more work but it will make more sense

fdurand:
Bump

julsemaan:
Looks good

Only little thing is we've been merging stuff like there is no tomorrow in the last weeks.

Maybe time for a meeting to put things into perspective and decide what should go in/out of 6.5 in the ton of PR that remain

PR #1895, opened 2016-12-11
Assign one haproxy process for the database and another one for http(s) stuff

fdurand:
# Description
Run 2 haproxy process, one for the dabase and the other one for http(s)

# Impacts
Haproxy will use 2 cpu...

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* 2 haproxy process , one for the db and teh other one for http(s)

### DIFF
--- a/lib/pf/services/manager/haproxy.pm
+++ b/lib/pf/services/manager/haproxy.pm
@@ -50,13 +50,14 @@ sub executable {
 
 sub _number_cpus {
     my ($self) = @_;
-    open my $la, '<', '/proc/cpuinfo';
+    open my $cpuinfo, '<', '/proc/cpuinfo';
     my $cpu_cores = 0;
-    foreach my $line (<$la>)  {
+    foreach my $line (<$cpuinfo>)  {
         if ($line =~ /^cpu\scores\s+:\s+(\d+)/) {
             $cpu_cores = $cpu_cores + $1;
         }
     }
+    close $cpuinfo;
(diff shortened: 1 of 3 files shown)


lib/pf/services/manager/haproxy.pm
```diff
@@ -58,6 +70,17 @@ sub generateConfig {
     $tags{'http'} = '';
     $tags{'mysql_backend'} = '';
     $tags{'var_dir'} = $var_dir;
+    my $bind_process = '';
+    if ($self->_number_cpus > 1) {
```
dwlfrth, line 75: Strange condition since there is no "else".
What happen in the (improbable) case that there is only one CPU ?

fdurand: else nothing

dwlfrth: Will it blend ?

fdurand: If there is only one cpu it will be like now, only one process to handle the db and the portal.

louismunro:
Good enough.
Let's start testing it for real.
Merging.

lib/pf/services/manager/haproxy.pm
dwlfrth: My question was more, if there is only 1 cpu, the CPU tag will not be defined...
Will the template be generated properly ?

PR #1899, opened 2016-12-12
Feature/pfmon registered tasks

dwlfrth:
# Description
Do not register a task in pfmon if not required.

# Impacts
pfmon tasks.
Will also require a restart of pfmon if a task need to be enabled.

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* pfmon will not register and start a process for disabled task


dwlfrth:
Review me

dwlfrth:
Bump

PR #1903, opened 2016-12-14
go firewall SSO - a.k.a our first shot at a Go service

julsemaan:
Things done:
 - pfconfig support for JSON replies
 - simple pfconfig driver for golang
 - Golang FirewallSSO «object» factory


lib/pfconfig/objects/Net/Netmask.pm
```diff
@@ -0,0 +1,23 @@
+package pfconfig::objects::Net::Netmask;
+
+use base qw(Net::Netmask);
+
+sub new {
+    my $package = shift;
+    my $o = Net::Netmask->new(@_);
+    return bless $o, $package;
```
jrouzierinverse, line 8: To allow also a bless object to be used properly as a constructor this should be rewritten as

```
return bless $o, ref($package) || $package;
```

julsemaan: do we really want to allow that ? 

don't think its in our usual practice to do it from the other constructors I looked at

jrouzierinverse: It is done inconsistently some does others do not.

It has more upsides than downsides in my opinion.
However it does not change much at the end of the day.

julsemaan:
Added a README on the go directory and reviewed the code a bit to make sure nothing is abnormal

Could be merged into devel to serve as a base go directory

Logging will likely change but at least everybody can have fun with the first draft of the pfconfig driver and can use a common environment (check README for setup)

go/README.md
```diff
@@ -0,0 +1,43 @@
+
+# PacketFence go library
+
+## Basic setup
+
+This is an initial draft on how to setup/use the PacketFence go libraries.
```
extrafu: go -> Go everywhere

go/pfconfigdriver/fetch.go
```diff
@@ -0,0 +1,223 @@
...
+		// Do nothing, cache is populated, will be returned below
+	} else if flag.Lookup("test.v") == nil {
+		pfconfigSocketPathCache = pfconfigSocketPath
+	} else {
+		pfconfigSocketPathCache = pfconfigTestSocketPath
+	}
+	return pfconfigSocketPathCache
+}
+
+// Struct that encapsulates the necessary informations to do a query to pfconfig
```
jrouzierinverse: informations -> information

extrafu:
@julsemaan Please merge together with instructions on how to pull/build/debug/replace/etc.

go/pfconfigdriver/structs.go
```diff
@@ -0,0 +1,41 @@
...
+	Element *json.RawMessage
+}
+
+// To be combined with another struct in order to give it a Type attribute common in PF
+type TypedConfig struct {
+	Type string `json:"type"`
+}
+
+// Represents the pf.conf general section
+type PfConfGeneral struct {
```
fdurand, line 24: PfconfigGeneral

julsemaan: I was thinking of it as pf.conf general as its not really tied to pfconfig but more to the structure of the file

go/firewallsso/lib/base.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Check if info["ip"] is part of the configured networks if any
+// If there isn't any network, all networks are allowed
+// Otherwise, if the IP is part of one of the networks, this succeeds, otherwise it fails
+func (fw *FirewallSSO) MatchesNetwork(ctx context.Context, info map[string]string) bool {
+	if len(fw.Networks) == 0 {
+		fw.logger(ctx).Debug("No network defined. Allowing all networks")
+		return true
+	}
+	ip := net.ParseIP(info["ip"])
```
jrouzierinverse: Add an error check for the ip address.

go/firewallsso/lib/base.go
```diff
@@ -0,0 +1,135 @@
...
+	Stop(ctx context.Context, info map[string]string) bool
+	GetFirewallSSO(ctx context.Context) *FirewallSSO
+	IsRoleBased(ctx context.Context) bool
+	MatchesRole(ctx context.Context, info map[string]string) bool
+	MatchesNetwork(ctx context.Context, info map[string]string) bool
+}
+
+// Basic struct for all firewalls
+type FirewallSSO struct {
+	PfconfigMethod string `val:"hash_element"`
```
jrouzierinverse: Maybe we should use a different tag instead of val to avoid clashing with a different library.

Maybe pfconfig_val or pf_val what do you think?

go/firewallsso/lib/base.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// Basic interface that all FirewallSSO must implement
+type FirewallSSOInt interface {
+	init(ctx context.Context) error
+	logger(ctx context.Context) log.Logger
+	Start(ctx context.Context, info map[string]string, timeout int) bool
+	Stop(ctx context.Context, info map[string]string) bool
+	GetFirewallSSO(ctx context.Context) *FirewallSSO
+	IsRoleBased(ctx context.Context) bool
```
jrouzierinverse: Having both IsRoleBased and MatchesRole seems redundant to me.

Just calling MatchesRole should be enough.

go/firewallsso/lib/base.go
julsemaan: I really doubt another library would lookup things in that field... but I can adjust it

go/firewallsso/lib/base.go
julsemaan: Done

julsemaan:
I've added more stuff in the readme.

I can't complete the whole loop on getting the service in our stack as the systemd integration isn't complete

I would suggest we get this merged so I can migrate @fdurand golang dispatcher to caddy

This PR itself doesn't actually replace the SSO process yet, it only creates the structure for golang HTTP driven services.

I'll work on getting SSO completed in another PR once @fdurand work will be «caddyfied»

extrafu:
Let's Go for it! No pun intended (even if @lzammit wouldn't get it).

julsemaan:
Alright, I'll just have @louismunro check it out and make sure that I didn't do any too stupid and I'll let him have fun pressing «zi merj battun» as the french say

go/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+If your middleware uses statsd, you don't have to configure statsd in your Caddyfile which will result in the packets just not being sent (a dummy statsd client will be created).
+
+You can start pfcaddy with your Caddyfile using the following command:
+
+```
+# /usr/local/pf/bin/pfcaddy -conf /usr/local/pf/conf/caddy-services/pfsso.conf
+```
+
+You should now create a systemd service for it calling the method above. This part is still to be written as @louismunro will have to show us how to do it.
```
louismunro: Let me add basic instructions on that.
Otherwise this is going to stay in the doc forever and I will keep hearing about it until I retire.

julsemaan: Its not that bad, imagine if Zayme would have to hear us ranting on him and his work all day until he retires

go/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+# cd $GOPATH/src/github.com/inverse-inc/packetfence/go/firewallsso/lib
+# go test
+```
+
+## TODO
+
+ * Integrate with pfconfig namespace expiration (for now any config change requires a restart)
+ * Integrate a vendoring solution
+ * I (Julien) don't like the name pfcaddy even though I came up with it. Suggestions on the name are welcome.
```
louismunro: pf-httpd?

go/caddy/pfsso/pfsso.go
```diff
@@ -0,0 +1,116 @@
...
+			firewallsso.ExecuteStart(ctx, firewall, info, int(timeout))
+		}()
+	}
+
+	w.WriteHeader(http.StatusAccepted)
+}
+
+func (h PfssoHandler) handleStop(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	ctx := r.Context()
+	defer statsd.NewStatsDTiming(ctx).Send("PfssoHandler.handleStart")
```
louismunro: Should handleStop be sending a statsd time for handleStart?
Am I missing something obvious?

go/README.md
extrafu: I would use a dash in the name, just like we do for our other services.

How about 'pfappserver' ?

go/caddy/pfsso/pfsso.go
julsemaan: You're catching me copy/pasting stuff without reading it :)

I'll change this

go/README.md
julsemaan: pfappserver would be confusing with the admin

pfhttpd sounds good to me

julsemaan:
Adjusted based on comments

@louismunro 
Let me know if there is anything else

louismunro:
I can see nothing wrong here.
Hit "zi merde boutonne".

PR #1906, opened 2016-12-15
Feature/dhcp rate limiting 6.0

julsemaan:
Since the devel based branch doesn't apply on 6.0, use this to patch setups with the DHCP rate limiting.

Should we want to send the rate limiting to maintenance, this can then be used.

Otherwise this should be closed when #1845 is merged

### DIFF
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -29,6 +29,8 @@ BEGIN {
     use pf::log(service => 'pfdhcplistener');
 }
 
+use Try::Tiny;
+use pf::util::dhcp;
 use pf::constants;
 use pf::clustermgmt;
 use pf::config qw(
@@ -248,6 +250,27 @@ sub process_pkt {
 
             # we send all IPv4 DHCPv4 codepath
             if ( $l2->{type} eq ETH_TYPE_IP ) {
+                my ($dhcp);
+
+                # we need success flag here because we can't next inside try catch
+                my $success;
+                try {
+                    $dhcp = decode_dhcp($l4->{'data'});
+                    $success = 1;
+                } catch {
+                    $logger->warn("Unable to parse DHCP packet: $_");
+                };
+                return if (!$success);
+                $args{dhcp} = $dhcp;
+
+                if ($is_inline_vlan) {
+                    my $ip = new NetAddr::IP::Lite clean_ip($l3->{'src_ip'});
+                    if ($net_addr->contains($ip)) {
+                        $args{inline_sub_connection_type} = $NET_TYPE_INLINE_L2;
+                    } else {
+                        $args{inline_sub_connection_type} = $NET_TYPE_INLINE_L3;
+                   }
+                }
                 my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
                 $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
             }
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -309,10 +309,13 @@ sub process_pkt {
                         $args{inline_sub_connection_type} = $NET_TYPE_INLINE_L3;
                    }
                 }
-                if(!$rate_limit_cache->get($rate_limit_key)) {
+                my $rate_limiting = $Config{network}->{dhcp_rate_limiting};
+                if($rate_limiting == 0 || !$rate_limit_cache->get($rate_limit_key)) {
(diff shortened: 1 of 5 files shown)


sbin/pfdhcplistener
```diff
@@ -248,8 +254,73 @@ sub process_pkt {
...
+                    $logger->debug("Source MAC is invalid. skipping");
+                    return;
+                }
+
+                if ( $dhcp->{'chaddr'} ne "00:00:00:00:00:00" && !valid_mac($dhcp->{'chaddr'}) ) {
+                    $logger->debug( sub {
+                        "invalid CHADDR value ($dhcp->{'chaddr'}) in DHCP packet from $dhcp->{src_mac} ($dhcp->{src_ip})"
+                    });
+                    return;
+                }
```
jrouzierinverse, line 295: Move any logic that would exit early before creating the %args hash and doing the base64 of the packet to avoid extra work.

sbin/pfdhcplistener
```diff
@@ -248,8 +254,73 @@ sub process_pkt {
...
+                $dhcp->{'dest_ip'} = $args{'dest_ip'};
+
+
+                # grab DHCP information
+                if ( !defined($dhcp->{'chaddr'}) ) {
+                    $logger->debug("chaddr is undefined in DHCP packet");
+                    return;
+                }
+
+                $dhcp->{'chaddr'} = clean_mac( substr( $dhcp->{'chaddr'}, 0, 12 ) );
```
jrouzierinverse, line 283: Create a temporary variable to hold $dhcp->{'chaddr'} to avoid accessing the hash multiple times for the same variable.

sbin/pfdhcplistener
```diff
@@ -248,8 +254,73 @@ sub process_pkt {
...
+                my $rate_limiting = $Config{network}->{dhcp_rate_limiting};
+                if($rate_limiting == 0 || !$rate_limit_cache->get($rate_limit_key)) {
+                    my $apiclient = pf::api::queue->new(queue => 'pfdhcplistener');
+                    $apiclient->notify('process_dhcp', %args, udp_payload_b64 => $udp_payload_b64);
+                    if($rate_limiting != 0) {
+                        $rate_limit_cache->set($rate_limit_key, 1, $rate_limiting);
+                    }
+                }
+                else {
+                    $logger->info("Ignoring packet due to rate-limiting");
```
jrouzierinverse, line 322: This could create a lot of noise in the log files can change to debug or trace.

sbin/pfdhcplistener
```diff
@@ -248,8 +254,73 @@ sub process_pkt {
...
+
+
+                my $dhcp_mac = $dhcp->{'chaddr'};
+                # If its a DHCPREQUEST, we take the IP address from the options
+                my $dhcp_ip = ( $dhcp->{'options'}{'53'} == 3 ) ? $dhcp->{'options'}{'50'} : $dhcp->{yiaddr};
+                # Key is with BOOT type, DHCP type, IP and MAC
+                my $rate_limit_key = $dhcp->{'op'} . "-" . $dhcp->{'options'}{'53'} . "-$dhcp_ip-" . $dhcp_mac;
+                $logger->trace("Rate limit key : $rate_limit_key");
+
+                if ($is_inline_vlan) {
```
jrouzierinverse, line 305: Why do this here since it is not needed for making a decisions to send or not send.
We should avoid any unnecessary work in the pfdhcplistener process.

sbin/pfdhcplistener
```diff
@@ -248,8 +254,73 @@ sub process_pkt {
...
+                    });
+                    return;
+                }
+
+
+                my $dhcp_mac = $dhcp->{'chaddr'};
+                # If its a DHCPREQUEST, we take the IP address from the options
+                my $dhcp_ip = ( $dhcp->{'options'}{'53'} == 3 ) ? $dhcp->{'options'}{'50'} : $dhcp->{yiaddr};
+                # Key is with BOOT type, DHCP type, IP and MAC
+                my $rate_limit_key = $dhcp->{'op'} . "-" . $dhcp->{'options'}{'53'} . "-$dhcp_ip-" . $dhcp_mac;
```
jrouzierinverse, line 302: Just a question is the rate limiting key too specific?

jrouzierinverse (dismissed):
Should move any logic that is used for deciding to send or not send early in the flow to avoid unnecessary work.

julsemaan:
Please comment on the appropriate PR #1845 

This one is for patching purposes only like said in the description

PR #1922, opened 2016-12-20
Make valid_from have a proper default value

### LINKED ISSUE #1920: User valid_from should have a default value
Since we don't define a default value for the valid_from in the password table and it is a required attribute when editing a user, changing another attribute of an already created user makes this attribute flag as required although it wasn't touched.

We should put a default value into it so that the value isn't null

julsemaan:
# Description
Make password.valid_from have a proper default value that fits with what the application expects

# Impacts
local accounts

# Issue
fixes #1920 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Changed password.valid_from default value to "0000-00-00 00:00:00" so its value is valid across the whole application (#1920)

### DIFF
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -380,7 +380,7 @@ CREATE TABLE `configfile` (
 CREATE TABLE `password` (
   `pid` varchar(255) NOT NULL,
   `password` varchar(255) NOT NULL,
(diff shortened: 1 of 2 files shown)


fdurand:
It miss modification in db/pf-schema-X.Y.Z.sql

julsemaan:
updated schema

PR #1923, opened 2016-12-20
validate date cannot exceed 2038-01-18 in admin interface

### LINKED ISSUE #1126: Prevent local user registration window from being after 2038-01-19
If a local user has a registration window after 2038-01-19 (2099-01-01 in my case) then it says the password is not valid.

We should add validation of the date there and in the authentication sources

julsemaan:
# Description
Limit the date inputs to 2038-01-18

# Impacts
Admin interface date editing

# Issue
fixes #1126 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Now limiting dates to 2038-01-18 in admin interface (#1126)


julsemaan:
Ready for review

PR #1926, opened 2016-12-21
Feature/passthrough ports

### LINKED ISSUE #1078: Add option for additionnal ports to open in pfdns passthroughs
Right now it opens 80 and 443.

We should allow additional ones via the configuration

julsemaan:
# Description
Allow to specify ports for DNS based passthroughs.

# Impacts
pfdns passthroughs

# Issue
fixes #1078

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Can now specify specific ports for passthroughs


julsemaan:
Ready for review

lib/pfconfig/namespaces/resource/passthroughs.pm
```diff
@@ -0,0 +1,113 @@
...
+    if($passthrough =~ /(.*?):(udp:|tcp:)?([0-9]+)/) {
+        my $domain = $1;
+        # NOTE: proto contains the ':' at the end
+        my $proto = $2;
+        my $port = $3;
+        if($proto) {
+            return ($domain, [$proto.$port]);
+        }
+        else {
+            return ($domain, ["udp:$port", "tcp:$port"]);
```
jrouzierinverse, line 109: Should default to tcp if none is defined

julsemaan: Why ?

jrouzierinverse: The element of least surprise.

julsemaan: I really don't see your point.

If no protocol is specified, it makes sense to open both protocols as this would be the behavior I expect

jrouzierinverse: If I add a port with no proto there three different way of interrupting it.

* Only tcp
* Only udp
* Both

Some people would think it will only open tcp since you only added a single port like I did.
Other would expect both
Almost none would expect udp 

In either case it is not properly documented behavior.

julsemaan: In our case it opens both protocols and I documented the behavior in the admin guide so I don't see the issue...

jrouzierinverse: It is not documented in documentation.conf

julsemaan: Done, now everybody is happy

jrouzierinverse: Not really

julsemaan: You should be, its Christmas time and as the catchy song says:
https://www.youtube.com/watch?v=gFtb3EtjEic&t=5

fdurand: IMO it's better to open both because there are some protocols that use both (not really a good example but dns do that and few people know that)

fdurand:
Looks good. merge

PR #1928, opened 2016-12-21
Ability to customize the OU in which the machine account will be created

### LINKED ISSUE #1927: domains: Ability to customize the OU in which the machine account will be created
We should add the ability to provide the OU in which the machine account will be created when joining the domain.

```
~$ net ads join --help
net ads join [options]
Valid options:
   createupn[=UPN]    Set the userPrincipalName attribute during the join.
                      The deault UPN is in the form host/netbiosname@REALM.
   createcomputer=OU  Precreate the computer account in a specific OU.
                      The OU string read from top to bottom without RDNs and delimited by a '/'.
                      E.g. "createcomputer=Computers/Servers/Unix"
```

dwlfrth:
# Description
Add the ability to customize the OU in which the machine account is created on domain join

# Impacts
Domain join

# Issue
fixes #1927 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Ability to customize the OU in which the machine account will be created (#1927)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
@@ -97,6 +97,18 @@ has_field 'dns_name' =>
              help => 'The DNS name (FQDN) of the domain.' },
   );
 
+has_field 'ou' => (
+    type        => 'Text',
+    label       => 'OU',
+    default     => 'Computers',
+    required    => 1,
+    message     => { required => 'Please specify a OU in which the machine account will be created' },
+    tags        => {
+        after_element   => \&help,
+        help            => 'Precreate the computer account in a specific OU. The OU string read from top to bottom without RDNs and delimited by a \'/\'. E.g. "Computers/Servers/Unix"',
+    },
+);
+
 has_field 'registration' =>
   (
    type => 'Checkbox',
--- a/html/pfappserver/root/config/domain/view.tt
+++ b/html/pfappserver/root/config/domain/view.tt
@@ -17,6 +17,7 @@
         [% form.field('dns_server').render | none %]
         [% form.field('bind_dn').render | none %]
         [% form.field('bind_pass').render | none %]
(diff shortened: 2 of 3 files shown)


dwlfrth:
Review me

html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
```diff
@@ -97,6 +97,18 @@ has_field 'dns_name' =>
              help => 'The DNS name (FQDN) of the domain.' },
   );
 
+has_field 'ou' => (
+    type        => 'Text',
+    label       => 'OU',
+    default     => 'Computers',
```
julsemaan, line 103: Should default to tcp

html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
```diff
@@ -97,6 +97,18 @@ has_field 'dns_name' =>
...
+    required    => 1,
+    message     => { required => 'Please specify a OU in which the machine account will be created' },
+    tags        => {
+        after_element   => \&help,
+        help            => 'Precreate the computer account in a specific OU. The OU string read from top to bottom without RDNs and delimited by a \'/\'. E.g. "Computers/Servers/Unix"',
```
julsemaan, line 108: Precreate or create ?

And not sure precreate is right, maybe be pre-create ?

I would also said 'server account' instead of computer

julsemaan:
Other than my comment on the help text, this looks good

html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
dwlfrth: It's from the 'net ads join --help' output :)

julsemaan: If its good for samba, its good for PacketFence

PR #1929, opened 2016-12-23
Feature/registration role violation

### LINKED ISSUE #1278: allow to whitelist unregistered devices from violations
Right now an unregistered device is considered the same as a registered device when looking at the whitelisted roles.

That means, unless a device has already been registered with a role and then unregistered it can't be matched in the whitelist.

We should enforce the 'registration' role in the matching of the whitelist when a node is unregistered ignoring any previous role it had and making sure it has one for matching

julsemaan:
# Description
Allow to define the 'registration' role in violation whitelistable roles so it matches unregistered devices
Also prevents user from selecting whitelistable roles that can never match (macDetection,voice,inline,isolation)

# Impacts
Violation role whitelisting

# Issue
fixes #1278 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Allow to whitelist unregistered devices from violations

# UPGRADE file entries

When whitelisting roles in a violation, the registration role will now match unregistered devices where before it would never match. Make sure to go through violations that may include this role to make sure it is relevant.

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Violation.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Violation.pm
@@ -106,7 +106,7 @@ has_field 'whitelisted_roles' =>
    type => 'Select',
    multiple => 1,
    label => 'Whitelisted Roles',
-   options_method => \&options_roles,
+   options_method => \&options_whitelisted_roles,
    element_class => ['chzn-select', 'input-xxlarge'],
    element_attr => {'data-placeholder' => 'Click to add a role'},
    tags => { after_element => \&help,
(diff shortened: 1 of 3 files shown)


lib/pf/violation.pm
```diff
@@ -885,11 +885,17 @@ sub _is_node_category_whitelisted {
         return 0;
     }
 
+    my $node_role = $node_info->{category};
+    # matching registration role for unregistered devices
+    if($node_info->{status} eq $pf::node::STATUS_UNREGISTERED) {
+        $node_role = "registration";
```
fdurand: Use constant $REGISTRATION_ROLE

html/pfappserver/lib/pfappserver/Form/Violation.pm
```diff
@@ -277,6 +277,31 @@ sub options_vclose {
...
+The options for whitelisted roles
+
+=cut
+
+sub options_whitelisted_roles {
+    my ($self) = @_;
+    # NOTE: options_roles is a method on form but that receives the field as the first argument
+    my %roles = options_roles($self);
+    # Roles that aren't technically roles (non-db), except for registration which matches unregistered devices
+    my %skip_roles = (
```
fdurand: Use constants

fdurand (changes requested):

julsemaan:
fixaide

fdurand:
sank iou

PR #1930, opened 2016-12-23
Guest pass printing improvement

### LINKED ISSUE #1409: Admin User create: Allow to print guest passes without the AUP
Right now, guest passes created via 'Users -> Create -> Multiple' are printing one per page with the AUP.

We should add an option to print them without the AUP and have ~5 passes per page

julsemaan:
# Description
Allow to print a guest pass (local account) without the AUP while having multiple per page
Also improves the general layout of printing those passes

# Impacts
Printing guest passes

# Issue
fixes #1409 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Can now print multiple guest passes per page without the AUP in the administration interface (#1409)

### DIFF
--- a/html/pfappserver/root/user/print.tt
+++ b/html/pfappserver/root/user/print.tt
@@ -1,85 +1,116 @@
 <style type="text/css">
 h3, .aup {
-        display: none;
+  display: none;
 }
 </style>
 
 <style type="text/css" media="print">
 .row-fluid .offset2:first-child {
-        margin-left: 0;
+  margin-left: 0;
 }
 .row-fluid .span8 {
-        width: auto;
+  width: auto;
 }
 hr {
   border: 0;
   border-bottom: 1px dashed #000;
 }
-h1, #navbar, .alert {
-        display: none;
+h1, #navbar, .alert, .print-menu {
+  display: none;
 }
 h3, .aup {
-        display: block;
+  display: block;
 }
 dd.aup {
-        font-size: 9pt;
-        line-height: 9pt;
+  font-size: 9pt;
+  line-height: 9pt;
 }
 .dl-horizontal dt {
-        float: none;
-        text-align: left;
-        width: auto;
+  float: none;
+  text-align: left;
+  width: auto;
 }
 .dl-horizontal dd {
-        margin-left: 0;
+  margin-left: 0;
 }
 .page-break {
-        display:block;
-        page-break-before:always;
+  display:block;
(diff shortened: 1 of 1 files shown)


extrafu:
@cgx Please have a quick look to make sure all is right. If so, merge it.

jrouzierinverse:
On the first page there is no divider for the last user. 
This was tested on a Mac with Chrome Version 55.0.2883.95 (64-bit)

[UserPasswordNoAup.pdf](https://github.com/inverse-inc/packetfence/files/714757/UserPasswordNoAUp.pdf)

jrouzierinverse:
No issues with firefox on a mac

jrouzierinverse:
The page break for safari on a mac is not working properly.

[User4PasswordsNoAUP.pdf](https://github.com/inverse-inc/packetfence/files/714785/User4PasswordsNoAUP.pdf)
[User4Passwords.pdf](https://github.com/inverse-inc/packetfence/files/714786/User4Passwords.pdf)

julsemaan:
Fixed the issue on Chrome by using a bordered div instead of an hr tag

Unfortunately Safari ignores the margin setting (actually the page rule as a whole) so there isn't anything I can do to make it respect the layout via CSS

The user will have to use a browser which respects standards (Safari is the only one who doesn't respect it, even IE respects that) or adjust the margins manually in the browser while printing the guest passes

html/pfappserver/root/user/print.tt
```diff
@@ -57,10 +64,12 @@ body {
   <div class="row-fluid">
     <div class="span8 offset2">
       <div class="print-menu">
-        <h1>[% l('Access Code') %]</h1>
-        <div>
-          <label>Print with AUP: <input autocomplete="off" type="checkbox" name="with-aup" value="yes" checked="checked"></label>
-        </div>
+        <h1>
+          <span>[% l('Access Code') %]</span>
+          <span class="print-with-aup" style="display:inline-block;">
```
cgx: Move the inline style to the CSS `.print-with-aup` class.

html/pfappserver/root/user/print.tt
```diff
@@ -57,10 +64,12 @@ body {
...
+            <label>Print with AUP: <input autocomplete="off" type="checkbox" name="with-aup" value="yes" checked="checked"></label>
```
cgx: Localize this label.

julsemaan:
Changes done

PR #1931, opened 2016-12-23
allow to use git in pf-maint (for ignoring files)

### LINKED ISSUE #807: pf maintenance script tries to patch non existing files
I guess this is a consequence of releasing from the maintenance.

We should add ignores to the comparison.

For now, they can be ignored so it doesn't break the flow. Can maybe confuse people.

julsemaan:
# Description
Allow to use git in pf-maint.pl so un-patchable files (like the ones in debian/) don't attempt to be patched and make users insecure with meaningless warnings

# Impacts
pf-maint.pl

# Issue
fixes #807 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Maintenance patching can now use git in order to ignore files that shouldn't be patched via the maintenance script (#807)

### DIFF
--- a/addons/pf-maint.pl
+++ b/addons/pf-maint.pl
@@ -45,8 +45,24 @@ our $COMMIT;
 our $BASE_COMMIT;
 our $NO_ASK;
 our $PATCH_BIN = '/usr/bin/patch';
+our $GIT_BIN = '/usr/bin/git';
 our $COMMIT_ID_FILE = catfile($PF_DIR,'conf','git_commit_id');
 our $test;
+
+# Files that should be excluded from patching
+# Will only work when using git to patch a server
+our @excludes = (
+    ".gitattributes",
+    ".gitconfig",
+    ".gitignore",
+    "packetfence.logrotate",
+    ".github/*",
+    ".tx/*",
+    "debian/*",
+    "docs/*",
+    "src/*",
+    "t/*",
+);
 
 GetOptions(
     "github-user|u=s" => \$GITHUB_USER,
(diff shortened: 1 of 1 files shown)


fdurand:
Looks good to me, agree to merge.

julsemaan:
@dwlfrth could you double verify this

Also, big question now, is whether we push this in the maintenance so that pf-maint patches itself

dwlfrth:
If it's tested and it works as expected, LGTM
As for the pf-maint patching itself, I'd say
e7633bedf897bb24ce668ac9c5df6bf88a58ff7e114d27606a756f4c4888a3f1

julsemaan:
Maybe we could push a binary via the maintenance that would patch the maintenance binary and then the new maintenance binary would remove the binary that patched the maintenance binary if it succeeded. Otherwise we could push another binary that would patch the patching binary and fix it, which could then be re-run to finalize the maintenance binary patching.

The more I think about it, the more it sounds great.
If it could open an rssh tunnel at the same time, I'd say we have some chances of stealing the 2017 http://womeninitawards.com/ to @lzammit

dwlfrth:
I'd say, go with pf-maint patching itself.

dwlfrth:
Actually, let's start by merging it in devel, we'll see if we cherry-pick it in maintenance branches. Will squash merge for easier cherry-picking. Sounds good ?

julsemaan:
squashaide

julsemaan:
Will cherry-pick in maintenance tomorrow and test it

PR #1934, opened 2016-12-23
Fix bad memory cache update of portal profile sources when editing an auth source

### LINKED ISSUE #1933: Changing an auth source doesn't expire the portal memcached sources
The $SOURCES_CACHE in pf::Portal::Profile only gets expired when config::Profiles changes but not when config::Authentication does.

This means that changing a source doesn't make this change effective on the portal until the forked process gets replaced or the process restarts

Need to add pfconfig::memory_cached the ability to watch more than 1 namespace and apply it for that object

julsemaan:
# Description
Fix bad memory cache update of portal profile sources when editing an auth source

# Impacts
Portal profile sources

# Issue
fixes #1933 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed bad memory cache refresh of the Portal Profile sources (#1933)

### DIFF
--- a/lib/pf/Connection/Profile.pm
+++ b/lib/pf/Connection/Profile.pm
@@ -38,7 +38,7 @@ use pf::config qw(
 use pfconfig::memory_cached;
 
 # a memory cache tied to the config::Profiles namepace
-our $SOURCES_CACHE = pfconfig::memory_cached->new('config::Profiles');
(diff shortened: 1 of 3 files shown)


dwlfrth:
@julsemaan rebase

julsemaan:
ribayzedh

PR #1935, opened 2016-12-23
Freeradius DHCP server

fdurand:
# Description

Introduce freeradius dhcp server in place of iscdhcpd
route(s) are now managed by his own service
replace pfdhcplistener (ipv4)
Ip address can be assigned by role in inline mode.

#Impacts 
DHCP server

# NEW Package(s) required

perl(Net::DHCP::Packet)
libnet-dhcp-perl
# Delete branch after merge

YES
# NEWS file entries
## New Features
- Introduce FreeRadius dhcp server
- route has now is own service
- Ip address per role in inline mode


julsemaan:
Will need to be rebased before the review

PR #1936, opened 2016-12-24
Time and bandwidth balance

fdurand:
# Description
Added a way to set the time balance and bandwidth balance as an action in a rule.
Need to enable accounting in order to make it work.

# Impacts
RADIUS accounting workflow

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Added a way to set time and bandwidth balance in action rule (need accounting)


extrafu:
Awesome stuff 👍

extrafu:
Rebaysse  + @lzammit test.

julsemaan:
Was this rebayzed ?

Because there is still a conflict

dwlfrth:
Rebase this little beauty so that review can go on !

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
```diff
@@ -252,7 +252,7 @@ sub apply_new_node_info {
     my ( $status, $status_msg );
     ( $status, $status_msg ) = pf::node::node_register($self->current_mac, $self->username, %{$self->new_node_info()});
     if ($status) {
-        $self->app->flash->{notice} = [ "Role %s has been assigned to your device with unregistration date : %s", $self->new_node_info->{category}, $self->new_node_info->{unregdate} ];
+        $self->app->flash->{notice} = [ "Role ".$self->new_node_info->{category}." has been assigned to your device with ". ( defined($self->new_node_info->{unregdate}) ? " unregistration date : ".$self->new_node_info->{unregdate} : "")."".( defined($self->new_node_info->{time_balance}) ? " time balance : ".$self->new_node_info->{time_balance} : "")."".( defined($self->new_node_info->{bandwidth_balance}) ? " bandwidth balance : ".$self->new_node_info->{bandwidth_balance} : "") ];
```
julsemaan: This here will break i18n and will never be localized since it will always be dynamic, this is why the %s was there before

Think about all your french PacketFence friends that will be like:
> OHHH SACREBLEU, MAIS QU'EST-CE QUE C'EST CE BORDEL. JE VAIS APPELLER ZAMMIT

lib/pf/radius.pm
```diff
@@ -391,9 +396,53 @@ sub accounting {
...
+                    $bandwidth_balance = 0 if ($bandwidth_balance < 0);
+                    if ($isStop && node_modify($mac, ('bandwidth_balance' => $bandwidth_balance))) {
+                        $logger->info("Session stopped: data was $total_octets octets ($bandwidth_balance octets left)");
+                    }
+                    elsif ($isUpdate) {
+                        $logger->info("Session status: data is $total_octets octets ($bandwidth_balance octets left)");
+                    }
+                    if ($bandwidth_balance == 0) {
+                        # Trigger violation
+                        #TODO CREATE A BANDWIDTH VIOLATION
```
julsemaan: do it ? or should the user be unregistered ?

lib/pf/password.pm
```diff
@@ -331,7 +331,14 @@ sub _update_from_actions {
         $data,$actions,$Actions::SET_ACCESS_DURATION,
         'access_duration',undef
     );
-
+    _update_field_for_action(
+        $data,$actions,$Actions::SET_TIME_BALANCE,
+        'time_balance',undef
```
julsemaan, line 336: This field seems to be missing from the schema and I can't find it in the existing table. 

Forgotten field or me looking like cr*p ?

lib/pf/password.pm
```diff
@@ -331,7 +331,14 @@ sub _update_from_actions {
...
+    );
+    _update_field_for_action(
+        $data,$actions,$Actions::SET_BANDWIDTH_BALANCE,
+        'bandwidth_balance',undef
```
julsemaan, line 340: This field seems to be missing from the schema and I can't find it in the existing table. 

Forgotten field or me looking like cr*p ?

julsemaan:
@fdurand 

Given that this requires minor rework and a rebase, and the merge window was yesterday, do we postpone it to 7.1 or do we do a last sprint to get this in ?

fdurand:
Let's shoot for another PacketFence version

fdurand:
Bump

lib/pf/password.pm
fdurand: https://github.com/inverse-inc/packetfence/blob/devel/db/pf-schema-7.0.0.sql#L149

lib/pf/password.pm
fdurand: https://github.com/inverse-inc/packetfence/blob/devel/db/pf-schema-7.0.0.sql#L150

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
```diff
@@ -260,8 +260,18 @@ sub apply_new_node_info {
     my ( $status, $status_msg );
     ( $status, $status_msg ) = pf::node::node_register($self->current_mac, $self->username, %{$self->new_node_info()});
     if ($status) {
-        if($self->new_node_info->{category} && $self->new_node_info->{unregdate}) {
-            $self->app->flash->{notice} = [ "Role ".$self->new_node_info->{category}." has been assigned to your device with ". ( defined($self->new_node_info->{unregdate}) ? " unregistration date : ".$self->new_node_info->{unregdate} : "")."".( defined($self->new_node_info->{time_balance}) ? " time balance : ".$self->new_node_info->{time_balance} : "")."".( defined($self->new_node_info->{bandwidth_balance}) ? " bandwidth balance : ".$self->new_node_info->{bandwidth_balance} : "") ];
+        $self->app->flash->{notice} = "";
+        if($self->new_node_info->{category}) {
+            $self->app->flash->{notice} .= [ "Role %s has been assigned to your device : %s", $self->new_node_info->{category} ];
```
julsemaan: You're concatenating an array

And you don't have enough variables for the amount of %s you have in the string

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
```diff
@@ -260,8 +260,18 @@ sub apply_new_node_info {
...
+        }
+        if($self->new_node_info->{unregdate}) {
+            $self->app->flash->{notice} .= [ " unregistration date : %s", $self->new_node_info->{unregdate} ];
```
julsemaan: you're concatenating a string + an array again

And that sentence won't make any sense from what I can tell

Its going to say 'Role default has been assigned to your device unregistration date 2018-01-01'

PR #1940, opened 2017-01-03
eduroam authentication source match

dwlfrth:
# Description
Ability to match rules in an eduroam source

# Impacts
* eduroam authentication sources
* RADIUS instances

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* eduroam authentication source can now match rules.

# UPGRADE file entries
* eduroam authentication source is now an "exclusive" authentication source rather than an "external" one. That being said, make sure to adjust portal profile accordingly (an "exclusive" authentication source can be the only one configured in a portal profile)

### DIFF
--- a/lib/pf/Authentication/Source/EduroamSource.pm
+++ b/lib/pf/Authentication/Source/EduroamSource.pm
@@ -47,6 +47,44 @@ sub available_actions {
 }
 
 
+=head2 available_attributes
+
+Allow to make a condition on the user's username.
+
+=cut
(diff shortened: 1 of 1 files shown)


dwlfrth:
Review me

julsemaan (commented):
It looks good, don't have any eduroam setup to try it out, so if you confirm its been tested and working we shall merge

dwlfrth:
It is working yes

PR #1946, opened 2017-01-05
CentOS repository package signing

tlaurion:
# Description
Build server now signs all packages prior to moving them to their final location. This doesn't impact anyone before packetfence-release is installed from RHEL7 repo, which enforces package signature validation.

Changes:
- packetfence-release.spec was merged with buildserver's version to be CENTO6/7 compliant, and add extra repository)
- CentOS6 packages are also signed.
- GPG public key added to validate package signing.
- GPG signature validation activated for main devel and extra repositories.
- pfbuilder scripts got modified to sign  packages with the same key used to sign Sogo RPM packages on the nightly and release.
 
# Impacts
Yum configuration file will be overwritten. (no noreplace)

# Delete branch after merge
YES 

# NEWS file entries
## New Features
* CentOS repositories (packetfence and packetfence-devel) packages are now signed. Please upgrade package-release in CentOS7!

### DIFF
--- a/addons/packages/packetfence-release.spec
+++ b/addons/packages/packetfence-release.spec
@@ -53,12 +53,14 @@ for the PacketFence RPM repository.
 [packetfence]
 name=PacketFence Repository
 baseurl=http://inverse.ca/downloads/PacketFence/RHEL$releasever/$basearch
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-PACKETFENCE-CENTOS
 gpgcheck=1
 enabled=0
 
 [packetfence-devel]
 name=PacketFence Devel Repository
 baseurl=http://inverse.ca/downloads/PacketFence/RHEL$releasever/devel/$basearch
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-PACKETFENCE-CENTOS
 gpgcheck=1
 enabled=0
 
--- a/addons/packages/packetfence-release.spec
+++ b/addons/packages/packetfence-release.spec
@@ -53,21 +53,21 @@ for the PacketFence RPM repository.
 ## PacketFence RPM Repository for RHEL/Centos
 [packetfence]
 name=PacketFence Repository
-baseurl=http://inverse.ca/downloads/PacketFence/RHEL$releasever/$basearch
+baseurl=http://inverse.ca/downloads/PacketFence/RHEL\$releasever/\$basearch
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-PACKETFENCE-CENTOS
 gpgcheck=1
 enabled=0
 
 [packetfence-devel]
 name=PacketFence Devel Repository
-baseurl=http://inverse.ca/downloads/PacketFence/RHEL$releasever/devel/$basearch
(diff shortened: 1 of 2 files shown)


tlaurion:
Questions to @fdurand:
-Why not signing packetfence-extra repository packages? Signing is the only way to make sure nothing is modified in transit over HTTP (MITM attacks).
-Is there any reason why we specify HTTP instead of HTTPS, anyway?

fdurand:
I am not against, we just have to find the way to sign the packages on the obs for extra repo.
For https, i just think about some setup where we have to use a proxy in order to download packages and sometimes we have issue.

addons/packages/packetfence-release.spec
```diff
@@ -44,41 +44,93 @@ Vendor: PacketFence, http://www.packetfence.org
...
 rm -rf $RPM_BUILD_ROOT
 
 %files -n %{real_name}
 %defattr(0755, root, root)
 %config /etc/yum.repos.d/packetfence.repo
+/etc/pki/rpm-gpg/RPM-GPG-KEY-PACKETFENCE-CENTOS
 
 
 %changelog
+* Thu Jan 05 2017 Inverse inc. <support@inverse.ca> - 1.2-6
```
fdurand: info@inverse.ca or tlaurion@inverse.ca

tlaurion:
@fdurand 

> i just think about some setup where we have to use a proxy in order to download packages and sometimes we have issue.

1. In your opinion, the default should still be HTTP or changed to HTTPS? I would go for the latter. 
1. What kind of issue not resolved through "export https_proxy=https://user:password@serverproxy:port" or yum proxy configuration?

fdurand:
@tlaurion
1 http
2 i think about transparent https proxy

tlaurion:
Packages coming from external repositories need to be resigned (addsign)

Exemple at fresh install when gpg signature checking is activated:

```
Install  1 Package (+751 Dependent packages)

Total size: 378 M
Installed size: 1.8 G
Is this ok [y/d/N]: y
Downloading packages:
warning: /var/cache/yum/x86_64/7/packetfence/packages/perl-Sub-Exporter-Progressive-0.001011-4.2.noarch.rpm: Header V3 DSA/SHA1 Signature, key ID a9ee32f3: NOKEY
Retrieving key from file:///etc/pki/rpm-gpg/RPM-GPG-KEY-PACKETFENCE-CENTOS


The GPG keys listed for the "PacketFence Repository" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.


 Failing package is: perl-Sub-Exporter-Progressive-0.001011-4.2.noarch
 GPG Keys are configured as: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-PACKETFENCE-CENTOS
```

tlaurion:
@jrouzierinverse 
[Mileage would vary ](http://stackoverflow.com/questions/30196903/rpm-resign-appending-signature-instead-of-replacing-one)depending of age of integrated packages packages. 

We could delsign and resign for all packages not being signed with our key?

tlaurion:
All packages (CentOS 6 and 7 and both extra ) are resigned and deployed to public repository.
- CentOS 7 was installed with gpgcheck enforced, works. eapol_test was installed from packetfence-extra, works.
- CentOS 6 was installed without gpgcheck enforced, still works.

PR #1948, opened 2017-01-09
Feature/node last seen

julsemaan:
# Description
Add a new way to track when a node was last seen active via a dedicated column in the node table.
At the same time, add a way to unregister a device that was inactive for x time
Also cleanup nodes based on this new column instead of last_dhcp

last_seen is updated on:
 * DHCP packet that generates a call to pf.dhcp.processor.update_iplog
 * RADIUS authentication + accounting
 * Port-security trap
 * Captive portal hit

# Impacts
Node cleanup

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Added way to unregister devices that were inactive for a certain amount of time (maintenance.node_unreg_window)

## Enhancements
* Added a new last_seen column to nodes to track their last activity (Authentication, HTTP portal, DHCP)
* Delete nodes based on the new last_seen column instead of looking at the last DHCP packet

### DIFF
--- a/db/upgrade-X.X.X-X.Y.Z.sql
+++ b/db/upgrade-X.X.X-X.Y.Z.sql
@@ -149,3 +149,9 @@ CREATE TABLE `chi_cache` (
   KEY chi_cache_expires_at (expires_at)
 );
 
+--
+-- Add last_seen column to node table
+--
+
+ALTER TABLE node ADD last_seen DATETIME NOT NULL DEFAULT "0000-00-00 00:00:00";
+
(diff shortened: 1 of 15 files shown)


julsemaan:
This is waydi for review

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -22,6 +22,12 @@ UPDATE password set valid_from="0000-00-00 00:00:00" WHERE valid_from IS NULL;
 ALTER TABLE password MODIFY valid_from DATETIME NOT NULL DEFAULT "0000-00-00 00:00:00";
 
 --
+-- Add last_seen column to node table
+--
+
+ALTER TABLE node ADD last_seen DATETIME NOT NULL DEFAULT "0000-00-00 00:00:00";
```
fdurand: AFTER `bypass_role_id`;

fdurand: Also don't we need an index on last_seen ?

julsemaan: bypass_role_id is already the last column of the table, so no need to say we want it after.

Unless my comprehension of the add column is equivalent to zayme's comprehension of email clients.

PR #1960, opened 2017-01-12
Fix/accounting session id cache cluster only

### LINKED ISSUE #1959: Accounting session id cache is messy in active-active
Sometimes the accounting session id is not updated properly in active-active cluster case, this means that the CoA will not work when the accounting session id is not updated in the cache.
We should disable this functionality for active-active

whitx:
# Description
Do not use the cache for node_accounting_dynauth_attr and node_accounting_current_sessionid in active-active cluster

# Impacts
Accounting, deauth

# Issue
fixes #1959 

# Delete branch after merge
NO

# NEWS file entries
## Enhancements
* Remove the cache for accounting session ID in active-active clustering

### DIFF
--- a/lib/pf/accounting.pm
+++ b/lib/pf/accounting.pm
@@ -60,6 +60,7 @@ use pf::db;
 use pf::violation;
 use pf::util;
 use pf::CHI;
+use pf::cluster qw($cluster_enabled);
 
 # The next two variables and the _prepare sub are required for database handling magic (see pf::db)
 our $accounting_db_prepared = 0;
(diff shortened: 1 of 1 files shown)


julsemaan:
I actually did the reverse which is to not populate the cache when in active/active

This way that doesn't make it so we have to handle that case in every accounting cache get

I obviously forgot to push it to devel after doing it so I'll do it this morning

Not sure we should handle it at the set and the get, one place should be enough

whitx:
Your call, I just didn't see the modification yesterday that's why I start doing, but if you have done you can close this PR.

julsemaan:
fixaide via 604147d0d2e95633bb57825912834a428ad347b7

PR #1968, opened 2017-01-19
Added a floored value for iplog lease time

### LINKED ISSUE #1965: iplog: Floored lease time for "tolerance"
When opening an iplog entry (either creating a new one or updating an existing one), PacketFence is using the actual lease time of a given DHCP lease (if provided) or set a NULL (which defaults to 0000-00-00 00:00:00) as the 'end_time' for the record.

It looks like more and more device "may not care" about the provided lease time or it may be some other reason, but the thing is, we are seeing more and more issues where "The device cannot be found in PacketFence database" when trying to reach the captive-portal, which is the result of the portal not being able to get an "open" iplog entry to do the IP <-> MAC lookup.

dwlfrth:
# Description
Add a default minimal value for iplog entries lease time (end_time)

# Impacts
* iplog
* portal
* dhcp

# Issue
fixes #1965 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* iplog: Floored lease time for "tolerance" (#1965)

### DIFF
--- a/lib/pf/ip4log.pm
+++ b/lib/pf/ip4log.pm
@@ -46,6 +46,7 @@ use constant IP4LOG                         => 'ip4log';
 use constant IP4LOG_CACHE_EXPIRE            => 60;
 use constant IP4LOG_DEFAULT_HISTORY_LIMIT   => '25';
 use constant IP4LOG_DEFAULT_ARCHIVE_LIMIT   => '18446744073709551615'; # Yeah, that seems odd, but that's the MySQL documented way to use LIMIT with "unlimited"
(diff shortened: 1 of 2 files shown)


dwlfrth:
Review me

extrafu:
Rebaisse

dwlfrth:
Review and merge

julsemaan:
Its all fine with me.

Do we cherry-pick in maintenance ?

If so, I'll let you squash merge it

dwlfrth:
Merged into 
* maintenance/6.5
* maintenance/6.4

PR #1977, opened 2017-01-23
Disable the default role in switch config

lzammit:
# Description
Disable the default role mapping under the switch configuration. It priorities the VLAN ID configuration.

# Impacts
Switches configuration

# Delete branch after merge
YES 

# NEWS file entries
Disable the default role mapping configuration

# UPGRADE file entries
Default RoleMap for the switches:
If you were using the default 'RoleMap = Y' in the conf/switches.conf it's disabled by default now. You will need to put 'RoleMap = Y' under your switches or switch group configuration.


extrafu:
Review ASAP please.

PR #1986, opened 2017-01-26
Feature/revamped captive portal

cgx:
# Description
Minor revamp of the captive portal so the default CSS is more neutral.

# Impacts
Changes the colours but not the layout.

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Revamped default colours of the captive portal to a more neutral/grayish theme.

# Preview

pf-cp-bg-frame

# New available classes

pf-cp
pf-cp-bg-frame-dark
pf-cp-border
pf-cp-bg frame border
pf-cp-bg


extrafu:
@julsemaan Can you test this and if all looks good (ie., CSS builds correctly), we can include this in v6.5

julsemaan:
Not sure I really like it that much. The footer font color is almost invisible and the green aup background looks a bit weird

image

julsemaan (commented):
Issue with footer color and not sure about the green in the AUP background..

cgx:
@julsemaan You don't seem to have the `frame` class on your main container.

cgx:
You don't have the new default logo neither.

julsemaan:
the logo is a setting I put so its normal its not the same.

Now for the frame class, I'll take any guidance you can give on how it can be missing. I'm running the templates provided in the PR

cgx:
Try with the new css build.

julsemaan:
Already looks better.

Now, for the green AUP, IMO that looks a bit weird for the color, I (still opinionated) preferred it with a neutral background

Don't know what the other guys think about it but it doesn't look like it fits for me

julsemaan:
Discussion with @dwlfrth and @lzammit : 

AUP background could use the same color as the background of the 'I accept the terms'

Green header + text could be the same color as the background (blueish)

cgx:
I think we must have an accent color for titles. How about this red?

screen shot 2017-01-30 at 12 02 32 pm

julsemaan:
I'm not really a reference for that.

But, I still prefer the green over the pink

dwlfrth:
I agree with the accent color but I think this red and the green stands out a bit too much.
Can we try with the PacketFence blue-ish ?

cgx:
Same color as the action buttons?

screen shot 2017-01-30 at 12 41 22 pm

dwlfrth:
Something like ... (warning... GIMP alert)

portal_blue
portal_grey

dwlfrth:
That last one looks good !

julsemaan:
same color as button looks good

julsemaan:
Found a visual glitch

When the 'restart' button appears, it actually brings the logo and the whole layout down

No reset:
image

With reset:
image

julsemaan:
Also, footer doesn't look good on mobile:
image

cgx:
I suggest to move the reset button near the navigation "dots", but in not in this release.

Footer has been fixed.

PR #1988, opened 2017-01-26
Fix/1971 ruckus deauth

### LINKED ISSUE #1971: Radius disconnect doesn't work on the Ruckus switch module
The deauthentication doesn't work if we send the accounting session ID:

#Fetching the acct-session-id
    my $dynauth = node_accounting_dynauth_attr($mac);

    $logger->debug("deauthenticate $mac using RADIUS Disconnect-Request deauth method");
    return $self->radiusDisconnect(
        $mac, { 'Acct-Session-Id' => $dynauth->{'acctsessionid'}, 'User-Name' => $dynauth->{'username'} },

The solution that works:

#Fetching the acct-session-id
    my $dynauth = node_accounting_dynauth_attr($mac);

    $logger->debug("deauthenticate $mac using RADIUS Disconnect-Request deauth method");
    return $self->radiusDisconnect(
    $mac, { 'User-Name' => $dynauth->{'username'} },

@dwlfrth suggested that we do check the ruckus version to keep the old deauth way for older Ruckus ZoneDirector.

dwlfrth:
# Description
Introduced a legacy class for old Ruckus equipment and modify the RADIUS Disconnect message behavior

# Impacts
Ruckus network equipment

# Issue
fixes #1971 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Radius disconnect doesn't work on the Ruckus switch module (#1971)

# UPGRADE file entries
* Ruckus RADIUS Disconnect message on newer version does not support the Accounting-Session-Id parameter to be sent anymore. That parameter have been removed. To make sure continuing supporting "legacy" equipment, please use the new "Ruckus Wireless Controller - Legacy" switch module if your Ruckus Wireless Controller needs that attribute.

### DIFF
--- a/lib/pf/Switch/Ruckus/Ruckus_Legacy.pm
+++ b/lib/pf/Switch/Ruckus/Ruckus_Legacy.pm
@@ -1,0 +1,80 @@
+package pf::Switch::Ruckus::Ruckus_Legacy;
+
+=head1 NAME
+
+pf::Switch::Ruckus::Ruckus_Legacy
+
+=head1 SYNOPSIS
+
+Override base methods to comply with legacy hardware
+
+=cut
+
+use strict;
+use warnings;
+
+use base ('pf::Switch::Ruckus');
+
+use pf::accounting qw(node_accounting_dynauth_attr);
+use pf::constants;
+use pf::util;
+
+sub description { 'Ruckus Wireless Controllers - Legacy' }
+
+=over
+
+=item deauthenticateMacDefault
+
+Overrides base method to send Acct-Session-Id withing the RADIUS disconnect request
+
+=cut
+
+sub deauthenticateMacDefault {
+    my ( $self, $mac, $is_dot1x ) = @_;
+    my $logger = $self->logger;
+
+    if ( !$self->isProductionMode() ) {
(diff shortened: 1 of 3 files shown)


dwlfrth:
Review me.

lib/pf/Switch/Ruckus/Ruckus_Legacy.pm
```diff
@@ -0,0 +1,80 @@
+package pf::Switch::Ruckus::Ruckus_Legacy;
```
rouzier: Why not just call it pf::Switch::Ruckus::Legacy

dwlfrth: Because I copy pasted it and been interrupted by @julsemaan at the same time ;)
Fixaide

julsemaan: I'll use similar excuse when people will be reviewing my stuff now

Like:
Oh I was going to write the code properly but then @extrafu and @lzammit transformed the zammitcorp office into a warzone. I really didn't meant to plug in a reboot in the middle of the pfdns boot process

extrafu:
@julsemaan I'm innocent - I don't own a gun!

PR #1989, opened 2017-01-27
mschap machine auth for not chrooted domains

whitx:
# Description
This is a fix to not stripped the user-name in case of machine authentication while domains are not in a chroot

# Impacts
RADIUS, mschap

# Delete branch after merge
NO

# NEWS file entries
## Enhancements
* Fix an issue where machine authentication would not work if domains are not in a chroot

### DIFF
--- a/raddb/mods-available/mschap
+++ b/raddb/mods-available/mschap
@@ -393,19 +393,12 @@ mschap chrooted_mschap_machine {
 }
 
 mschap mschap_machine {
-
         use_mppe = yes
-
         require_encryption = yes
-
-
         require_strong = yes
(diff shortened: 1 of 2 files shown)


whitx:
@julsemaan @fdurand @dwlfrth Quick review to add this to 6.5? might also need to be pushed in maintenance

julsemaan:
Looks OK from my point of view.

Can't go in maintenance as is, because I think the packetfence-mschap-authenticate doesn't exist prior to this release

whitx:
I can do another one for prior version

extrafu:
I think you should strip all blank lines, they don't bring anything.

dwlfrth:
Strip blank lines as @extrafu said and we're good.

PR #1998, opened 2017-01-31
feature/proxy-auth

jrouzierinverse:
Description
-----------
New Source that allows users to log in using an http proxy

Impacts
-----------
Admin GUI login

NEWS file entries
-------

New Features
++++++++++++
* New source for logging into the Admin GUI using an http proxy

Delete branch after merge
-------------------------
NO

### DIFF
--- a/html/pfappserver/lib/pfappserver/Authentication/Credential/Proxy.pm
+++ b/html/pfappserver/lib/pfappserver/Authentication/Credential/Proxy.pm
@@ -62,7 +62,7 @@ Inverse inc. <info@inverse.ca>
 
 =head1 COPYRIGHT
 
-Copyright (C) 2005-2015 Inverse inc.
+Copyright (C) 2005-2017 Inverse inc.
 
 =head1 LICENSE
 
--- a/html/pfappserver/lib/pfappserver/Authentication/Credential/Proxy.pm
+++ b/html/pfappserver/lib/pfappserver/Authentication/Credential/Proxy.pm
@@ -23,11 +23,21 @@ use List::Util qw(first);
 has realm => (is => 'rw');
 has _config => (is => 'rw');
 
+=head2 BUILDARGS
+
+=cut
+
 sub BUILDARGS {
     my ($class, $config, $app, $realm) = @_;
 
     return { _config => $config, realm => $realm};
 }
+
+=head2 authenticate
+
+authenticate using an admin proxy
+
+=cut
 
 sub authenticate {
     my ($self, $c, $realm, $authinfo) = @_;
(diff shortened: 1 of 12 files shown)


jrouzierinverse:
Review me

julsemaan:
This needs a documentation section in the admin guide

html/pfappserver/lib/pfappserver/Authentication/Credential/Proxy.pm
```diff
@@ -0,0 +1,87 @@
+package pfappserver::Authentication::Credential::Proxy;
+
```
julsemaan, line 2: Missing perldoc on subs

html/pfappserver/lib/pfappserver/Form/Config/Authentication/Source/AdminProxy.pm
```diff
@@ -0,0 +1,66 @@
...
+use strict;
+use warnings;
+use HTML::FormHandler::Moose;
+extends 'pfappserver::Form::Config::Authentication::Source';
+
+has_block definition => (
+    render_list => [qw(proxy_addresses user_header group_header)],
+);
+
+has_field 'proxy_addresses' => (
```
julsemaan, line 26: Can there be multiple addresses ? If so, this needs a validator as well as help text on how to separate the addresses

html/pfappserver/lib/pfappserver/I18N/i_default.po
```diff
@@ -1 +0,0 @@
-en.po
```
julsemaan: Not sure that you should modify this to put the strings back in it. My understanding was that en.po was the only place where they needed to be

lib/pf/Authentication/Source/AdminProxySource.pm
```diff
@@ -0,0 +1,153 @@
+package pf::Authentication::Source::AdminProxySource;
```
julsemaan, line 1: I might be wrong, but I don't see anything preventing this source from being put on a portal profile by mistake

If its only for use in the admin, it should be removed from the portal profiles sources choices

julsemaan (changes requested):

html/pfappserver/lib/pfappserver/I18N/i_default.po
jrouzierinverse: That was a rebase issue I will fix that

jrouzierinverse:
Bump

julsemaan:
Please rebase onto devel and I'll take a look at it for the required changes

jrouzierinverse:
Done

PR #2008, opened 2017-02-03
trigger a violation when node is discovered (added)

julsemaan:
# Description
Allow to trigger a violation when a node is discovered (added)

# Impacts
node_add flow

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Allow to trigger a violation and its actions when a node is discovered

### DIFF
--- a/conf/pfqueue.conf.example
+++ b/conf/pfqueue.conf.example
@@ -18,7 +18,6 @@ redis_every=100
 #
 [queue general]
 workers=4
-has_delayed_queue=enabled
 delayed_queue_batch=100
 delayed_queue_sleep=100
 delayed_queue_workers=1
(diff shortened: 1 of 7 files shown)


conf/pfqueue.conf.example
```diff
@@ -18,7 +18,7 @@ redis_every=100
 #
 [queue general]
 workers=4
-has_delayed_queue=disabled
+has_delayed_queue=enabled
```
jrouzierinverse: Should be in conf/pfqueue.conf.defaults

jrouzierinverse (changes requested):
Looks good outside of the configuration issue.

fdurand:
Except the comment of James i am ok to merge.

julsemaan:
Fixed according to james comment

PR #2010, opened 2017-02-06
Feature/systemd managed

louismunro:
# Description
**BROKEN**
Adds systemd support to PacketFence services.

# Impacts
This will most likely break PF to some degree at this point.
The code to handle the configurator and GUI has not yet been written.
The packaging is not yet done.
Run the addons/enable_systemd.sh script to create the target directories and generate the unit files.
Not tested on Debian (yet).
Winbindd only partly handled.

# Delete branch after merge
NO

# NEWS file entries

## Enhancements
* All services are now handled by systemd.


lib/pf/services/manager.pm
```diff
@@ -66,13 +67,11 @@ has shouldCheckup => ( is => 'rw', default => sub { 1 } );
...
 
-sprintf-formatted string that control how the services should be started
-  %1$s: is the service executable
-  %2$s: optional parameters
+Method that launches the service.
 
 =cut
 
-has launcher => ( is => 'rw', lazy => 1);
+has launcher => ( is => 'rw', lazy => 1, builder => '_build_launcher' );
```
louismunro: Useless.

PR #2012, opened 2017-02-07
Fixed wrong test

fdurand:
# Description
Fix for firewall sso modules ( BarracudaNG,PaloAlto).

# Impacts
Firewall SSO

# Issue
Fix sso stop.

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Fix sso stop for PaloAlto and BarracudaNG firewall

### DIFF
--- a/lib/pf/firewallsso/PaloAlto.pm
+++ b/lib/pf/firewallsso/PaloAlto.pm
@@ -71,7 +71,7 @@ sub action {
         if (
             defined($node_info) &&
             (ref($node_info) eq 'HASH') &&
-            $node_info->{'status'} eq $pf::node::STATUS_REGISTERED &&
+            $node_info->{'status'} eq $pf::node::STATUS_UNREGISTERED &&
             (grep $_ eq $node_info->{'category'}, @categories)
         ){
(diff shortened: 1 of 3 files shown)


lib/pf/firewallsso/PaloAlto.pm
```diff
@@ -71,7 +71,7 @@ sub action {
         if (
             defined($node_info) &&
             (ref($node_info) eq 'HASH') &&
-            $node_info->{'status'} eq $pf::node::STATUS_REGISTERED &&
+            $node_info->{'status'} eq $pf::node::STATUS_UNREGISTERED &&
```
julsemaan: I would say we shouldn't have any assumption on the registered vs unregistered status of the node...

If we send a stop, we send it whether the node is reg or unreg...

Otherwise if a user changes its IP address, the stop won't be sent for his old IP if he is registered

fdurand: Yes good point

extrafu:
@julsemaan Relevant with the new firewall SSO code? If so, review and merge please.

PR #2019, opened 2017-02-09
Make Sourcefire great again.

### LINKED ISSUE #2018: Sourcefire module is no longer compatible with pf::api
The current code for the pf::api expects events to be a hash, containing an event element that points to a hashref of type and id.

The old Sourcefire module just sends a list with the time, sid, and ip.

louismunro:
# Description
Fixes the Sourcefire client to be compatible with the current pf::api.

# Impacts
The current pf::api expects a hash of events.
This fixes the Sourcefire module to comply with the API.

# Issue
fixes #2018

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes
* Restores pf::api compatibility to Sourcefire module (#2048)

### DIFF
--- a/addons/sourcefire/pfdetect.pl
+++ b/addons/sourcefire/pfdetect.pl
@@ -90,7 +90,14 @@ eval {
 
    my $date = POSIX::strftime("%m/%d-%H:%M:%S",localtime(time));
 
-   my $result = $soap->event_add($date, $src_ip, "detect", $sid);
+   my %event = (
+       events => {
+           detect => $sid,
+       },
+       srcip => $src_ip,
+       date  => $date,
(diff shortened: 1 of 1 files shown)


julsemaan:
Looks good

As discussed, just not sure if the SOAP call will serialise the attributes correctly but that should be it

julsemaan (commented):
Looks good

As discussed, just not sure if the SOAP call will serialise the attributes correctly but that should be it

jrouzierinverse:
Quick test shows that the soap api deserialize everything correctly

louismunro:
Cool.
I knew XML would not let a feature go unimplemented...

PR #2020, opened 2017-02-09
feature/pf-orm

jrouzierinverse:
Description
-----------
A simple ORM for packetfence

Delete branch after merge
-------------------------
NO


extrafu:
@louismunro @julsemaan Please review as soon as you can so we merge this to start testing it for PF7 ASAP.

julsemaan (changes requested):
This only adds the ORM but doesn't change any of the code from what I can tell.

We need this to change the actual calls at least for the RADIUS, portal and pfsetvlan flows. Otherwise this doesn't bring us anything currently.

Also, there isn't the read-only mode but that can always wait

louismunro:
I still think it's worthwhile to have this in devel.
Otherwise the PR would be unmanageable.
It does mean we'll have to start working on the radius and pfsetvlan code soon.

julsemaan:
@louismunro 

In that case, I'll stretch my luck and ask for unit tests before we merge in devel so we can know it works

louismunro:
Agreed. We need some kind of testing before we merge.
It will be to tempting to indefinitely postpone the tests otherwise.

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -1,3 +1,33 @@
...
+-- Add a primary key to the iplog_archive
+--
+
+ALTER TABLE `iplog_archive` ADD COLUMN `iplog_archive_id` int NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST;
+
+--
+-- Add a primary key to the locationlog
+--
+
+ALTER TABLE `locationlog` ADD COLUMN `locationlog_id` int NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST;
```
julsemaan: In your testing, did this add a value to the pre-existing rows ?

If it doesn't, how will the ORM behave ?

lib/pf/dal/_action.pm
```diff
@@ -0,0 +1,181 @@
+package pf::dal::_action;
+
+=head1 NAME
+
+pf::dal::_action -
```
julsemaan: I think you might need to adjust the template so that a meaningful name and description show up

lib/pf/dal/_auth_log.pm
```diff
@@ -0,0 +1,232 @@
...
+pf::dal::_auth_log -
+
+=cut
+
+use strict;
+use warnings;
+
+###
+### pf::dal::_auth_log is auto generated any change to this file will be lost
+### Make and changes to this file in pf::dal::auth_log
```
julsemaan: Make and changes ?

lib/pf/dal/wrix.pm
```diff
@@ -0,0 +1,46 @@
+package pf::dal::wrix;
+
+=head1 NAME
+
+pf::dal::wrix -
```
julsemaan: Same comment on templating

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -1,3 +1,33 @@
 --
 -- PacketFence SQL schema upgrade from X.X.X to X.Y.Z
 --
+
+--
+-- Add a primary key to the radacct_log
+--
+
+ALTER TABLE `radacct_log` ADD COLUMN `radacct_log_id` int NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST;
```
julsemaan: Why not just call the columns id instead of tablename_id ?

jrouzierinverse: Since there is no real standard I choose the most explicit naming convention.

So there is no confusion on what that field is for.

Also it allow to write joins using 'USING(tablename_id)' syntax.

db/upgrade-X.X.X-X.Y.Z.sql
jrouzierinverse: Yes it add the value to pre-existing rows.

db/upgrade-X.X.X-X.Y.Z.sql
julsemaan: I'd personally stick with the simple `id` as it tends to be the default standard for MySQL tables but if I'm the only one that likes it better with the short format, we can leave it as is

louismunro: Sort of agree with @julsemaan here.
There is no standard, but everyone seems to use just table.id.

jrouzierinverse:
Check documentation update please

julsemaan:
Looked at it quickly and all seems fine

Just missing the unit tests and we're good for merge

extrafu:
Bump - where are the unit test? We need to merge this ASAP.

jrouzierinverse:
Sorry I thought I pushed it is coming

jrouzierinverse:
It is there

extrafu:
@julsemaan Review the unit test and merge if all is good.

julsemaan:
TBH, I expected a bit more unit tests for a feature that has this much complexity, use cases and corner cases. 

Anyway, I'll remove my changes requested but IMO we should have more than that

julsemaan (approved):

jrouzierinverse:
@julsemaan I add some additional tests for saving invalid data such as an invalid enum and null in a non nullable field.
With that said the use cases are pretty straight forward CRUD operations.
Can you please provide some possible corner cases that you can think of to have a more complete test suite.

julsemaan:
Given the nature of a commit that can occur up to a second after the object creation here are the ones I thought of (quickly):
- Creating a node via ORM then the node gets deleted in the DB (without using ORM, like in the cleanup), then the ORM 'commits' the changes. Check that it can recuperate from this
- Race condition test of creation

julsemaan:
Scanned the code a bit and found those as well:
- Test the copy function of dal
- Test the validation

jrouzierinverse:
Validation is already tested

jrouzierinverse:
The copy function may not be really useful.
I think it would be better to remove it since it only has very limited uses which be handled manually in code.

jrouzierinverse:
Julien your previous two examples are really questions of usage.

The answer to both of those questions is that it will fail.

The question is how to we want to handle it those types of failures.

Do we handle it automatically or manually in code?

Please let me know your thoughts.

julsemaan:
I think we should handle the case of the creation race condition by having a create which is new+immediate save

As for the deletion mid-request, it would be nice if it could re-create the entry and update the appropriate data.

One thing I've noted as well (correct me if I'm wrong), is that failures don't generate a meaningful return, they just return false without a message, so a validation failure and a DB gone failure seem to generate the same return which doesn't seem right.

jrouzierinverse:
> I think we should handle the case of the creation race condition by having a create which is new+immediate save
>
> As for the deletion mid-request, it would be nice if it could re-create the entry and update the appropriate data.

So what you are saying we should automatically handle those type of scenarios in the save method.
Or would it make sense to have a separate function to handle those cases.
 
> One thing I've noted as well (correct me if I'm wrong), is that failures don't generate a meaningful return, they just return false without a message, so a validation failure and a DB gone failure seem to generate the same return which doesn't seem right.

It is fine only if we do not expect to handle errors differently in code.

However If we going to handle those types of errors differently in code then yes it make sense to provide more detailed error messages.

Should we do like go and c when returning an error message.
If an error is a false value then it succeeded.

```perl
my $err = $node->save;
if ($err) {
 ..handle error
}
```
Or should we use the status codes do this.

```perl
my $status = $node->save;
if (is_error($status)) {
 ..handle error
}
```

I would say status codes since we already are doing that in for other parts of the code in perl.
Even though I prefer the go and c style of doing this.

julsemaan:
I believe we'll have to handle errors differently.

There is a big difference between an invalid value and a DB that is unavailable

As for the issue handling, I'd handle it directly in the save method so that its built-in the «official» codepath

jrouzierinverse:
Status code are now returned for publicly consumed methods.
I also now use the upsert pattern (which combines an insert and an update in the same operation) for updating records.

jrouzierinverse:
Please rereview

extrafu:
@julsemaan Required changes were pushed - if all good, merge so we can start using/testing the beast.

julsemaan:
Just 2 last comments and we're good to merge:
- Document what upsert does in the perldoc and what implementation logic it follows (in the SQL::Abstract extension I would believe it belongs) - going all Yoda here
- A test that creates 2 new nodes with the same MAC, saves the first one, then saves the second one and just validate that all is graceful which was the race condition case we discussed in the comments above

jrouzierinverse:
done

t/unittest/dal.t
```diff
@@ -120,6 +120,27 @@ is($node->voip, "yes", "Voip was saved");
 
 pf::dal::node->remove_by_id({mac => $test_mac});
 
+$node = pf::dal::node->new({ mac => $test_mac });
```
julsemaan: Can you set like computername = "zams-computer" here and validate that it will still be there when node2 gets saved.

Just to test that conflicting fields are overwritten but non-conflicting fields are untouched

extrafu:
@julsemaan All good?

PR #2023, opened 2017-02-13
MariaDB Galera cluster - phase #1

julsemaan:
# Description
Adds a MariaDB wrapper for managing MariaDB and everything necessary to generate the MariaDB configuration and galera scripts
Adds the database section in the admin GUI
Adds a maintenance mode for cluster nodes

Still to be done:
* systemd integration
* Adjust the clustering guide to describe the cluster behavior and how to build the cluster
* Packaging to have people install/update to MariaDB 10.1
* Handle sequential full crash of the DB servers (partial gvwstate), see internal wiki for details

# Impacts
The whole clustering stack

# NEW Package(s) required
MariaDB-Server-10.1 and connected packages

# Delete branch after merge
NO

# NEWS file entries
## New Features
* Added built-in support for MariaDB Galera Cluster
* Added a maintenance mode for cluster nodes

## Enhancements
* Added the database section in the admin GUI

# UPGRADE file entries

To be written


bin/mariadb
```diff
@@ -0,0 +1,280 @@
...
+=cut
+
+sub ping_quorum {
+    my $alive = 0;
+    foreach my $server (@cluster_servers) {
+        if(ping($server->{management_ip})) {
+            $alive ++;
+        }
+    }
+    return ($alive > (scalar(@cluster_servers) / 2));
```
fdurand, line 134: test if $alive is even should be enough

bin/mariadb
```diff
@@ -0,0 +1,280 @@
...
+
+exit(0);
+
+sub normal_sighandler {
+    $RUNNING = 0;
+    if(!$IS_CHILD) {
+        kill(SIGKILL, $MYSQLD_PID);
+        `pkill mysqld`;
+        deletepid();
+        $logger->debug( "pfdns: caught SIG" . $_[0] . " - terminating" );
```
fdurand: pfdns

bin/mariadb
```diff
@@ -0,0 +1,280 @@
...
+=cut
+
+sub db_quorum {
+    my $alive = 0;
+
+    foreach my $server (@cluster_servers) {
+        my $host = $server->{management_ip};
+        $alive ++ if(test_db($host));
+    }
+    return ($alive > (scalar(@cluster_servers) / 2));
```
fdurand, line 170: test if $alive is even should be enough

julsemaan: What if we have a cluster of 5 servers

2 is even but isn't a quorum

extrafu:
Bump - we need to merge this quickly to get the ball rolling on testing.

.gitignore
```diff
@@ -141,6 +141,7 @@ conf/templates/emails-guest_sponsor_activation.txt.tt
 conf/templates/emails-guest_sponsor_confirmed.txt.tt
 conf/templates/emails-guest_sponsor_preregistration.txt.tt
 conf/dns_filters.conf
+conf/mariadb/*.tt
```
louismunro, line 144: Why ignore that file?

bin/mariadb
```diff
@@ -0,0 +1,280 @@
...
+
+# Time to wait between attempts to discover alive servers
+# TODO: move this to configuration
+our $WAIT_TIME_ALIVE = 10;
+
+our $failed_counts = {};
+
+# init signal handlers
+POSIX::sigaction(
+    &POSIX::SIGHUP,
```
louismunro: If we are going to define a sighandler for SIGHUP, might it not do something more useful than killing the process?

bin/mariadb
```diff
@@ -0,0 +1,280 @@
...
+
+sub launch_mysql {
+    my (@args) = @_;
+    
+    my $args = join(' ', @ARGV, @args);
+    # Must not lookup in /sbin for mysqld, but in /usr/sbin
+    $ENV{PATH} = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin:/root/bin";
+    my $cmd = "mysqld_safe --defaults-file=/usr/local/pf/var/conf/mariadb.conf $args";
+    print "Starting MySQL with command: $cmd \n";
+    `$cmd`;
```
louismunro, line 93: Are we sure this does not call /bin/sh in the background?
Meaning we have an additional process between us?

.gitignore
julsemaan: Because they are templates, we should track the .example

bin/mariadb
julsemaan: Thats how all of our processes behave, I can simply remove it if you want

louismunro: I would remove it.
Let's define it for real when we actually want to implement dynamic configuration reloading.
Otherwise that signal breaks user expectations. That is unfortunately true of all our other processes.

conf/documentation.conf
```diff
@@ -1011,6 +1011,48 @@ description=<<EOT
...
+description=<<EOT
+The max_connections MySQL configuration attribute. Only change if you know what you are doing. Will only affect a locally running MySQL server.
+EOT
+
+[database.table_cache]
+type=numeric
+description=<<EOT
+The table_cache MySQL configuration attribute. Only change if you know what you are doing. Will only affect a locally running MySQL server.
+EOT
+
```
louismunro, line 1055: We are missing (at least) two important values:
max_allowed_packet=16M
thread_cache_size=50

conf/mariadb/mariadb.conf.tt.example
```diff
@@ -0,0 +1,160 @@
...
+socket    = /var/lib/mysql/mysql.sock
+skip-external-locking
+key_buffer_size = [% key_buffer_size %]M
+max_allowed_packet = 16M
+table_open_cache = 512
+sort_buffer_size = 2M
+read_buffer_size = 2M
+read_rnd_buffer_size = 8M
+myisam_sort_buffer_size = 64M
+thread_cache_size = 10
```
louismunro: I suggest making this configurable.
10 may not be enough for everyone.
Since it's so cheap, I would at least raise the default to 25 or even 50.

louismunro:
Apart from the startup script which we must discuss, I see no problem.
I suggested a few minor changes to add two managed variables to mysql but that's all.

bin/mariadb
jrouzierinverse: You are correct exec would better

extrafu:
So we're ok to merge?

louismunro:
@julsemaan and I need to have a discussion about the startup script.
Then we can merge.

julsemaan:
@extrafu 

I'll sit down with @louismunro and fix this in the next hour and I guess we'll be good to merge

extrafu:
Great 👍

louismunro:
Mergèd.

PR #2036, opened 2017-02-22
feature/pfdhcplistener-master

jrouzierinverse:
Description
-----------
pfdhcplistner now run as a master-worker style service

Impacts
-----------
pfdhcplistner, Admin GUI, pfcmd service

NEWS file entries
-------

Enhancements
++++++++++++
* pfdhcplistner now run as a master-worker style service

Delete branch after merge
-------------------------
NO

### DIFF
--- a/lib/pf/services/manager/pfdhcplistener.pm
+++ b/lib/pf/services/manager/pfdhcplistener.pm
@@ -21,8 +21,6 @@ extends 'pf::services::manager';
 
 has '+name' => (default => sub { 'pfdhcplistener'} );
 
-has '+launcher' => (default => sub { '%1$s -d' } );
-
 sub isManaged {
     my ($self) = @_;
     return isenabled($Config{'network'}{'dhcpdetector'}) && $self->SUPER::isManaged();
--- a/lib/pf/services/manager/pfdhcplistener.pm
+++ b/lib/pf/services/manager/pfdhcplistener.pm
@@ -14,71 +14,19 @@ pf::services::manager::pfdhcplistener
 use strict;
 use warnings;
 use Moo;
-use pf::config qw(
-    %Config
-    @listen_ints
-    @dhcplistener_ints
-);
-use pf::file_paths qw(
-    $var_dir
-    $install_dir
-    $systemd_unit_dir
-);
-use pf::util;
-use List::MoreUtils qw(any all uniq);
 use pf::log;
+use pf::config qw(%Config);
 
-extends 'pf::services::manager::submanager';
-
-has pfdhcplistenerManagers => ( is => 'rw', builder => 1, lazy => 1);
+extends 'pf::services::manager';
 
 has '+name' => (default => sub { 'pfdhcplistener'} );
 
-has 'int' => (is => 'ro');
-
-sub _cmdLine {
-    my $self = shift;
-    "$install_dir/sbin/pfdhcplistener -i " . $self->int;
-}
-
-sub _build_pfdhcplistenerManagers {
-    my ($self) = @_;
-    my @managers = map {
-        pf::services::manager::pfdhcplistener->new(
-            {   name                    => "pfdhcplistener_$_",
-                forceManaged            => $self->isManaged,
-                systemdTemplateFilePath => $self->systemdTemplateFilePath,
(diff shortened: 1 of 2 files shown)


lib/pf/services/manager/pfdhcplistener.pm
```diff
@@ -14,72 +14,20 @@ pf::services::manager::pfdhcplistener
...
-            )
-    } uniq @listen_ints, @dhcplistener_ints;
-    return \@managers;
-}
-
-sub managers {
-    my ($self) = @_;
-    return @{$self->pfdhcplistenerManagers};
-}
+has '+launcher' => (default => sub { '%1$s -d' } );
```
louismunro: Don't bother with laucher anymore.
It's not used by systemd.

sbin/pfdhcplistener
```diff
@@ -42,61 +43,50 @@ use pf::config qw(
...
+$SIG{INT}  = \&normal_sighandler;
+$SIG{TERM} = \&normal_sighandler;
+$SIG{CHLD} = \&child_sighandler;
+$SIG{ALRM} = \&alarm_sighandler;
+
+$SIG{PIPE} = 'IGNORE';
 
 my %args;
-getopts( 'dhi:', \%args );
+getopts( 'dh', \%args );
```
louismunro, line 81: We may want to remove the -d option.
Systemd will not use it.

sbin/pfdhcplistener
```diff
@@ -343,10 +360,214 @@ sub process_pkt {
...
+    }
+    elsif ($pid == 0) {
+        $SIG{CHLD} = 'DEFAULT';
+        $IS_CHILD = 1;
+        Log::Log4perl::MDC->put('tid', $$);
+        setup_global($task);
+        reload_config();
+        _runtask($interface);
+    }
+    else {
```
louismunro: why the final else?

sbin/pfdhcplistener
```diff
@@ -180,7 +197,7 @@ sub dhcp_detector {
     # updating process name so we know what interface we are listening on
     # WARNING: the format is expected by watchdog in pf::services. Don't change lightly.
     $PROGRAM_NAME = basename($PROGRAM_NAME) . ": listening on $interface";
-    $pcap = Net::Pcap::pcap_open_live( $interface, $Config{advanced}{pfdhcplistener_packet_size}, 1, 0, \$err );
+    $pcap = Net::Pcap::pcap_open_live( $interface, $Config{advanced}{pfdhcplistener_packet_size}, 1, 10 * 1000, \$err );
```
extrafu, line 200: Why a read timeout now?

sbin/pfdhcplistener
```diff
@@ -343,10 +360,214 @@ sub process_pkt {
...
+
+=head2 registertasks
+
+    Register all tasks
+
+=cut
+
+sub registertasks {
+    for my $task (get_tasks()) {
+        register_task($task);
```
louismunro, line 504: minor nitpick here, but can we have a better name than register_task?
even _doRegisterTask would be better, or else it's confusing to have two subs with pretty much the same name.

sbin/pfdhcplistener
julsemaan: Don't believe we should, it can come in handy at some point, especially in testing and when in 5 years we get zammitsd as a new init system which will make us go back to daemonized processes, we'll be ready

So lesson here, always be ready for zammitsd

louismunro: Nobody fucks with zammitsd.
Agreed.

sbin/pfdhcplistener
jrouzierinverse: This was done so a child process can see if the parent process is still running when there was no traffic coming in and shut itself down properly.

sbin/pfdhcplistener
louismunro: Forget it.
It's not worth the trouble.

louismunro:
Apart from my minor requests (including the launcher) this looks good.
I just tested it and it works with systemd.
Please make the few minor changes (launcher and "else") and I will merge.

PR #2037, opened 2017-02-23
pfsso + pfconfig resource pool

### LINKED ISSUE #1144: firewall sso modules need a cleanup
Firewall SSO modules are all checking if the user belongs to a role instead of checking it in a super class.

Also, Start + Stop are mixed in the same subroutine.

Both should be extracted into their own sub.

julsemaan:
# Items for discussion:
* Do we remove lib/pf/firewallsso along with its associated code or do we wait in case we want to rollback. That code can stay in place if we want as both codepaths are not incompatible

# Missing
* Packaging of pfhttpd

# Description
Adds pfhttpd binary (service wrapper that will include all caddy based pf services)
Adds pfsso service (to serve firewall SSO requests)
Adds pfconfig resource pool for resource reloading
Adds pfconfigpool caddy middleware to reload the pfconfig resources in caddy
PacketFence API (pf::api) will now talk to pfsso for firewall SSO requests

# Impacts
Firewall SSO

# NEW Package(s) required
None

# Issue
fixes #1144

# Delete branch after merge
NO, we will use it for patching

# NEWS file entries
## New Features
* Added a new Golang based pfsso service to handle the Firewall SSO requests


julsemaan:
Other than the packaging, this is ready

louismunro:
That's a big one you got there.
I didn't see anything objectionable, but I will feel better when we test it against a real FW.
Let's remove the old code. We can add it back easily if it comes to that and otherwise it's just going to be forgotten and rot.

PR #2038, opened 2017-02-23
Feature/winbindd wrapper

julsemaan:
# Description
Adds a winbindd wrapper for the PacketFence managed winbindd processes

# Impacts
winbindd processes managed by PacketFence

# Delete branch after merge
YES


sbin/winbindd-wrapper
```diff
@@ -0,0 +1,310 @@
...
+
+start();
+
+while($running){
+    sleep 1;
+    foreach my $pid (keys %CHILDREN){
+        unless(kill(0,$pid)){
+            $logger->error("Child $pid ($CHILDREN{$pid}) is dead. Respawning it.");
+            run_winbindd($CHILDREN{$pid});
+            delete $CHILDREN{$pid};
```
jrouzierinverse: This will cause a race condition if that pid was ever reused by another process since it may have been already cleaned in the child handler.

Use the pattern in pfmon, pffilter, pfdhcplistener and pfqueue where they keep track of which child processes have died and restart them when they have died.

sbin/winbindd-wrapper
```diff
@@ -0,0 +1,310 @@
...
+=head1 SUBROUTINES
+
+=head2 cleanup
+
+cleans after children
+
+=cut
+
+sub cleanup {
+    signal_children('INT');
```
jrouzierinverse, line 149: Are we going to handle the case where if the child processes take too long to die we send a KILL signal?

sbin/winbindd-wrapper
```diff
@@ -0,0 +1,310 @@
...
+sub run_all_winbindd {
+    my $mask = POSIX::SigSet->new(POSIX::SIGCHLD());
+    sigprocmask(SIG_BLOCK,$mask);
+    foreach my $id (keys(%ConfigDomain)) {
+        run_winbindd($id);
+    }
+    sigprocmask(SIG_UNBLOCK,$mask);
+}
+
+=head2 runtask
```
jrouzierinverse: run_winbindd

sbin/winbindd-wrapper
```diff
@@ -0,0 +1,310 @@
...
+    &POSIX::SIGALRM,
+    POSIX::SigAction->new(
+        'alarm_sighandler', POSIX::SigSet->new(), &POSIX::SA_NODEFER
+    )
+) or die("winbindd-wrapper could not set SIGALRM handler: $!");
+
+POSIX::sigaction(
+    &POSIX::SIGHUP,
+    POSIX::SigAction->new(
+        'reload_config' , POSIX::SigSet->new(), &POSIX::SA_NODEFER
```
jrouzierinverse: There is no matching reload_config function

sbin/winbindd-wrapper
julsemaan: I chose to not handle it

sbin/winbindd-wrapper
julsemaan: Unfortunately I couldn't do that because the exec stays alive as a zombie and that made the detection not work given its still alive and I couldn't find a way to determine its a zombie.

If you have a way let me know, but that was the only way I could get it to work

sbin/winbindd-wrapper
louismunro: We can kill all those process with systemd if they take too long to die.

jrouzierinverse: Can systemd monitor the children of a process?

sbin/winbindd-wrapper
jrouzierinverse: Using the pattern used by pfmon, pffilter, pfdhcplistener and pfqueue will avoid the zombies from happening.

sbin/winbindd-wrapper
julsemaan: @louismunro, we just tested and it works so we're good

PR #2044, opened 2017-02-24
feature/pfdhcplistener-process-readonly-mode

jrouzierinverse:
Description
-----------
Do not process dhcp packets when the database is in readonly

Impacts
-----------
dhcp processing

NEWS file entries
-------

New Features
++++++++++++
* Do not process dhcp packets when the database is in readonly

Delete branch after merge
-------------------------
NO

### DIFF
--- a/lib/pf/dhcp/processor.pm
+++ b/lib/pf/dhcp/processor.pm
@@ -150,6 +150,10 @@ Process a packet
 sub process_packet {
     my $timer = pf::StatsD::Timer->new();
     my ( $self ) = @_;
+    if (db_check_readonly()) {
+        $logger->trace("The database is in readonly mode skipping processing the database");
+        return;
+    }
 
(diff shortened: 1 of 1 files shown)


louismunro:
Frankly, this is so small as to not be worth a PR.

lib/pf/dhcp/processor.pm
```diff
@@ -150,6 +150,10 @@ Process a packet
 sub process_packet {
     my $timer = pf::StatsD::Timer->new();
     my ( $self ) = @_;
+    if (db_check_readonly()) {
```
louismunro, line 153: Where is that defined?

jrouzierinverse: pf::db

louismunro: ok then.

PR #2067, opened 2017-03-01
Feature/go dispatcher reload

julsemaan:
# Description
Rework the httpdispatcher so that it reloads its config dynamically

# Impacts
httpdispatcher

# Delete branch after merge
YES


louismunro:
good enough.

PR #2080, opened 2017-03-08
Feature/encrypt quorum comm

julsemaan:
# Description
Encrypt the communications to acquire MariaDB quorum using the clustering password

# Impacts
MariaDB crash recovery

# Delete branch after merge
YES

### DIFF
--- a/lib/pf/cluster.pm
+++ b/lib/pf/cluster.pm
@@ -647,6 +647,7 @@ sub deactivate_maintenance {
 
 Takes the active_active.password and ensures its a 56 bytes password
 Will pad the password with zeros if its less than that
+Will strip the password to 56 bytes if its more than that
 
 =cut
 
@@ -666,6 +667,12 @@ sub encryption_password {
 
     return $password;
 }
+
+=head2 cipher
+
+Get the cipher to encrypt/decrypt cluster communications
+
+=cut
 
 sub cipher {
     return Crypt::CBC->new(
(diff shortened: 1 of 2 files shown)


lib/pf/cluster.pm
```diff
@@ -644,6 +645,66 @@ sub deactivate_maintenance {
...
+    require pf::config;
+    my $password = $pf::config::Config{active_active}{password};
+    
+    my $desired_length = 56;
+
+    # Ensure it isn't more than 56 bytes
+    $password = substr($password, 0, $desired_length);
+
+    my $missing = $desired_length - length($password) - 1;
+    for my $i (0..$missing) {
```
extrafu, line 664: eww! how about something cleaner like we could do in C:

foo = calloc(56, sizeof(char));
strncpy(foo, password, strlen(password));

calloc == zero'ed buffer.

julsemaan: Doesn't seem like what you said exists in a simple form in perl (from my vast knowledge was acquired across 2 minutes of googling this)

We could do:
```
$password .= '0' x ($desired_length - length($password) - 1)
```

Which would be on a single line and be cleaner

extrafu: For advanced Perl questions, ask @lzammit

extrafu:
Reebayise

julsemaan:
ribaysaide

PR #2079, opened 2017-03-07
Feature/clustering upgrade doc

julsemaan:
# Description
Upgrade documentation for the new clustering stack

# Impacts
Documentation

# Delete branch after merge
YES

PR #2082, opened 2017-03-08
Feature/disable cluster node

julsemaan:
# Description
Allow to disable a cluster node on a specific cluster node (my head hurts)
This is required so #2079 works fine

# Impacts
Whole clustering stack

# Delete branch after merge
YES


julsemaan:
bump

PR #2088, opened 2017-03-13
Feature/webauth redirect map

julsemaan:
# Description
Support a Ruckus SmartZone dedicated landing URL for web auth
Requires to add support for mapping URLs to modules since the SmartZone cannot accept colons in its redirect URL

# Impacts
Web-auth / Ruckus SmartZone

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed SmartZone specific handling of redirect URL


lib/pf/web/externalportal.pm
```diff
@@ -34,6 +35,12 @@ use pf::web::constants;
 use pf::web::util;
 use pf::constants;
 
+# Some vendors don't support some charatcters in their redirect URL
+# This here below allows to map some URLs to a specific switch module
+Readonly our $SWITCH_REWRITE_MAP => {
+    'RuckusSmartZone' => 'Ruckus::SmartZone',
```
extrafu, line 41: Shouldn't be the opposite? It looks to me we want to rewrite Ruckus::SmartZone to RuckusSmartZone.

julsemaan: No, we want to the URL /RuckusSmartZone to the name of the module (Ruckus::SmartZone) since you can't put `:` in the redirect URL in the SmartZone itself

PR #2090, opened 2017-03-13
feature/radius-readonly

jrouzierinverse:
Description
-----------
Allow radius to operate in readonly mode

Impacts
-----------
RADIUS authorize workflow

NEWS file entries
-------

New Features
++++++++++++
*  Allow radius authoize to work in readonly mode

Delete branch after merge
-------------------------
NO

### DIFF
--- a/lib/pf/dal.pm
+++ b/lib/pf/dal.pm
@@ -601,7 +601,7 @@ merge fields into object
 sub merge {
     my ($self, $vals) = @_;
     return unless defined $vals && ref($vals) eq 'HASH';
-    foreach my $field ($self->field_names) {
+    foreach my $field ( @{$self->field_names} ) {
         next unless exists $vals->{$field};
         $self->{$field} = $vals->{$field};
     }
--- a/lib/pf/dal.pm
+++ b/lib/pf/dal.pm
@@ -122,16 +122,22 @@ Find the pf::dal object by it's primaries keys
 =cut
 
 sub find {
-    my ($proto, @ids) = @_;
-    my $sql = $proto->_find_one_sql;
-    my ($status, $sth) = $proto->db_execute($sql, @ids);
+    my ($proto, $ids) = @_;
+    my $where = $proto->build_primary_keys_where_clause($ids);
+    my $sqla = $proto->get_sql_abstract;
+    my ($sql, @bind) = $sqla->select(
+        -columns => $proto->field_names,
+        -from => $proto->find_from_tables,
+        -where => $where,
+    );
+    my ($status, $sth) = $proto->db_execute($sql, @bind);
     return $status, undef if is_error($status);
     my $row = $sth->fetchrow_hashref;
+    $sth->finish;
     unless ($row) {
         return $STATUS::NOT_FOUND, undef;
     }
     my $dal = $proto->new_from_table($row);
-    $sth->finish;
     return $STATUS::OK, $dal;
 }
 
@@ -533,6 +539,17 @@ sub create {
     return $obj->insert;
 }
 
+=head2 find_from_tables
+
+find_from_tables
+
+=cut
+
+sub find_from_tables {
+    my ($proto) = @_;
+    return $proto->table;
+}
+
 =head2 find_or_create
 
 finds a table record or creates it
@@ -545,7 +562,7 @@ sub find_or_create {
     my $sqla = $proto->get_sql_abstract;
     my ($sql, @bind) = $sqla->select(
         -columns => $proto->field_names,
-        -from => $proto->table,
+        -from => $proto->find_from_tables,
         -where => $obj->primary_keys_where_clause,
     );
     my ($status, $sth) = $proto->db_execute($sql, @bind);
--- a/lib/pf/dal.pm
+++ b/lib/pf/dal.pm
@@ -594,7 +594,7 @@ sub find_or_create {
 
 =head2 merge_fields
 
-merge_fields
+An array ref of the fields to merge
 
 =cut
 
--- a/lib/pf/dal.pm
+++ b/lib/pf/dal.pm
@@ -592,6 +592,22 @@ sub find_or_create {
     return $status, $obj;
 }
 
+=head2 merge
+
+merge fields into object
+
+=cut
+
+sub merge {
+    my ($self, $vals) = @_;
+    return unless defined $vals && ref($vals) eq 'HASH';
+    foreach my $field ($self->field_names) {
+        next unless exists $vals->{$field};
+        $self->{$field} = $vals->{$field};
+    }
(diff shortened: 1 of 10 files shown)


extrafu:
@julsemaan Review and if/when it's fine, please merge to get the ball rolling on testing.

lib/pf/radius.pm
```diff
@@ -232,16 +232,11 @@ sub authorize {
...
-        # automatic registration
-        my %autoreg_node_defaults = $role_obj->getNodeInfoForAutoReg($args);
-        $args->{'node_info'} = merge($args->{'node_info'}, \%autoreg_node_defaults);
-        $logger->debug("[$mac] auto-registering node");
-        ( $status, $status_msg ) = pf::node::node_register($mac, $autoreg_node_defaults{'pid'}, %autoreg_node_defaults);
-        if (!$status) {
-            $logger->error("auto-registration of node failed");
-        }
+        $node_obj->autoreg("yes");
+        %autoreg_node_defaults = $role_obj->getNodeInfoForAutoReg($args);
```
julsemaan: Correct me if I'm wrong but the node attributes / node info will not be modified during the autoreg process here but only at the end of the request.

That means the changes done to the node for auto-registration won't apply (status, role, ...) until the end of the request which means the node will not be auto-registered in this request and thus won't have the right RADIUS attributes sent back

From what I can see, the autoreg process in the cleanup should be done here except for the save which can be done in the cleanup

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+    if ($status == $STATUS::CREATED ) {
+        pf::lookup::person::async_lookup_person($pid, $info->{'source'});
+    }
+    if ($person) {
+        $person->source($info->{source});
+        $person->portal($info->{portal});
+        $person->save;
+    }
+
+    # if it's for auto-registration and mac is already registered, we are done
```
julsemaan: Says we are done but it seems it does stuff

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+            $status_msg = '';
+            if (is_error($status)) {
+                $status_msg = "modify of node $mac failed";
+                $logger->error($status_msg);
+            }
+            return ($status, $status_msg);
+        }
+    }
+    else {
+    # do not check for max_node if it's for auto-register
```
julsemaan: That comment doesn't look like it belongs with the code below

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+        if ( is_max_reg_nodes_reached($mac, $pid, $node->category, $node->category_id) ) {
+            $status_msg = "max nodes per pid met or exceeded";
+            $logger->error( "$status_msg - registration of $mac to $pid failed" );
+            return ($STATUS::PRECONDITION_FAILED, $status_msg);
+        }
+    }
+
+    $node->status('reg');
```
julsemaan: use the pf::node constant

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+    $node->regdate(mysql_date());
+    $status = $node->save;
+
+    if (is_error($status)) {
+        $status_msg = "modify of node $mac failed";
+        $logger->error($status_msg);
+        return ($status, $status_msg);
+    }
+    $pf::StatsD::statsd->increment( called() . ".called" );
```
julsemaan: Would be cleaner if you increment it at the top of the sub instead of the middle

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+
+    # Closing any parking violations
+    # loading pf::violation here to prevent circular dependency
```
julsemaan: Doesn't look like you're loading it

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+    pf::violation::violation_force_close($mac, $PARKING_VID);
+
+    do_violation_scans($mac);
+
+    return ($STATUS::OK, "");
+}
+
+=head2 do_violation_scans
+
+do_violation_scans
```
julsemaan: Could use a bit more description

lib/pf/registration.pm
```diff
@@ -0,0 +1,154 @@
...
+use pf::violation;
+use pf::util;
+use pf::dal::person;
+use pf::Portal::ProfileFactory; 
+use pf::constants::scan qw($SCAN_VID $POST_SCAN_VID);
+use pf::constants::parking qw($PARKING_VID);
+
+=head2 register_node
+
+register_node
```
julsemaan: A bit more description would be good

lib/pf/dal.pm
```diff
@@ -564,6 +592,33 @@ sub find_or_create {
     return $status, $obj;
 }
 
+=head2 merge_fields
+
+merge_fields
```
julsemaan: Add a bit more description here

lib/pf/radius.pm
julsemaan: On top of this, role.pm also seems to still call node_modify

It would be better if getNodeInfoForAutoReg worked on the DAL object because right now autoreg_node_defaults will not contain everything that is set by getNodeInfoForAutoReg (missing unregdate and PID for instance)

jrouzierinverse:
Please rereview

lib/pf/registration.pm
```diff
@@ -0,0 +1,163 @@
...
+use pf::constants::node qw($STATUS_REGISTERED);
+use pf::util;
+use pf::dal::person;
+use pf::Portal::ProfileFactory; 
+use pf::constants::scan qw($SCAN_VID $POST_SCAN_VID);
+use pf::constants::parking qw($PARKING_VID);
+
+=head2 setup_node_for_registration
+
+setup a node for registration and do the following actions
```
julsemaan: which actions ?

lib/pf/registration.pm
```diff
@@ -0,0 +1,163 @@
...
+    }
+    $node->status($STATUS_REGISTERED);
+    $node->regdate(mysql_date());
+
+    return ($STATUS::OK, "");
+}
+
+=head2 finalize_node_registration
+
+do the node registration after being saving
```
julsemaan: after saving

julsemaan:
Codewise this looks good.

@jrouzierinverse, can you just confirm the smokes pass for this and we'll merge

extrafu:
@jrouzierinverse How about the smoke tests?

jrouzierinverse:
They just finished running and it is "good" as in it did not cause a new test to fail.

extrafu:
Perfect, merging.

PR #2093, opened 2017-03-15
add option to single fetch users during ntlm cache job

julsemaan:
# Description
Added option to fetch users indiviually during the NTLM auth cache job due to side effects a job with thousands of users can have on an AD

# Impacts
NTLM auth caching

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Added option to fetch users one by one in the NTLM cache instead of all together

### DIFF
--- a/lib/pf/domain/ntlm_cache.pm
+++ b/lib/pf/domain/ntlm_cache.pm
@@ -142,7 +142,7 @@ sub fetch_hashes_one_at_a_time {
     my @users = split(/\n/, read_file($users_file));
     foreach my $user (@users) {
 
-        my ($ntds_file, $msg) = secretsdump($domain, $source, "-just-dc-user $user");
+        my ($ntds_file, $msg) = secretsdump($domain, $source, "-just-dc-user '$user'");
         if ($ntds_file) {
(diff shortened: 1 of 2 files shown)


lib/pf/domain/ntlm_cache.pm
```diff
@@ -110,9 +164,8 @@ Fetch the NT hashes of all the valid users of a domain
 sub fetch_all_valid_hashes {
     my ($domain) = @_;
     my $logger = get_logger;
+    
     my $config = $ConfigDomain{$domain};
-    my $source = getAuthenticationSource($config->{ntlm_cache_source});
```
dwlfrth, line 114: Can't this stays here instead of duplicating the logic in the callee ?

dwlfrth (commented):
Looks good other than the one comment

lib/pf/domain/ntlm_cache.pm
julsemaan: I'd have to pass the source object in the other subs and change their signatures.

Nothing against that, but I was trying to keep the subroutines with simple parameters since its easier to call them from the command line when you troubleshoot it

PR #2100, opened 2017-03-17
Fix accounting_unique_session_id to not be too much unique

fdurand:
# Description
We have multiples radacct entries open for the same mac address because the accounting_unique_session_id is too unique (if the device roam)

# Impacts
Reduce the number of entries in the radacct table

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fix multiples radacct open entries because the accounting unique session id is too much unique

### DIFF
--- a/conf/radiusd/packetfence-tunnel.example
+++ b/conf/radiusd/packetfence-tunnel.example
@@ -242,13 +242,6 @@ session {
 #  the ONLY safe way is to set "use_tunneled_reply = yes", and
 #  then update the inner-tunnel reply.
 post-auth {
-	#  Insert class attribute (with unique value) into response,
-	#  aids matching auth and acct records, and protects against duplicate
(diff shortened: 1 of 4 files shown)


julsemaan:
Personally, I'd like to see this reach v7 since that should solve most of our accounting scaling issues

fdurand:
I'd like to fix another issue with the accounting sql procedure

dwlfrth:
@fdurand Add missing commit and check backporting for 6.5.0 & 6.5.1

fdurand:
Commit done, sql procedure will be in another branch

julsemaan:
We never stated on whether or not it should reach the maintenance given it changes a behavior.

What do we do ?

louismunro:
If this part does not require the SQL stored proc and it helps I am in favour of adding it to maintenance.

PR #2111, opened 2017-03-22
X-Frame-Options SAMEORIGIN enforced for browsers that supports it.

tlaurion:
# Description
X-Frame-Options SAMEORIGIN informs the browsers that only frames coming from the same origin of the main site are permitted to be generated in a frame. We now instruct that only the domain associated with PF should render pages in a frame only for its own domain.

This mitigates clickjacking attacks.

Might impact Billing:
 * Mirapay  ( @jrouzierinverse: I would love to test this with you when you have the time.)
 

# Impacts
Opening PF from an iframe won't be possible ifthe original request doesn't come from PF.

# Delete branch after merge
YES

## Enhancements
* Clickjacking attack prevention enforcement for recent browsers.

### DIFF
--- a/conf/httpd.conf.d/httpd.proxy.tt.example
+++ b/conf/httpd.conf.d/httpd.proxy.tt.example
@@ -152,7 +152,6 @@ SSLStaplingCache        shmcb:/var/run/ocsp(128000)
 [% END %]
 
 SSLHonorCipherOrder  on
-Header always append X-Frame-Options SAMEORIGIN
 ErrorLog  [% install_dir %]/logs/[% name %].error
 
 ServerAdmin [% server_admin %]
--- a/conf/httpd.conf.d/httpd.proxy.tt.example
+++ b/conf/httpd.conf.d/httpd.proxy.tt.example
@@ -152,7 +152,7 @@ SSLStaplingCache        shmcb:/var/run/ocsp(128000)
 [% END %]
 
 SSLHonorCipherOrder  on
-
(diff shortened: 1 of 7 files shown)


conf/httpd.conf.d/httpd.webservices.tt.example
```diff
@@ -159,6 +159,7 @@ SSLHonorCipherOrder  on
 
 TraceEnable Off
 
+Header always append X-Frame-Options SAMEORIGIN
```
jrouzierinverse: We do not need it here

conf/httpd.conf.d/httpd.proxy.tt.example
```diff
@@ -145,7 +145,7 @@ SSLCipherSuite ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-R
 [% END %]
 
 SSLHonorCipherOrder  on
-
+Header always append X-Frame-Options SAMEORIGIN
```
jrouzierinverse: I am not sure if this is should be here as it may break external sites

jrouzierinverse (changes requested):

jrouzierinverse:
rebase

tlaurion:
rabayzed.

tlaurion:
@jrouzierinverse : Could you validate that Mirapay was manually tested against a recent PF7 deployment just to make sure that the headers didn't break that feature? (it shouldn't but we were unsure)

PR #2113, opened 2017-03-23
Updated procedure in order to detect Interim update just after stop

fdurand:
# Description
Some NAS send an interim update just after a stop for the same session, it occur that we open a new radacct entry with wrong accounting values. In order to fix it we try to detect if a stop match the interim update before trying to open a new radacct entry.

# Impacts
RADIUS accounting workflow

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Avoid opening a double entry with wrong accounting values


fdurand:
Bump

dwlfrth:
Ready for review ?

fdurand:
Yeah !!!

louismunro (approved):

PR #2114, opened 2017-03-23
Enforces block behavior when a browser detects an XSS attempt targeting PF

tlaurion:
# Description
IE, Chrome and Safari enforces X-XSS filter. 

Rather than sanitize the page, when a XSS attack is detected, the browser is instructed to prevent rendering of the page completely and serves an empty page instead.

# Impacts
None

# Delete branch after merge
YES

# News
Cross-site scripting (XSS) filtering is now requested from your browser.


jrouzierinverse:
Please add the same headers to 
 conf/httpd.conf.d/httpd.parking.tt.example
 conf/httpd.conf.d/httpd.graphite.tt.example

PR #2124, opened 2017-03-27
Feature/rsyslog config

louismunro:
# Description
Configures all logging to be sent to syslog.
Configures rsyslog to then writes the messages to the right files in the pf/logs direstory.

# Impacts
Logging should still work as before.
The only difference ought to be that it's rsyslog that writes to the files instead of the PF processes directly.

# NEW Package(s) required
perl-Log-Dispatch

# Delete branch after merge
YES

# NEWS file entries

## New Features
* All logging is now done through syslog.

# UPGRADE file entries
Since all logging now goes through rsyslog, if you had edited the logging configuration (e.g. to forward logs to a centralized syslog server) make sure that the new logging rules in /etc/rsyslog.d/packetfence.conf do not conflict with your changes.
Take a look at /usr/local/pf/conf/log.conf and /usr/local/pf/conf/log.conf.d/* for the detailed configuration of the PacketFence services.


conf/log.conf.d/httpd.aaa.conf.example
```diff
@@ -1,21 +1,6 @@
 ### httpd.aaa logger ###
-log4perl.rootLogger = INFO, HTTPD_AAA
```
julsemaan: We can't control the log level of this anymore ?

conf/radiusd/acct.conf.example
```diff
@@ -86,7 +86,7 @@ log {
 	#  The exact values permitted here are OS-dependent.  You probably
 	#  don't want to change this.
 	#
-	syslog_facility = daemon
+	syslog_facility = local2
```
julsemaan, line 89: why not local5 like the other services ?

conf/radiusd/auth.conf.example
```diff
@@ -93,7 +93,7 @@ log {
 	#  The exact values permitted here are OS-dependent.  You probably
 	#  don't want to change this.
 	#
-	syslog_facility = daemon
+	syslog_facility = local1
```
julsemaan, line 96: local5 like others ?

conf/radiusd/cli.conf.example
```diff
@@ -85,7 +85,7 @@ log {
 	#  The exact values permitted here are OS-dependent.  You probably
 	#  don't want to change this.
 	#
-	syslog_facility = daemon
+	syslog_facility = local3
```
julsemaan, line 88: same, same

conf/radiusd/eduroam.conf.example
```diff
@@ -86,7 +86,7 @@ log {
 	#  The exact values permitted here are OS-dependent.  You probably
 	#  don't want to change this.
 	#
-	syslog_facility = daemon
+	syslog_facility = local4
```
julsemaan, line 89: again

conf/radiusd/load_balancer.conf.example
```diff
@@ -88,7 +88,7 @@ log {
 	#  The exact values permitted here are OS-dependent.  You probably
 	#  don't want to change this.
 	#
-	syslog_facility = daemon
+	syslog_facility = local5
```
julsemaan, line 91: you, I like you

conf/radiusd/sql.conf.example
```diff
@@ -442,3 +442,248 @@ sql sql_reject {
         # Read database-specific queries
         $INCLUDE ${modconfdir}/${.:name}/main/mysql/reject.conf
 }
+
+sql sqlaudit {
```
julsemaan: Sure this belongs in this PR ?

packetfence.rsyslog
```diff
@@ -0,0 +1,103 @@
+# RADIUS
+if $syslogtag contains "auth" and $syslogfacility-text == "local1" then -/usr/local/pf/logs/radius.log
```
julsemaan, line 2: AHHHH, I think I get it about what I said above

But, is there a way to accomplish the same thing without doing the local changing dance ?

conf/log.conf.d/httpd.portal.conf.example
```diff
@@ -1,56 +1,31 @@
 ### httpd.portal logger ###
-log4perl.rootLogger = INFO, HTTPD_PORTAL
+#log4perl.rootLogger = INFO, PFSYSLOG
```
julsemaan: Why commented ?

And shouldn't this be packetfence instead of pfsyslog ?

conf/log.conf.d/httpd.aaa.conf.example
louismunro: Good point.
This uses info as default.
I will add the config lines to change it.

packetfence.rsyslog
louismunro: Not really.
I can't tag the syslog messages from FreeRADIUS the way I can do it from apache.
It's a kludge, but it works.

conf/radiusd/sql.conf.example
louismunro: Now that you mention it, no...

conf/log.conf.d/httpd.portal.conf.example
louismunro: Because it inherits from the root logger in log.conf.
This is just a place to override it if required.
But I can see your point.
I will give it a separate configuration from the start, even if it's the same as the root.

louismunro: I thought PFSYSLOG made it clearer that this is sent to syslog and not appended to a file.

louismunro:
Fixed according to comments.
Please take another look.

julsemaan:
pfhttpd (pfsso/httpd.dispatcher) is still logging in /var/log/messages from what I can see here

Could you put the output of all processes named pfhttpd in packetfence.log and we'll filter by pfsso and httpd.dispatcher when we set their process name if need be

louismunro:
Ahhh, pfhttpd...
Let me test that and see what I can do.

louismunro:
Frickin' rebase...
Added pfhttpd config. Should be good now.

PR #2135, opened 2017-03-28
Feature/ntp addition in packaging

tlaurion:
# Description
NTP is added in both Debian and CentOS packages.

# Impacts
Nightmare stories mitigated with -L option (do not deal with virtual interfaces)
Will attempt to sync time with OS default's NTP servers.

# NEW Package(s) required
ntp

# Delete branch after merge
YES

# NEWS file entries
* time synchronisation at install out of the box.

# UPGRADE file entries
* /etc/ntp.conf need to point to your own internal NTP servers.

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -762,7 +762,10 @@ sed -i 's/\%.*$//g' /etc/resolv.conf
 echo "net.ipv4.ip_forward = 1" >> /etc/sysctl.conf
 sysctl -p /etc/sysctl.conf
 
-#enabling and starting ntp, required for Active Directory communications. Will talk to ntp.pool.org by default.
+#Fixing ntp to not deal with virtual interfaces
+sed -i '/OPTIONS/d' /etc/sysconfig/ntpd
+echo 'OPTIONS="-g -L"' >> /etc/sysconfig/ntpd
+#enabling and starting ntp, required for Active Directory communications. Will talk to centos.pool.ntp.org by default.
 /bin/systemctl enable ntp
 /bin/systemctl start ntp
 
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -90,6 +90,7 @@ Requires: libpcap, libxml2, zlib, zlib-devel, glibc-common,
 Requires: httpd, mod_ssl
 Requires: mod_perl, mod_proxy_html
 requires: libapreq2
+Requires: ntp
 Requires: dhcp
 Requires: redis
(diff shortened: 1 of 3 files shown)


dwlfrth:
Need to make sure it does not interfere with network interfaces.
I remember one or two stories about @lzammit customers where seeing network interfaces removed/disabled.

julsemaan:
@dwlfrth is right

It was messing up the domain interfaces

dwlfrth:
Not sure the ntp server is required on the PacketFence server itself or should we just ship a cronjob synching once a day

julsemaan:
Cronjob would be the best IMO but only thing is some admins won't like the fact we're putting an ntpdate pointing to a public server without them knowing

We'd have to make this configurable in someplace in PacketFence if we're going to ship it by default

tlaurion:
@dwlfrth "[...]the clock is off by about 43 seconds per day. Only poor old mechanical wristwatches are worse." --[source](http://www.ntp.org/ntpfaq/NTP-s-sw-clocks-quality.htm) 

It is general good practice to have all systems in sync to ease debugging with time being all the same between systems at all times...when possible.

@julsemaan @dwlfrth  @lzammit : "When the interface is down or removed, ntpd updates the listen interface according to the current interface status." 

The solution is known and is to tell ntpd to not deal with virtual interfaces through "-L" switch. Will fix.

jrouzierinverse:
rebase me

tlaurion:
@julsemaan

> Cronjob would be the best IMO but only thing is some admins won't like the fact we're putting an ntpdate pointing to a public server without them knowing. 
> We'd have to make this configurable in someplace in PacketFence if we're going to ship it by default

A Small note in the Admin guide to change the defaults would do?

julsemaan:
@tlaurion, I guess so, anyway, we're continuously moving towards PacketFence taking more and more place in the OS to the point where we behave almost like an appliance so if we think the «appliance way», there should be no issues putting it in a cron and adding a note in the admin guide

tlaurion:
After discussion with @julsemaan, the default configuration for ntp pointing to OS defaults is not a problem if there is a note in the Admin guide to change it to client's servers.

That ntp communicates with defaults servers before the client change the configuration to its servers it is not a problem.

TODO: 

- verify that the program launch options are not modified per ntp package upgrades (that those files are considered configuration files)
 DEBIAN: "sudo dpkg-reconfigure ntp" and "sudo apt-get --reinstall install ntp" didn't modify /etc/default/ntpd
  In fact, those files are protected by [debhelper](http://unix.stackexchange.com/questions/112178/debian-package-created-by-checkinstall-overwrites-a-config-file-in-etc-default)
- modify Admin guide.

dwlfrth:
Should we really get into manage system behavior / configuration that is not related to PacketFence ?
Maybe use a different approach like have an optional maintenance task that could be enabled to sync time.

PR #2146, opened 2017-03-29
So that PF upgrades do not create duplicate ip_forward entries.

### LINKED ISSUE #2145: ip_forward is added even if it's already present
When adding ip_forward to /etc/sysctl.conf, it is added even if it exist. So on each update or reinstall it will be added.

tlaurion:
# Description
Title says it all.

If ip_forward is also existing in /etc/sysctl.conf (PF 6.x installed on CentOS7), it will still duplicate. Each new upgrade will overwrite its /etc/sysctl.d/99-ip_forward.conf file.

# Impacts
none 

# Delete branch after merge
YES

fixes #2145 

# News
ip forwarding is now activated by default per packetfence package installation.

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -758,8 +758,9 @@ sed -i 's/^SELINUX=.*/SELINUX=disabled/g' /etc/selinux/config
 sed -i 's/\%.*$//g' /etc/resolv.conf
 
 # Enabling ip forwarding
-echo "net.ipv4.ip_forward = 1" >> /etc/sysctl.conf
-sysctl -p /etc/sysctl.conf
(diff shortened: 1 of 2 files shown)


louismunro:
I fail to see how this will not create duplicates.
You merely changed the file where the duplicates are going to end.

Can you overwrite the file instead of appending?

fdurand:
Approved

tlaurion:
@louismunro  
- The first echo redirects (>) to create/overwrite the file.
- The second echo appends (>>) to add content to file.

True that if there is another ip_forward entry in sysctl.conf (created manually from PF6), that one will be duplicated under another file autmagically by packetfence package.

louismunro:
Good points. I failed to notice the first echo.
LGTM then.

Merging.

PR #2173, opened 2017-03-31
More fine grained stats for the portal

### LINKED ISSUE #1962: Portal / HTTP metrics
We should have nice metrics for the portal / HTTP services in general.

jrouzierinverse:
# Description
More fine grained stats for the portal

# Impacts
The Portal

# Issue
Fixes #1962

# Delete branch after merge
NO

# NEWS file entries

## Enhancements
* More fine grain stats for the captive portal

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
@@ -55,6 +55,7 @@ captiveportal::PacketFence::Controller::Root - Root Controller for captiveportal
 
 sub auto : Private {
     my ( $self, $c ) = @_;
+    $c->stash->{statsd_timer} = pf::StatsD::Timer->new({ 'stat' => 'captiveportal/' . $c->request->path });
     $c->forward('setupLanguage');
     $c->forward('setupDynamicRouting');
     $c->forward('checkReadonly');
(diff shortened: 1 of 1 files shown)


html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
```diff
@@ -316,6 +316,17 @@ sub getRequestLanguages : Private{
     return \@l;
 }
 
+=head2 setupStatsd
+
+setup statsd
+
+=cut
+
+sub setupStatsd : Private {
```
julsemaan: Is it just me or that just do nothing at all ?

Maybe I'm missing some perl magic

html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
```diff
@@ -55,6 +55,7 @@ captiveportal::PacketFence::Controller::Root - Root Controller for captiveportal
 
 sub auto : Private {
     my ( $self, $c ) = @_;
+    $c->stash->{statsd_timer} = pf::StatsD::Timer->new({ 'stat' => 'captiveportal/' . $c->request->path });
```
julsemaan: that would be better if put on level 6 

Otherwise that will create a DB for every URL we have by default and we're not currently displaying those

jrouzierinverse:
done

James Rouzier

PR #2207, opened 2017-04-05
Maintain Captive Portal Interface Seperation

tallwireless:
# Description
One of the problems with the captive portal is that on a html redirect to the captive portal the hostname of the management interface is used. This is fine if the user subnet has access to the management interface, but poses a problem when it does not. An use case for this is our guest network resides in a L3VPN, and doesn't have access to the management interface, but shares the same DNS server with the rest of campus preventing the use of different DNS entries.

This enhancement adds a hostname variable to the interface sections of the pf.conf file. When a request comes in to httpd.portal the pf::web::dispatcher does a lookup of the incoming server IP address, and uses the hostname variable for the $captive_portal_domain if it exists. This in turn returns the correct hostname for isolated networks.

# Impacts
This code makes changes to the pf::web::dispatcher::html_redirect subroutine adding a lookup to code to redirect the incoming request to the appropriate interface.

# Code / PR Dependencies
See diff.

# NEW Package(s) required
None.

# Issue
None.

# Delete branch after merge
YES

# NEWS file entries
* Ability to redirect captive portals to incoming interface

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -11,6 +11,17 @@ This is a list of noteworthy changes across releases.
 For more details and developer visible changes see the ChangeLog file.
 For a list of compatibility related changes see the UPGRADE.asciidoc file.
 
+Version 6.5.1 released on XXXX-XX-XX
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Bug Fixes
++++++++++
+ * Fix incorrect node cleanup job handling.
+ * Fix multiple firewall SSO not working when cached updates were enabled.
+ * Removed usage of pf_memoize which could create a race condition when adding a node.
+ * Fix incorrect locationlog informations because of a null role.
+ * Fixed syntax error in generated Suricata rules
+
 Version 6.5.0 released on 2016-01-30
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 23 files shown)


dwlfrth:
That PR seems bogus

louismunro:
Perhaps not bogus, but certainly misguided as far as merging and rebasing.
As it is, it's not acceptable.

dwlfrth:
What I meant ;)

julsemaan:
And I hate to break your dream, but the dispatcher was rewritten in go so this change is not relevant for devel

Also, the hostname would make more sense in the networks configuration more than the interface configuration

Nonetheless, this need to be ported to the Golang code so it can reach devel (7+)

julsemaan:
@louismunro, @dwlfrth 

This seems to have been based off maintenance/6.5 instead of devel which explains the weird merge attempt

tallwireless:
I'm currently in the process of deploying 6.5.1 in production, and ran into this problem. I wasn't sure if there was going to be another maintenance release of 6.5 before 7 was deployed.

I'll take a pull of the devel branch, and look into merging it into the dispatcher written in go.

As for the placement of the hostname, tying it with the interface seemed appropriate. Unless there have been changes to the networks.conf file and what is contained there, networks.conf was for the list of routed subnets, not the interfaces.

dwlfrth:
What are we doing with this ?
@tallwireless, do you want to submit a reworked PR ?

dwlfrth:
@tallwireless, can you let us know what you think about the comments. If we don't get any reply, we gonna go ahead and close

julsemaan:
no love given to the PR based on our comments, will close this, @tallwireless feel free to reopen if you want to adjust it

PR #2268, opened 2017-04-24
Fix/2222 locationlog externalportal

### LINKED ISSUE #2222: When performing a CoA, we should open a locationlog entry with the new role
While doing WebAuth and a CoA, after the registration, if there is no VLAN number for the new role, the reevaluate_access is not trigger. This issue is only for CoA and not for radiusDisconnect.

dwlfrth:
# Description
Switch module is responsible to instruct whether or not to update locationlog on externalportal requests

# Impacts
External portal flow

# Issue
fixes #2222

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* When performing a CoA, we should open a locationlog entry with the new role (#2222)

### DIFF
--- a/lib/pf/Switch/Ruckus.pm
+++ b/lib/pf/Switch/Ruckus.pm
@@ -182,11 +182,12 @@ sub parseExternalPortalRequest {
     my %params = ();
 
     %params = (
-        switch_id       => $req->param('sip'),
-        client_mac      => clean_mac($req->param('client_mac')),
-        client_ip       => defined($req->param('uip')) ? $req->param('uip') : undef,
-        ssid            => $req->param('ssid'),
-        redirect_url    => $req->param('url'),
+        switch_id               => $req->param('sip'),
+        client_mac              => clean_mac($req->param('client_mac')),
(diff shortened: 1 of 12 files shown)


dwlfrth:
Review me. Please pay attention to $TRUE or $FALSE depending on the module

lib/pf/Switch/Ruckus.pm
```diff
@@ -182,11 +182,12 @@ sub parseExternalPortalRequest {
...
-        client_mac      => clean_mac($req->param('client_mac')),
-        client_ip       => defined($req->param('uip')) ? $req->param('uip') : undef,
-        ssid            => $req->param('ssid'),
-        redirect_url    => $req->param('url'),
+        switch_id               => $req->param('sip'),
+        client_mac              => clean_mac($req->param('client_mac')),
+        client_ip               => defined($req->param('uip')) ? $req->param('uip') : undef,
+        ssid                    => $req->param('ssid'),
+        redirect_url            => $req->param('url'),
+        synchronize_locationlog => $TRUE,
```
julsemaan: i'd remove that

PR #2287, opened 2017-05-01
Feature/add brands icons

### LINKED ISSUE #2228: Adding brand icons
Would be nice to have matching brand icons for specific sources: for instance oauth2/billing icons

whitx:
# Description
Adding brands icons next to sources name and in creation of a source. Available only if the icons is existing in http://fontawesome.io/icons/

# Impacts
Authentication sources list, and source edit.

# Issue
fixes #2228 

# Delete branch after merge
YES 

# NEWS file entries
## Enhancements
* Adding brands icons to authentication source (i.e Twitter, PayPal etc ..) in the administration interface

### DIFF
--- a/html/pfappserver/root/authentication/source/read.tt
+++ b/html/pfappserver/root/authentication/source/read.tt
@@ -12,7 +12,7 @@
 
 <div class="card">
 <div class="card-title">
-<h3><i class="icon-2x icon-[% source.type | html | lower %]"></i>[% heading | html %] <small>[% source.type | html %]</small></i></h3>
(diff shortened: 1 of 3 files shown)


whitx:
auth source
twitsource

jrouzierinverse:
The icons should be on the left of the name

jrouzierinverse:
We should also consider having some type of icons for the other sources

cgx:
@jrouzierinverse I would not recommend trying to add an icon for each authentication source. Let's keep the icons for branded OAuth sources only.

cgx (approved):
👍

PR #2316, opened 2017-05-10
Feature/2227 lldpdetection

### LINKED ISSUE #2227: VoIP detection: LLDP lookup enhancement
Currently, 'httpd.aaa' process will query the switch over SNMP to get the lldpLocPortDesc table (all the switch ports) to be able to match if there is a phone connected to the switch port. That query occurs everytime a device plugs in to detect whether or not PacketFence should take the VoIP workflow.

On a 24 - 48 switch, it can be acceptable but it is another story with a stack of 10 48 ports switches.

Getting the lldpLocPortDesc table for 480 ports at each request only to check whether or not it is matching an ifIndex does not make sense, thinking that this table only change if a new switch is added to the stack or if the switch get replaced.

Solution:
 * Have a 'pfmon' job that will cache the lldpLocPortDesc table
 * Modify the VoIP detection flow to use the cached data
 * 'pfmon' will cache only for switches that have VoIP LLDP detection / VoIP CDP detection enabled
 * Add a new button in the switch config to allow invalidation of switch cache

dwlfrth:
# Description
- Use 'get-next-request' to query big SNMP tables
- Cache the result
- Provide a pfmon job to populate the cache to avoid the load in httpd.aaa

# Impacts
VoIPLLDPDetect

# Issue
fixes #2227 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* VoIP detection: LLDP lookup enhancement (#2227)


conf/pfmon.conf.defaults
```diff
@@ -541,3 +541,30 @@ batch=1000
...
+type=switch_cache_lldpLocalPort_description
+#
+# description
+#
+# The description of task to perform
+description=Cache switches LLDP local port descriptions
+#
+# status
+#
+# If the task auth_log_cleanup is enabled
```
julsemaan: not auth_log_cleanup

conf/pfmon.conf.defaults
```diff
@@ -541,3 +541,30 @@ batch=1000
...
+status=enabled
+#
+# interval
+#
+# At which interval to run the auth log cleanup pfmon task
```
julsemaan: No, you lie

lib/pf/Switch.pm
```diff
@@ -1817,23 +1817,6 @@ sub reverseBitmask {
     return $flippedBitMask;
 }
 
-=item getSysUptime - returns the sysUpTime
-
-=cut
-
-sub getSysUptime {
```
julsemaan, line 1824: Why removing this ?

julsemaan:
Also, we talked about having a visible button while editing the switch in the admin to flush the cache.

lib/pf/Switch.pm
dwlfrth: It is not used anywhere. Removed because I originally reworked it but since we are no longer automatically invalidating the cache i removed both (the original and the rework)

dwlfrth:
Review me

PR #2319, opened 2017-05-13
Set GOPATH while setting up go environment

tribut:
# Description
Fix a bug when setting up the go environment for the first time.

# Impacts
The script run by `make go-env` currently creates the symlink under `/` instead of `~/gospace`.

# NEWS file entries
## Bug Fixes
* Set up variable GOPATH correctly while setting up developer environment for go.

### DIFF
--- a/addons/dev-helpers/setup-go-env.sh
+++ b/addons/dev-helpers/setup-go-env.sh
@@ -11,9 +11,12 @@ else
   tar -C /usr/local -xzf /tmp/$GOVERSION.linux-amd64.tar.gz
   rm /tmp/$GOVERSION.linux-amd64.tar.gz
 
-  echo 'export PATH=$PATH:/usr/local/go/bin' >> ~/.bashrc
-  echo 'export GOPATH=~/gospace' >> ~/.bashrc
-  echo 'export PATH=~/gospace/bin:$PATH' >> ~/.bashrc
+  SETUP='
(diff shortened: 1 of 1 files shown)


addons/dev-helpers/setup-go-env.sh
```diff
@@ -14,6 +14,7 @@ else
   echo 'export PATH=$PATH:/usr/local/go/bin' >> ~/.bashrc
   echo 'export GOPATH=~/gospace' >> ~/.bashrc
   echo 'export PATH=~/gospace/bin:$PATH' >> ~/.bashrc
+  GOPATH=~/gospace
```
julsemaan: I suggest we just 'source ~/.bashrc' instead so it setups everything that was added above.

tribut: Not sure if this could have any side effects. If you're looking for something generic, how about:
~~~
SETUP='
export PATH=$PATH:/usr/local/go/bin
export GOPATH=~/gospace
export PATH=~/gospace/bin:$PATH
'
echo "$SETUP" >> ~/.bashrc
eval "$SETUP"
~~~

julsemaan: What you suggested will also do the trick.
Adjust your PR and we're good to merge this

tribut: Done.

PR #2322, opened 2017-05-15
force the mobileprovider field when there is only one carrier

### LINKED ISSUE #1958: Portal: Option to force (and hide) SMS mobile provider
We should find a way to 'hardcode' the provider for the user based on an admin requirement.

Ex: A home made SMTP interface which should handle all requests

User shouldn't have the option to choose a mobileprovider and the server should select it automatically.

Now, only little question is, should this be in the source (by leaving only 1 provider) or a configuration of the portal module where you could force a provider

julsemaan:
# Description
Force the mobileprovider field when there is only one carrier

# Impacts
SMS authentication

# Issue
fixes #1958 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Don't display the mobileprovider field when doing SMS authentication with only one carrier enabled


dwlfrth:
LGTM

PR #2327, opened 2017-05-16
Added dependencies on sd_notify version of FreeRADIUS.

louismunro:
# Description
Adds package dependencies on sd_notify compliant versions of FreeRADIUS.

# Impacts
Packaging: packetfence.spec, and debian/control

# NEW Package(s) required
Depends on the correct versions of the packages being available, which they should be by now, i.e. freeradius 3.0.14.2 and upwards.

# Delete branch after merge
YES 

## Enhancements
* Adds support for sd_notify in FreeRADIUS


fdurand:
Ok for me but you will need to update the debianize.patch

PR #2328, opened 2017-05-17
Isolation passthrough

fdurand:
# Description
Allow passthrough for device in isolation state. 

# Impacts
No impacts

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Allow passthrough for devices in violation state


dwlfrth:
Please rebase

conf/documentation.conf
```diff
@@ -534,6 +534,24 @@ You can specify the same domain with a different port multiple times and they wi
 The configuration parameter "passthrough" must be enabled for passthroughs to be effective.
 EOT
 
+[fencing.isolation_passthrough]
```
julsemaan: Can you adjust the passthough configuration to make it explicit that they are only honored in registration

lib/pf/util/dns.pm
```diff
@@ -56,6 +57,42 @@ sub matches_passthrough {
...
+
+Whether or not a domain matches a configured DNS isolation passthrough
+
+Returns 2 values
+- Whether or not it matched
+- List of ports that should be opened for this domain
+
+=cut
+
+sub matches_isolation_passthrough {
```
julsemaan: We need to create a function that will match any passthroughs to a domain so that the code isn't a copy/paste
Let me know if you want me to do it or if you'll take care of it

lib/pfconfig/namespaces/config/Pf.pm
```diff
@@ -113,6 +113,11 @@ sub build_child {
         $Config{$group}{$item} = [ split( /\s*,\s*/, $Config{$group}{$item}  // '' ) ];
     }
 
+    foreach my $val ("fencing.isolation_passthroughs") {
```
julsemaan: just add "fencing.isolation_passthrougs" to the foreach loop above, they do the same thing

lib/pfconfig/namespaces/resource/isolation_passthroughs.pm
```diff
@@ -0,0 +1,144 @@
+package pfconfig::namespaces::resource::isolation_passthroughs;
```
julsemaan, line 1: Same idea as above, we need to remove duplicated logic.
Let me know if you want me to do it

sbin/pfdns
```diff
@@ -442,6 +458,39 @@ sub regzone_non_filtered {
     return \%results;
 }
 
+=head2 isolzone_non_filtered
+
+Non filtered isolation zone answered
+
+=cut
+
+sub isolzone_non_filtered {
```
julsemaan: Again, we'll need to reduce duplication of the code

fdurand:
Bump

PR #2329, opened 2017-05-17
fix/double_nodes

### LINKED ISSUE #1848: Admin: Double nodes still showing up when there is a duplicate locationlog
There is two issues with this, first, the duplicate inline L3 locationlog (see #1847) and second we are still showing duplicate records when there are duplicate locationlogs (this issue is for the latest).

I believe we should address this duplicate locationlog issue directly in the admin so that we stop having this recurring problem to show double nodes when there are duplicate locationlogs.

image

image

P.S. : This is on 6.3

jrouzierinverse:
Description
===========
Fixes the double nodes in the admin view

Impacts
=======
Admin node searching

Issue
=====
fixes #1848

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------
* Fixes duplicate nodes displaying when there are multiple locationlog entries (#1848)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Search/Node.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Search/Node.pm
@@ -87,7 +87,10 @@ sub add_searches {
         $builder->where(')');
         $builder->where('AND');
     }
+    $builder->where('(');
     $builder->where( { table => 'r2', name => 'radacctid' }, 'IS NULL' );
+    $builder->where( { table => 'locationlog2', name => 'id' }, 'IS NULL' );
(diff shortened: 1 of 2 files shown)


jrouzierinverse:
I would like someone to review/ test this on a decently sized set of nodes with for speed.

dwlfrth:
SQL query error when doing a search

dwlfrth:
Basically tested and does what it is supposed to.

dwlfrth:
Tested with a DB with the following (relevant) specs:
node: 40771
locationlog: 48735

PR #2336, opened 2017-05-18
Introduce CSP headers

tlaurion:
# Description
CSP headers enforces a policy that tells the browser what to trust from within a specific web page. Currently, the common enforced policy for the Admin and Portal is really basic and distrusts inline stuff (style, javascript, etc) which needs to be moved to files. This is a pretty good measure against XSS and a necessary step toward XSRF.

The portal adds an img-src statement that points to the configured network-detection server (Inverse IP address), which hosts a 0px image that is downloaded to validate network access from within the portal before redirecting to the requested URI.

# Impacts
- Check that there is no error raised in browser dev tools.
- To report but not block, one could change "Content-Security-Policy" to "Content-Security-Policy-Report-Only"
- documentation.conf and pf.default.conf were modified. @whitx said that there was a problem in the generation of the langage file?

# Issues (2017-07-25)
There's only one JavaScript library that we used in the Web admin that is not CSP-compliant: [Ace](https://ace.c9.io/). We should look for an alternative.

Library | CSP-ready?
--------|------------
[Ace](https://ace.c9.io/) | [no](https://github.com/ajaxorg/ace/issues/3260)
[CodeMirror](http://codemirror.net/) | [no](https://discuss.codemirror.net/t/inline-styles-and-content-security-policy/1311)
[Monaco Editor](https://microsoft.github.io/monaco-editor/) | [no](https://github.com/Microsoft/monaco-editor/issues/271)
[CodeFlask.js](https://kazzkiq.github.io/CodeFlask.js/) | yes but rendering is buggy
[LDT](http://kueblc.github.io/LDT/) | maybe be unmaintained

Another idea: put the code editor inside a frame and disable CSP for this page only.

**Implemented solution** (2017-08-25): be more permissive on the configuration page to handle ACE requirements.

# Delete branch after merge
YES 

# NEWS file entries
- Content Security Policy (CSP) enforced through the browser for Admin and Portal interfaces. Can be disabled through Admin->System Configuration: Main Configuration: Advanced: CSP security headers.

## Enhancements
- Content Security Policy (CSP) enforced through the browser for Admin and Portal interfaces. Can be disabled through Advanced System Configuration.


dwlfrth:
Is this ready to be merged ?
Can this be rebased ? (looking at you @cgx )

cgx:
No.

dwlfrth:
@cgx can we aim for 7.3 ?

dwlfrth:
@cgx : What is the status for this one ?
@extrafu : Do we still aim at 7.3 ?

dwlfrth:
As per conversation with @cgx, the only remaining issue with the CSP headers would be the ACE editor (used to edit portal profile pages in the GUI). The possible solution would be to use iframe for this very specific use case.
If possible to achieve before 7.3 feature freeze, it will be done, otherwise we will merge everything CSS / JS related and adjust.

cgx:
Now ready for review.

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2846,6 +2846,12 @@ Action -
 
     modify_node:  mac, $mac, notes, $notes
 
+Content-Security-Policy (CSP)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The Content-Security-Policy HTTP response header tells modern browsers what can be accessed from a generated web page. The default policy is pushed for both the captive portal and the admin interfaces and enforces that everything the browser executes comes from within PacketFence, with the exeption of the configured network detection host that is by default Inverse IP address.
```
julsemaan: *exception

julsemaan: *is by default the Inverse IP address

html/captive-portal/lib/captiveportal/PacketFence/Controller/Root.pm
```diff
@@ -324,7 +324,11 @@ Attempt to render a view, if needed.
 
 sub end : ActionClass('RenderView') {
     my ( $self, $c ) = @_;
-
+    
+    if($Config{'advanced'}{'csp_security_headers'} eq 'enabled'){
```
julsemaan: should use pf::util::is_enabled

html/pfappserver/lib/pfappserver/PacketFence/Controller/Root.pm
```diff
@@ -84,6 +84,11 @@ Attempt to render a view, if needed.
 
 sub end : ActionClass('RenderView') {
     my ( $self, $c ) = @_;
+     
+    if($Config{'advanced'}{'csp_security_headers'} eq 'enabled'){
```
julsemaan: Should be using pf::util::is_enabled

julsemaan:
Also this needs a rebase

cgx:
@julsemaan, please apply your proposed changes. I can take care of the rebase after. Thanks.

jrouzierinverse (commented):
There is a missing call to csp_server_headers in the end sub in html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Pf.pm

cgx:
@jrouzierinverse CSP is useless when returning JSON.

jrouzierinverse:
Ok

PR #2337, opened 2017-05-19
Feature/new status page

whitx:
# Description
Status page and device-registration are merged together, and improvements to device-registration

# Require
https://github.com/fingerbank/perl-client/pull/29 to be merged

# Impacts
Status page
Device Registration page

# Delete branch after merge
YES 

# NEWS file entries
## New Features
* Re-factor of the status page, to offer more options and be configurable.


conf/documentation.conf
```diff
@@ -573,6 +573,12 @@ description=<<EOT
 The role to assign to devices registered from the specific portal. If none is specified, the role of the registrant is used.
 EOT
 
+[device_registration.oses]
+type=fingerbank_select
+description=<<EOT
+Lists of OSES where the MAC vendor will be allowed to be registered via the device registration page.
```
extrafu: Shouldn't we just say "List of allowed devices to be registered via the device registration page." ? If we use Fingerbank's device, we don't need to tie this with OS/MAC vendors.

conf/pf.conf.defaults
```diff
@@ -158,6 +158,11 @@ status=disabled
 #
 # The role to assign to devices registered from the specific portal. If none is specified, the role of the registrant is used.
 role=
+#
+# device_registration.oses
+#
+# Lists of OSES where the MAC vendor will be allowed to be registered via the device registration page.
+oses=
```
extrafu: I would prefer "operating_systems" or "operatingsystems" or I as said before, maybe just "allowed_devices"

conf/violations.conf.defaults
```diff
@@ -251,6 +251,16 @@ template=bandwidth_expiration
...
 enabled=N
 
+[1300005]
+priority=1
+actions=email_user,email_admin,log,role
+desc=Lost or Stolen
+enabled=Y
+template=banned_devices
+auto_enable=N
+user_mail_message=Your devices %mac as been declared as lost or stolen. Please contact your system administrator to be able to use this device again on the network.
```
julsemaan: *Your device

html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm
```diff
@@ -178,7 +178,7 @@ sub registerNode : Private {
             reevaluate_access($mac, 'manage_register');
         }
     } else {
-        $self->showError($c,"Please verify the provided MAC address.");
+        $self->showError($c,"The provided MAC address is not allowed to be register.");
```
julsemaan: *to be registered using this self-service page.

html/captive-portal/lib/captiveportal/PacketFence/Controller/LostStolen.pm
```diff
@@ -0,0 +1,85 @@
...
+=head2 index
+
+=cut
+
+sub index : Path : Args(1) {
+    my ( $self, $c, $mac ) = @_;
+    my $node = node_view($mac);
+    my $owner = lc($node->{pid});
+    my $username = lc($c->user_session->{username});
+    my $vid = "1300005";
```
julsemaan: Put that in a constant in the violations constants

html/captive-portal/lib/captiveportal/PacketFence/Controller/LostStolen.pm
```diff
@@ -0,0 +1,85 @@
...
+
+=head2 index
+
+=cut
+
+sub index : Path : Args(1) {
+    my ( $self, $c, $mac ) = @_;
+    my $node = node_view($mac);
+    my $owner = lc($node->{pid});
+    my $username = lc($c->user_session->{username});
```
julsemaan: We'll need to see how we handle the case where the device was registered using a realm

Ex:
bobby@inverse.ca is the owner of the device
But when bobby uses the self-service portal, he logs in as bobby

I think the easiest way would be to strip $owner and $username and ignore the realm

html/captive-portal/lib/captiveportal/PacketFence/Controller/LostStolen.pm
```diff
@@ -0,0 +1,85 @@
...
+    my $vid = "1300005";
+    if ( $username eq $owner ) {
+        my $trigger = violation_add($mac, $vid);
+
+        if ($trigger) {
+            $c->stash(
+                mac => $mac,
```
julsemaan: stash mac + template outside of your if and then just stash the status in the if{}else{}

Also stash a $TRUE/$FALSE versus a string and handle it accordingly in your templates/rendering

julsemaan: Just re-read the code and it seems you indeed need a string status since its not only boolean, so forget about my second comment

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -45,20 +46,38 @@ sub index : Path : Args(0) {
...
     $c->stash(
-        title => "State - Network Access",
+        title => "Status - Network Access",
         template => 'status.html',
         billing  => $c->profile->hasBilling(),
         access_registration_when_registered => $c->profile->canAccessRegistrationWhenRegistered(),
     );
 }
 
+sub has_open_violation : Path : Args(1) {
```
julsemaan: This checks for lost/stolen violation only so change it to is_lost_stolen so its clearer

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -45,20 +46,38 @@ sub index : Path : Args(0) {
...
+    my ( $mac ) = @_;
+   
+    my @violations = violation_view_open($mac);
+    if ( grep {$_->{'vid'} eq "1300005"} @violations ) {
```
julsemaan: constant for the violation

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -45,20 +46,38 @@ sub index : Path : Args(0) {
...
+
 sub userIsAuthenticated : Private {
     my ( $self, $c ) = @_;
     my $pid   = $c->user_session->{"username"};
     my @nodes = person_nodes($pid);
     foreach my $node (@nodes) {
         setExpiration($node);
+        my $mac = $node->{'mac'};
+        if (has_open_violation($mac)) {
+            $node->{lostOrStolen} = 1;
```
julsemaan: 1 -> $TRUE

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -114,6 +132,42 @@ sub login : Local {
     $c->forward('index');
 }
 
+sub person : Local {
+    my ( $selfc, $c ) = @_;
```
julsemaan: $selfc ?

html/captive-portal/templates/device-registration/registration.html
```diff
@@ -43,14 +43,18 @@
...
+        <div class="flag__img">[% svgIcon(id='ic_devices_black_24px',size='small') %]</div>
+        <p class="flag__body">[% i18n("Manage my devices") %]</p>
       </div>
     </a>
-  </div>
-</div>
-<div class="layout layout--center">
-  <div class="layout__item u-mt u-2/3 u-1/1-palm">
+[% IF hasLocalAccount %]
+    <a href="/status/person" class="btn btn--full btn--light">
```
julsemaan: Given this is only used to change password, this should be called /status/reset_password or something like this

html/captive-portal/templates/lost_stolen.html
```diff
@@ -0,0 +1,40 @@
+[% IF status == "success" %]
+<div class="media media--notice u-p u-mb">
+  <div class="media__img">[% flashIcon(level='notice') %]</div>
+  <p class="media__body">[% i18n_format("Your device %s has been declared as lost or stolen.", mac) | none %]</p>
+</div>
+[% ELSIF status == "error" %]
+<div class="media media--error u-p u-mb">
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n_format("An error occured while trying to delcare the device %s as lost or stolen.", mac) | none %]</p>
```
julsemaan: *declare

html/captive-portal/templates/lost_stolen.html
```diff
@@ -0,0 +1,40 @@
...
+</div>
+[% ELSIF status == "notowned" %]
+<div class="media media--error u-p u-mb">
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n_format("You do not own this device %s, therefor you can not decalre it as lost or stolen.", mac) | none %]</p>
```
julsemaan: *declare

html/captive-portal/templates/lost_stolen.html
```diff
@@ -0,0 +1,40 @@
...
+<div class="layout layout--center">
+  <div class="layout__item u-mt u-2/3 u-1/1-palm">
+    <a href="/status" class="btn btn--full btn--light">
+      <div class="flag">
+        <div class="flag__img">[% svgIcon(id='ic_devices_black_24px',size='small') %]</div>
+        <p class="flag__body">[% i18n("Manage my devices") %]</p>
+      </div>
+    </a>
+[% IF hasLocalAccount %]
+    <a href="/status/person" class="btn btn--full btn--light">
```
julsemaan: meaningful URI

html/captive-portal/templates/lost_stolen.html
```diff
@@ -0,0 +1,40 @@
...
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n_format("An error occured while trying to delcare the device %s as lost or stolen.", mac) | none %]</p>
+</div>
+[% ELSIF status == "notowned" %]
+<div class="media media--error u-p u-mb">
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n_format("You do not own this device %s, therefor you can not decalre it as lost or stolen.", mac) | none %]</p>
+</div>
+[% END %]
+<div class="layout layout--center">
```
julsemaan: Please put the menu in a partial that you reuse across pages so that there is no duplication

lib/pf/person.pm
```diff
@@ -443,6 +449,19 @@ sub person_cleanup {
     }
 }
 
+=head2 person_has_local_account
+
+Query the DB to see if the pid is a local account
+
+=cut
+
+sub person_has_local_account {
```
julsemaan: Why not just use the functions in lib/pf/password.pm ?

lib/pf/web/device_registration.pm
```diff
@@ -18,65 +18,95 @@ use warnings;
...
-        open(FILE,$file_name);
-        @items =
-            grep {$_}
-            map {
-                #Getting rid of newlines, comments and trimming spaces
-                chomp; s/#.*$//; s/^\s+//; s/\s+$//;
-                $_;
-            } <FILE>;
-        close(FILE);
+sub device_id {
```
julsemaan: Should be called, device_id_from_mac_vendor

julsemaan: And might as well return the device object for extended usage so:
device_from_mac_vendor

lib/pf/web/device_registration.pm
```diff
@@ -18,65 +18,95 @@ use warnings;
...
+    my ($mac_vendor_id) = @_; 
+    my $logger = get_logger();
+
+    my ($status, $result) = fingerbank::Model::Combination->find([{ mac_vendor_id => $mac_vendor_id }, {columns => ['device_id']}]);
+
+    if(is_success($status)){
+        return $result->device_id;
```
julsemaan: So here we would return the device instead of just he ID

lib/pf/web/device_registration.pm
```diff
@@ -18,65 +18,95 @@ use warnings;
...
-                #Getting rid of newlines, comments and trimming spaces
-                chomp; s/#.*$//; s/^\s+//; s/\s+$//;
-                $_;
-            } <FILE>;
-        close(FILE);
+sub device_id {
+    my ($mac_vendor_id) = @_; 
+    my $logger = get_logger();
+
+    my ($status, $result) = fingerbank::Model::Combination->find([{ mac_vendor_id => $mac_vendor_id }, {columns => ['device_id']}]);
```
julsemaan: Here it seems you're just grabing the first combination that has the mac vendor you want.

What you should do instead is find the most popular device ID for the MAC vendor you provide, not just any of them

So replicating the logic from this query:
select device_id from combination where mac_vendor_id=$mac_vendor_id group by device_id order by count(device_id) desc limit 1;

lib/pf/web/device_registration.pm
```diff
@@ -18,65 +18,95 @@ use warnings;
...
-Validate if mac address is an allowed device mac
+Verify 
 
-=cut
+=cut 
 
 sub is_allowed {
     my ($mac) = @_;
-    return 1 unless @DEVICE_OUI;
-    $mac =~ s/O/0/i;
```
julsemaan: That wasn't a bad «dumb catcher», I'd leave it there for people who don't make the difference between a 0 and an O (i.e. Zayme's cousins)

lib/pf/web/device_registration.pm
```diff
@@ -18,65 +18,95 @@ use warnings;
...
+    my $device_id = device_id($mac_vendor_id);
+    my ($status, $result) = fingerbank::Model::Device->find([{ id => $device_id}, {columns => ['name']}]);
+
+    # We are loading the fingerbank endpoint model to verify if the device id is matching as a parent or child
+    if (is_success($status)){
+        my $device_name = $result->name;
+        my $endpoint = fingerbank::Model::Endpoint->new(name => $device_name, version => undef, score => undef);
+
+        for my $id (@oses) {
+            $logger->debug("The devices type ".$device_name." is authorized to be registered via the device-registration module");
```
julsemaan: That logging statement is only true if the condition on the line below is true so that statement is false...

Wrap your return + this logging in the if block

conf/pf.conf.defaults
julsemaan: +1 for allowed_devices

conf/violations.conf.defaults
```diff
@@ -251,6 +251,16 @@ template=bandwidth_expiration
 auto_enable=N
 enabled=N
 
+[1300005]
+priority=1
+actions=email_user,email_admin,log,role
```
julsemaan: You need the reevaluate action on this instead of setting the role

conf/violations.conf.defaults
```diff
@@ -251,6 +251,16 @@ template=bandwidth_expiration
...
+desc=Lost or Stolen
+enabled=Y
+template=banned_devices
+auto_enable=N
+user_mail_message=Your devices %mac as been declared as lost or stolen. Please contact your system administrator to be able to use this device again on the network.
+target_category=isolation
```
julsemaan: And you can remove this when you use the reevaluate_access action

lib/pf/web/device_registration.pm
```diff
@@ -18,65 +18,95 @@ use warnings;
...
 use pf::web::custom;    # called last to allow redefinitions
 
 use pf::authentication;
 use pf::Authentication::constants;
 use List::MoreUtils qw(any);
 
-Readonly our @DEVICE_OUI => _load_file_into_array($allowed_device_oui_file);
-Readonly our @DEVICE_TYPES => _load_file_into_array($allowed_device_types_file);
-
 =head1 SUBROUTINES
```
julsemaan: A :nail_care: comment:
We're trying to move away from the lib/pf/web/ directory and remove things that are in it.
I think your code for the device validation would be better placed in the DeviceRegistration controller

julsemaan (changes requested):

whitx:
@cgx  might have some CSS improvements to add

dwlfrth:
@whitx rebase

html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm
```diff
@@ -176,9 +174,106 @@ sub registerNode : Private {
...
+=cut 
+
+sub is_allowed {
+    my ($mac) = @_;
+    $mac =~ s/O/0/i;
+    my $logger = get_logger();
+    my @oses = @{$Config{'device_registration'}{'allowed_devices'}};
+
+    # If no oses are defined then it will not match any oses
+    return $FALSE if @oses == 0;
```
julsemaan: Shouldn't it be allowed if no OS is defined in the configuration ?

html/captive-portal/templates/status/reset_password.html
```diff
@@ -0,0 +1,39 @@
+[% IF status == "success" %]
+<div class="media media--notice u-p u-mb">
+  <div class="media__img">[% flashIcon(level='notice') %]</div>
+  <p class="media__body">[% i18n("Your password have been updated") %]</p>
```
julsemaan: has been updated

html/captive-portal/templates/status/reset_password.html
```diff
@@ -0,0 +1,39 @@
...
+</div>
+[% ELSIF status == "error_match" %]
+<div class="media media--error u-p u-mb">
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n("The passwords you type do not match") %]</p>
```
julsemaan: The two entered passwords did not match

html/captive-portal/templates/status/reset_password.html
```diff
@@ -0,0 +1,39 @@
...
+</div>
+[% ELSIF status == "error_fill" %]
+<div class="media media--error u-p u-mb">
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n("A password field has not been fill") %]</p>
```
julsemaan: One of the password fields hasn't been filled

html/captive-portal/templates/status/reset_password.html
```diff
@@ -0,0 +1,39 @@
...
+</div>
+[% END %]
+
+[% UNLESS ShowLogin %]
```
julsemaan: This variable doesn't seem to exist anymore

html/captive-portal/templates/status/reset_password.html
```diff
@@ -0,0 +1,39 @@
...
+  <p class="media__body">[% i18n("The two entered passwords did not match") %]</p>
+</div>
+[% ELSIF status == "error_fill" %]
+<div class="media media--error u-p u-mb">
+  <div class="media__img">[% flashIcon(level='error') %]</div>
+  <p class="media__body">[% i18n("One of the password fields hasn't been filled") %]</p>
+</div>
+[% END %]
+
+[% UNLESS showLogin %]
```
julsemaan: Instead of not showing the form in the template if the user isn't logged in, it should instead redirect to the login page if he isn't logged in and then here you unconditionally display the form

whitx: Make sense doing that right now

PR #2339, opened 2017-05-19
Use only internal authentication sources if none specified

### LINKED ISSUE #1784: device-registration: 'Null' source type should not be used
Device registration module uses `captiveportal::PacketFence::Controller::Authenticate::authenticateLogin` to authenticate a user.

`https://github.com/inverse-inc/packetfence/blob/packetfence-6.3.0/html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm#L102`

We should not allow the use of a configured 'Null' source for this kind of scenario.

Current ugly fix:

Right after
`https://github.com/inverse-inc/packetfence/blob/packetfence-6.3.0/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm#L106`

add

`
if ( $request->{'match'} eq "device-registration" ) {
    my @indexes_to_delete = grep { $sources[$_]->{'type'} eq 'Null' } 0..$#sources;
    @indexes_to_delete = reverse(@indexes_to_delete);
    foreach my $index ( @indexes_to_delete ) {
        splice(@sources, $index, 1);
    }
}
`

And change

`https://github.com/inverse-inc/packetfence/blob/packetfence-6.3.0/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm#L163`

by

`if ($c->request->{'match'} =~ "activate/email" || $c->request->{'match'} =~ "device-registration") {`

### LINKED ISSUE #2338: authentication: Too large scoping of authentication sources
When calling `pf::authentication::authenticate` without specifying a source array, all configured authentication source (except the "exclusive" ones) are used.
Since we match on a first hit based, some authentication sources do not have any specific" logic to return a "TRUE" for authenticate and we might wrongly authenticate.

ie. Email source which is provided by default

dwlfrth:
# Description
`pf::authentication::authenticate` uses only "internal" authentication sources if none are specified

# Impacts
- httpd.portal
- WISPr

# Issue
fixes #2338 
fixes #1784 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* authentication: Too large scoping of authentication sources (#2338)
* device-registration: 'Null' source type should not be used (#1784)

### DIFF
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -221,7 +221,7 @@ sub authenticate {
 
     # If no source(s) provided, all (except 'exclusive' ones) configured sources are used
     unless (@sources) {
-        @sources = grep { $_->class ne 'exclusive'  } @authentication_sources;
+        @sources = @{pf::authentication::getInternalAuthenticationSources()};
     }
     my $display_username = (defined $username) ? $username : "(anonymous)";
 
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -219,7 +219,7 @@ sub authenticate {
     my $username = $params->{'username'};
     my $password = $params->{'password'};
 
-    # If no source(s) provided, all (except 'exclusive' ones) configured sources are used
+    # If no source(s) provided, all 'internal' configured sources are used
     unless (@sources) {
         @sources = @{pf::authentication::getInternalAuthenticationSources()};
     }


dwlfrth:
Review me

lib/pf/authentication.pm
```diff
@@ -221,7 +221,7 @@ sub authenticate {
 
     # If no source(s) provided, all (except 'exclusive' ones) configured sources are used
```
julsemaan: Fix the comment here

julsemaan:
Also given the potential destructive impact of this, I'm not sure I would cherry pick this in maintenance just yet.
When 7.1 is released and this is confirmed as tested, we could cherry-pick it

dwlfrth:
Review me

PR #2347, opened 2017-05-25
use MAC based SID in pf::Portal::Session when possible

### LINKED ISSUE #2352: Web-authentication session leak
When creating a web-authentication session (using pf::Portal::Session), then if the endpoint doesn't reuse the session ID we gave him in the next requests, we will create a session for nothing. This is especially damageable when a device hammers the portal with multiple web-auth redirect requests per second which may end up taking a lot of place in the redis instance

We should make this session use a MAC based session ID so its consistently tied to the device even though it doesn't reuse its session ID

julsemaan:
# Description
Use a MAC based session for pf::Portal::Session when possible to prevent an endpoint creating thousands of useless sessions by not reusing the session ID

# Impacts
All web-authentication and wispr

# Delete branch after merge
YES

# Issue
fixes #2352 

# NEWS file entries
## Enhancements
* Limit the amount of portal sessions an endpoint can create when using web authentication

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
@@ -251,9 +251,18 @@ sub _build_profile {
 
 sub _build_dispatcherSession {
     my ($self) = @_;
-    my $session = new pf::Portal::Session()->session;
-    my %session_data;
     my $logger = get_logger();
+
+    # Restore with a dummy MAC since we don't care about what contains the session if it can't be restored from the session ID
+    my $dummy_mac = "ff:ff:ff:ff:ff:ff";
+    my $session = new pf::Portal::Session(client_mac => $dummy_mac)->session;
+
+    if($session->param('_client_mac') eq $dummy_mac) {
+        $logger->debug("Ignoring dispatcher session as it wasn't restored from a valid session ID");
+        return {};
+    }
+
+    my %session_data;
     foreach my $key ($session->param) {
         my $value = $session->param($key);
         $logger->debug( sub { "Adding session parameter from dispatcher session to Catalyst session : $key : " . ($value // 'undef') });
--- a/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
@@ -27,6 +27,7 @@ use URI::URL;
 use URI::Escape::XS qw(uri_escape uri_unescape);
 use HTML::Entities;
 use List::MoreUtils qw(any);
+use pf::constants::Portal::Session qw($DUMMY_MAC);
 
 =head1 NAME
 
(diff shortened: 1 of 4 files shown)


julsemaan:
@dwlfrth, @jrouzierinverse, @fdurand 

Please take a look at this and ensure I'm not doing anything crazy.
This will be cherry picked up to maintenance/6.0

lib/pf/Portal/Session.pm
```diff
@@ -94,10 +96,11 @@ sub _initialize {
 
     $self->{'_cgi'} = $cgi;
 
-    my $sid = $cgi->cookie(SESSION_ID) || $cgi->param(SESSION_ID) || $cgi;
+    my $md5_mac = defined($mac) ? md5_hex($mac) : undef;
+    my $sid = $cgi->cookie(SESSION_ID) || $cgi->param(SESSION_ID) || $md5_mac || md5_hex(word(8, 12));
     $logger->debug("using session id '$sid'" );
     my $session;
-    $self->{'_session'} = $session = new CGI::Session( "driver:chi", $sid, { chi_class => 'pf::CHI', namespace => 'httpd.portal' } );
+    $self->{'_session'} = $session = new CGI::Session( "driver:chi;id:static", $sid, { chi_class => 'pf::CHI', namespace => 'httpd.portal' } );
```
dwlfrth, line 104: Does this require `CGI::Session::ID::static` ?

jrouzierinverse: No it is already included in CGI::Session versions greater the 4. We would need to verify the version for older 6 install before we cherry pick it into maintence

jrouzierinverse (changes requested):
The only issue I see is that since we did not specify which version of CGI::Session to use in the spec file there may be older version of CGI::Session that would not have CGI::Session::ID::static.

Can you please verify the version of CGI::Session we had in the obs for the 6.Y.Z versions of packetfence.

julsemaan:
@jrouzierinverse 

From what I can tell, we're shipping version 4 and above in both CentOS 6 and 7 
As for Debian, we're not shipping it so we're using whatever version is shipped by Debian

julsemaan:
Also, I've added a second commit, please check it out as well

dwlfrth:
Since 'dummy_mac' is used in two different places, it should be a constant

lib/pf/Portal/Session.pm
```diff
@@ -106,7 +109,10 @@ sub _initialize {
         }
     );
 
-    $self->{'_client_mac'} = $mac || $self->session->param("_client_mac") || $self->_restoreFromSession("_client_mac",sub {
+    $self->{_dummy_session} = (defined($mac) && $mac eq "ff:ff:ff:ff:ff:ff");
```
dwlfrth: Constant !

lib/pf/Portal/Session.pm
```diff
@@ -106,7 +109,10 @@ sub _initialize {
...
+
+    # Don't assign $mac if the dummy MAC was used for restoring the session
+    $self->{'_client_mac'} = ((defined($mac) && $mac ne "ff:ff:ff:ff:ff:ff") ? $mac : undef) || $self->session->param("_client_mac") || $self->_restoreFromSession("_client_mac",sub {
```
dwlfrth: Constant !

html/captive-portal/lib/captiveportal/PacketFence/Model/Portal/Session.pm
```diff
@@ -251,9 +251,21 @@ sub _build_profile {
 
 sub _build_dispatcherSession {
     my ($self) = @_;
-    my $session = new pf::Portal::Session()->session;
-    my %session_data;
     my $logger = get_logger();
+
+    # Restore with a dummy MAC since we don't care about what contains the session if it can't be restored from the session ID
+    my $dummy_mac = "ff:ff:ff:ff:ff:ff";
```
dwlfrth: Constant !

julsemaan:
All requested changes should be there.

I would suggest we merge this in devel, test it thoroughly for 7.1, then cherry-pick it in maintenance once we confirm it works fine.

Side note, this is already working in a production environment using web-auth with no complaints yet

jrouzierinverse (approved):

PR #2369, opened 2017-05-31
Use returned radius attributes in order to tests them in rules

fdurand:
# Description
When you do radius authentication (radius source) you can use the radius attributes returned in the answer in order to use it to compute some rules.

# Impacts
Authentication flow

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Use the attributes returned by a radius use source as attributes to compute the rules.

### DIFF
--- a/lib/pf/Authentication/Source/RADIUSSource.pm
+++ b/lib/pf/Authentication/Source/RADIUSSource.pm
@@ -12,6 +12,7 @@ use pf::constants qw($TRUE $FALSE);
 use pf::Authentication::constants qw($LOGIN_CHALLENGE);
 use pf::constants::authentication::messages;
 use pf::log;
+use pf::config qw(%Config);
 
 our $RADIUS_STATE = 'State';
 our $RADIUS_REPLY_MESSAGE = 'Reply-Message';
@@ -48,9 +49,9 @@ sub available_attributes {
   my $self = shift;
 
   my $super_attributes = $self->SUPER::available_attributes;
-  my $own_attributes = [{ value => "username", type => $Conditions::SUBSTRING }];
-
-  return [@$super_attributes, @$own_attributes];
+  my @attributes = @{$Config{advanced}->{radius_attributes}};
+  my @radius_attributes = map { { value => $_, type => $Conditions::SUBSTRING } } @attributes;
+  return [@$super_attributes, @radius_attributes];
 }
 
 =head2  authenticate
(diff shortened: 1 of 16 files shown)


dwlfrth:
@jrouzierinverse see if it conflicts with authentication sources rework. Let @fdurand know and adjust.
Depending on the work, we aim at 7.2

jrouzierinverse:
There are no conflicts

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -167,15 +180,23 @@ sub check_radius_password {
...
     $params->{'username'} = $params->{'stripped_user_name'} if (defined($params->{'stripped_user_name'} ) && $params->{'stripped_user_name'} ne '' && isenabled($self->{'stripped_user_name'}));
     my $username =  $params->{'username'};
+
     foreach my $condition (@{ $own_conditions }) {
         if ($condition->{'attribute'} eq "username") {
             if ( $condition->matches("username", $username) ) {
                 push(@{ $matching_conditions }, $condition);
             }
         }
+        for my $attribute (@{ $extra->{attributes}} ) {
```
jrouzierinverse: Add a check to make sure extra is defined

jrouzierinverse (changes requested):
Extra is not used for matching here
html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm

Extra is not save in the session here
html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm

In create_local_account extra is not used for matching there

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication.pm

Should we use extra here even though these sources do not return the extra field
html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm

fdurand:
Extra is save in the session here
html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Login.pm +207

Also not sure we have to use the extra field in email and sponsor. Right now only radius source support this extra field.

lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -167,15 +180,25 @@ sub check_radius_password {
...
     $params->{'username'} = $params->{'stripped_user_name'} if (defined($params->{'stripped_user_name'} ) && $params->{'stripped_user_name'} ne '' && isenabled($self->{'stripped_user_name'}));
     my $username =  $params->{'username'};
+
     foreach my $condition (@{ $own_conditions }) {
         if ($condition->{'attribute'} eq "username") {
             if ( $condition->matches("username", $username) ) {
                 push(@{ $matching_conditions }, $condition);
             }
         }
+        if defined($extra->{attributes}) {
```
jrouzierinverse: That would not compile you forgot to put () around your condition. 
Yes switching between go and perl is annoying
Just check if $extra is defined
if (defined $extra) {

jrouzierinverse (commented):
Has compile errors

jrouzierinverse (approved):

PR #2387, opened 2017-06-08
Added sd_notify support to most sbin/ daemons.

louismunro:
# Description
Adds systemd sd_notify support to 

- sbin/pfbandwidthd
- sbin/pfconfig 
- sbin/pfdetect
- sbin/pfdhcplistener
- sbin/pffilter
- sbin/pfmon
- sbin/pfqueue
- sbin/pfsetvlan
- sbin/winbindd-wrapper

Also reconfigure Mariadb to use sd_notify.

# Impacts
All the above services should now require a "READY" notification before systemd considers them started.

# Delete branch after merge
YES 

## Enhancements
* PF sbin/ daemons now include support for systemd notify service type

### DIFF
--- a/sbin/pfsetvlan
+++ b/sbin/pfsetvlan
@@ -45,7 +45,7 @@ use Getopt::Long;
 use Net::SMTP;
 use Pod::Usage;
 use POSIX;
-
+use Systemd::Daemon qw{ -soft -notify };
 
 use constant {
     PFCMD_FILE => INSTALL_DIR . "/bin/pfcmd"
@@ -217,6 +217,8 @@ sub add_new_threads_to_list {
 # }}}1
 
 # main program - read trap lines and addTrapLineToQueue($trapLine) {{{1
+
+&Systemd::Daemon::notify( READY => 1, STATUS => "Ready", unset => 1 );
 my $currentTrapLine = undef;
 my $completeTrapLine;
 my $inMultiLineTrap = 0;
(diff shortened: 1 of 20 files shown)


sbin/pfsetvlan
```diff
@@ -45,7 +45,7 @@ use Getopt::Long;
 use Net::SMTP;
 use Pod::Usage;
 use POSIX;
-
+use Systemd::Daemon qw{ -soft -notify };
```
jrouzierinverse: Change this to 
```Systemd::Daemon qw{ -soft -notify };```
This
``` use Systemd::Daemon(); ```
Since you use the full name of methods

File: sbin/pfsetvlan [hunk 2]
```diff
@@ -2109,6 +2111,7 @@ sub add_to_switch_locker {
 }
 
 END {
+    &Systemd::Daemon::notify( STOPPING => 1 );
     if ( ( !$man ) && ( !$help ) ) {
```
jrouzierinverse: Move notify in the if statement

sbin/winbindd-wrapper
```diff
@@ -127,6 +129,7 @@ start();
 cleanup();
 
 END {
+    &Systemd::Daemon::notify( STOPPING => 1 );
```
jrouzierinverse: Notify should be done only when you are doing the delete pid.
To avoid calling notify if you run the help

jrouzierinverse (commented):
The notify for stopping should only be called.
When the service was running.
My suggestion is to move the notify-stop to same place where deletepid is done.

jrouzierinverse:
```use Systemd::Daemon qw{ -soft -notify };``` is needed to preload the XS

PR #2389, opened 2017-06-12
Limit journal maximum size on disk.

### LINKED ISSUE #2380: journald: limit logfile size
Current limit is set to 4Go, should limit to 1

louismunro:
# Description
Adds configuration to ensure journald never uses more than 1GB on disk, and always leaves a minimum of 1Gb free disk space.

# Impacts
/var/log/journald should not grow over 1Gb of usage.

# Issue
fixes #2380 

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Journald total file size is now capped at 1Gb

### DIFF
--- a/debian/packetfence.preinst
+++ b/debian/packetfence.preinst
@@ -64,6 +64,8 @@ case "$1" in
         echo "RateLimitInterval=0" >> /etc/systemd/journald.conf
         echo "RateLimitBurst=0" >> /etc/systemd/journald.conf
         echo "ForwardToWall=no" >> /etc/systemd/journald.conf
+        echo "SystemMaxUse=1G" >> /etc/systemd/journald.conf
+        echo "SystemKeepFree=1G" >> /etc/systemd/journald.conf
         echo "Restarting journald to enable persistent logging"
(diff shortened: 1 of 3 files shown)


debian/packetfence.preinst
```diff
@@ -64,6 +64,8 @@ case "$1" in
         echo "RateLimitInterval=0" >> /etc/systemd/journald.conf
         echo "RateLimitBurst=0" >> /etc/systemd/journald.conf
         echo "ForwardToWall=no" >> /etc/systemd/journald.conf
+        echo "SystemMaxUse=1G" >> /etc/systemd/journald.conf
```
dwlfrth, line 67: Can't we use the same logic as RHEL/CentOS and use the provided file rather than writing a new one ?
That would avoid having to "manage" two separate configurations

louismunro: Unfortunately no.
Debian uses an older version of systemd than RHEL which does not allow overriding journald.conf in a separate file.
It sucks tremendously, I know.

dwlfrth (approved):
LGTM

PR #2391, opened 2017-06-12
Server certs and keys are now never overwritten if they exist.

### LINKED ISSUE #2366: TLS cert is overwritten if server.crt has been removed
Due to the way the packaging handles certificate creation, if the server.crt file has been moved or deleted a new server.pem will be created, overwriting the previous one.

We should never overwrite certificate files.

louismunro:
# Description
Adds targets to the Makefile to create the TLS certs and key if and only if the files do not already exist.

# Impacts
server.key, server.crt, and server.pem will only get created if they don't exist already.
Otherwise they will remain untouched.

# Issue
fixes #2366 

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes
* TLS certificates and keys will no longer be overwritten (#2366)


dwlfrth:
LGTM

PR #2393, opened 2017-06-13
redirect to plain HTTP for access detection in portal

julsemaan:
# Description
Redirect to plain HTTP for network access detection to prevent mixed content warnings

# Impacts
The **whole** network detection flow, so we need to test this against as many devices we can

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Use plain HTTP for network access detection page

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
@@ -86,7 +86,7 @@ sub release {
     # One last check for the violations
     return unless($self->handle_violations());
 
-    return $self->app->redirect("/access") unless($self->app->request->path eq "access");
+    return $self->app->redirect("http://" . $self->app->request->header("host") . "/access") unless($self->app->request->path eq "access");
 
     get_logger->info("Releasing device");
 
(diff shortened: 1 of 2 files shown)


html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Root.pm
```diff
@@ -86,7 +86,7 @@ sub release {
     # One last check for the violations
     return unless($self->handle_violations());
 
-    return $self->app->redirect("/access") unless($self->app->request->path eq "access");
+    return $self->app->redirect("http://" . $self->app->request->header("host") . "/access") unless($self->app->request->path eq "access");
```
dwlfrth, line 89: Is the `unless` condition still required ?

julsemaan: Yes, otherwise it will always redirect to access even though you're still on access so that doesn't change

dwlfrth (approved):
LGTM

fdurand (approved):

whitx (approved):

dwlfrth:
@julsemaan can you squash merge so that you keep credit. I will cherry-pick afterward.

julsemaan:
@dwlfrth done

dwlfrth:
Cherry-picked into:

- `maintenance/7.1`
- `maintenance/7.0`
- `maintenance/6.5`

PR #2400, opened 2017-06-15
Added redirection template in violation

fdurand:
# Description
Added a violation template to redirect the user to the redirect url specified in the violation config.

# Impacts
Violation flow

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Redirect template for violation

### DIFF
--- a/html/captive-portal/templates/violations/redirect.html
+++ b/html/captive-portal/templates/violations/redirect.html
@@ -1,1 +1,1 @@
-<meta http-equiv="refresh" content="0; url=[% IF redirect_url %][% redirect_url %][% ELSE %]packetfence.org[% END %]" />
+<meta http-equiv="refresh" content="0; url=[% IF redirect_url %][% redirect_url %][% ELSE %]http://packetfence.org/[% END %]" />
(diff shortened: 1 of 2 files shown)


html/captive-portal/templates/violations/redirect.html
```diff
@@ -0,0 +1 @@
+<meta http-equiv="refresh" content="0; url=[% redirect_url %]" />
```
jrouzierinverse: Check if redirect_url is defined before displaying it.

jrouzierinverse (changes requested):

html/captive-portal/templates/violations/redirect.html
```diff
@@ -1 +1 @@
-<meta http-equiv="refresh" content="0; url=[% redirect_url %]" />
+<meta http-equiv="refresh" content="0; url=[% IF redirect_url %][% redirect_url %][% END %]" />
```
jrouzierinverse: If there is no redirect url where should it go packetfence.org or will there always be a redirect_url?

fdurand: I can set packetfence.org by default

jrouzierinverse (approved):

PR #2409, opened 2017-06-19
Added a check to ignore broadcasts in cluster when not holding the VIP.

### LINKED ISSUE #2408: DHCP broadcast packets should not be handled by all servers in a cluster

louismunro:
# Description
Adds a check to pfdhcplistener to ignore broadcast dhcprequests in clustered mode if this server is hot currently holding the VIP.

# Impacts
DHCPv4 requests to the broadcast ip (255.255.255.255) should be ignored in a cluster when the node does not hold the VIP.

# Issue
fixes #2408 

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes
* Fixed issue where dhcp broadcast requests are handled multiple times in a cluster (#2408)

### DIFF
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -97,6 +97,8 @@ my $pcap;
 my $net_type;
 my $process;
 my $interface;
+my $is_clustered = $FALSE;
+my $holds_vip = $FALSE;
 
 my $rate_limit_hash = {};
 my $rate_limit_cache = CHI->new( driver => 'Memory', datastore => $rate_limit_hash );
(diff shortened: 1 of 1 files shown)


sbin/pfdhcplistener
```diff
@@ -107,6 +109,8 @@ sub reload_config {
     }
     elsif ( pf::cluster::is_vip_running($interface) ) { 
         $process = $TRUE;
+        $is_clustered = $TRUE;
+        $holds_vip = $TRUE;
```
julsemaan, line 113: You're never setting this variable back to false so it seems that if you own the VIP once, this flag will stay true even if you lose the VIP

julsemaan:
Also, IMO this belongs in maintenance

sbin/pfdhcplistener
louismunro: You can get the VIP back when you pry it from my cold, dead hands...

All right, I'll give it back...
Fixing this.

louismunro:
Agreed, this should be cherry picked.

PR #2405, opened 2017-06-16
remove "password server" from samba chroot config

kkplein:
Removed the "password server" from the generated chroot config
as recommended by the samba core team members Andrew Bartlett
and Rowland Penny here:
https://lists.samba.org/archive/samba/2017-June/208999.html

The "password server" option exists, to be able to *restrict* samba from using all available servers, but rather make it stick to one or more specific DCs.

# Description
After some discussion on the samba mailinglist on the packetfence generated smb.conf in the chroot, it was concluded that removing the "password server = " option from smb.conf is the better way of doing things. With the line removed, it will default to "password server = *" and that will use the samba builtin-mechanisms to locate all available DCs.
Specially in the case of using 802.1x auth, the DC selection/location should be as robust and fool-proof as possible.

# Impacts
I am not aware of any other impact, rather than improving the DC location/selection behaviour.

# Delete branch after merge
YES

# NEWS file entries
none

## Enhancements
* use the regular samba mechanisms to locate and select DCs

# UPGRADE file entries
none

PR #2410, opened 2017-06-19
Ability to use a sticky DC on domain join

dwlfrth:
# Description
Address #2405 but in a configurable way. See that PR (#2405) for details


dwlfrth:
@julsemaan

PR #2412, opened 2017-06-19
Rework pfdhcplistener to conditionally handle broadcasts

### LINKED ISSUE #2408: DHCP broadcast packets should not be handled by all servers in a cluster

louismunro:
NOT READY
# Description
Reworks the pfdhcplistener logic to handle broadcast requests only on the VIP node of a cluster.

# Impacts
Reworks the logic of dhcp request processing.
From now on, the only packets that are not processed are the DHCP broadcast sent to a cluster node that does not hold the VIP.

# Issue
fixes #2408 

# Delete branch after merge
YES 

# NEWS file entries
## Bug Fixes
* Fixed an issue where DHCP broadcast were treated more than once in clustered mode

### DIFF
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -96,27 +96,22 @@ my $interface_ip;
 my $interface_vlan;
 my $pcap;
 my $net_type;
-my $process;
 my $interface;
+my $process_broadcast = $TRUE;
 
 my $rate_limit_hash = {};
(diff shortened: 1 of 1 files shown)


julsemaan:
It looks good but with the indent change its hard to be sure of what changed but that does look legit

@louismunro, can you confirm this PR was tested and we'll go ahead and merge it as all looks good

julsemaan (approved):

louismunro:
It's not good.
I messed up a merge, some code is duplicated.

Fixing it.

PR #2413, opened 2017-06-19
Ignore DHCP broadcasts on the node that does not hold the VIP

### LINKED ISSUE #2408: DHCP broadcast packets should not be handled by all servers in a cluster

louismunro:
# Description
Reworks the pfdhcplistener logic to handle broadcast requests only on the VIP node of a cluster.

# Impacts
Reworks the logic of dhcp request processing.
From now on, the only packets that are not processed are the DHCP broadcast sent to a cluster node that does not hold the VIP.

# Issue
fixes #2408 

# Delete branch after merge
YES 

# NEWS file entries
## Bug Fixes
* Fixed an issue where DHCP broadcast were treated more than once in clustered mode

### DIFF
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -217,7 +217,7 @@ sub process_pkt {
             my $l2 = NetPacket::Ethernet->decode($pkt);
 
             if ( ! $process_broadcast ) { 
-                if ( $l2->{'dest_mac'} eq 'ffffffffff' ) { 
+                if ( $l2->{'dest_mac'} eq 'ffffffffffff' ) { 
                     $logger->trace("Skipping broadcast request.");
                     return;
                 } 
--- a/sbin/pfdhcplistener
+++ b/sbin/pfdhcplistener
@@ -96,27 +96,21 @@ my $interface_ip;
 my $interface_vlan;
 my $pcap;
 my $net_type;
-my $process;
+my $process_broadcast;
 my $interface;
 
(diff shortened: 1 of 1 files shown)


sbin/pfdhcplistener
```diff
@@ -219,10 +213,15 @@ sub setup_pcap {
 
 sub process_pkt {
     my ( $user_data, $hdr, $pkt ) = @_;
-    if ($process){
         eval {
             my $l2 = NetPacket::Ethernet->decode($pkt);
 
+            if ( ! $process_broadcast ) { 
+                if ( $l2->{'dest_mac'} eq 'ffffffffff' ) {
```
julsemaan: Either I should go back to 1st grade or you're missing two f here

louismunro: ff it.

julsemaan:
Merging this but I feel it might be better to have this tested in a cluster prior to cherry-picking it

louismunro:
I agree.
I think we should refactor this whole mess, but that is not something I want to cherry pick.

julsemaan:
Thing is we «have» to cherry-pick since that will cause duplicate ip4log entries to be opened which will lead to big fat ip4log tables (like in the movie My big fat greek database)

louismunro:
I mean I did not want to cherry pick the refactor I believe we should do.

The current code is a nest of conditions and long running subs.
It should be broken up in more manageable subs and the conditions reconsidered to see if they still even make sense.

PR #2414, opened 2017-06-20
feature/normalize_radius_username

jrouzierinverse:
Description
===========
Normalize the radius username for machine auth when using TLS

Impacts
=======
RADIUS authorize workflow

Delete branch after merge
=========================
NO

NEWS file entries
=================

Enhancements
------------
* Have the ability to normalize the machine auth username when using TLS

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1127,7 +1127,7 @@ EOT
 type=toggle
 options=enabled|disabled
 description=<<EOT
-Use the radius username instead of the TLS certificate common name when doing machine auth.
+Use the RADIUS username instead of the TLS certificate common name when doing machine authentication.
 EOT
 
 [omapi.ip2mac_lookup]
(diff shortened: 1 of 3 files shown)


louismunro:
The code is fine, but the documentation does not explain what the option is doing, or why one would want to use it. That should be made more explicit otherwise people will just toggle it on and off for no reason.

jrouzierinverse:
Where in the documentation do you want to place it?

louismunro:
Just in the help string.
The word "normalize" can mean anything depending on the context.
Just make it explicit what we mean by that.

jrouzierinverse:
Updated documentation.
Provide any recommendation.

conf/documentation.conf
```diff
@@ -1123,6 +1123,13 @@ description=<<EOT
 Ability to manage all active devices from a same switch port
 EOT
 
+[advanced.normalize_radius_machine_auth_username]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Use the RADIUS username instead of the TLS certificate common name when doing machine authentication.
```
louismunro, line 1130: Much better.

PR #2422, opened 2017-06-21
make haproxy rate limiting configurable

### LINKED ISSUE #2418: haproxy ACLs: allow to disable the abuse ACLs
Some people prefer user experience over improved performance and can prefer disabling the abuser ACL in haproxy especially since it returns a simple 403 without many explanations which can cause calls to their help desk.

We should provide an option to disable these ACLs in haproxy

julsemaan:
# Description
make haproxy rate limiting configurable and add a more meaningful message when a user is rate-limited

# Impacts
haproxy rate limiting

# Issue
fixes #2418 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Allowed haproxy captive portal rate-limiting to be configurable

### DIFF
--- a/lib/pf/services/manager/haproxy.pm
+++ b/lib/pf/services/manager/haproxy.pm
@@ -158,9 +158,15 @@ frontend portal-http-$cluster_ip
         http-request set-header Host %[var(req.host)] if host_exist
         http-request lua.select
         acl action var(req.action) -m found
+EOT
+            if($rate_limiting) {
+            $tags{'http'} .= <<"EOT";
         acl unflag_abuser src_clr_gpc0 --
         http-request allow if action unflag_abuser
         http-request deny if { src_get_gpc0 gt 0 }
+EOT
+            }
+            $tags{'http'} .= <<"EOT";
         reqadd X-Forwarded-Proto:\\ http
         use_backend %[var(req.action)]
         default_backend $cluster_ip-backend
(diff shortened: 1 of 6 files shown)


fdurand:
It's optional but it should be better to remove that from the generated haproxy config if the rate limiting is disabled.
 
`       acl action var(req.action) -m found
        acl unflag_abuser src_clr_gpc0 --
        http-request allow if action unflag_abuser
        http-request deny if { src_get_gpc0 gt 0 }
`

julsemaan:
@dwlfrth also made me notice:
https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/services/manager/haproxy.pm#L159

julsemaan:
@fdurand + @dwlfrth 

I think I've fixed all the required places, please confirm

fdurand:
Looks good to me

lib/pf/services/manager/haproxy.pm
```diff
@@ -30,6 +30,7 @@ use pf::file_paths qw(
     $install_dir
     $conf_dir
     $var_dir
+    $captiveportal_templates_path
```
dwlfrth, line 33: I think you are missing the `pf::file_paths` path !

dwlfrth (dismissed):

lib/pf/services/manager/haproxy.pm
dwlfrth: Forget it !

dwlfrth (approved):
LGTM

PR #2443, opened 2017-07-03
Systemd will no longer kill only master httpd process.

### LINKED ISSUE #2439: Potential issue while restarting processes
It seems that the killmode and the way we kill processes via systemd kill only the parent, which is supposed to kill their childs. Sometimes it does appear that the parent do not do as expected. This will block a process in a state with a new parent and old childs. Have to manually kill the childs.

Have seen it happen with apache processes.

louismunro:
# Description
This changes the way systemd kills httpd processes.
It will no longer only kill the parent process.
I will now kill all processes left after a maximum of TimeoutStopSec.

# Impacts
When restarting httpd processes through systemd there should be no left over child processes.

# Issue
fixes #2439 

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes
* Fixes leftover httpd processes when restarting (#2439)


dwlfrth (approved):
LGTM

PR #2449, opened 2017-07-04
Fix/fdb issues bypass

### LINKED ISSUE #2448: Our critical processes aren't protected against a Fingerbank issue
Our portal flow and RADIUS flow aren't protected against a potential issue with the Fingerbank DB and requests will fail due to it even though they don't actually require device profiling info to grant/deny access to the network

We need to protect the following flows from this issue:
- RADIUS 802.1x
- RADIUS MAB
- Portal registration
- iplog update

julsemaan:
# Description
Protect RADIUS and portal authentication against a potential issue with the Fingerbank DB

# Impacts
RADIUS + portal flows that use device profiling

# Issue
fixes #2448 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Ensure RADIUS and portal authentication won't be affected if the Fingerbank database has issues

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Application.pm
@@ -134,41 +134,6 @@ sub reached_retry_limit {
     return $retries > $max;
 }
 
-=head2 process_user_agent
-
-Process the user agent
-
-=cut
-
-sub process_user_agent {
-    my ( $self ) = @_;
-    my $user_agent    = $self->current_user_agent;
-    my $logger        = get_logger();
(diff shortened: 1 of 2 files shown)


dwlfrth (commented):
LGTM,
One question: we no longer want to "process_user_agent" at all ?

julsemaan:
@dwlfrth I don't think so, don't really see where this is used anymore but then, I didn't look that much but I'm 95% sure its not related to device profiling anymore

dwlfrth:
Aight.
Squash merge then I will backport

dwlfrth:
Cherry-picked in

- `maintenance/7.1`
- `maintenance/7.0`
- `maintenance/6.5`
- `maintenance/6.4`
- `maintenance/6.3`

PR #2471, opened 2017-07-06
Feature/select role

julsemaan:
# Description
Create a portal module that allows a certain role to select another role for his device (like a technician registering devices with the portal)

# Impacts
Portal

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Added portal module for selecting a role for the device being registered on the portal

### DIFF
--- a/html/captive-portal/lib/captiveportal/DynamicRouting/Module/SelectRole.pm
+++ b/html/captive-portal/lib/captiveportal/DynamicRouting/Module/SelectRole.pm
@@ -1,0 +1,47 @@
+package captiveportal::DynamicRouting::Module::SelectRole;
+use Moose;
+
+BEGIN { extends 'captiveportal::PacketFence::DynamicRouting::Module::SelectRole'; }
+
+=head1 NAME
+
+captiveportal::DynamicRouting::Module::SelectRole - SelectRole Controller for captiveportal
+
+=head1 DESCRIPTION
+
+[enter your description here]
+
+=cut
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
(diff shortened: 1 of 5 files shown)


fdurand:
Applied on a setup and it works.
Not against merging for 7.3

html/captive-portal/lib/captiveportal/DynamicRouting/Module/SelectRole.pm
```diff
@@ -0,0 +1,47 @@
...
+
+BEGIN { extends 'captiveportal::PacketFence::DynamicRouting::Module::SelectRole'; }
+
+=head1 NAME
+
+captiveportal::DynamicRouting::Module::SelectRole - SelectRole Controller for captiveportal
+
+=head1 DESCRIPTION
+
+[enter your description here]
```
fdurand, line 12: Module to select a new role for the device being registered

julsemaan: Given its the «extension» point, we usually don't put a description there
At least, thats how the others are, IDK if it was the way it was meant to be but I find it to be a good idea

dwlfrth:
Push to 7.4 to allow documentation writeup

dwlfrth:
@whitx test and check if you can get some documentation

fdurand:
Agree to merge

julsemaan:
@fdurand then you can hit that beautiful merge button.

Since @whitx has a commit in this branch the CLA check won't pass but I think that doesn't matter.
I guess @cgx can confirm

julsemaan:
image

PR #2476, opened 2017-07-07
add long description to custom reports

### LINKED ISSUE #2022: Admin: Custom reports should have a long description
Custom reports should have a long description.

That will be especially handy when we will start shipping 'built-in' reports so that we can add some context around each report

julsemaan:
# Description
add long description to custom reports

# Impacts
Custom reports

# Issue
fixes #2022 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Added parameter to have extended descriptions in the custom reports

### DIFF
--- a/html/pfappserver/root/dynamicreport/index.tt
+++ b/html/pfappserver/root/dynamicreport/index.tt
@@ -1,7 +1,7 @@
 <div class="card-title">
 <h2>[% IF report.description ; report.description ; ELSE ; report.id ; END %]</h2>
 [% IF report.long_description %]
-<p>[% report.long_description.replace('\n', '<br>') | none %]</p>
+<p>[% l(report.long_description.replace('\n', '<br>')) | none %]</p>
 [% END %]
 </div>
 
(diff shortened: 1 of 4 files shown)


html/pfappserver/root/dynamicreport/index.tt
```diff
@@ -1,4 +1,9 @@
-<h2 class="card-title">[% IF report.description ; report.description ; ELSE ; report.id ; END %]</h2>
+<div class="card-title">
+<h2>[% IF report.description ; report.description ; ELSE ; report.id ; END %]</h2>
+[% IF report.long_description %]
+<p>[% report.long_description.replace('\n', '<br>') | none %]</p>
```
fdurand: localize

julsemaan: Should we really localize a user driven value ?
long_description is set by the user and isn't static

I guess that would be only good for the ones we ship by default but if that does make sense, I'll add it

fdurand: At least the stuff you defined in report.conf.defaults

julsemaan:
Re-review

PR #2490, opened 2017-07-17
FixedRole portal module

fdurand:
# Description
Portal module to keep the previous role defined if this role is in a specific list.

# Impacts
PORTAL authentication workflow

# Delete branch after merge
YES

# NEWS file entries
## New Features
* New portal module to keep the previous role if this role is in a list.

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/PortalModule/RoleInStone.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/PortalModule/RoleInStone.pm
@@ -1,0 +1,75 @@
+package pfappserver::Form::Config::PortalModule::RoleInStone;
+
+=head1 NAME
+
+pfappserver::Form::Config::PortalModule:RoleInStone
+
+=head1 DESCRIPTION
+
+Form definition to create or update a Role in stone portal module.
+
+=cut
+
+use HTML::FormHandler::Moose;
+extends 'pfappserver::Form::Config::PortalModule';
+with 'pfappserver::Base::Form::Role::Help';
+
+use pf::nodecategory;
+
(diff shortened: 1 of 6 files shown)


html/pfappserver/lib/pfappserver/Form/Config/PortalModule/RoleInStone.pm
```diff
@@ -0,0 +1,75 @@
...
+use captiveportal::DynamicRouting::Module::RoleInStone;
+sub for_module {'captiveportal::PacketFence::DynamicRouting::Module::RoleInStone'}
+## Definition
+
+has_field 'stone_roles' =>
+  (
+   type => 'Select',
+   multiple => 1,
+   label => 'Roles',
+   options_method => \&options_admin_role,
```
julsemaan: that would be options_stone_role

html/pfappserver/lib/pfappserver/Form/Config/PortalModule/RoleInStone.pm
```diff
@@ -0,0 +1,75 @@
...
+   element_attr => {'data-placeholder' => 'Click to add a role'},
+   tags => { after_element => \&help,
+             help => 'Nodes with the selected roles will be affected' },
+  );
+
+sub child_definition {
+    return qw(stone_roles);
+}
+
+sub options_admin_role {
```
julsemaan: options_stone_role

julsemaan (commented):
I'm not sure about the RoleInStone name

Sounds fancy and all but not sure a non-Inverse user would understand what it is.

Maybe FixedRole or something like that would be more accurate for a name

PR #2499, opened 2017-07-18
Feature/sso username format

### LINKED ISSUE #2495: Normalize username sent via pfsso
Most of the time in PacketFence we have usernames that aren't following the same standard (`example\bobby`, `bobby@example`, `bobby`) and they are sent as is to the firewall.

We should have a way to format the way the username is sent to the firewall

I suggest the following:
We first do a strip_username on the username and extract the realm + username

Then we allow the user to configure the format he wishes those to be sent where $u is the username and $r is the realm

Examples:
- `$r\$u` would send example\bobby
- `$u@$r` would send bobby@example
- `zammitcorp\$u` would send zammitcorp\bobby and would ignore the realm

Then we could have the ability to define the default realm when there is none defined

With these two parameters we should be able to send the username the way the firewall expects it

julsemaan:
# Description
Allow the formatting of the username sent to firewalls in SSO

# Impacts
Firewall SSO

# Issue
fixes #2495 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Allow to specify the format of the username that is being sent in firewall SSO

### DIFF
--- a/go/caddy/pfsso/pfsso.go
+++ b/go/caddy/pfsso/pfsso.go
@@ -98,6 +98,11 @@ func (h PfssoHandler) parseSsoRequest(ctx context.Context, body io.Reader) (map[
 		return nil, 0, err
 	}
 
+	if info["stripped_username"] == "" {
+		log.LoggerWContext(ctx).Warn("No stripped_username set in the request, using the username as the stripped_username and no realm")
+		info["stripped_username"] = info["username"]
+	}
+
 	return info, int(timeout), nil
 }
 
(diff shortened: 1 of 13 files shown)


extrafu:
Looks good to me.

julsemaan:
Ready for review

julsemaan:
Should add a way to format or not the machine accounts usernames (host/*)

julsemaan:
This is ready for review now, the host account part was added

julsemaan:
@dwlfrth could you give this some love like zayme does when he prepares his sandwiches

dwlfrth (approved):
LGTM

julsemaan:
@dwlfrth 

Merge that beauty

PR #2501, opened 2017-07-19
Feature/timezone select

### LINKED ISSUE #2483: Timezone should be selectable
Currently the Timezone is a text field.
We even point to the list of supported zones.

Why not make it a pull down menu where the user selects the right one?
The current field only makes for odd issues when someone makes a typo in the Timezone.

whitx:
# Description
Allow timezone to be selected instead of written manually.

# Impacts
PF general configuration

# Issue
Fixes #2483 

# Delete branch after merge
YES 

# NEWS file entries
## Enhancements
* Allow timezone to be selectable


html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
```diff
@@ -141,6 +141,16 @@ sub field_list {
                 $field->{options} = \@options;
                 last;
             };
+            $type eq 'timezone' && do {
+                $field->{type} = 'Select';
+                $field->{element_class} = ['chzn-deselect'];
+                $field->{element_attr} = {'data-placeholder' => 'Select a timezone'};
+                my @timezones = DateTime::TimeZone->all_names();
```
fdurand, line 148: Just in case i prefer to add DateTime::TimeZone as a dep for Debian and maybe for centos

PR #2506, opened 2017-07-20
document galera max connections reached

### LINKED ISSUE #2503: Document max connections reached behavior in clustering doc
When the max connections are reached on the MariaDB Galera server, it locks down all the synchronization.

Now, if you restart the packetfence-mariadb service while its in that state on any of the node, a certain amount of events will not have been synced which will lead to a split brain and a manual recovery to be necessary.

We need to document that behavior since its impossible to connect to MySQL during that time to up the amount of connections so its inevitable that a split brain will occur when we reach max connections.

We should also research solutions to that problem if they exist (like reserving connections for the exclusive usage of the replication)

julsemaan:
# Description
Document what happens and how to resolve issues with Galera when MySQL max connections was reached

# Impacts
Documentation

# Issue
fixes #2503 

# Delete branch after merge
YES

### DIFF
--- a/docs/PacketFence_Clustering_Guide.asciidoc
+++ b/docs/PacketFence_Clustering_Guide.asciidoc
@@ -913,7 +913,7 @@ MariaDB Galera cluster troubleshooting
 Max connections reached
 +++++++++++++++++++++++
 
-In the event that one of the 3 servers reaches the maximum amount of connections (defaults to 1000), this will dead-lock the Galera cluster synchronization. In order to resolve this, you should first increase database_advanced.max_connections, then stop packetfence-mariadb on all 3 servers, and follow the steps in the section 'Recovering from a split brain' of this document using anyone of the 3 servers as your source of truth.
+In the event that one of the 3 servers reaches the maximum amount of connections (defaults to 1000), this will dead-lock the Galera cluster synchronization. In order to resolve this, you should first increase database_advanced.max_connections, then stop packetfence-mariadb on all 3 servers, and follow the steps in the section 'Recovering from a split brain' of this document. Note that you can use any of the database servers as your source of truth.
 
 Investigating further
 ^^^^^^^^^^^^^^^^^^^^^
--- a/docs/PacketFence_Clustering_Guide.asciidoc
+++ b/docs/PacketFence_Clustering_Guide.asciidoc
@@ -907,3 +907,29 @@ Once this is done, restart pfmon on all nodes using:
 
   # /usr/local/pf/bin/pfcmd service pfmon restart
 
+MariaDB Galera cluster troubleshooting
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Max connections reached
++++++++++++++++++++++++
+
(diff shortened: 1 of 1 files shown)


docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -902,3 +902,29 @@ Once this is done, restart pfmon on all nodes using:
 
 NOTE: It is normal that pfmon will refuse to start on nodes B and C. Nonetheless you should execute the command on them to make sure you catch any stale process.
 
+MariaDB Galera cluster troubleshooting
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Max connections reached
```
extrafu: I would write Max -> Maximum

docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -902,3 +902,29 @@ Once this is done, restart pfmon on all nodes using:
...
++++++++++++++++++++++++
+
+In the event that one of the 3 servers reaches the maximum amount of connections (defaults to 1000), this will dead-lock the Galera cluster synchronization. In order to resolve this, you should first increase database_advanced.max_connections, then stop packetfence-mariadb on all 3 servers, and follow the steps in the section 'Recovering from a split brain' of this document using anyone of the 3 servers as your source of truth.
```
extrafu: Should we briefly mention how to identify the source of truth? Or point the user in an other section from our doc that says how?

louismunro: using any

docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -902,3 +902,29 @@ Once this is done, restart pfmon on all nodes using:
...
+
+Investigating further
+^^^^^^^^^^^^^^^^^^^^^
+
+The limit of 1000 connections is fairly high already so if you reached the maximum number of connections, this might indicate an issue with your database cluster. If this issue happens often, you should monitor the active connections and their associated queries to find out what is using up your connections.
```
extrafu, line 921: Shouldn't we put a best practice script that monitors that too?

It would be nice to be warned before it explodes.

louismunro (approved):
LGTM with minor nitpicking.

docs/PacketFence_Clustering_Guide.asciidoc
julsemaan: Not sure its the job of our best practices script to monitor that, especially since they run at every hour and this can explode in much less than that.

We'd need something a bit more proactive than those scripts like a dedicated monit check IMO

docs/PacketFence_Clustering_Guide.asciidoc
julsemaan: Any of the 3 servers can be the source of truth given the transactions post dead-lock will be lost anyway. I guess it must not be clear how I wrote it. Will try to rephrase that

docs/PacketFence_Clustering_Guide.asciidoc
extrafu: Monit check sounds good to me!

julsemaan: Addressed in #2507

julsemaan:
Adjustments done, ready for another review

Issue #2378, opened 2017-06-02
Permission issue with /usr/local/pf/var/config_version 


julsemaan:
In the case where the first ever configreload hard of the install runs as root, then this file will be created by root and have him as a user. This means changes from the admin will fail because of that.

We should:
1. Pre-create this file during packaging
2. Run configreload as the pf user

julsemaan:
Actually not the configreload that is the issue, its the upgrade scripts that run as root as the configreload doesn't write that file.

I would suggest that any commit going through the configstore chowns the file to the pf user in order to ensure all configstore files are always owned by pf

julsemaan:
made fixpermissions fix this as a start: b064d6effb49282b0e2454b6db91e460b6adf711

dwlfrth:
upgrade scripts should run as pf

PR #2509, opened 2017-07-20
run the 6+ upgrade scripts as the PF user

julsemaan:
# Description
run the 6+ upgrade scripts as the PF user

# Impacts
Upgrade scripts

# Issue
part of the fix for #2378 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Run the 6.0+ upgrade scripts as the pf user to prevent permissions issues after running them

### DIFF
--- a/lib/pf/util.pm
+++ b/lib/pf/util.pm
@@ -22,7 +22,7 @@ use POSIX::2008;
 use Net::MAC::Vendor;
 use Net::SMTP;
 use File::Path qw(make_path remove_tree);
-use POSIX();
+use POSIX qw(setuid setgid);
 use File::Spec::Functions;
 use File::Slurp qw(read_dir);
 use List::MoreUtils qw(all any);
@@ -89,6 +89,7 @@ BEGIN {
         pf_chown
         user_chown
         ping
+        run_as_pf
     );
 }
 
@@ -1352,6 +1353,17 @@ sub ping {
     return $p->ping($host);
 }
 
+=head2 run_as_pf
+
+Sets the UID and GID of the currently running process to pf
+
+=cut
+
+sub run_as_pf {
+    my (undef, undef,$uid,$gid) = getpwnam('pf');
+    setgid($gid);
+    setuid($uid);
+}
 
 =back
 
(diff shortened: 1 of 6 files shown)


fdurand:
Agree to merge

lib/pf/util.pm
```diff
@@ -1352,6 +1353,17 @@ sub ping {
...
 }
 
+=head2 run_as_pf
+
+Sets the UID and GID of the currently running process to pf
+
+=cut
+
+sub run_as_pf {
+    my (undef, undef,$uid,$gid) = getpwnam('pf');
```
jrouzierinverse, line 1363: Check if current user if already pf before running setgid and setuid.

lib/pf/util.pm
```diff
@@ -1352,6 +1353,17 @@ sub ping {
...
+    setgid($gid);
```
jrouzierinverse: Check for errors and log them

lib/pf/util.pm
```diff
@@ -1352,6 +1353,17 @@ sub ping {
...
+    setuid($uid);
```
jrouzierinverse: Check for errors and log them

jrouzierinverse (changes requested):

julsemaan:
Should be all polished up, ready for another review

lib/pf/util.pm
```diff
@@ -1352,6 +1353,43 @@ sub ping {
...
+
+Sets the UID and GID of the currently running process to pf
+
+=cut
+
+sub run_as_pf {
+    # Early return if we're already running as pf
+    return $TRUE if(running_as eq "pf");
+
+    my (undef, undef,$uid,$gid) = getpwnam('pf');
```
jrouzierinverse: Just compare $< to $uid instead of using running_as since the environmental variables could be modified.

jrouzierinverse (changes requested):

julsemaan:
done

jrouzierinverse (approved):

PR #2510, opened 2017-07-20
run the configreload as the pf user

julsemaan:
# Description
Runs the configreload as the pf user

# Impacts
Command line config reload
It shouldn't cause any issue to do it as pf as the admin already runs as pf and reloads the config but unexpected stuff may occur so we will want to test this thoroughly.

# Code / PR Dependencies
#2509 because run_as_pf is defined in that PR

# Issue
Part of the fix for #2378

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Run the configreload as the pf user when done through pfcmd


fdurand:
agree to merge

jrouzierinverse (approved):
Looks good check will merge when other pull request is taken care of

PR #2528, opened 2017-08-01
fix/fingerbank-update

### LINKED ISSUE #2514: pfmon: fingerbank update job breaks DB cluster sync
**tldr;** Make `127.0.0.1` as the default DB host for:
- `pf.conf`
- `pfconfig.conf`
- `fingerbank.conf`

First occurence of the issue appeared when using Fingerbank DB update while having an active-active architecture with a galera cluster DB backend.

The Fingerbank DB update job is managed by `pfmon` and notify all members of the cluster to update themselves; including the DB...
When the issue occurs, `fingerbank.conf` was not configured to use `127.0.0.1` as the DB backend (for an unknown reason). `pfmon` was starting the update job on each of the cluster members and each of them was updating their local instance of the database, which was leading to split-brain and awful lock of tables followed by processes (connections) pilling-up due to none of them being able to finish (`query end` status due to incapacity of flushing the tx). Once that state is reached, the only way to go back is to force a new DB cluster.

Since there is apparently no way of restricting write access through local instance, avoiding mis-configuration by enforcing some defaults is probably our best way.

Action items:
- change the default DB from `localhost` to `127.0.0.1` in `pf.conf`, `pfconfig.conf`, `fingerbank.conf`;
- add a checkup script to make sure that latter settings is configured;
- create a monitoring script;
- modify fingerbank db update procedure to update only on the "master" instance when in cluster;

jrouzierinverse:
Description
===========
Fixes the fingerbank update process to avoid updating the mysql database from bring down a cluster

Impacts
=======
Fingerbank update process

Issue
=====
Fixes #2514

Delete branch after merge
=========================
YES

NEWS file entries
=================

Bug Fixes
------------
* Avoid breaking the DB cluster when updating the fingerbank data (#2514)

UPGRADE file entries
====================
* Avoid breaking the DB cluster when updating the fingerbank data (#2514)

### DIFF
--- a/lib/pf/pfmon/task/fingerbank_data_update.pm
+++ b/lib/pf/pfmon/task/fingerbank_data_update.pm
@@ -24,6 +24,8 @@ use pf::log;
 use Moose;
 extends qw(pf::pfmon::task);
 
+our $MYSQL_ACTION = 'update-mysql-db';
+
 =head2 run
 
 Update all the Fingerbank data from the API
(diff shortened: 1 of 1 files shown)


jrouzierinverse:
@fdurand said it is good I placed in maintenance

fdurand:
Tested on multiples setup and works as expected.

PR #2530, opened 2017-08-02
Add support for Roles and MAC AAA

garci66:
Roles can be returned (besides vlan) using Ruckus' Ruckus-User-Group attribute and support basic SNMP traps

# Description
Allows Ruckus devices (Both ZoneDirector and SmartZone) controllers to return a role together with the vlan for the access-accept. It allows the PF defined roles to be matched to ZD/SZ roles (user groups) either with the same names or different names by using the switch object's role-mapping table.

# Impacts
RADIUS authorize workflow for ruckus controllers

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Add support for roles (RBAC) for Ruckus WLAN controllers


lib/pf/Switch/Ruckus.pm
```diff
@@ -113,12 +115,18 @@ sub parseTrap {
     my $trapHashRef;
     my $logger = $self->logger;
 
-    $logger->debug("trap currently not handled.  TrapString was: $trapString");
-    $trapHashRef->{'trapType'} = 'unknown';
-
+    # Handle WIPS Trap
+    if ( $trapString =~ /\.1\.3\.6\.1\.4\.1\.25053\.2\.2\.2\.20 = STRING: \"([a-f0-9]{2}:[a-f0-9]{2}:[a-f0-9]{2}:[a-f0-9]{2}:[a-f0-9]{2}:[a-f0-9]{2})/ ) {
```
fdurand, line 123: Useless, parsetrap is use for port sec or link up link down

lib/pf/Switch/Ruckus.pm
```diff
@@ -166,6 +174,20 @@ sub deauthTechniques {
...
+=item returnRoleAttribute
+
+What RADIUS Attribute (usually VSA) should the role returned into.
+
+=cut
+
+sub returnRoleAttribute {
+    my ($self) = @_;
+
+    return 'Ruckus-User-Groups';
```
fdurand, line 191: Is it supported from a specific version ?

garci66: The earliest I could find it in the guides was Release 9.8  for zone director (from early 2014). But the attribute itself is supported in Radius since release 9.1 which is the first release for any of these controllers. The attribute was just used for admin-user's group identification and not for wifi users, so I doubt it's presence would break anything.  On releases prior to 9.8 groups tied to users seem to be limited to only allowing attachment to specific SSIDs, while on 9.8 onwards they allow specifying rate-limits, vlan-overrides, device identification, etc.

lib/pf/Switch/Ruckus.pm
garci66: Fine to delete. I got it form an older bit of code for ruckus integration. Can be deleted. I can amend the commit if needed.

fdurand: wirelessIPS has been removed https://github.com/inverse-inc/packetfence/commit/80516cb3b75da06f1fa0e8f0475f9b02aa69392a . So go ahead and remove it.

fdurand:
It just miss a little bit of documentation in the Network Device Configuration guide.

garci66:
Will add docs there as well. Thanks.  Wasn't sure if they had to be
provided in the same PR or not.

extrafu:
@garci66 Can you please sign the CLA and also add a snippet of documentation? We'll merge the PR after.

fdurand:
Still missing the documentation

garci66:
Yes. I will doxument this this week as I've been putting a new system in
production.

fdurand:
Do we merge as is without the doc ? (have to fix the conflict)

julsemaan:
> Do we merge as is without the doc ? (have to fix the conflict)

Probably better to have this than not have it so I'd say lets get it in

garci66:
I have added PR #6141 with additional details on how the Roles can be used in Ruckus Legacy  (Unleashed and zone director). Zone director in particular has support for RBAC-based ACL/Rate limit assignment based on RADIUS roles. I will also add the documentation for SmartZone roles

PR #2545, opened 2017-08-10
feature/decrease-loading-speed

### LINKED ISSUE #2484: httpd.admin is too slow to start
On my system it consistently takes 45-50 seconds to start.
Those numbers seem to be consistent with what others are seeing.

Disabling switch modules preloading does not seem to help.

Considering that with the current configuration systemd will time it out after 60 seconds, this is cutting it awfully close...

jrouzierinverse:
Description
===========
Decreasing loading time

Impacts
=======
Moose meta data immutability

Issue
=====
Fixes #2484

Delete branch after merge
=========================
NO

NEWS file entries
=================

Enhancements
------------
* Decrease the loading of the admin


dwlfrth:
Looks good for now, we should simply add a way to skip it only in the admin (we want runtime performance for the portal)

jrouzierinverse:
Done!

PR #2547, opened 2017-08-11
Feature/admin role pf only

julsemaan:
# Description
Create an administrative role that has all the admin roles for PacketFence but doesn't have switch login rights when they are set via PacketFence

# Impacts
Admin roles

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Added administrative role for all PacketFence permissions excluding switch login


fdurand:
Good to me.
Merge !

PR #2555, opened 2017-08-16
feature/reset-search

### LINKED ISSUE #1581: Admin Searches: Should have option to reset the search
There should be an option to reset the search back to empty whether it is saved or not.

In: 
- RADIUS audit log
- Node search
- Users search

jrouzierinverse:
Description
===========
Reset search forms

Issue
=====
fixes #1581

Delete branch after merge
=========================
NO

NEWS file entries
=================

Enhancements
------------
* Allow the ability reset the search forms

### DIFF
--- a/html/pfappserver/root/admin/nodes.tt
+++ b/html/pfappserver/root/admin/nodes.tt
@@ -351,6 +351,7 @@
                         </td>
                         <td>
                           <div class="btn-group pull-right">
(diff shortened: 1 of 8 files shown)


jrouzierinverse:
When testing this you would need to run ```grunt js```

julsemaan (commented):
Just tested it and the button indeed resets the search but then it leaves the current view unchanged which means the view isn't consistent with the current search. I believe we should submit the new empty form so the view is consistent with the action that was taken

jrouzierinverse:
OK

jrouzierinverse:
Done

PR #2557, opened 2017-08-16
Fix/pfsso logging

### LINKED ISSUE #2553: pfsso: Logging is in default location (/var/log/messages)
We should be handling it and sending it to the packetfence.log instead

louismunro:
# Description
Allows pfsso to log to a separate file.
Adds a command line option to pfhttpd to set the process name sent to syslog (does not actually set the process name as seen from ps).

# Impacts
Adds a pfsso.log destination to rsyslog.
Changes the pfsso unitfile to add the -log-name option to the command.

# Issue
fixes #2553 

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes

* pfsso now logs to logs/pfsso.log (#2553)

### DIFF
--- a/conf/systemd/packetfence-pfsso.service
+++ b/conf/systemd/packetfence-pfsso.service
@@ -7,7 +7,7 @@ Before=packetfence-httpd.portal.service
 [Service]
 Type=notify
 LimitNOFILE=8192
-ExecStart=/usr/local/pf/bin/pfhttpd -conf /usr/local/pf/conf/caddy-services/pfsso.conf
+ExecStart=/usr/local/pf/bin/pfhttpd -conf /usr/local/pf/conf/caddy-services/pfsso.conf -log-name pfsso
 Restart=on-failure
 Slice=packetfence.slice
 
--- a/packetfence.rsyslog
+++ b/packetfence.rsyslog
@@ -136,3 +136,7 @@ if $programname == "pfdetect" then -/usr/local/pf/logs/pfdetect.log
 if $programname == "pfbandwidthd" then -/usr/local/pf/logs/pfbandwidthd.log
 & stop
 
+# PFSSO
+if $programname == "pfsso" then -/usr/local/pf/logs/pfsso.log
+& stop
+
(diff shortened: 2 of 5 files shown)


packetfence.rsyslog
```diff
@@ -136,3 +136,7 @@ if $programname == "pfdetect" then -/usr/local/pf/logs/pfdetect.log
 if $programname == "pfbandwidthd" then -/usr/local/pf/logs/pfbandwidthd.log
 & stop
 
+# PFSSO
+if $programname == "pfsso" then -/usr/local/pf/logs/pfsso.log
+& stop
+
```
julsemaan: Not directly related to this, but I think it would make sense to redirect all of pfhttpd logs into packetfence .log

packetfence.rsyslog
```diff
@@ -136,3 +136,7 @@ if $programname == "pfdetect" then -/usr/local/pf/logs/pfdetect.log
 if $programname == "pfbandwidthd" then -/usr/local/pf/logs/pfbandwidthd.log
 & stop
 
+# PFSSO
+if $programname == "pfsso" then -/usr/local/pf/logs/pfsso.log
```
julsemaan: I'd put it in packetfence.log personally

conf/systemd/packetfence-pfsso.service
```diff
@@ -7,7 +7,7 @@ Before=packetfence-httpd.portal.service
 [Service]
 Type=notify
 LimitNOFILE=8192
-ExecStart=/usr/local/pf/bin/pfhttpd -conf /usr/local/pf/conf/caddy-services/pfsso.conf
+ExecStart=/usr/local/pf/bin/pfhttpd -conf /usr/local/pf/conf/caddy-services/pfsso.conf -log-name pfsso
```
julsemaan, line 10: Might as well change http-dispatcher at the same time ?

louismunro: Should it also be sent to packetfence.log in your masterful opinion?

fdurand:
Just make sure that when you modify the systemd init script the debianize patch still work.

louismunro:
This should be ready to merge.
All concerns have been addressed as far as I can tell.
Please hit the big "MERGE" button.

PR #2560, opened 2017-08-17
feature/ldap-source-cache-by-rule

### LINKED ISSUE #2346: LDAP authentication source cache match key
The key used to cache a match when querying an LDAP authentication source should be review and adjusted to match the  configured rule it matches. The cache would then be on a "per rule" basis and would avoid keeping in cache a result that was computed for an old rule.

jrouzierinverse:
Description
===========
Cache ldap matches by source id, rule id, and filter id

Impacts
=======
Ldap sources when mathing rules

Issue
=====
Fixes #2346

Delete branch after merge
=========================
NO

NEWS file entries
=================

Enhancements
------------
* Make LDAP Source cache more granular

### DIFF
--- a/lib/pf/Authentication/Source/LDAPSource.pm
+++ b/lib/pf/Authentication/Source/LDAPSource.pm
@@ -215,32 +215,6 @@ sub _connect {
 }
 
 
-=head2 match
-
-    Overrided to add caching to avoid a hit to the database
-
-=cut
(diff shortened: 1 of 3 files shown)


jrouzierinverse:
Please review

louismunro:
LGTM, let's merge and test.

PR #2566, opened 2017-08-18
fix/readonly-pf-conf

### LINKED ISSUE #2464: read-only mode: Saving configuration parameters
Even by being in a read-only mode on a database point-of-view, it should be possible to save some configuration changes.

Some of them needs to be excluded from that logic since requires duplication in the database:
- switches
- roles
- violations

See #2458

jrouzierinverse:
Description
===========
Allow saving configuration files in readonly mode except violation, switches, roles

Issue
=====
Fixes #2464

Delete branch after merge
=========================
NO

NEWS file entries
=================

Enhancements
------------
* Allow saving configuration files in readonly mode except violation, switches, role

### DIFF
--- a/lib/pf/freeradius.pm
+++ b/lib/pf/freeradius.pm
@@ -161,7 +161,15 @@ Populates the radius_nas table with switches in switches.conf.
 # First, we aim at reduced complexity. I prefer to dump and reload than to deal with merging config vs db changes.
 sub freeradius_populate_nas_config {
     my $logger = get_logger();
-    return unless db_ping;
+    unless(db_ping()){
+        $logger->error("Can't connect to db");
+        return;
+    }
+
+    if (db_readonly_mode()) {
+        print STDERR "Cannot reload radius nas table when the database is in read only mode\n";
+        return;
+    }
     my ($switch_config,$timestamp) = @_;
     my %skip = (default => undef, '127.0.0.1' => undef );
     my $radiusSecret;
--- a/lib/pf/freeradius.pm
+++ b/lib/pf/freeradius.pm
@@ -167,7 +167,9 @@ sub freeradius_populate_nas_config {
     }
 
     if (db_readonly_mode()) {
-        print STDERR "Cannot reload radius nas table when the database is in read only mode\n";
+        my $msg = "Cannot reload the RADIUS clients table when the database is in read only mode\n";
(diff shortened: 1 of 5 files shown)


jrouzierinverse:
Please review

julsemaan:
@lzammit will test

lib/pf/freeradius.pm
```diff
@@ -161,7 +161,15 @@ Populates the radius_nas table with switches in switches.conf.
...
 sub freeradius_populate_nas_config {
     my $logger = get_logger();
-    return unless db_ping;
+    unless(db_ping()){
+        $logger->error("Can't connect to db");
+        return;
+    }
+
+    if (db_readonly_mode()) {
+        print STDERR "Cannot reload radius nas table when the database is in read only mode\n";
```
julsemaan: Lets just say 'the RADIUS clients table' instead of 'radius nas table' so that its «Zammit readable»

julsemaan: I'd also add a logging statement so that if we go back in the logs for support, we could know that their reload didn't work in case the person who did the reload just ignored the warning and is now complaining about things not working

lib/pf/nodecategory.pm
```diff
@@ -106,6 +106,17 @@ It will simply do an upsert of all the roles in the configuration
...
     my ($config) = @_;
+    my $logger = get_logger;
+
+    unless(db_ping()){
+        $logger->error("Can't connect to db");
+        return;
+    }
+
+    if (db_readonly_mode()) {
+        print STDERR "Cannot reload roles when the database is in read only mode\n";
```
julsemaan: same about logging statement

lib/pf/violation_config.pm
```diff
@@ -42,6 +42,12 @@ sub loadViolationsIntoDb {
         $logger->error("Can't connect to db");
         return;
     }
+
+    if (db_readonly_mode()) {
+        print STDERR "Cannot reload violations when the database is in read only mode\n";
```
julsemaan: same about logging statement

lib/pf/freeradius.pm
jrouzierinverse: OK

jrouzierinverse:
Done

PR #2572, opened 2017-08-22
fix/duplicate-locationlog

### LINKED ISSUE #2527: duplicate locationlog entries
The battle of the duplicate locationlog entries needs to end.

Here are some solutions.

1. Treat locationlog entries like iplog entries. 
   * Where there can only be only one locationlog entry.
   * The primary key is the mac address.
   * A trigger moves the older locationlog entries to a history table.

2. Have a pfmon job cleanup stale locationlog entries. (This would be less than ideal)

3. Fix the root cause of the duplicate entries.

jrouzierinverse:
Description
===========
Additional fixes to node and location log logic to reduce the chances of duplicate locationlog

Impacts
=======
locationlog and node

Issue
=====
fixes #2527

Delete branch after merge
=========================
NO

NEWS file entries
=================

* Fix voip logic to reduce the chance of duplicate locationlog entries (#2527)

### DIFF
--- a/db/upgrade-X.X.X-X.Y.Z.sql
+++ b/db/upgrade-X.X.X-X.Y.Z.sql
@@ -46,4 +46,11 @@ END
 DELIMITER ;
 call ValidateVersion;
 
+
+--
+-- Normalize node.voip column
+--
+
+UPDATE node SET voip = 'no' WHERE voip != 'yes';
+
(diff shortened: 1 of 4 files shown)


julsemaan:
Looks good to me, I know @fdurand initially reported that issue, so if he says its good, we shall merge

fdurand (approved):

PR #2577, opened 2017-08-25
Few fix for inline

fdurand:
# Description
Fix incorrect inline sub typedetection.
Fix update ipset with the old ip address

# Impacts
Inline flow

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Fix incorrect inline sub type detection
* Fix ipset update with the incorrect ip address (the old one)


dwlfrth (approved):
LGTM. Exactly what we discussed

PR #2584, opened 2017-08-28
update golang dependencies for fixed RADIUS lib

### LINKED ISSUE #2561: pfsso times out on Accounting read
The library we are using layeh.com/radius has an issue where if its not able to authenticate the packet received for some reason (it looks like it has a bug about it too), it will continue reading data from the wire although the packet has been completely read. This generates timeout errors in the log although the SSO actually completed fine.

I suggest:
1. We fork the lib and fix it dirty for now by not looking at the message authenticator in the reply
1. Find out what is going on with the authentication method of the lib
1. Provide a fix to layeh.com/radius
1. Reintegrate the official lib once our fix is accepted

One other issue is that the lib has gone through a rewrite and its API changed a lot so having to maintain a fork might be necessary for now

julsemaan:
# Description
Dirty fix for #2561 where layeh.com/radius never considers Accounting packets as valid

# Impacts
pfsso

# Issue
fixes #2561 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed issue were pfsso would timeout reading accounting replies from the firewalls

### DIFF
--- a/go/firewallsso/base.go
+++ b/go/firewallsso/base.go
@@ -13,7 +13,7 @@ import (
 	"github.com/inverse-inc/packetfence/go/log"
 	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
 	"github.com/inverse-inc/packetfence/go/sharedutils"
-	"layeh.com/radius"
+	"github.com/julsemaan/radius"
 )
 
 // Basic interface that all FirewallSSO must implement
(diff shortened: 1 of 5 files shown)


louismunro (approved):
Looks good enough for now, but we should keep in mind that we need to upgrade to the latest version of the lib.

PR #2591, opened 2017-08-30
Fixed useless check for SNMP return on Extricom controllers.

### LINKED ISSUE #2590: Extricom controllers always fail SNMP connections
When connecting to an Extricom controller using SNMP the return value is always false, even when the connection is successfully established.

louismunro:
# Description
Extricom controllers never return the right value when connecting using SNMP.
This removes a check that will always fail.

# Impacts
Any SNMP calls to the controller should now work.

# Issue
fixes #2590

# Delete branch after merge
YES 

# NEWS file entries

## Bug Fixes
* Fixed SNMP connection issues on Extricom controllers

### DIFF
--- a/lib/pf/Switch/Extricom.pm
+++ b/lib/pf/Switch/Extricom.pm
@@ -169,6 +169,7 @@ sub connectWrite {
                 ]
             );
        }
+       # We no longer check the $result here since it is always false even when the call succeeds.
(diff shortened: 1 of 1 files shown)


extrafu:
Perhaps add a comment in the code why we don't check $result?

PR #2594, opened 2017-08-31
Fix/galera readonly strategy

### LINKED ISSUE #2593: mariadb clustering: RO mode
As per @julsemaan:
We currently have issues where a small percentage of reconnects between Galera cluster instances fails and has a member excluded or the database mistakenly put into read-only mode.

We are currently using a home-made script to set the read only flag which requires to temporarily disable wsrep and then turn it back on right after.

I'm suspecting that when the cluster is unstable for a very short amount of time, this can cause a member to change its view of the cluster state mistakenly and then the only way to recover is to do a full DB resync.

The more I look into this, the more I realize we don't need to put the DB in read only more and could simply look if wsrep_ready is on or off since it represents how galera sees its state in the cluster.

This will obviously require testing to confirm it doesn't have side effects but it all seems possible and would allow us to rely on a galera cluster stack that is closer to the one that is official (the read-only script we use is unofficial and contains adjustments I did which may not be compatible with some galera internal logic)

julsemaan:
# Description
Use the wsrep_ready state of the DB as a read only indicator
Stop putting the DB in read only on quorum + primary loss of MariaDB and trust wsrep_ready instead

# Impacts
The galera cluster stack
## This will require a lot of testing before its put in official production, there shouldn't be anything that goes wrong but we should test it thoroughly and BETA try it in a couple of production environments before merging it

# Issue
fixes #2593 

# Delete branch after merge
NO

# NEWS file entries
## Bug Fixes
* Trust the wsrep_ready flag of MariaDB Galera cluster for read only detection as putting the DB in read-only can result in occasional de-synchronization between members.

# UPGRADE file entries

Ensure you merge changes in the galera section of conf/mariadb/mariadb.conf.tt.rpmnew into conf/mariadb/mariadb.conf.tt

### DIFF
--- a/lib/pf/db.pm
+++ b/lib/pf/db.pm
@@ -463,11 +463,10 @@ sub db_readonly_mode {
     my $row = $sth->fetch;
     $sth->finish;
     my $readonly = $row->[0];
-
-    my $wsrep_healthy = db_wsrep_healthy();
-
-    # If the read_only flag is set or wsrep isn't ready, we are in read only
-    return $readonly || !$wsrep_healthy;
+    # If readonly no need to check wsrep health
+    return 1 if $readonly;
+    # If wsrep is not healthly then it is in readonly mode
+    return !db_wsrep_healthy();
 }
 
 =head2 db_wsrep_healthy
--- a/lib/pf/db.pm
+++ b/lib/pf/db.pm
@@ -40,6 +40,7 @@ use constant PREPARED_VAR      => '_db_prepared'; # prepare flag with <modulenam
 use constant PREPARE_PF_PREFIX => 'pf::';         # prefix to access exported _prepare(d) variables
 
 our $MYSQL_READONLY_ERROR = 1290;
+our $WSREP_NOT_READY_ERROR = 1047;
 
 our ( $DBH, $LAST_CONNECT, $DB_Config, $NO_DIE_ON_DBH_ERROR );
 
@@ -151,7 +152,7 @@ db_handle_error
 
 sub db_handle_error {
     my ($err) = @_;
-    if ($err == $MYSQL_READONLY_ERROR) {
+    if ($err == $MYSQL_READONLY_ERROR || $err == $WSREP_NOT_READY_ERROR) {
         db_set_readonly_mode(1);
     }
     return ;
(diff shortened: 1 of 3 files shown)


julsemaan:
Just thought about a case we'll need to handle where when we're doing an update and putting a single node in production in standalone (first step of golive) wsrep_ready will be off because galera will be disabled. Need to think about a way around that

julsemaan:
Regarding the point above, we would just have to check for wsrep_provider_name and check only the wsrep_ready if its there

lib/pf/db.pm
```diff
@@ -452,11 +456,54 @@ sub db_readonly_mode {
...
+    if(defined($row) && $row->[1] ne "") {
+        $logger->debug("There is a wsrep provider, checking the wsrep_ready flag");
+        # check if the wsrep_ready status is ON
+        $sth = $dbh->prepare_cached('show status like "wsrep_ready";');
+        return 0 unless $sth->execute;
+        $row = $sth->fetch;
+        $sth->finish;
+        # If there is no wsrep_ready row, then we're not in read only because we don't use wsrep
+        # If its there and not set to ON, then we're in read only
+        my $wsrep_ready = (defined($row) && $row->[1] eq "ON");
```
jrouzierinverse: There is no return

jrouzierinverse (changes requested):

jrouzierinverse (approved):

lib/pf/db.pm
```diff
@@ -452,11 +456,54 @@ sub db_readonly_mode {
...
     };
     return 0 unless $dbh;
+
+    # check if the read_only flag is set
     my $sth = $dbh->prepare_cached('SELECT @@global.read_only;');
     return 0 unless $sth->execute;
     my $row = $sth->fetch;
     $sth->finish;
-    return $row->[0];
+    my $readonly = $row->[0];
```
jrouzierinverse, line 465: Return early if readonly if true

sbin/pf-mariadb
```diff
@@ -173,16 +175,22 @@ Test the connection to the database for a specific host using the replication cr
...
-    if ($mydbh) {
-        return 1;
-    }
-    else {
-        return 0;
-    }
+    eval {
+        local $SIG{ALRM} = sub {die "Timeout connecting to MySQL for host $host\n";};
+        alarm($MYSQL_CONNECT_TIMEOUT);
+        my $mydbh = DBI->connect( "dbi:mysql:dbname=mysql;host=$host;port=3306",
```
jrouzierinverse: There is a built in option to do this mysql_connect_timeout.

http://search.cpan.org/~capttofu/DBD-mysql-4.023/lib/DBD/mysql.pm#Class_Methods

This should be more reliable since it is handled in mysql client instead using unix signals.

julsemaan: fixaide

jrouzierinverse:
It seems ready to merge. Will wait to merge until confirmed it works as advertised.

dwlfrth:
LGTM. Are we merging it and start test phase in preparation of 7.3 ?

louismunro:
Let's merge.
We need to test with it if we are to use it in prod.

julsemaan:
I'm with you on that @louismunro 

I've had success in lab testing that + @whitx also had some

@lzammit had issues setting it up but most likely because he restored the wrong innobackup

I'd say we merge !

PR #2598, opened 2017-08-31
fix/reuse-dot1x-status-page

### LINKED ISSUE #2542: Reuse dot1x credentials breaks status page
Having Reuse dot1x credentials enabled on the portal profile where the status page is being hit makes the login completely fail without any warnings.

IMO, we shouldn't honor that anymore in the catalyst controllers or at least make sure it won't be the case for the status page

jrouzierinverse:
Description
===========
Fix issue with combining reuse dot1x and logging into the status page

Impacts
=======
Logging into status when reuse dot1x is enabled

Issue
=====
Fixes #2542

Delete branch after merge
=========================
NO

Bug Fixes
------------
* Logging into the status page when reuse dot1x is enabled is no longer broken (#2542)

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Authenticate.pm
@@ -113,35 +113,23 @@ sub authenticationLogin : Private {
         $username = $stripped_username;
     }
 
-    if(isenabled($profile->reuseDot1xCredentials)) {
-        my $mac       = $portalSession->clientMac;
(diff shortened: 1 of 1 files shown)


julsemaan (commented):
I'm not sure why you're deleting the whole email + sponsor activation routes + modules

To my knowledge, they're still in use and haven't been replaced by anything in the DynamicRouting part of the portal

jrouzierinverse:
Yes you are right I will fix that

jrouzierinverse:
Please review

PR #2602, opened 2017-09-04
fix/use-server-time

### LINKED ISSUE #2559: Refer last 30 minutes on the current time of the packetfence server
Searching for the last 30 minutes entries use the browser time and not the packetfence's server time.

jrouzierinverse:
Description
===========

Impacts
=======
Admin

Issue
=====
fixes #2559

Delete branch after merge
=========================
NO

NEWS file entries
=================
Bug Fixes
------------

* Using the system when instead of browser time when using preset time values (#2559)

UPGRADE file entries
====================


dwlfrth:
Is this ready to be merged ?
If not, can we make it ready (and merge) before end of day.

jrouzierinverse:
Yes it is ready who is going to review it?

html/pfappserver/lib/pfappserver/PacketFence/Controller/Admin.pm
```diff
@@ -350,6 +352,34 @@ sub configuration :Chained('object') :PathPart('configuration') :Args(0) {
 
 }
 
+=head2 time_offset
+
+time_offset
```
dwlfrth: That is a name, not a description

dwlfrth (changes requested):

html/pfappserver/lib/pfappserver/PacketFence/Controller/Admin.pm
louismunro: Agreed, we should have at least one sentence to describe what the method is supposed to do and what the params are.

jrouzierinverse:
Done

louismunro:
Looks good then.

PR #2614, opened 2017-09-11
fix/ldap-timeouts

### LINKED ISSUE #2613: LDAP Source: authenticate+authorize never ending connection
Right now in LDAP we have a connection timeout which is used passed onto IO::Socket according to Net::LDAP perldoc.

I have seen an issue where the pfqueue workers are not recovering from a temporary LDAP connection issue and are stuck reading from the LDAP TCP socket (strace reports its reading from the file descriptor associated to that TCP connection). That connection is maintaining as ESTABLISHED but seems none of the two parties in the connection are saying a thing.

We should:

- [ ] See if we can find anything about the underlying bug which is the never ending connection
- [ ] Have a global timeout when doing any LDAP query in Authentication::Source::LDAPSource to have a full safety net in case anything goes wrong during that process

jrouzierinverse:
Description
===========

Set read and write timeouts on the LDAP socket

Impacts
=======
LDAP connections

Issue
=====
Fixes #2613

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------

* Set read and write timeouts for LDAP connections (#2613)


julsemaan:
IMO, I'd see those timeouts a bit lower probably around 30 seconds and even that looks enormous.

Or, we could honor the connection timeout in those maybe ?

@louismunro being the initiator of the connection timeout parameter, I'm sure he'll have an opinion on that

jrouzierinverse:
Lowering the timeouts make sense however having them as the same value can hide problems.

Keeping them as different values helps deal with different scenarios.

Connection timeout
  - Should timeout fast 
  - It is just the TCP handshake
  - If we cannot connect to a server quickly then there is a problem.
  - Better to know it early.

Sending a request
  - Should timeout reasonably quick 
  - Provide enough time for the server to receive the request

Receiving a request 
  - Should be the longest timeout since it does the most work
  - Provide enough time for the server for the follow
    - deserialize the request
    - perform the lookup 
    - serialize the response
    - send it back.

julsemaan:
I'd gauge the following:
- connection: 5 seconds
- sending: 5 seconds
- receiving: 10 seconds

With those being configurable via the source for advanced zammit editing

jrouzierinverse:
Done

fdurand:
Just tested, no issue to report.

extrafu:
Please merge.

html/pfappserver/lib/pfappserver/Form/Config/Source/LDAP.pm
```diff
@@ -44,6 +46,24 @@ has_field 'connection_timeout' =>
     },
     default => $META->get_attribute('connection_timeout')->default,
   );
+has_field 'write_timeout' =>
```
julsemaan: Add help for this field

html/pfappserver/lib/pfappserver/Form/Config/Source/LDAP.pm
```diff
@@ -44,6 +46,24 @@ has_field 'connection_timeout' =>
...
+  (
+    type         => 'PosInteger',
+    label        => 'Socket write timeout',
+    element_attr => {
+        'placeholder' => $META->get_attribute('write_timeout')->default
+    },
+    default => $META->get_attribute('write_timeout')->default,
+  );
+has_field 'read_timeout' =>
```
julsemaan: Add help for this field

julsemaan (commented):
Just needs contextual help for the new fields and its ready for merge

jrouzierinverse:
Done

Issue #2627, opened 2017-09-12
question: enabled vs disabled CSP headers in the 7.3


julsemaan:
Seems the CSP headers brought more breaking changes than expected (at least in my tests)

Lets discuss in the next meeting whether we should ship them disabled for 7.3 and aim to enable them by default in 7.4 instead as I'm pretty sure we'll find CSP related issues post 7.3 release considering the amount of pages we have in the admin

julsemaan:
we said:

- [ ] CSP disabled for 7.3 then reenabled in devel right after
- [ ] Split portal + admin CSP parameter
- [ ] Add way to whitelist resources

dwlfrth:
Disable right before building process

dwlfrth:
Disable it by default for 7.3. If we want to test CSP header in 7.3 release cycle, enable it manually

PR #2638, opened 2017-09-14
feature/split-csp-header-toggle

jrouzierinverse:
Description
===========
Allow the admin and the portal enable/disable the CSP headers seperatly

Impacts
=======
Admin / Captive portal

Issue
=====
Related to #2627

Delete branch after merge
=========================
NO

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1049,11 +1049,18 @@ Amount of requests on invalid URLs after which the rate limiting will kick in fo
 Requires to restart haproxy in order to apply the change.
 EOT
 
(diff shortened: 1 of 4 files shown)


cgx:
I would vote to merge this pull request for 7.3. It totally makes sense to be able to enforce CSP for the Web admin but not for the captive portal.

julsemaan:
@cgx We decided to disable it in 7.3 since we felt it needed more testing.

And I think its a bit too late to add stuff in 7.3 as all of us have completed their testing and there isn't going to be a full re-test before release

PR #2645, opened 2017-09-18
fix/rework-realm-matching

### LINKED ISSUE #2644: Adding a EAPTLS source breaks admin login
If you add an EAPTLS source you are not able to login on the admin interface via the source local. No logs, no nothing. Only error message on the admin interface "Error! Unexpected error. See server-side logs for details." No logs anywhere about that.

Example of the source which is causing issue:

[tt]
set_access_level_action=
description=test
type=EAPTLS

[tt rule base]
action0=set_role=default
match=all
class=authentication
action1=set_access_duration=12h
description=lul

jrouzierinverse:
Description
===========
Reworking on the logic of filtering sources based of a realm

Impacts
=======
Filtering sources by realm in authentication

Issue
=====
Fixes #2644

Delete branch after merge
=========================
NO


dwlfrth:
@whitx test with dot1x

whitx:
peap and eaptls works.

jrouzierinverse:
Then push the green button

PR #2649, opened 2017-09-19
fix/auth-log-email

### LINKED ISSUE #2648: Auth log is not set to completed after email registration

jrouzierinverse:
Description
===========
Auth log is not set to completed after email registration

Impacts
=======
Email registration workflow

Issue
=====
fixes #2648

NEWS file entries
=================

Bug Fixes
------------

* Auth log is not set to completed after email registration (#2648)

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
@@ -79,6 +79,7 @@ sub code : Path : Args(2) {
             get_logger->info("Extending duration to $unregdate");
             node_modify($node_mac, unregdate => $unregdate);
             pf::activation::set_status_verified($GUEST_ACTIVATION, $code);
+            pf::auth_log::record_completed_guest($activation_record->{source_id}, $node_mac, $pf::auth_log::COMPLETED);
             $c->stash(
                 title => "Access granted",
                 template => "activation/email.html",
@@ -217,6 +218,7 @@ sub doSponsorRegistration : Private {
         }
 
         pf::activation::set_status_verified($SPONSOR_ACTIVATION, $code);
+        pf::auth_log::record_completed_guest($activation_record->{source_id}, $node_mac, $pf::auth_log::COMPLETED);
 
         # send to a success page
         $c->stash(
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
@@ -139,18 +139,13 @@ sub doSponsorRegistration : Private {
         'username'   => $pid,
         'user_email' => $email
     };
-
+    my $sponsor_source_id = $activation_record->{source_id};
     my $profile = $c->profile;
-    my $sponsor_type =
-      pf::Authentication::Source::SponsorEmailSource->getDefaultOfType;
-    my $source = $profile->getSourceByType($sponsor_type);
-
+    my $source = getAuthenticationSource($sponsor_source_id);
     if ($source) {
-
         # if we have a username in session it means user has already authenticated
         # so we go ahead and allow the guest in
         if ( !defined( $c->user_session->{"username"} ) ) {
-
             # User is not logged and didn't provide username or password: show login form
             if (!(  $request->param("username") && $request->param("password")
                 )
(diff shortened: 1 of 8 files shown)


louismunro (approved):
Tested at a client site.
Does what it says.

louismunro:
@dwlfrth can we merge (even if a bit late)?
This fixes an actual bug.
As opposed to, say CSP Headers...

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
```diff
@@ -79,6 +80,9 @@ sub code : Path : Args(2) {
             get_logger->info("Extending duration to $unregdate");
             node_modify($node_mac, unregdate => $unregdate);
             pf::activation::set_status_verified($GUEST_ACTIVATION, $code);
+            my $sponsor_type = pf::Authentication::Source::EmailSource->getDefaultOfType;
```
julsemaan: We can't assume that anymore, we should find out what is the actual source that was used during the sponsor registration

julsemaan (changes requested):

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
julsemaan: s/sponsor/email/ also

jrouzierinverse:
The real fix would require us to store the source in the activation record.
This would require a changes to the DB and other parts of the code.

We should wait for the next release for the proper fix.

Also I would prefer migrate this out of the controller and to a DynamicRouting::Module

julsemaan:
@jrouzierinverse I partly agree with you for bringing this out of the controllers as one of the use cases that needs to be handled is the fact that the link for activation may be opened from another device than the one that is performing the registration. In that case, restoring the proper context in the DynamicRouting could be tricky as the state of the flow is stored in the session tied to the MAC address (and the device opening the link obviously doesn't share the MAC). That's why it was left in the controllers as the DynamicRouting assumes a single device is performing a chain of even which isn't the case for the email activation. I'm obviously opened to bringing it in DynamicRouting but that case will need to be properly handled.

And +1 for waiting for next release

jrouzierinverse:
Should we wait till the next release for this one as the pull request needs to be adjusted.

julsemaan:
@jrouzierinverse if you have time, IMO it would be good to adjust it since its a bug and not an enhancement.

jrouzierinverse:
I will work on it after the dal is in

jrouzierinverse:
Reworked please review

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
```diff
@@ -217,6 +218,7 @@ sub doSponsorRegistration : Private {
         }
 
         pf::activation::set_status_verified($SPONSOR_ACTIVATION, $code);
+        pf::auth_log::record_completed_guest($activation_record->{source_id}, $node_mac, $pf::auth_log::COMPLETED);
```
julsemaan: Sponsor was already working to my knowledge since the record completed is made in the client portal directly

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
```diff
@@ -179,8 +179,8 @@ sub do_sponsor_registration {
     }
 
     $info{'sponsor'} = $sponsor;
-
     $info{'subject'} = $self->app->i18n_format("%s: Guest access request", $Config{'general'}{'domain'});
+    $info{'source_id'} = $source->id;
```
julsemaan, line 183: No need for that I believe, for the reason outlined above

html/captive-portal/lib/captiveportal/PacketFence/Controller/Activate/Email.pm
jrouzierinverse: I will verify if the sponsor is working as you say.
Then I will just remove that dead code to avoid future confusion.

julsemaan:
@jrouzierinverse 

This has pending comments + requires a rebase

extrafu:
@jrouzierinverse Please rebase.

jrouzierinverse:
Done

extrafu:
@julsemaan Merge if alright.

PR #2659, opened 2017-09-28
feature/db-max-statement-time

### LINKED ISSUE #2630: Admin: Prevent DB denial of service via queries
As most of the @inverse-inc/packetfence team knows, the admin is home of a lot of CPU and I/O intensive queries that can bring a setup down if the tables grow a lot or have dirty data in them (like Zayme's panties).

An admin requesting to view a report on accounting or more recently searching nodes with doubled locationlog entries results in a long running query. If that admin clicks multiple times on the report or search because its taking too long, then he effectively brings the database down by sending that query multiple times.

We should:

- [ ] Have a timeout (even if its high like 5 minutes) on the admin interface requests so that those long running queries get eventually killed without the need to connect to MySQL and kill the queries manually
- [ ] Establish the acceptable amount of rows in the different tables for proper operation (locationlog, radacct, radacct_log, etc) and warn if it goes beyond that threshold. (*decide if built-into PF or via best practices scripts*)
- [ ] Track **all** the SQL related issues we have in a best-practices script (voip flag not being yes/no, radacct tables containing entries that haven't been closed, etc) so proper actions can be taken
- [ ] Track the current queries running in MySQL and warn when queries have been running longer than for X amount of time

Not sure if we should aim to have this in 7.3 so putting +1 for now and we'll discuss during the next team meeting

jrouzierinverse:
Description
===========
Set a timeout for database queries for the admin

Impacts
=======
The Admin Portal and database connections

Issue
=====
fixes #2630

Delete branch after merge
=========================
NO

NEWS file entries
=================

Enhancements
------------
* Set a timeout for database queries for the admin to avoid long running queries slowing the system.

### DIFF
--- a/lib/pf/WebAPI/InitHandler.pm
+++ b/lib/pf/WebAPI/InitHandler.pm
@@ -60,8 +60,9 @@ sub post_config {
     my ($class, $conf_pool, $log_pool, $temp_pool, $s) = @_;
     pf::StatsD->closeStatsd;
     db_disconnect();
-    preloadSwitches();
+    $class->preloadSwitches();
     pf::CHI->clear_memoized_cache_objects;
+    $class->post_config_hook();
     return Apache2::Const::OK;
 }
 
@@ -75,6 +76,14 @@ sub preloadSwitches {
     my ($class) = @_;
     pf::SwitchFactory->preloadConfiguredModules();
 }
+
+=head2 post_config_hook
+
+post config hook allow child class to add additional actions
+
+=cut
+
+sub post_config_hook { }
 
 =head1 AUTHOR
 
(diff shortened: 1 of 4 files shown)


julsemaan:
Code looks good.

Just before the merge, we'll give it a test in the admin with a low timeout and then its ready for merging

jrouzierinverse:
Should this be a configurable value?

julsemaan:
Don't really think so, it could lead a user to set this to a value that would kill his server.

That would be a case of "we know better than the user" IMO

jrouzierinverse:
Anything need on this one?

julsemaan:
Nope, good enough, merging

PR #2662, opened 2017-10-02
Enable firewall sso on accounting packet

fdurand:
# Description
Allow to update the firewall sso based on radius accounting packet.

# Impacts
RADIUS accounting workflow

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Update the Firewall SSO on accounting packet

### DIFF
--- a/docs/PacketFence_Administration_Guide.asciidoc
+++ b/docs/PacketFence_Administration_Guide.asciidoc
@@ -2856,12 +2856,13 @@ If, for some reason the portal is modified with content that needs to be accesse
 Firewall SSO Integration
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-PacketFence is able to update some Firewalls based on device informations, like the IP address, the username connected on it. (Look for Integration guides to see how you can configure your Firewall with PacketFence)
-By default PacketFence use the dhcp traffic to trigger an update on the firewall but it's also possible to do it with the RADIUS Accounting traffic.
+PacketFence is able to update some firewall based on device informations, like the IP address, the username connected on it. (Look for Integration guides to see how you can configure your Firewall with PacketFence)
+By default PacketFence uses the DHCP traffic to trigger an update on the firewall but it's also possible to do it with the RADIUS Accounting traffic.
 
 In order to manage the way you want to update the Firewall, go in Configuration -> System Config -> Main Configuration -> Advanced,  then there is two choices,
 Trigger Single-Sign-On on dhcp and Trigger Single-Sign-On on accounting.
 
+You can use both methods at the same time but this will result in duplicate SSO requests if you receive the DHCP and accounting of the same device which can cause unexpected load on your firewall.
 
 Suricata example:
 -----------------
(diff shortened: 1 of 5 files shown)


lib/pf/api.pm
```diff
@@ -1360,9 +1361,50 @@ sub handle_accounting_metadata : Public {
...
+=cut
+
+sub firewallsso_accounting : Public {
+    my ($class, %RAD_REQUEST) = @_;
+    my $logger = pf::log::get_logger();
+    if ($RAD_REQUEST{'Calling-Station-Id'} && $RAD_REQUEST{'Framed-IP-Address'} && pf::util::isenabled($pf::config::Config{advanced}{sso_on_accounting})) {
+        my $mac = pf::util::clean_mac($RAD_REQUEST{'Calling-Station-Id'});
+        my $node = pf::node::node_attributes($mac);
+        my $ip = $RAD_REQUEST{'Framed-IP-Address'};
+        if($ip){
```
julsemaan: Seems this can never be false as you're already checking it in the top condition

IMO you should split the top condition to output a meaningful warn message when not finding an IP or MAC so we can know what is up

lib/pf/api.pm
```diff
@@ -1360,9 +1361,50 @@ sub handle_accounting_metadata : Public {
...
+            if ($node->{status} eq $pf::node::STATUS_REGISTERED) {
+                $firewallsso_method = "Update";
+                if ($node->{unregdate} ne '0000-00-00 00:00:00') {
+                    my $time = DateTime::Format::MySQL->parse_datetime($node->{unregdate});
+                    $time->set_time_zone("local");
+                    my $now = DateTime->now(time_zone => "local");
+                    $timeout = $time->epoch - $now->epoch;
+                }
+            }
+
```
julsemaan: You don't seem to be checking for MAC change by an IP like in DHCP where a device would take the IP of another device because the first became inactive

julsemaan:
Also requires documentation in the admin guide

julsemaan:
Other than documentation this is all good. IDK where we could put this in the docs since we don't have a dedicated section for SSO, so we might skip that.

What do the other @inverse-inc/packetfence guys think ?

docs/PacketFence_Administration_Guide.asciidoc
```diff
@@ -2856,12 +2856,13 @@ If, for some reason the portal is modified with content that needs to be accesse
 Firewall SSO Integration
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-PacketFence is able to update some Firewalls based on device informations, like the IP address, the username connected on it. (Look for Integration guides to see how you can configure your Firewall with PacketFence)
-By default PacketFence use the dhcp traffic to trigger an update on the firewall but it's also possible to do it with the RADIUS Accounting traffic.
+PacketFence is able to update some firewall based on device informations, like the IP address, the username connected on it. (Look for Integration guides to see how you can configure your Firewall with PacketFence)
```
julsemaan: Sorry my previous comments weren't clear.

its "some firewalls" and then later its "how you can configure your firewall with PacketFence"

PR #2666, opened 2017-10-04
Minor changes in the documentation

fdurand:
# Description
Reworked/Removed documentation

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Removed useless documentation


fdurand:
Don't we have to remove MySQL optimizations section ?

julsemaan:
@fdurand yes I think you're right since they are mostly changes to the config in that section although we could still leave the sections that talk about using the optimizer and the i/o

PR #2683, opened 2017-10-17
fix/authentication-rule-add-delete-button-2678

### LINKED ISSUE #2678: (+) and (-) can not be seen in new authentication rule is created
When screen resolution is less than 1423xXXX the two add and remove buttons are not visibles.
image

The problem can be solved by reducing the left margin or the size of the box on the right.

jrouzierinverse:
Description
===========
Moves the add delete buttons to the left instead of the right

Impacts
=======
The Admin GUI

Issue
=====
Fixes #2678

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------

* Move add delete buttons to the left to avoid the being cutoff (#2678)


fdurand:
Tested and works. Merge

PR #2687, opened 2017-10-18
feature/move-to-pf-dal

jrouzierinverse:
Description
===========
Move to use pf::dal and SQL::Abstract::More for database access

Impacts
=======
Database access and sql generation

NEW Package(s) required
=======================
A patched version of SQL::Abstract::More (I submitted the patch to the project but they have not updated CPAN as of yet)
perl-SQL-Abstract-Plugin-InsertMulti

Delete branch after merge
=========================
NO

TODO
=========================
 - [x] Move node search to the dal
 - [x] Have search and find use the same columns


julsemaan:
Seems the node search wasn't changed in this PR so it goes through DAL. That means it will need to be adjusted for a tenant manually ? Or was that just forgotten ?

julsemaan:
As jazzed, search should default to the same columns as find to ease the usage

jrouzierinverse:
Node search was missed

jrouzierinverse:
search uses the same columns as find now

julsemaan:
This is still missing node search via the DAL according to the todo of the PR

lib/pf/Report.pm
```diff
@@ -37,15 +38,13 @@ has 'person_fields', (is => 'rw', isa => 'ArrayRef[Str]');
...
-# empty since no queries are prepared upfront
-sub Report_db_prepare {}
-
 sub generate_sql_query {
     my ($self, %infos) = @_;
     my $logger = get_logger;
 
     my $sqla = SQL::Abstract::More->new();
     my $and = [];
+    $infos{search}{type} //= $self->base_conditions_operator;
```
julsemaan, line 47: Any specific reason for adding this? Doesn't seem related to DAL work

lib/pf/activation.pm
```diff
@@ -248,13 +236,11 @@ update the status of a given pending activation record
...
     my ($status, $rows) = pf::dal::activation->update_items(
-        {
-            -set => {
-                status => $new_status,,
-            },
-            -where => {
-                code_id => $code_id,
-            }
+        -set => {
+            status => $new_status,,
```
julsemaan: Seems like there a useless comma **#**:nail_care:

lib/pf/activation.pm
```diff
@@ -591,17 +571,15 @@ sub set_status_verified_by_mac {
...
-            -columns => [\1],
-            -limit => 1,
-        }
+        -where => {
+            mac => $mac,
+            type => $type,
+            status => $UNVERIFIED,
+            expiration => { ">=" => \['NOW()']},
+        },
+        -from => pf::dal::activation->find_from_tables(),
```
julsemaan: Not 100% sure, but I believe we said this shouldn't be used anymore because it should already do this behavior by default

lib/pf/violation.pm
```diff
@@ -178,16 +179,14 @@ sub violation_grace {
     my ( $mac, $vid ) = @_;
 #        qq [ select unix_timestamp(start_date)+grace_period-unix_timestamp(now()) from violation v left join class c on v.vid=c.vid where mac=? and v.vid=? and status="closed" order by start_date desc ]);
```
julsemaan: not really useful anymore

lib/pf/violation.pm
```diff
@@ -197,11 +196,9 @@ sub violation_count {
     my ($mac) = @_;
 #        qq [ select count(*) from violation where mac=? and status="open" ]);
```
julsemaan: Not really useful anymore

lib/pf/violation.pm
```diff
@@ -212,15 +209,13 @@ sub violation_count_reevaluate_access {
     my ($mac) = @_;
 #       qq [ select count(*) from violation, action where violation.vid=action.vid and action.action='reevaluate_access' and mac=? and status="open" ]);
```
julsemaan: Not really useful anymore

lib/pf/violation.pm
```diff
@@ -230,11 +225,9 @@ sub violation_count_vid {
     my ( $mac, $vid ) = @_;
 #        qq [ select count(*) from violation where mac=? and vid=? ]);
```
julsemaan: Not really useful anymore

lib/pf/violation.pm
```diff
@@ -245,12 +238,10 @@ sub violation_count_open_vid {
     my ( $mac, $vid ) = @_;
 #        qq [ select count(*) from violation where mac=? and vid=? and status="open" ]);
```
julsemaan: Not really useful anymore

lib/pf/violation.pm
```diff
@@ -267,6 +258,7 @@ sub violation_exist {
         },
         -columns => [qw(id mac vid start_date release_date status ticket_ref notes)],
         -limit => 1,
+        -from => pf::dal::violation->table,
```
julsemaan: useful ?

lib/pf/violation.pm
```diff
@@ -278,6 +270,7 @@ sub violation_exist_id {
...
```
julsemaan: useful ?

lib/pf/violation.pm
```diff
@@ -290,6 +283,7 @@ sub violation_exist_open {
             status => 'open',
         },
         -columns => [qw(id mac vid start_date release_date status ticket_ref notes)],
+        -from => pf::dal::violation->table,
```
julsemaan: useful ?

lib/pf/violation.pm
```diff
@@ -330,7 +324,8 @@ sub violation_view_open {
             mac => $mac,
         },
         -columns => [qw(id mac vid start_date release_date status ticket_ref notes)],
-        -order_by => '-start_date',
+        -order_by => { -desc => 'start_date' },
+        -from => pf::dal::violation->table,
```
julsemaan: useful ?

lib/pf/violation.pm
```diff
@@ -814,6 +805,7 @@ sub violation_exist_acct {
                 "<=" => \'NOW()',
             },
         },
+        -from => pf::dal::violation->table,
```
julsemaan: useful ?

lib/pf/violation.pm
```diff
@@ -832,8 +824,9 @@ sub violation_view_last_closed {
             vid => $vid,
             status => "closed",
         },
-        -order_by => '-release_date',
+        -order_by => {-desc => 'release_date'} ,
         -columns => [qw(mac vid release_date)],
+        -from => pf::dal::violation->table,
```
julsemaan: useful ?

lib/pf/violation.pm
```diff
@@ -894,15 +887,17 @@ sub violation_maintenance {
...
-               -limit => $batch,
-               -columns => [qw(id mac vid notes status)],
-            }
+            -where => {
+                status => ["open", "delayed"],
+                release_date => [-and => {"!=" => $ZERO_DATE}, {"<=" => \'NOW()'}],
+            },
+           -limit => $batch,
+           -columns => [qw(id mac vid notes status)],
+           -from => pf::dal::violation->table,
```
julsemaan: useful ?

lib/pf/Report.pm
jrouzierinverse: Yes there a undef bug when I was refactoring.
I will added add that to devel.

lib/pf/violation.pm
jrouzierinverse: Yes it is I did not want to hard code the table name

julsemaan: What I mean is that shouldn't this already be the case when using the violation DAL object ?

julsemaan: As jazzed verbally, if joins should be skipped, it should be done via an explicit parameter instead of this workaround so its more explicit to the untrained eye

lib/pf/violation.pm
```diff
@@ -258,7 +253,7 @@ sub violation_exist {
         },
         -columns => [qw(id mac vid start_date release_date status ticket_ref notes)],
         -limit => 1,
-        -from => pf::dal::violation->table,
+        -no_default_join => 1,
```
julsemaan: use $TRUE constant for this and the ones below

jrouzierinverse:
Finished pf::accounting

julsemaan:
Alright, not sure I'll be able to keep the little sanity I have remaining if I go through and read all of this code again.

Could you rebase this PR and we'll go ahead and merge?

extrafu:
👍 Good job guys.

PR #2742, opened 2017-10-31
feature/multitenant

jrouzierinverse:
Description
===========
Add multi tenancy to packect fence

Impacts
=======

Code / PR Dependencies
======================
depend on move-to-dal

New Features
------------
* multi tenancy


db/pf-schema-X.Y.Z.sql
```diff
@@ -20,12 +20,14 @@ CREATE TABLE `tenant` (
   id int NOT NULL AUTO_INCREMENT,
   name VARCHAR(255) NOT NULL,
   portal_domain_name VARCHAR(255),
+  domain_name VARCHAR(255),
   PRIMARY KEY (id),
   UNIQUE KEY tenant_name (`name`),
-  UNIQUE KEY tenant_portal_domain_name (`portal_domain_name`)
+  UNIQUE KEY tenant_portal_domain_name (`portal_domain_name`),
+  UNIQUE KEY tenant_domain_name (`domain_name`)
```
jrouzierinverse: Please add this to the update sql

lib/pf/UnifiedApi.pm
```diff
@@ -0,0 +1,183 @@
...
+        return $c->render(json => { message => "Unknown path", errors => [] }, status => 404);
+    });
+}
+
+sub api_v1_routes {
+    my ($self) = @_;
+    return @API_V1_ROUTES, $self->api_v1_custom_routes;
+}
+
+sub api_v1_custom_routes {
```
julsemaan: @jrouzierinverse 

Do we need this or should this be stripped out from the code?

lib/pf/UnifiedApi.pm
```diff
@@ -0,0 +1,183 @@
...
+
+}
+
+sub custom_startup_hook {
```
julsemaan, line 153: @jrouzierinverse 

Same here

lib/pf/UnifiedApi/Controller/Nodes.pm
```diff
@@ -0,0 +1,65 @@
...
+
+use strict;
+use warnings;
+use Mojo::Base 'pf::UnifiedApi::Controller::Crud';
+use pf::dal::node;
+use pf::locationlog qw(locationlog_history_mac locationlog_view_open_mac);
+
+has dal => 'pf::dal::node';
+has id_key => 'id';
+has resource_id => 'pid';
```
julsemaan: @jrouzierinverse 

Is the resource id really the pid ?

lib/pf/UnifiedApi/Controller/Violations.pm
```diff
@@ -0,0 +1,80 @@
...
+use warnings;
+use Mojo::Base 'pf::UnifiedApi::Controller::Crud';
+use pf::violation;
+
+has dal => 'pf::dal::violation';
+has id_key => 'id';
+has resource_id => 'id';
+
+
+sub list {
```
julsemaan: @satkunas 

Not sure why listing violations would only list the open ones.
This should list all the violations and then we could have /violations/opened to list the opened violations

What do you think?

lib/pf/UnifiedApi/Controller/Violations.pm
```diff
@@ -0,0 +1,80 @@
...
+    my $search = $self->param('search');
+    return $self->_list_by_mac($search) if pf::util::valid_mac($search);
+    return $self->_list_by_id($search);
+}
+
+sub _list_by_mac {
+    my ($self, $mac) = @_;
+    my @violations = pf::violation::violation_view_desc($mac);
+    return $self->render(json => { items => \@violations }) if scalar @violations > 0 and defined($violations[0]);
+    return $self->render(json => undef);
```
julsemaan, line 43: @satkunas 

Should render an empty array instead

julsemaan (changes requested):
A few things to polish up

@satkunas and @jrouzierinverse could you review my questions/comments 

Then @jrouzierinverse can perform his super rebase de la muerte and we can merge that

lib/pf/UnifiedApi/Controller/Violations.pm
satkunas: Render an empty array "[]" on a collection, or "{items:[]}"?

...and render an empty object "{}" on a row?

lib/pf/UnifiedApi/Controller/Violations.pm
satkunas: I didn't see an existing method in the base class that returned all (including non-open) violations. Please correct me if I'm wrong.

lib/pf/UnifiedApi/Controller/Violations.pm
julsemaan: When a row isn't found it should send a 404 with with a message that says it hasn't been found

When its an empty collection, it should be a 200 with an empty array

lib/pf/UnifiedApi/Controller/Violations.pm
julsemaan: If you just go through the DAL, that will perform an unfiltered select

@jrouzierinverse can confirm this.

So basically, you'd just have to delete this subroutine and we'd be golden (again @jrouzierinverse  can confirm)

lib/pf/UnifiedApi/Controller/Violations.pm
satkunas: @julsemaan please confirm again, I updated my last comment above at the same time as your last response.

julsemaan: @satkunas it is "{items:[]}" you are right for a collection

satkunas: Adjusted over in https://github.com/inverse-inc/packetfence/pull/2920

lib/pf/UnifiedApi.pm
jrouzierinverse: Yes we need this because if we do not have it then we cannot allow custom route/controllers in our API.

lib/pf/UnifiedApi.pm
jrouzierinverse: Yes we need this because if we do not have it then we cannot allow custom route/controllers in our API.

lib/pf/UnifiedApi/Controller/Nodes.pm
julsemaan: @jrouzierinverse 

Could you confirm if this needs to be changed?

jrouzierinverse: It needs to be changed will do so after the merging in feature/api-restructure

addons/packages/packetfence.spec
```diff
@@ -118,6 +118,8 @@ Requires: perl(Data::Phrasebook), perl(Data::Phrasebook::Loader::YAML)
 Requires: perl(DBI)
 Requires: perl(Rose::DB)
 Requires: perl(Rose::DB::Object)
+Requires: perl(Lingua::EN::Nums2Words) >= 1.16
```
julsemaan, line 121: These need to be put in devel

lib/pf/UnifiedApi/Controller/Crud.pm
```diff
@@ -20,8 +20,9 @@ use pf::error qw(is_error);
 use pf::log;
 
 has 'dal';
-has 'id_key';
-has 'resource_id' => 'id';
+has 'url_param_name'; # param_name
```
julsemaan: remove comment

lib/pf/UnifiedApi/Controller/Crud.pm
```diff
@@ -20,8 +20,9 @@ use pf::error qw(is_error);
...
+has 'primary_key' => 'id'; # primary_key
```
julsemaan: remove comment

lib/pf/UnifiedApi/Controller/Crud.pm
julsemaan: Also perldoc for this field would be nice

lib/pf/UnifiedApi/Controller/Crud.pm
julsemaan: perldoc for this too

lib/pf/UnifiedApi/Controller/Crud.pm
```diff
@@ -20,8 +20,9 @@ use pf::error qw(is_error);
...
+has 'parent_primary_key_map' => sub { {} };
```
julsemaan: And perdoc for this too

julsemaan (changes requested):

julsemaan (changes requested):
Missing before the merge:
- Rebase the branch since its now in conflict
- Put the missing dependencies in the PacketFence devel repo

jrouzierinverse:
The missing dependencies are in the PacketFence devel repo

jrouzierinverse:
Rebased

julsemaan (approved):

alinahaider96:
Hi Julsemaan,
Can you please tell me how can we apply multi tenancy using PacketFence. I can see the  filter of tenants in connection profile but I don't know how to set that tenant filter.The only value appear in front of it is default.I wanted to make a product that if two organization using my product they couldn't access or see each other's data.
Thanks and regards,
Alina

julsemaan:
@alinahaider96 

This question should be asked on the PF mailing list, not in this PR

PR #2757, opened 2017-11-03
Add i18n english strings to admin violations

satkunas:
# Description
* More user-friendly strings on Admin > Compliance > Violations > Actions
# Impacts
* None

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Enhanced English i18n strings in admin CMS violation actions.

### DIFF
--- a/html/pfappserver/lib/pfappserver/I18N/en.po
+++ b/html/pfappserver/lib/pfappserver/I18N/en.po
@@ -9435,7 +9435,7 @@ msgstr "Auto Registration"
 
 # pf::action (VIOLATION_ACTIONS)
 msgid "autoreg_action"
-msgstr "Register node"
+msgstr "Register node on violation"
 
 # pfappserver::Model::Node::Tab::MSE (MSE Tab)
 msgid "band"
@@ -9531,7 +9531,7 @@ msgstr "cell phone"
 
 # pf::action (VIOLATION_ACTIONS)
 msgid "close_action"
-msgstr "Close a violation"
+msgstr "Close violation"
 
 # pf::Authentication::Source::ADSource (available_attributes)
 # pf::Authentication::Source::LDAPSource (available_attributes)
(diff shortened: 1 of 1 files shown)


html/pfappserver/lib/pfappserver/I18N/en.po
```diff
@@ -10312,7 +10312,7 @@ msgstr ""
 
 # pf::action (VIOLATION_ACTIONS)
 msgid "log_action"
-msgstr "Log message"
+msgstr "Log violation"
```
extrafu, line 10315: Or "Log violation message" ?

html/pfappserver/lib/pfappserver/I18N/en.po
```diff
@@ -10954,7 +10954,7 @@ msgstr "Redis Port"
 
 # pf::action (VIOLATION_ACTIONS)
 msgid "reevaluate_access_action"
-msgstr "Reevaluate Access Action"
+msgstr "Isolate device on violation"
```
extrafu: I don't know if we always "isolate" - we could also just switch the user to an other VLAN, which isn't necessarily the isolation/quarantine VLAN.

html/pfappserver/lib/pfappserver/I18N/en.po
satkunas: IMO "message" seems excessive and is implied with "Log".

html/pfappserver/lib/pfappserver/I18N/en.po
satkunas: How does "Change device VLAN on violation" work?

extrafu: Might not always be a VLAN :-)  It can be an ACL, a controller role, etc.

How about "Reevaluate network access" ?

satkunas: "Modify device network access on violation"?

extrafu: or "Change network access on violation" ? I hesitate between Change or Update or Modify. I would skip "device" to make it shorter.

satkunas:
@extrafu changes pushed based on comments above.

julsemaan:
@satkunas I think you need to accept the CLA before we merge this one

PR #2760, opened 2017-11-03
Feature/improve pfcmd import node

julsemaan:
# Description
Improve the features of pfcmd node import so its more in line with the ones available in the admin

# Impacts
pfcmd import nodes

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Improved the features of pfcmd node import so its more in line with the ones available in the admin

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Node.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node.pm
@@ -39,6 +39,7 @@ use pf::config::util;
 use pf::violation;
 use pf::SwitchFactory;
 use pf::Connection;
+use Text::CSV;
 
 =head1 METHODS
 
--- a/html/pfappserver/lib/pfappserver/Model/Node.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node.pm
@@ -411,6 +411,7 @@ sub importCSV {
         else {
             $logger->debug("Modify already registered MAC $mac ($pid)");
             $result = node_modify($mac, %data);
+            node_update_last_seen($mac);
         }
         if ($result) {
             $count++;
--- a/html/pfappserver/lib/pfappserver/Model/Node.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node.pm
@@ -312,6 +312,7 @@ sub importCSV {
     my $default_node_pid = $data->{default_pid};
     my $default_category_id = $data->{default_category_id};
     my $default_voip = $data->{default_voip};
+    my $default_unregdate = $data->{default_unregdate};
 
     $logger->debug("CSV file import nodes from $tmpfilename ($filename, \"$delimiter\")");
 
@@ -380,7 +381,7 @@ sub importCSV {
            'pid'         => $pid,
            'category'    => $index{'category'}  ? $row->[$index{'category'}]  : undef,
            'category_id' => $index{'category'}  ? undef                       : $default_category_id,
-           'unregdate'   => $index{'unregdate'} ? $row->[$index{'unregdate'}] : undef,
+           'unregdate'   => $index{'unregdate'} ? $row->[$index{'unregdate'}] : $default_unregdate,
            'voip'        => $index{'voip'}      ? $row->[$index{'voip'}]      : $default_voip,
            'notes'       => $index{'notes'}     ? $row->[$index{'notes'}]     : undef,
           );
--- a/html/pfappserver/lib/pfappserver/Model/Node.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node.pm
@@ -302,11 +302,11 @@ See pf::import::nodes
 =cut
 
 sub importCSV {
-    my ($self, $data, $user, $allowed_roles) = @_;
+    my ($self, $file, $data, $user, $allowed_roles) = @_;
     my $logger = get_logger();
     my ($status, $message);
-    my $filename = $data->{nodes_file}->filename;
-    my $tmpfilename = $data->{nodes_file}->tempname;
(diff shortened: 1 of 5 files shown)


julsemaan:
@jrouzierinverse 

Any comments on that one ? Or will it be good for merging by Friday ?

html/pfappserver/lib/pfappserver/Model/Node.pm
```diff
@@ -302,16 +303,16 @@ See pf::import::nodes
 =cut
 
 sub importCSV {
```
jrouzierinverse, line 306: The importCSV logic should be moved to pf::import::nodes to improve code sharing.

lib/pf/cmd/pf/import/nodes.pm
```diff
@@ -15,24 +15,94 @@ use strict;
...
+Run the import
+
+=cut
+
 sub _run {
     my ($self) = @_;
-    require pf::import;
-    pf::import::nodes($self->{file});
-    print "Import process complete\n";
+    use lib "$install_dir/html/pfappserver/lib";
```
jrouzierinverse: If logic is moved to pf::import::nodes this is no longer needed.

jrouzierinverse (changes requested):
The importCSV should be moved to pf::import::nodes.

html/pfappserver/lib/pfappserver/Model/Node.pm
julsemaan: Done

julsemaan:
Reworked, please re-review

lib/pf/cmd/pf/import/nodes.pm
```diff
@@ -15,24 +15,93 @@ use strict;
...
+        $params->{filename},
+        {
+            columns => $params->{columns},
+            delimiter => $params->{delimiter},
+            default_pid => $params->{"default-owner"},
+            default_category_id => pf::nodecategory::nodecategory_view_by_name($params->{"default-role"})->{category_id},
+            default_voip => $params->{"default-voip"},
+            default_unregdate => $params->{"default-unregdate"},
+        },
+    );
```
jrouzierinverse, line 103: No error checking of the status

lib/pf/cmd/pf/import/nodes.pm
```diff
@@ -15,24 +15,93 @@ use strict;
...
+    my %params = (
+        'columns'           => "mac",
+        'default-role'      => "default",
+        'default-unregdate' => "2038-01-01 00:00:00",
+        'default-voip'      => "no",
+        'default-owner'     => "default",
+        'delimiter'         => "comma",
+    );
+    foreach my $param ( @params ) {
+        my @data = split('=', $param);
```
jrouzierinverse: The following would be more clear
```perl
my ($name, $val) = split('=', $param);
```

lib/pf/import.pm
```diff
@@ -41,79 +42,141 @@ use pf::util;
...
+Handle bulk importation of nodes. They are automatically registered.
 
 =cut
 
-# TODO support more fields
-# TODO instead of printout to STDOUT directly, we should provide a handle to print to in pfcmd
-#      this would allow callers to determine where the output should go. pfcmd -> STDOUT, SOAP -> Apache
 sub nodes {
-    my ($filename) = @_;
+    my ($file, $data, $user, $allowed_roles) = @_;
```
jrouzierinverse, line 50: $user does not seem to be used

jrouzierinverse (changes requested):

lib/pf/import.pm
julsemaan: Kept the same signature as when it was in pfappserver as my guess is that this could be useful at some point for other checks so IMO I'd leave it

lib/pf/cmd/pf/import/nodes.pm
julsemaan: It prints to STDOUT the result of the import

lib/pf/cmd/pf/import/nodes.pm
julsemaan: This was your code I copy/pasted from another file BTW ;) but I'll fix it

julsemaan:
re-check please

lib/pf/cmd/pf/import/nodes.pm
```diff
@@ -15,24 +15,93 @@ use strict;
...
+            delimiter => $params->{delimiter},
+            default_pid => $params->{"default-owner"},
+            default_category_id => pf::nodecategory::nodecategory_view_by_name($params->{"default-role"})->{category_id},
+            default_voip => $params->{"default-voip"},
+            default_unregdate => $params->{"default-unregdate"},
+        },
+    );
+    
+    print "Import process complete. Imported $info->{count} and skipped $info->{skipped}\n";
     return $EXIT_SUCCESS;
```
jrouzierinverse: It will return a success even if there is a problem with the import.
```perl
  return $status ? $EXIT_SUCCESS : $EXIT_FAILURE;
```

lib/pf/cmd/pf/import/nodes.pm
jrouzierinverse: Which file so we can improve readability there also.

julsemaan:
@jrouzierinverse 

Done

lib/pf/cmd/pf/import/nodes.pm
julsemaan: Its in lib/pf/cmd/pf/ipmachistory.pm

I wouldn't "pollute" this PR with code cleanup in case we use it again for patching a customer so we could do it in devel in a separate task

PR #2771, opened 2017-11-09
Add SSID suggestion in Admin > Policy > Connection Profiles > Filters

### LINKED ISSUE #2758: Admin: Should provide suggestions for SSID filter
Sometimes people are tired and do a typo on the SSID and may troubleshoot it for a while before figuring out they screwed up the case of a character in the profile filter.

We should provide suggestions for the SSID filter based on the SSIDs we've seen in the locationlog while still allowing someone to go 'full beast mode' and enter it manually

satkunas:
# Description
Add SSID suggestion to Admin > Policy > Connection Profiles > Filters

# Impacts
May affect existing Input filter/match form entry pairs. Testing needed in alt sections.

# Delete branch after merge
YES

## Enhancements
* Adds intermediary Select box on Filter > SSID >
* Fixes #2758

### DIFF
--- a/html/pfappserver/root/static/admin/config/connection_profile.js
+++ b/html/pfappserver/root/static/admin/config/connection_profile.js
@@ -224,21 +224,20 @@ function initReadPage(element) {
         select.find("option:selected").removeAttr("selected");
         var options = select.find('option[value!=""]');
         // Select the next option that was not yet selected
-        try {
-            options.each(function(index,element) {
-                var selector = '[value="' + element.value + '"]';
-                if(selected_options.filter(selector).length === 0) {
-                    $(element).attr("selected", "selected");
-                    throw "";
-                }
-            });
-        }
-        catch(e) {}
+        options.each(function(index,element) {
+            var selector = '[value="' + element.value + '"]';
+            if(selected_options.filter(selector).length === 0) {
+                $(element).attr("selected", "selected");
+                return false;
+            }
+        });
+        
         // If all options have been added, remove the add button
         var rows = row.siblings(':not(.hidden)').addBack();
         if (rows.length == options.length) {
             rows.find('[href="#add"]').addClass('hidden');
         }
+        select.trigger('change');
     });
     // When selecting an exclusive source, remove all other sources
     $('#sources').on('change', 'select', function(event) {
@@ -266,9 +265,10 @@ function initReadPage(element) {
     $('select[name$=".type"]:not(:disabled)').each(function(i,e){
         updateFilterMatchInput($(e),true);
     });
+    $('#filter select[name$=".type"]').trigger('change');
 }
 
-function updateFilterMatchInput(type_input, keep) {
+function updateFilterMatchInput(type_input, keep_value) {
     var match_input = type_input.next();
     var type_value = type_input.val();
     var match_input_template_id = '#' + type_value + "_filter_match";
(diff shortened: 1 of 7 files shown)


cgx:
You don't have to update the French *po* since we manage it from [transifex](https://www.transifex.com/inverse/packetfence/).

html/pfappserver/root/static/admin/config/connection_profile.js
```diff
@@ -277,8 +277,18 @@ function updateFilterMatchInput(type_input, keep) {
...
+            $(type_input.next()).on('change', function(event){
+                /* Replace SELECT with INPUT */
+                select_element = $(event.currentTarget);
+                select_element.replaceWith('<input type="text" id="'+select_element.attr('id')+'" name="'+select_element.attr('name')+'" value="'+select_element.val()+'" data-required="'+(select_element.attr('required')||'0')+'" placeholder="'+select_element.attr('placeholder')+'">');
+                if(select_element.val() == '') {
+                    type_input.next().focus();
+                }
+            });
+        }
+        if(type_value == "switch") {
```
cgx: `else if` would be more appropriate here.

html/pfappserver/root/static/admin/config/connection_profile.js
```diff
@@ -277,8 +277,18 @@ function updateFilterMatchInput(type_input, keep) {
...
     if ( match_input_template.length ) {
-        changeInputFromTemplate(match_input, match_input_template, keep);
-        if (type_value == "switch") {
+        changeInputFromTemplate(match_input, match_input_template, keep_value);
+        if(type_value == "ssid") {
+            $(type_input.next()).on('change', function(event){
+                /* Replace SELECT with INPUT */
+                select_element = $(event.currentTarget);
+                select_element.replaceWith('<input type="text" id="'+select_element.attr('id')+'" name="'+select_element.attr('name')+'" value="'+select_element.val()+'" data-required="'+(select_element.attr('required')||'0')+'" placeholder="'+select_element.attr('placeholder')+'">');
+                if(select_element.val() == '') {
```
julsemaan: Should use a === according to jshint while doing a grunt dist on my machine:
```
Running "jshint:files" (jshint) task

   admin/config/connection_profile.js
    286 |                if(select_element.val() == '') {
                                                 ^ Use '===' to compare with ''.

>> 1 error in 25 files
Warning: Task "jshint:files" failed. Use --force to continue.
```

julsemaan:
As discussed verbally, if you have an SSID configured that isn't in the DB as a filter, when looking at a persisted profile, it doesn't show the "custom" configured SSID

lib/pf/locationlog.pm
```diff
@@ -241,6 +243,9 @@ sub locationlog_db_prepare {
         FROM locationlog
         WHERE mac = ?
         ORDER BY start_time DESC LIMIT 1 ]);
+        
+    $locationlog_statements->{'locationlog_unique_ssids_sql'} = get_db_handle()->prepare(
+        qq [ SELECT DISTINCT `ssid` FROM `locationlog` WHERE `ssid` != '' AND `ssid` IS NOT NULL ORDER BY `ssid` ASC]);
```
extrafu, line 248: Check with @jrouzierinverse  with the upcoming multi-tenant code if it's safe or will need to be adjusted.

julsemaan: @extrafu is right, this will have to use the DAL, you can check with James whether this should be done before or after the merge of #2687

jrouzierinverse: Would need to be adjusted but it is not a big deal.

html/pfappserver/root/static/admin/config/connection_profile.js
satkunas: Deprecated in https://github.com/inverse-inc/packetfence/pull/2771/commits/b04723b22d2b5c6800edf242375d6f727d15bb07

html/pfappserver/root/static/admin/config/connection_profile.js
satkunas: Deprecated in https://github.com/inverse-inc/packetfence/pull/2771/commits/b04723b22d2b5c6800edf242375d6f727d15bb07

lib/pf/locationlog.pm
satkunas: @extrafu We'll ( @jrouzierinverse and I ) will address after DAL is merged.

PR #2776, opened 2017-11-14
Add access_duration on Admin > Compliance > Violations with 'Register…

### LINKED ISSUE #1333: Set access duration action for violation
Right now when we register a node using a violation, it gets registered for life.

We should add the ability to set an unreg date or access duration

satkunas:
… node' (autoreg)

Add unregdate on node registration to support Registered node access duration.

# Description
Allow packetfence to assign an access duration when a node is automatically registered through a violation.

# Issue
* Fixes #1333

# Delete branch after merge
YES


lib/pf/action.pm
```diff
@@ -353,15 +355,15 @@ sub action_reevaluate_access {
 sub action_autoregister {
     my ($mac, $vid) = @_;
     my $logger = get_logger();
-
+    my $unregdate = access_duration($pf::violation_config::Violation_Config{$vid}{access_duration});
```
julsemaan, line 358: What if its not defined ? Is it still making it registered for life ?

html/pfappserver/lib/pfappserver/Form/Violation.pm
```diff
@@ -320,6 +337,34 @@ sub options_template {
...
 }
 
+=head2 options_durations
+
+Populate the access duration select field with the available values defined
+in the pf.conf configuration file.
+
+=cut
+
+sub options_durations {
```
julsemaan: Instead of copy/pasting the subroutine, it would be better to reuse the existing one in the other form or create a util function called by both forms

extrafu:
@satkunas Please review.

satkunas:
Conflicts resolved

lib/pf/action.pm
jrouzierinverse: The question is do we allow this behavior or not.
@julsemaan

html/pfappserver/lib/pfappserver/Form/Violation.pm
```diff
@@ -12,13 +12,17 @@ Form definition to create or update a violation.
...
 use HTTP::Status qw(:constants is_success);
-
+use List::MoreUtils qw(uniq);
+use pf::config qw(%Config);
+use pf::web::util;
+use pf::admin_roles;
 use pf::action;
 use pf::log;
 use pf::constants::violation qw($MAX_VID %NON_WHITELISTABLE_ROLES);
+use pfappserver::Base::Form::Authentication::Action qw(options_durations);
```
jrouzierinverse: Remove this please.

jrouzierinverse (changes requested):

extrafu:
@satkunas Please review and proceed with the changes.

extrafu:
@jrouzierinverse Merge if alright.

jrouzierinverse (approved):

Issue #2772, opened 2017-11-10
Admin: 2 new charts/reports


veedubbs:
I was forced to sit through another NAC product demo and they had two reports (and pie charts):

- Top reasons for authentication failures
- Top users with most authentications

Might be nice to have in PF!

extrafu:
👍

extrafu:
@satkunas Get your hands more dirty!

veedubbs:
If it helps, the product I saw did both of these as "top 10".

fdurand:
i can't imagine the trauma, are you ok ?

veedubbs:
I'll survive, but I'll need counseling.

fdurand:
Yeah i can imagine ....
![hrpu3vivopal](https://user-images.githubusercontent.com/1553962/32762454-5d8209de-c8c7-11e7-9de5-1229401f426b.jpg)

satkunas:
@veedubbs Do you recall how the "top 10" report for each was laid out? Better yet, do you have any screenshots?

veedubbs:
I'm hoping to be able to grab some screenshots tomorrow...

veedubbs:
Sorry @satkunas, I wasn't able to get screenshots.  I recall it was something very similar to the current reports in PF; for example, the SSIDs report (a simple pie chart with a table):


![pf](https://user-images.githubusercontent.com/382007/33040670-67412614-ce01-11e7-888b-5c3b9170961c.JPG)

PR #2785, opened 2017-11-20
Add new AAA Auth Reports (7) for Successes (4) and Failures (3)

satkunas:
# Description
Added 7 new Reports in Admin > Reports > Accounting
- Top Auth Failures by MAC
- Top Auth Failures by SSID
- Top Auth Failures by Username
- Top Auth Successes by MAC
- Top Auth Successes by SSID
- Top Auth Successes by Username
- Top Auth Successes by Computername

# Impacts
May be slow on large data sets.

# Delete branch after merge
YES 

# NEWS file entries
## New Features
* Implements #2772


extrafu:
👍

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -49,3 +49,16 @@ call ValidateVersion;
...
+
+
+
+
+
+
+---
+--- https://github.com/inverse-inc/packetfence/pull/2785
+--- 
+ALTER TABLE `radius_audit_log` ADD KEY (`auth_status`, `created_at`);
```
cgx, line 61: You'll need to also modify `pf-schema-X.Y.Z.sql`. The `created_at` key seems already defined.

html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
```diff
@@ -907,6 +907,141 @@ sub osclassbandwidth :Local :AdminRole('REPORTS') {
...
 
+=head2 topauthenticationfailures_by_mac
+
+Radius AAA errors by mac for a specific period.
+
+Defined as a report.
+
+=cut
+
+sub topauthenticationfailures_by_mac :Local :AdminRole('REPORTS') {
```
julsemaan, line 918: This naming convention makes my eyes bleed. IDK if I'm the only one

satkunas: I followed existing conventions. Feel free to open a new item to modify the conventions throughout.

db/upgrade-X.X.X-X.Y.Z.sql
satkunas: The single KEY (`created_at`) isn't useful when using a compound WHERE:

        WHERE `auth_status`='Accept'
            AND `created_at` BETWEEN ? AND ?

A second compound KEY (`auth_status`,`created_at`) is useful for the compounded WHERE above.

html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
satkunas: Schema added in https://github.com/inverse-inc/packetfence/pull/2785/commits/e1e09a6cde3ecceea23e7da6eeb30590e011136c

db/upgrade-X.X.X-X.Y.Z.sql
satkunas: Schema added in https://github.com/inverse-inc/packetfence/pull/2785/commits/e1e09a6cde3ecceea23e7da6eeb30590e011136c and https://github.com/inverse-inc/packetfence/pull/2785/commits/ea40bb4680215d3acfac0a5940edfcec5b0cd03c

html/pfappserver/lib/pfappserver/PacketFence/Controller/Graph.pm
julsemaan: The convention in that file seems to be everything together so it would be topauthenticationfailuresbymac

Would still make my eyes bleed but it would follow the "convention" of that file.

IMO, it would be better to just make it underscore separated right away

julsemaan:
Minor detail, right now the reports are under the accounting section, they would fit better under a new "Authentication" section or under the existing "Connection section" since these have nothing to do with data accounting.

julsemaan:
Also, there seems to be a lot of records returned (more than the scoped top 10).

Maybe we could reduce it to 10 or 20 to make the pages smaller since right now the tables are as long as @lzammit lunchtime stories

julsemaan:
Other :nail_care: comment, the default scope is 30 days which is much more than the default retention and if a large deployment has a big retention going in this report will probably make the machine spin as hard as @lzammit in his spartan classes. I'd suggest we scope it to a default of 1 day to keep it as light as possible while keeping the ability to extend it but at least accessing the page will have a lighter query be ran.

satkunas:
@julsemaan comments above addressed in https://github.com/inverse-inc/packetfence/pull/2785/commits/03b5d30fb1803193da68c0a0b25a6f6b4745b7a4

julsemaan:
Alright, so the only thing remaining is scoping this to 1 day by default

satkunas:
Conflicts resolved.

julsemaan:
@satkunas 

Has this been set to 1 day default or not yet ?

db/pf-schema-X.Y.Z.sql
```diff
@@ -1114,6 +1114,7 @@ CREATE TABLE radius_audit_log (
   KEY `created_at` (created_at),
   KEY `mac` (mac),
   KEY `ip` (ip),
+  KEY `auth_status` (auth_status, created_at),
```
jrouzierinverse: Add this after user_name so the upgraded schema and the a new created schemas are the same.

extrafu:
@satkunas Review James' comment and fix them today please.

extrafu:
@jrouzierinverse Merge if alright.

PR #2787, opened 2017-11-20
Use $db var for expansion on the db_fingerbank configuration name

### LINKED ISSUE #2665: Configurator: when using a different DB name, the fingerbank.conf mysql section isn't updated
Basically, if you call your DB pf_dinde in the configurator, it will setup the user rights for pf_dinde_fingerbank but fingerbank.conf is still using pf_fingerbank so we should update the fingerbank.conf so it points to the right DB

satkunas:
# Description
Removes static fingerbank database name, adds dynamic $db var expansion

# Impacts
n/a

# Issue
Fixes #2665

# Delete branch after merge
YES

## Bug Fixes
* Configurator: when using a different DB name, the fingerbank.conf mysql section isn't updated (#2665)


extrafu:
@TomInverse Please review and merge.

PR #2794, opened 2017-11-22
Keepalived unicast feature

fdurand:
# Description
Added a way to enable keepalived to do unicast instead of multicast

# Impacts
Keepalived

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Keepalived can be configured for unicast traffic

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1318,6 +1318,13 @@ description=<<EOT
 The virtual router id for keepalive. Leave untouched unless you have another keepalive cluster in this network. Must be between 1 and 255.
 EOT
 
+[active_active.vrrp_unicast]
+type=toggle
+options=enable|disabled
+description=<<EOT
+Turn keepalived in unicast mode instead of multicast
+EOT
+
 [active_active.dns_on_vip_only]
 type=toggle
 options=enabled|disabled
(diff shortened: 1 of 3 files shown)


conf/documentation.conf
```diff
@@ -1318,6 +1318,13 @@ description=<<EOT
 The virtual router id for keepalive. Leave untouched unless you have another keepalive cluster in this network. Must be between 1 and 255.
 EOT
 
+[active_active.vrrp_unicast]
+type=toggle
+options=enable|disabled
+description=<<EOT
+Turn keepalived in unicast mode instead of multicast
```
extrafu: Enable keepalived in unicast instead of multicast

conf/pf.conf.defaults
```diff
@@ -996,6 +996,11 @@ password=1234
 # Shared KEY for vrrp protocol (Must be the same on all members).
 virtual_router_id=50
 #
+# active_active.vrrp_unicast
+#
+# Trun keepalived in unicast mode instead of multicast
```
extrafu: Enable keepalived in unicast instead of multicast

julsemaan:
Code looks simple & good, if it works, then just fix the docs and we'll merge

PR #2793, opened 2017-11-22
Removed duplicate field 'device_type' from nodes. Moved 'device_class…

satkunas:
# Description
Removed duplicate field 'device_type' from nodes. 
Moved 'device_class' beside sibling.

# Impacts
Ensure device_type is singleton choice in node types.

# Delete branch after merge
YES

## Bug Fixes
* Fixes "Admin: Multiple 'Device Type' options in Nodes tab" #2789

PR #2795, opened 2017-11-22
Add switch description in nodes side navigation

### LINKED ISSUE #2791: Switch Groups view under Nodes...
When viewing switches under Nodes and then Switch Groups on the left hand column it shows the Switch ID.  I'd like the ability to be able to have that show the Switch Description if at all possible.  It would save a lot of back and forth time trying to figure out what switch is what when troubleshooting client issues.

satkunas:
# Description
Adds a description to the switches in the Nodes side navigation.

# Impacts
UX/UI only

# Delete branch after merge
YES

## Bug Fixes
* Fixes #2791

### DIFF
--- a/html/pfappserver/root/admin/nodes.tt
+++ b/html/pfappserver/root/admin/nodes.tt
@@ -6,12 +6,13 @@
     <option value="bypass_vlan">[% l('Bypass VLAN') %]</option>
     <option value="computername">[% l('Computer Name') %]</option>
     <option value="connection_type">[% l('Connection Type') %]</option>
-    <option value="device_class">[% l('Device Class') %]</option>
     <option value="device_type">[% l('Device Type') %]</option>
     <option value="last_ip">[% l('Node IP') %]</option>
     <option value="category">[% l('Node Role') %]</option>
     <option value="notes">[% l('Notes') %]</option>
     <option value="online">[% l('Online Status') %]</option>
+    <option value="device_class">[% l('Device class') %]</option>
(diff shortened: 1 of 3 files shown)


html/pfappserver/root/admin/nodes.tt
```diff
@@ -6,13 +6,12 @@
...
     <option value="computername">[% l('Computer Name') %]</option>
     <option value="connection_type">[% l('Connection Type') %]</option>
+    <option value="device_class">[% l('Device Class') %]</option>
     <option value="device_type">[% l('Device Type') %]</option>
     <option value="last_ip">[% l('Node IP') %]</option>
     <option value="category">[% l('Node Role') %]</option>
     <option value="notes">[% l('Notes') %]</option>
     <option value="online">[% l('Online Status') %]</option>
-    <option value="device_class">[% l('Device class') %]</option>
-    <option value="device_type">[% l('Device type') %]</option>
```
julsemaan: Not sure this belongs in this PR as you're addressing it in #2793

julsemaan:
@satkunas 

For next time, since this is a "feature" more than a bugfix, the milestone should have been 7.4 (same as the issue it addresses in 99% of the cases)

julsemaan:
And also, the "NEWS file entry" for the PR needs to be a user friendly description of what you did (refer to NEWS.asciidoc for examples)

In this case, it will simply be the same as your description: "Adds a description to the switches in the Nodes side navigation."

PR #2807, opened 2017-11-28
Add violation suggestion for next VID pseudo auto_increment_id

### LINKED ISSUE #2804: Admin Violations: When creating a new violation, we should autofill the ID
Doesn't make sense to ask the user to decide of a DB numeric ID for the Identifier field.

We should take the latest violation ID incremented by 1 and autofill the field on creation

image

satkunas:
# Description
Populates the VID with the next auto_increment_id on Violation -> Add Violation
dB doesn't use auto_increment, so a pseudo value is calc'd using MAX(vid)+1 

# Impacts
None

# Issue
Fixes #2804

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Suggest VID in Admin Violations -> Add Violation

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Violation.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Violation.pm
@@ -19,6 +19,7 @@ use HTTP::Status qw(:constants is_success);
 use pf::action;
 use pf::log;
 use pf::constants::violation qw($MAX_VID %NON_WHITELISTABLE_ROLES);
+use pf::class qw(class_next_vid);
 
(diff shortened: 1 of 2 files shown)


julsemaan:
Missing the news file entry in your description

julsemaan:
Also, please validate that the error handling will provide a clear message if 2 users create a violation at the same time with the same auto-incremented violation ID

satkunas:
@julsemaan validation confirmed: 

> Error! Section 3000006 already exists in the configuration

extrafu:
@julsemaan Please merge.

PR #2812, opened 2017-12-01
Feature/auth log profile

julsemaan:
# Description
Record the connection profile in the auth_log table

# Impacts
auth_log, so pretty much all authentications

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* The connection profile is now recorded in the authentication log (auth_log table)

### DIFF
--- a/conf/report.conf.defaults
+++ b/conf/report.conf.defaults
@@ -7,7 +7,7 @@ EOT
 # The table to search from
 base_table=auth_log
 # The columns to select
-columns=auth_log.process_name as "Process Name", auth_log.mac "MAC Address", auth_log.pid as "Owner", auth_log.status as "Status", auth_log.attempted_at as "Attempted at", auth_log.completed_at as "Completed at", auth_log.source as "Source"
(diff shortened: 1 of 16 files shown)


julsemaan:
Not sure if it should hit 7.4 or not but I put the milestone anyway since I like living dangerously

Did that make sense, no... But this is still a great PR

julsemaan:
@jrouzierinverse 

Could you take a look at that one and particular attention to the usage of the DAL in auth_log.pm

jrouzierinverse (commented):
I have a few questions.

Why record_oauth_attempt record_guest_attempt does not record the profile name?

Does it make sense to use the mac profile and source as the "unique id" for the auth_log record?

julsemaan:
@jrouzierinverse 

Nothing against recording it on start+finish but I see this as complimentary information that should occur when the authentication is "sealed" but I may be biased by the use-case for which I did this PR initially.

I'd leave the "unique id" as is since I see this as complimentary information, again might be biased. If it would be put on open as well, I'd definitely make it tied to the profile.

jrouzierinverse:
I have no problem with leaving the "unique id" as is.
The only issue I can see is we can lose the complimentary profile information.
For instance If a person starts the oauth process but never completes it.
We do not know which profile they hit in to the auth_log.

julsemaan:
Re-review

PR #2817, opened 2017-12-05
feature/test_db

jrouzierinverse:
Description
===========
Create a dedicated user and table from testing database

To run tests run the following commands

```bash
mysql -uroot -p < /usr/local/pf/t/db/smoke_test.sql
/usr/local/pf/t/db/setup_test_db.pl
```

Impacts
=======
Testing

Delete branch after merge
=========================
NO

NEWS file entries
=================

New Features
------------
* Dedicated user and table for testing

### DIFF
--- a/t/db/setup_test_db.pl
+++ b/t/db/setup_test_db.pl
@@ -1,0 +1,73 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+setup_test_db -
+
+=cut
+
+=head1 DESCRIPTION
+
+setup_test_db
+
+=cut
+
+use strict;
+use warnings;
+use lib qw(/usr/local/pf/lib);
+BEGIN {
+    use lib '/usr/local/pf/t';
+    use setup_test_config;
+}
+
+use pf::db;
+use DBI;
+
+my $config = pf::db::db_config();
+
+if ($config->{user} ne 'pf_smoke_tester') {
+   die "Not using the standard testing user for db\n"; 
+}
+
+my $dbh = test_db_connection($config) or die "Cannot connection to db with test user please run\nmysql -uroot -p < t/smoke_test.sql\n";
+my $db = $config->{db};
+
+$dbh->do("DROP DATABASE IF EXISTS $db;") or die "Cannot drop database $db\n";;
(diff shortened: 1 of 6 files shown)


t/db/setup_test_db.pl
```diff
@@ -0,0 +1,80 @@
...
+my $db = $config->{db};
+
+$dbh->do("DROP DATABASE IF EXISTS $db;") or die "Cannot drop database $db\n";;
+$dbh->do("CREATE DATABASE $db;") or die "Cannot create database $db\n";
+system("mysql -u$config->{user} -p$config->{pass} $db < /usr/local/pf/db/pf-schema-X.Y.Z.sql");
+
+sub test_db_connection {
+    my ($config) = @_;
+    my $dsn = "dbi:mysql:;host=$config->{host};port=$config->{port};mysql_client_found_rows=0";
+    return  DBI->connect($dsn, $config->{user}, $config->{pass}, { RaiseError => 0, PrintError => 0, mysql_auto_reconnect => 1 });
```
julsemaan, line 49: Shouldn't we connect with the same parameters as in pf::db ?

Maybe a quick util function to create the DB connection would be useful ?

t/setup_test_config.pm
```diff
@@ -52,6 +46,15 @@ BEGIN {
     `rm -fr /tmp/chi/*`;
 }
 
+sub new_db_config {
```
julsemaan: perldoc

t/smoke.t
```diff
@@ -82,6 +85,19 @@ die(sprintf(
 ) if $num_bad;
 
 
+
+sub create_test_db {
```
julsemaan: perldoc

t/db/setup_test_db.pl
jrouzierinverse: I am connecting without a database.

jrouzierinverse: This is a special case where I want to connect to mysql without setting the current database.
Because if I try and connect with a current database that does exists it will fail.
So no need for a util function at the moment.
If we need it later I will add it.

julsemaan:
Skipped news file entry since tests aren't packaged in the releases

PR #2829, opened 2017-12-12
Fix portal default PID overwriting a valid previous PID

### LINKED ISSUE #2825: Default user is set for a device when the device hit another time the portal (inline case)
A device register on the captive portal and the pid is set to the username and the unregistration date is set to 1 year.
The day after it reconnect on the network and because his ip changed it hit the portal one more time but this time because the pid is not in the session PacketFence take the default pid and save it for the device.

julsemaan:
# Description
Fix portal default PID overwriting a valid previous PID

# Impacts
Portal node assignments (particularly the PID)

# Issue
fixes #2825 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Prevent a valid PID from being overwritten at the end of the portal registration if the new PID is default (#2825)


julsemaan:
@fdurand 

Do you want to try it out in the problematic environment and let me know if it fixes it

extrafu:
@fdurand Please merge + cherry pick in maintenance.

PR #2842, opened 2017-12-18
Juniper SRX Firewall SSO 

fdurand:
# Description
Firewall SSO module for Juniper SRX

# Impacts
Firewall SSO

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Juniper SRX Firewall SSO module

### DIFF
--- a/docs/firewall/junipersrx.asciidoc
+++ b/docs/firewall/junipersrx.asciidoc
@@ -1,0 +1,56 @@
+// to display images directly on GitHub
+ifdef::env-github[]
+:imagesdir: ../images
+endif::[]
+
+Configuration of the Juniper SRX in PacketFence
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Go to *Configuration -> Integration -> Firewall SSO -> Add Firewall -> JuniperSRX*.
+
+[options="compact"]
+* *Hostname or IP Address*: IP of your JuniperSRX
+* *Firewall type*: JuniperSRX (JuniperSRX = HTTPS requests)
+* *Password*: secret
+* *Port*: 8443
+* *Roles*: add the roles that you want to do SSO
+
+Step 1: webapi configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You need to setup webapi management as follows
+
+set system services webapi user PF
+set system services webapi user password YOURPASSWORD
+set system services webapi client PF_MANAGEMENT_IP_ADDRESS
+
+set system services webapi https port PORT_YOU_WANT_TO_USE i.e. 8443
+set system services webapi https default-certificate
+
+Next setup user entry settings
+
+set services user-identification authentication-source aruba-clearpass authentication-entry-timeout 120
+set services user-identification authentication-source aruba-clearpass no-user-query
+set services user-identification device-information authentication-source network-access-controller
+
+Then you need to allow traffic from packetfence management interface to port you set up on webapi settings (i.e. 8443) on SRX device.
+
+
+Step 2: Verification
+^^^^^^^^^^^^^^^^^^^^
+
+For debuging webapi set (disable it when you won't need it anymore):
+
+set system services webapi debug-log api-log
+set system services webapi debug-level notice
+
+To check registered device entries on SRX use
+
+show services user-identification authentication-table authentication-source all ( extensive for more detailed informations)
+
+or
+
+run show services user-identification device-information table all extensive
+
+to see more details about OS, device type etc.
+
--- a/docs/firewall/junipersrx.asciidoc
+++ b/docs/firewall/junipersrx.asciidoc
@@ -33,13 +33,13 @@ set services user-identification authentication-source aruba-clearpass authentic
 set services user-identification authentication-source aruba-clearpass no-user-query
 set services user-identification device-information authentication-source network-access-controller
 
-Then you need to allow traffic from packetfence management interface to port you set up on webapi settings (i.e. 8443) on SRX device.
+Then you need to allow traffic from the PacketFence management interface to port you set up on webapi settings (i.e. 8443) on SRX device.
 
 
 Step 2: Verification
 ^^^^^^^^^^^^^^^^^^^^
 
-For debuging webapi set (disable it when you won't need it anymore):
+For debuging the webapi set (disable it when you won't need it anymore):
 
 set system services webapi debug-log api-log
 set system services webapi debug-level notice
--- a/go/firewallsso/junipersrx.go
+++ b/go/firewallsso/junipersrx.go
@@ -31,7 +31,6 @@ func (fw *JuniperSRX) Start(ctx context.Context, info map[string]string, timeout
 func (fw *JuniperSRX) startHttp(ctx context.Context, info map[string]string, timeout int) (bool, error) {
 
 	req, err := http.NewRequest("POST", "https://"+fw.PfconfigHashNS+":"+fw.Port+"/api/userfw/v1/post-entry", bytes.NewBuffer([]byte(fw.startHttpPayload(ctx, info))))
-	// req, err := http.NewRequest("GET", "https://"+fw.PfconfigHashNS+":"+fw.Port+"/rpc/request-userfw-local-auth-table-add?user-name="+info["username"]+"&ip-address="+info["ip"]+"&roles="+info["role"], nil)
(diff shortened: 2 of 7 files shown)


go/firewallsso/base.go
```diff
@@ -230,29 +239,34 @@ func (fw *FirewallSSO) logger(ctx context.Context) log15.Logger {
 // Makes sure to call FirewallSSO.Start and to validate the network and role if necessary
 func ExecuteStart(ctx context.Context, fw FirewallSSOInt, info map[string]string, timeout int) (bool, error) {
 	ctx = log.AddToLogContext(ctx, "firewall-id", fw.GetFirewallSSO(ctx).PfconfigHashNS)
-	log.LoggerWContext(ctx).Info("Processing SSO Start")
 
-	if !fw.MatchesRole(ctx, info) {
-		log.LoggerWContext(ctx).Debug(fmt.Sprintf("Not sending SSO for user device %s since it doesn't match the role", info["role"]))
-		return false, nil
-	}
+	if fw.CheckStatus(ctx, info) {
```
julsemaan: To keep it a bit simpler and not nest too many conditions together, I'd suggest to apply the same recipe as with MatchesRole and MatchesNetwork which would be to just have:
```
if !fw.CheckStatus(ctx, info) {
    log.LoggerWContext(ctx).Debug("Not sending SSO for user device since it isn't registered")
    return false, nil
}
```

And that way the `else` isn't needed and adding new conditions doesn't create deeper if conditions with a new else everytime

fdurand:
BUMP

go/firewallsso/junipersrx.go
```diff
@@ -0,0 +1,95 @@
...
+		resp.Body.Close()
+	}
+
+	return err == nil, err
+}
+
+func (fw *JuniperSRX) startHttpPayload(ctx context.Context, info map[string]string) string {
+	info["Time"] = time.Now().Format(time.RFC3339)
+	t := template.New("JuniperSRX.startHttp")
+	t.Parse(`<?xml version="1.0"?><userfw-entries><userfw-entry><source>PacketFence</source><timestamp>{{.Time}}</timestamp><operation>logon</operation><IP>{{.Ip}}</IP><domain>{{.Realm}}</domain><user>{{.Username}}</user><role-list><role>{{.Role}}</role></role-list><posture>Healthy</posture><end-user-attribute><device-identity><value>{{.Computername}}</value><groups><group>{{.Role}}</group></groups></device-identity><device-category>{{.Device_class}}</device-category><device-os>{{.Device_type}}</device-os><device-os-version>"{{.Device_version}}"</device-os-version></end-user-attribute></userfw-entry></userfw-entries>`)
```
julsemaan: Could we structure this in a multi-line string?

That would make it a lot easier to read and potentially troubleshoot

go/firewallsso/junipersrx.go
```diff
@@ -0,0 +1,95 @@
...
+
+	b := new(bytes.Buffer)
+	t.Execute(b, fw.InfoToTemplateCtx(ctx, info, 0))
+	return b.String()
+}
+
+func (fw *JuniperSRX) stopHttpPayload(ctx context.Context, info map[string]string) string {
+	info["Time"] = time.Now().UTC().Format(time.RFC3339)
+	t := template.New("JuniperSRX.startHttp")
+	t.Parse(`<?xml version="1.0"?><userfw-entries><userfw-entry><source>PacketFence</source><timestamp>{{.Time}}</timestamp><operation>logoff</operation><IP>{{.Ip}}</IP></userfw-entry></userfw-entries>`)
```
julsemaan: Same comment about multi-line

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JuniperSRX.pm
```diff
@@ -0,0 +1,115 @@
...
+   label => 'Hostname or IP Address',
+   required => 1,
+   messages => { required => 'Please specify the hostname or IP of the firewall' },
+  );
+has_field 'username' =>
+  (
+   type => 'Text',
+   label => 'Username',
+   required => 1,
+   messages => { required => 'Please specify the username for the Barracuda' },
```
julsemaan: Didn't know Barracuda owned Juniper.

Seems like something recent:
http://cnn.com/news/economy/barracuda-firewall-buys-juniper-firewalls-division

Well as Mike would say: "I will coucher moins niaiseux tonight"

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JuniperSRX.pm
```diff
@@ -0,0 +1,115 @@
...
+  );
+has_field 'password' =>
+  (
+   type => 'ObfuscatedText',
+   label => 'Secret or Key',
+   required => 1,
+   messages => { required => 'You must specify the password' },
```
julsemaan, line 43: Its half-password, half-secret, half-key, like the [manbearpig](https://www.youtube.com/watch?v=h-moEu4_hEw)

I think we should use one of these words based on what is in the Juniper FW

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JuniperSRX.pm
```diff
@@ -0,0 +1,115 @@
...
+  );
+has_field 'port' =>
+  (
+   type => 'PosInteger',
+   label => 'Port of the service',
+   default => 80,
```
julsemaan: This defaults to port 80 but in the Golang file it uses the HTTPS protocol

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JuniperSRX.pm
```diff
@@ -0,0 +1,115 @@
...
+   tags => { after_element => \&help,
+             help => 'If you use an alternative port, please specify' },
+  );
+has_field 'type' =>
+  (
+   type => 'Hidden',
+  );
+has_field 'categories' =>
```
julsemaan: No need to define this field here since its defined in the parent form. At least, that works for the PA module

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JuniperSRX.pm
```diff
@@ -0,0 +1,115 @@
...
+has_block definition =>
+  (
+   render_list => [ qw(id type username password port categories networks cache_updates cache_timeout username_format default_realm) ],
+  );
+
+=head2 options_categories
+
+=cut
+
+sub options_categories {
```
julsemaan: No need to define this field here since its defined in the parent form. At least, that works for the PA module

julsemaan (changes requested):

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/JuniperSRX.pm
brucehubbert: I tried looking this up and could find no data to support that claim.

olmanolsztyn: SRX is still Junipers FW

fdurand:
BUMP

docs/firewall/junipersrx.asciidoc
```diff
@@ -0,0 +1,56 @@
...
+set system services webapi https port PORT_YOU_WANT_TO_USE i.e. 8443
+set system services webapi https default-certificate
+
+Next setup user entry settings
+
+set services user-identification authentication-source aruba-clearpass authentication-entry-timeout 120
+set services user-identification authentication-source aruba-clearpass no-user-query
+set services user-identification device-information authentication-source network-access-controller
+
+Then you need to allow traffic from packetfence management interface to port you set up on webapi settings (i.e. 8443) on SRX device.
```
julsemaan: the PacketFence management interface

docs/firewall/junipersrx.asciidoc
```diff
@@ -0,0 +1,56 @@
...
+
+
+Step 2: Verification
+^^^^^^^^^^^^^^^^^^^^
+
+For debuging webapi set (disable it when you won't need it anymore):
```
julsemaan: For debugging the webapi, set...

fdurand:
BUMP

PR #2848, opened 2017-12-21
Fix race condition when formatting username in pfsso

### LINKED ISSUE #2847: pfsso: Username formatting causes concurrent read/write when using multiple firewalls
Lets say you have 2 Fortigate configured, then that would cause a FortiRaceCondition since the SSO information is passed by reference to a goroutine.

That means both goroutines for each firewall will target modifying the information map to format the username:
https://github.com/inverse-inc/packetfence/blob/devel/go/firewallsso/base.go#L246

We need to perform a copy of the SSO information before spawning the SSO goroutine for each firewall

julsemaan:
# Description
Fix race condition when formatting username in pfsso

# Impacts
Firewall SSO

# Issue
fixes #2847 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fix race condition when formatting username in pfsso (#2847)

### DIFF
--- a/go/caddy/pfsso/pfsso.go
+++ b/go/caddy/pfsso/pfsso.go
@@ -118,10 +118,16 @@ func (h PfssoHandler) validateInfo(ctx context.Context, info map[string]string) 
 }
 
 // Spawn an async SSO request for a specific firewall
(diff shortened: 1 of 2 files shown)


julsemaan:
@fdurand 

Could you take a look at this and merge when if it looks good

PR #2893, opened 2018-01-15
Don't allow the portal FQDN as a passthrough

### LINKED ISSUE #2815: Forbid the usage of passthrough as the same captive portal domain name
If you have the packetfence url domain name declared in the passthrough configuration, you won't be able to see the captive portal.

We should add a message in the Web admin when you save the passthrough and also a checkup on the services check.

julsemaan:
# Description
Don't allow the portal FQDN as a passthrough

# Impacts
Passthroughs + the admin interface

# Issue
fixes #2815 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Don't allow the portal domain name as a passthrough (#2815)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
@@ -250,8 +250,10 @@ sub field_list {
 }
 
 our %FIELD_VALIDATORS = (
-    "fencing.passthroughs" => sub { validate_fqdn_not_in_passthroughs(@_, "passthroughs") },
-    "fencing.isolation_passthroughs" => sub { validate_fqdn_not_in_passthroughs(@_, "isolation_passthroughs") },
+    "general.hostname" => sub { validate_fqdn_not_in_passthroughs(@_, [ "passthroughs", "isolation_passthroughs" ]) },
+    "general.domain" => sub { validate_fqdn_not_in_passthroughs(@_, [ "passthroughs", "isolation_passthroughs" ]) },
+    "fencing.passthroughs" => sub { validate_fqdn_not_in_passthroughs(@_, ["passthroughs"]) },
+    "fencing.isolation_passthroughs" => sub { validate_fqdn_not_in_passthroughs(@_, ["isolation_passthroughs"]) },
 );
 
 =head2 validator_for_field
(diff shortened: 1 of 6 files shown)


extrafu:
Looks good to me. @jrouzierinverse please also review.

html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
```diff
@@ -235,11 +241,50 @@ sub field_list {
...
+Get the validator for a field
+
+=cut
+
+sub validator_for_field {
+    my ($self, $field) = @_;
+
+    return {
+        "fencing.passthroughs" => sub { validate_fqdn_not_in_passthroughs(@_, "passthroughs") },
+        "fencing.isolation_passthroughs" => sub { validate_fqdn_not_in_passthroughs(@_, "isolation_passthroughs") },
```
jrouzierinverse: This would be better as static lookup hash to avoid creating the same data multiple times.

lib/pf/util/dns.pm
```diff
@@ -33,31 +33,55 @@ Returns 2 values
 
 sub matches_passthrough {
     my ($domain, $zone) = @_;
+
+    return _matches_passthrough(eval("\\%$zone"), $domain);
```
jrouzierinverse: Avoid the use of eval as it is unsafe.

A safer approach is to create a lookup table for which hash to use.

t/unittest/util/dns.t
```diff
@@ -32,7 +32,7 @@ use_ok("pf::util::dns");
 
 my ($match, $ports);
 
-($match, $ports) = pf::util::dns::matches_passthrough();
+($match, $ports) = pf::util::dns::matches_passthrough(undef, 'passthroughs');
```
jrouzierinverse, line 35: Add a test for undef zone also.

jrouzierinverse: Add a test for wild-isolation.zammitcorp.com

html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
jrouzierinverse: Also there should also be a check if someone changes the hostname or domain to a current passthrough domain.

jrouzierinverse (changes requested):

lib/pf/util/dns.pm
jrouzierinverse: Example

```bash
perl -I/usr/local/pf/lib -Mpf::util::dns -e'pf::util::dns::matches_passthrough(undef, "passthroughs,print(\"gotcha\\n\"),\%passthroughs");' 
```

julsemaan:
Performed the requested adjustments, please re-review

jrouzierinverse (approved):

PR #2907, opened 2018-01-19
Multitenant /nodes endpoints

satkunas:
# Description
Add /api/v1/nodes/:mac/locationlog endpoint
Add /api/v1/nodes/:mac/latest_locationlog endpoint
Add testunits for both endpoints

# Impacts
root endpoint /nodes/ or /nodes/:mac/ not yet supported

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Add node to API


lib/pf/UnifiedApi/Controller/Nodes.pm
```diff
@@ -0,0 +1,65 @@
...
+
+=cut
+
+use strict;
+use warnings;
+use Mojo::Base 'pf::UnifiedApi::Controller::Crud';
+use pf::dal::locationlog;
+use pf::locationlog qw(locationlog_history_mac locationlog_view_open_mac);
+
+has dal => 'pf::dal::locationlog';
```
julsemaan: @jrouzierinverse can confirm, but the Nodes controller shouldn't be using the locationlog DAL

The locationlog will be a sub resources of the node so this will likely not work

lib/pf/UnifiedApi/Controller/Nodes.pm
```diff
@@ -0,0 +1,65 @@
...
+has id_key => 'id';
+has resource_id => 'pid';
+
+sub latest_locationlog_by_mac {
```
julsemaan: Missing the spec for this

lib/pf/UnifiedApi/Controller/Nodes.pm
```diff
@@ -0,0 +1,65 @@
...
+    my ($self) = @_;
+    my $mac = $self->param('mac');
+    $self->render(json => locationlog_view_open_mac($mac));
+}
+
+sub locationlog_by_mac {
```
julsemaan, line 31: Missing the spec for this as well

t/unittest/UnifiedApi/Controller/Nodes.t
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+Locationlog
```
julsemaan: This will be for nodes, so the name shouldn't be tied to the locationlog

t/unittest/UnifiedApi/Controller/Nodes.t
```diff
@@ -0,0 +1,131 @@
...
+
+=cut
+
+=head1 DESCRIPTION
+
+unit test for Locationlog
```
julsemaan: Same as my comment above

lib/pf/UnifiedApi/Controller/Nodes.pm
satkunas: Added in https://github.com/inverse-inc/packetfence/pull/2907/commits/d3626bfd39fc1e8f0d9246d979ffe96dcab1b029

PR #2909, opened 2018-01-23
Realm based username stripping

### LINKED ISSUE #2588: Status page: we are stripping the username + realm for local accounts
For example if you would use an email address that is a local account on PF, that would strip the username and prevent authentication to be correct.

Potential fix would be to have the local source declare it doesn't want a stripped username.
Not sure if it would work but it technically should

julsemaan:
# Description
Delegate the decision to strip the username for authentication+authorization to the realm configuration. Also allow to fine tune the stripping based on the context in which the request is made (portal, admin, RADIUS)

# Impacts
All the authentication on portal, admin, RADIUS CLI. 
All the authorisation on portal, admin, RADIUS CLI, RADIUS 802.1x

# Issue
fixes #2588 

# Delete branch after merge
NO

# NEWS file entries
## New Features
* Username stripping (removing the realm) is now configurable via the realms instead of the sources

# UPGRADE file entries

The way PacketFence detects whether or not it should strip the realm out of a username when performing authentication and authorisation has been moved to the realms. Moreover, it is now configurable based on the context (login on the captive portal or administration interface, as well as when performing authorization in RADIUS 802.1x)

In order to migrate your configuration, you can use the following script which will guide you through the migration: 

```
# /usr/local/pf/addons/upgrade/to-8.0-authentication-conf.pl
```


lib/pf/authentication.pm
```diff
@@ -17,6 +17,7 @@ use pf::log;
 
 use pf::constants;
 use pf::config;
+use pf::config::util;
```
fdurand: use pf::config::util qw(strip_username_if_needed);

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/SAML.pm
```diff
@@ -16,6 +16,8 @@ with 'captiveportal::Role::Routed';
 
 use pf::util;
 use pf::auth_log;
+use pf::config::util;
```
fdurand: use pf::config::util qw(strip_username_if_needed);

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/SAML.pm
```diff
@@ -74,10 +76,8 @@ sub assertion {
 
     my ($username, $msg) = $self->source->handle_response($self->app->request->param("SAMLResponse"));
 
-    # We strip the username if the authorization source requires it.
-    if(isenabled($self->source->authorization_source->{stripped_user_name})){
-        ($username, undef) = strip_username($username);
-    }
+    # We strip the username if required
+    ($username, undef) = strip_username_if_needed($username, $pf::constants::realm::PORTAL_CONTEXT);
```
fdurand: &pf::config::util::strip_username_if_needed

lib/pf/authentication.pm
```diff
@@ -221,6 +222,11 @@ sub authenticate {
     my $username = $params->{'username'};
     my $password = $params->{'password'};
 
+    #TODO: evaluate if we want to die or have another behavior
+    my $context = $params->{'context'} // die "No authentication context provided";
+    
+    ($username, undef) = strip_username_if_needed($username, $context);
```
fdurand: &pf::config::util::strip_username_if_needed

lib/pf/authentication.pm
```diff
@@ -289,6 +295,11 @@ sub match {
         return undef;
     }
 
+    #TODO: evaluate if we want to die or have another behavior
+    my $context = $params->{'context'} // die "No authentication context provided";
+
+    ($params->{'username'}, undef) = strip_username_if_needed($params->{'username'}, $context);
```
fdurand: &pf::config::util::strip_username_if_needed

lib/pf/authentication.pm
```diff
@@ -379,6 +390,11 @@ sub match2 {
         $logger->warn("Calling match with empty/invalid rule class. Defaulting to '" . $params->{'rule_class'} . "'");
     }
 
+    #TODO: evaluate if we want to die or have another behavior
+    my $context = $params->{'context'} // die "No authentication context provided";
+
+    ($params->{'username'}, undef) = strip_username_if_needed($params->{'username'}, $context);
```
fdurand: &pf::config::util::strip_username_if_needed

fdurand (changes requested):

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/SAML.pm
julsemaan: I'm not sure I really get why we should use the full name including package name

fdurand: I applied the PR on my server and got an issue with that.

PR #2920, opened 2018-01-31
Feature/multitenant api violations

satkunas:
# Description
Refactor violation(s) plurality on collections

# Impacts
None

# Delete branch after merge
YES


extrafu:
@julsemaan Please review so we can complete the API branch.

lib/pf/UnifiedApi/Controller/Violations.pm
```diff
@@ -21,33 +21,45 @@ has dal => 'pf::dal::violation';
...
+    ));
+    $data->{id} = $response;
+    return $self->render(status => 201, json => { status => 201, data => $data, message => "violation created" }) if $data->{id};
+    return $self->render(status => 400, json => { status => 400, message => $self->status_to_error_msg(400) });
+}
 
-sub list {
+sub read_list {
     my ($self) = @_;
     my @violations = pf::violation::violation_view_open_uniq();
```
julsemaan: As commented in the other branch, this subroutine shouldn't be overriden and should use the result coming from the CRUD controller since we don't only want the open violations

jrouzierinverse:
The urls rework is not ready yet lets hold off on adding more things that will need to change.

lib/pf/UnifiedApi.pm
```diff
@@ -107,6 +105,10 @@ our @API_V1_ROUTES = (
 
     { controller => 'nodes', collection_v2a => { get => 'latest_locationlog_by_mac' }, resource_v2a => {}, path => '/nodes/:mac/latest_locationlog' },
     { controller => 'nodes', collection_v2a => { get => 'locationlog_by_mac' }, resource_v2a => {}, path => '/nodes/:mac/locationlog' },
+
+    { controller => 'violations', resource_v2a => {}, collection_v2a => { get => 'read_list', post => 'create' }, path => '/violations' },
```
julsemaan: I don't think you need those

First, read_list shouldn't be overridden (see other comment)

Then no need to say that post goes to create since it already does in the routing so you only need to override the create subroutine like you did in the controller and we should be golden

lib/pf/UnifiedApi/Controller/Violations.pm
```diff
@@ -21,33 +21,45 @@ has dal => 'pf::dal::violation';
...
-    my @violations = pf::violation::violation_view_desc($mac);
+    my @violations = pf::violation::violation_view_desc($search);
     return $self->render(json => { items => \@violations }) if scalar @violations > 0 and defined($violations[0]);
-    return $self->render(json => undef);
+    return $self->render(json => { items => [] });
 }
 
-sub _list_by_id {
-    my ($self, $id) = @_;
+sub read_row_by_id {
```
julsemaan: I don't think we need this, simply letting the CRUD controller read the row would be enough to read the violation

lib/pf/UnifiedApi.pm
```diff
@@ -107,6 +105,10 @@ our @API_V1_ROUTES = (
...
+    { controller => 'violations', resource_v2a => {}, collection_v2a => { get => 'read_list_by_search' }, path => '/violations/:search' },
+    { controller => 'violations', resource_v2a => {}, collection_v2a => { get => 'read_row_by_id' }, path => '/violation/:id' },
```
julsemaan: Based on my other comment, this doesn't have much use since the CRUD controller will already handle reading a row by its id

lib/pf/UnifiedApi.pm
satkunas: I wonder what I'm missing. POST didn't work with this line omitted.

satkunas: I just reconfirmed that nothing works with `t/unittest/UnifiedApi/Controller/Violations.t`

extrafu:
Please do like explained in https://github.com/inverse-inc/packetfence/pull/2937

PR #2922, opened 2018-01-31
feature/api-restructure

jrouzierinverse:
Description
===========
Restructure the rest routes conventions

Impacts
=======
pf::UnifiedApi routing.

NEW Package(s) required
=======================
Lingua::EN::Inflexion
Lingua::EN::Nums2Words


julsemaan:
Ready for review ?

jrouzierinverse:
Yes

addons/packages/packetfence.spec
```diff
@@ -119,6 +119,8 @@ Requires: perl(Data::Phrasebook), perl(Data::Phrasebook::Loader::YAML)
 Requires: perl(DBI)
 Requires: perl(Rose::DB)
 Requires: perl(Rose::DB::Object)
+Requires: perl(Lingua::EN::Nums2Words) >= 1.16
```
julsemaan, line 122: Are these packages built and in the packetfence-devel repo?

julsemaan:
Too a quick look at it, just need confirmation for the packages, then we merge this and then we get onto the final multitenant work for merging in devel

addons/packages/packetfence.spec
jrouzierinverse: They are built not but not in packetfence-devel

PR #2953, opened 2018-02-14
Feature/api violation

satkunas:
# Description
Adds violations in API

# Impacts
Exposes violations to API

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Add violations to API

### DIFF
--- a/docs/api/spec/paths/violations.yaml
+++ b/docs/api/spec/paths/violations.yaml
@@ -4,7 +4,6 @@
     description: The list of vilations will be limited to 100 and a cursor will be returned for iterating on other violations. Iteration is only needed if the nextCursor is not an empty value.
     parameters:
       - $ref: '#/components/parameters/cursor'
-      - $ref: '#/components/parameters/additional_fields'
     responses:
       '200':
         description: List of violations
@@ -30,14 +29,10 @@
       '422':
         $ref: "#/components/responses/UnprocessableEntity"
 
-'/violation/mac/{mac}':
+'/violations/by_mac/{mac}':
   get:
     summary: Allows to search through violations by MAC address
     description: The list of violations will be limited to 100 and a cursor will be returned for iterating on other violations. Iteration is only needed if the nextCursor is not an empty value.
-    parameters:
-      - $ref: '#/components/parameters/search'
-      - $ref: '#/components/parameters/logic_operator'
-      - $ref: '#/components/parameters/additional_fields'
     responses:
       '200':
         description: List of violation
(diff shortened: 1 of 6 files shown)


lib/pf/UnifiedApi.pm
```diff
@@ -69,7 +69,14 @@ our @API_V1_ROUTES = (
     { controller => 'ApiUsers' },
     { controller => 'Locationlogs' },
     { controller => 'Config::ConnectionProfiles' },
-    { controller => 'Violations', },
+    { 
+        controller => 'Violations',
+        collection => {
+            subroutes    => {
+                ':search' => { get => 'search' },
```
julsemaan: I think based off the discussions with @jrouzierinverse this should be named differently

extrafu:
@satkunas Please adjust so we can merge this ASAP.

docs/api/spec/paths/violations.yaml
```diff
@@ -0,0 +1,106 @@
...
+      '422':
+        $ref: "#/components/responses/UnprocessableEntity"
+
+'/violation/mac/{mac}':
+  get:
+    summary: Allows to search through violations by MAC address
+    description: The list of violations will be limited to 100 and a cursor will be returned for iterating on other violations. Iteration is only needed if the nextCursor is not an empty value.
+    parameters:
+      - $ref: '#/components/parameters/search'
+      - $ref: '#/components/parameters/logic_operator'
```
julsemaan: I don't think we need the search payload and logic operator since from what I can see the only thing it searches with is the MAC in the URL

docs/api/spec/paths/violations.yaml
```diff
@@ -0,0 +1,106 @@
...
+      '201':
+        $ref: "#/components/responses/Created"
+      '400':
+        $ref: "#/components/responses/BadRequest"
+      '409':
+        $ref: "#/components/responses/Duplicate"
+      '422':
+        $ref: "#/components/responses/UnprocessableEntity"
+
+'/violation/mac/{mac}':
```
julsemaan: maybe by_mac instead of just mac so that its clearer

lib/pf/UnifiedApi/Controller/Violations.pm
```diff
@@ -21,32 +21,25 @@ has dal => 'pf::dal::violation';
...
-
-sub list_by_search {
+sub mac {
     my ($self) = @_;
     my $search = $self->param('search');
-    return $self->_list_by_mac($search) if pf::util::valid_mac($search);
-    return $self->_list_by_id($search);
+    return $self->_search_by_mac($search) if pf::util::valid_mac($search);
+    $self->render_error(404, "Resource not available");
+    #return $self->_search_by_id($search);
```
julsemaan: remove that line if its not necessary anymore

PR #2959, opened 2018-02-16
Feature/go api namespace

satkunas:
# Description
API namespace munging in pfdhcp and pfipset/pfhttpd

# Impacts
Changes API workflow for pfdhcp and pfipset consumers

# NEW Package(s) required
`ipset-devel` needed with `go# make pfhttpd`

# Delete branch after merge
YES 

## Enhancements
* API namespace munging in pfdhcp and pfipset

### DIFF
--- a/docs/api/spec/components/parameters/base.yaml
+++ b/docs/api/spec/components/parameters/base.yaml
@@ -62,4 +62,5 @@ local:
   required: false
   schema:
     type: integer
+    default: 1
   description: The local is used to determine if the request should be forwarded to the rest of the cluster. A value of '0' forwards the request and either a missing or alternate value does not forward the request to the rest of the cluster.
--- a/docs/api/spec/components/parameters/base.yaml
+++ b/docs/api/spec/components/parameters/base.yaml
@@ -56,3 +56,10 @@ search_params:
   - $ref: '#/components/parameters/fields'
   - $ref: '#/components/parameters/sort'
 
+local:
+  in: query
+  name: local
+  required: false
+  schema:
+    type: integer
+  description: The local is used to determine if the request should be forwarded to the rest of the cluster. A value of '0' forwards the request and either a missing or alternate value does not forward the request to the rest of the cluster.
--- a/docs/api/spec/components/parameters/base.yaml
+++ b/docs/api/spec/components/parameters/base.yaml
@@ -62,5 +62,5 @@ local:
   required: false
   schema:
     type: integer
-    default: 1
+    default: 0
   description: The local is used to determine if the request should be forwarded to the rest of the cluster. A value of '0' forwards the request and either a missing or alternate value does not forward the request to the rest of the cluster.
--- a/docs/api/spec/components/schemas/ipset.yaml
+++ b/docs/api/spec/components/schemas/ipset.yaml
@@ -3,14 +3,14 @@ IpsetNetwork:
     properties:
         network:
             type: string
-            description: The IPv4 address of the network
+            description: The IPv4 network identifier
 
 IpsetRoleId:
     type: object
     properties:
         role_id:
             type: integer
-            description: The role of the clients node
+            description: The client's node
 
 IpsetIp:
     type: object
@@ -24,25 +24,26 @@ IpsetPort:
     properties:
         port:
             type: string
-            description: The port type:integer example: tcp:65535
+            description: The port to allow through. Can be prefixed with a protocol `tcp:` or `udp:`. When the protocol is omited, then the port will be opened for both protocols. example: tcp:65535
 
 IpsetMac:
     type: object
     properties:
         mac:
             type: string
-            description: The clients MAC address example: 01:23:45:67:89:AB
+            description: The clients MAC address. example: 01:23:45:67:89:AB
 
 IpsetType:
     type: object
     properties:
         type:
             type: string
-            description: The Ipset type values(Reg|Unreg)
+            description: The Ipset type.
+            enum: [Reg, Unreg]
             
 IpsetStatus:
     type: object
     properties:
         type:
             type: string
-            description: The response status value: "ACK"
+            description: The response status. value: "ACK"
--- a/docs/api/spec/components/schemas/ipset.yaml
+++ b/docs/api/spec/components/schemas/ipset.yaml
@@ -1,0 +1,48 @@
+IpsetNetwork:
+    type: object
+    properties:
+        network:
+            type: string
+            description: The IPv4 address of the network
+
+IpsetRoleId:
+    type: object
+    properties:
+        role_id:
+            type: integer
+            description: The role of the clients node
+
+IpsetIp:
+    type: object
+    properties:
+        ip:
+            type: string
+            description: The IPv4 address of the client
+
+IpsetPort:
+    type: object
+    properties:
+        port:
+            type: string
+            description: The port type:integer example: tcp:65535
+
+IpsetMac:
+    type: object
+    properties:
+        mac:
+            type: string
+            description: The clients MAC address example: 01:23:45:67:89:AB
+
+IpsetType:
+    type: object
+    properties:
+        type:
+            type: string
+            description: The Ipset type values(Reg|Unreg)
+            
+IpsetStatus:
+    type: object
+    properties:
+        type:
+            type: string
+            description: The response status value: "ACK"
--- a/docs/api/spec/components/schemas/ipset.yaml
+++ b/docs/api/spec/components/schemas/ipset.yaml
@@ -46,4 +46,5 @@ IpsetStatus:
     properties:
         type:
             type: string
-            description: The response status. value: "ACK"
+            description: The response status.
+            enum: [ACK, NAK]
--- a/go/caddy/pfipset/handlers.go
+++ b/go/caddy/pfipset/handlers.go
@@ -39,8 +39,16 @@ func handlePassthrough(res http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Port := o.Port
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Port, valid_port := validate_port(o.Port)
+	if !valid_port {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	IPSET.jobs <- job{"Add", "pfsession_passthrough", IP + "," + Port}
@@ -73,8 +81,16 @@ func handleIsolationPassthrough(res http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Port := o.Port
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Port, valid_port := validate_port(o.Port)
+	if !valid_port {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	IPSET.jobs <- job{"Add", "pfsession_isol_passthrough", IP + "," + Port}
@@ -107,11 +123,31 @@ func handleLayer2(res http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Type := o.Type
-	Network := o.Network
-	Roleid := o.RoleID
-	Mac := o.Mac
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Type, valid_type := validate_type(o.Type)
+	if !valid_type {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
+	if !valid_roleid {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Mac, valid_mac := validate_mac(o.Mac)
+	if !valid_mac {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	// Update locally
@@ -175,9 +211,21 @@ func handleMarkIpL2(res http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Network := o.Network
-	Roleid := o.RoleID
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
+	if !valid_roleid {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	// Update locally
@@ -208,7 +256,7 @@ func handleMarkIpL3(res http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
 
 	IPSET := pfIPSETFromContext(req.Context())
-    
+  
 	body, err := ioutil.ReadAll(req.Body)
 	if err != nil {
 		panic(err)
@@ -218,9 +266,21 @@ func handleMarkIpL3(res http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Network := o.Network
-	Roleid := o.RoleID
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
+	if !valid_roleid {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	// Update locally
@@ -261,10 +321,26 @@ func handleLayer3(res http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Type := o.Type
-	Network := o.Network
-	Roleid := o.RoleID
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Type, valid_type := validate_type(o.Type)
+	if !valid_type {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
+	if !valid_roleid {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	// Update locally
@@ -322,7 +398,11 @@ func (IPSET *pfIPSET) handleUnmarkMac(res http.ResponseWriter, req *http.Request
 	if err != nil {
 		panic(err)
 	}	
-	Mac := o.Mac
+	Mac, valid_mac := validate_mac(o.Mac)
+	if !valid_mac {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	for _, v := range IPSET.ListALL {
@@ -363,8 +443,12 @@ func (IPSET *pfIPSET) handleUnmarkIp(res http.ResponseWriter, req *http.Request)
 	err = json.Unmarshal(body, &o)
 	if err != nil {
 		panic(err)
-	}	
-	IP := o.IP
+	}
+	IP, valid_ipv4 := validate_ipv4(o.IP)
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
 	Local := req.URL.Query().Get("local")
 
 	for _, v := range IPSET.ListALL {
@@ -391,4 +475,17 @@ func (IPSET *pfIPSET) handleUnmarkIp(res http.ResponseWriter, req *http.Request)
 	if err := json.NewEncoder(res).Encode(result); err != nil {
 		panic(err)
 	}
+}
+
+func handleError(res http.ResponseWriter, code int) {
+	var result = map[string][]*Info{
+		"result": {
+			&Info{Status: "NAK"},
+		},
+	}
+	res.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	res.WriteHeader(code)
+	if err := json.NewEncoder(res).Encode(result); err != nil {
+		panic(err)
+	} 
 }
--- a/go/caddy/pfipset/handlers.go
+++ b/go/caddy/pfipset/handlers.go
@@ -23,7 +23,7 @@ func handlePassthrough(res http.ResponseWriter, req *http.Request) {
 	vars := mux.Vars(req)
 	IP := vars["ip"]
 	Port := vars["port"]
-	Local := vars["local"]
+	Local := req.URL.Query().Get("local")
 
 	IPSET.jobs <- job{"Add", "pfsession_passthrough", IP + "," + Port}
 	if Local == "0" {
@@ -48,7 +48,7 @@ func handleIsolationPassthrough(res http.ResponseWriter, req *http.Request) {
 	vars := mux.Vars(req)
 	IP := vars["ip"]
 	Port := vars["port"]
-	Local := vars["local"]
+	Local := req.URL.Query().Get("local")
 
 	IPSET.jobs <- job{"Add", "pfsession_isol_passthrough", IP + "," + Port}
 	if Local == "0" {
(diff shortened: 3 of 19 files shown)


julsemaan:
That looks good from a code perspective.

@satkunas could you look into getting the spec done, @fdurand can help with the descriptions

satkunas:
@julsemaan did we want the spec within the existing /docs/api/spec/ or create a separate spec in /docs/go/spec/?

julsemaan:
It will use the same spec so we can put it with the rest of it

go/caddy/pfipset/pfipset.go
```diff
@@ -96,14 +96,14 @@ func buildPfipsetHandler(ctx context.Context) (PfipsetHandler, error) {
...
-	api.HandleFunc("/ipset/passthrough_isolation/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}/{local:[0-1]}", handleIsolationPassthrough).Methods("POST")
+	api.HandleFunc("/ipset/mark_layer3/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{type:[a-zA-Z]+}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleLayer3).Methods("POST")
+	api.HandleFunc("/ipset/mark_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{type:[a-zA-Z]+}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", handleLayer2).Methods("POST")
+	api.HandleFunc("/ipset/unmark_mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", pfipset.IPSET.handleUnmarkMac).Methods("POST")
+	api.HandleFunc("/ipset/unmark_ip/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", pfipset.IPSET.handleUnmarkIp).Methods("POST")
+	api.HandleFunc("/ipset/mark_ip_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleMarkIpL2).Methods("POST")
+	api.HandleFunc("/ipset/mark_ip_layer3/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleMarkIpL3).Methods("POST")
```
julsemaan: IMO, the network and category_id should go in a JSON body

go/caddy/pfipset/pfipset.go
```diff
@@ -96,14 +96,14 @@ func buildPfipsetHandler(ctx context.Context) (PfipsetHandler, error) {
...
-	api.HandleFunc("/ipset/passthrough/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}/{local:[0-1]}", handlePassthrough).Methods("POST")
-	api.HandleFunc("/ipset/passthrough_isolation/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}/{local:[0-1]}", handleIsolationPassthrough).Methods("POST")
+	api.HandleFunc("/ipset/mark_layer3/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{type:[a-zA-Z]+}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleLayer3).Methods("POST")
+	api.HandleFunc("/ipset/mark_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{type:[a-zA-Z]+}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", handleLayer2).Methods("POST")
+	api.HandleFunc("/ipset/unmark_mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", pfipset.IPSET.handleUnmarkMac).Methods("POST")
+	api.HandleFunc("/ipset/unmark_ip/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", pfipset.IPSET.handleUnmarkIp).Methods("POST")
+	api.HandleFunc("/ipset/mark_ip_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleMarkIpL2).Methods("POST")
```
julsemaan: IMO, the network and category_id should go in a JSON body

go/caddy/pfipset/pfipset.go
```diff
@@ -96,14 +96,14 @@ func buildPfipsetHandler(ctx context.Context) (PfipsetHandler, error) {
...
-	api.HandleFunc("/ipset/unmark_mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}/{local:[0-1]}", pfipset.IPSET.handleUnmarkMac).Methods("POST")
-	api.HandleFunc("/ipset/unmark_ip/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{local:[0-1]}", pfipset.IPSET.handleUnmarkIp).Methods("POST")
-	api.HandleFunc("/ipset/mark_ip_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{local:[0-1]}", handleMarkIpL2).Methods("POST")
-	api.HandleFunc("/ipset/mark_ip_layer3/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{local:[0-1]}", handleMarkIpL3).Methods("POST")
-	api.HandleFunc("/ipset/passthrough/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}/{local:[0-1]}", handlePassthrough).Methods("POST")
-	api.HandleFunc("/ipset/passthrough_isolation/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}/{local:[0-1]}", handleIsolationPassthrough).Methods("POST")
+	api.HandleFunc("/ipset/mark_layer3/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{type:[a-zA-Z]+}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleLayer3).Methods("POST")
```
julsemaan: IMO, the network, category_id and ip should go in a JSON body

go/caddy/pfipset/pfipset.go
```diff
@@ -96,14 +96,14 @@ func buildPfipsetHandler(ctx context.Context) (PfipsetHandler, error) {
...
+	api.HandleFunc("/ipset/mark_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{type:[a-zA-Z]+}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", handleLayer2).Methods("POST")
```
julsemaan: IMO, the network, category_id and the mac should go in a JSON body

go/caddy/pfipset/pfipset.go
```diff
@@ -96,14 +96,14 @@ func buildPfipsetHandler(ctx context.Context) (PfipsetHandler, error) {
...
+	api.HandleFunc("/ipset/unmark_mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", pfipset.IPSET.handleUnmarkMac).Methods("POST")
+	api.HandleFunc("/ipset/unmark_ip/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", pfipset.IPSET.handleUnmarkIp).Methods("POST")
+	api.HandleFunc("/ipset/mark_ip_layer2/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleMarkIpL2).Methods("POST")
+	api.HandleFunc("/ipset/mark_ip_layer3/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{category_id:[0-9]+}/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleMarkIpL3).Methods("POST")
+	api.HandleFunc("/ipset/passthrough/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}", handlePassthrough).Methods("POST")
```
julsemaan: IMO, the ip + port should go in a JSON body

go/caddy/pfipset/pfipset.go
```diff
@@ -96,14 +96,14 @@ func buildPfipsetHandler(ctx context.Context) (PfipsetHandler, error) {
...
+	api.HandleFunc("/ipset/passthrough_isolation/{ip:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{port:(?:udp|tcp):[0-9]+}", handleIsolationPassthrough).Methods("POST")
```
julsemaan: IMO, the ip + port should go in a JSON body

julsemaan:
We should also rename the category_id parameter by role_id since that is the new way of calling it

go/caddy/pfipset/pfipset.go
julsemaan: Also it seems the type should be in the body as well

go/caddy/pfipset/pfipset.go
julsemaan: I don't really get what is the difference between this and mark_layer2

Maybe we could get @fdurand to shed some light on this

fdurand: It's because you have 2 ipset , one with ip/mac for reg/isol(mark_layer2) and another one like network_ip_role for traffic shapping per example (mark_ip_layer2).

go/caddy/pfipset/handlers.go
```diff
@@ -70,20 +97,29 @@ func handleIsolationPassthrough(res http.ResponseWriter, req *http.Request) {
...
+	body, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		panic(err)
+	}
+	var o Options
+	err = json.Unmarshal(body, &o)
+	if err != nil {
+		panic(err)
+	}	
+	IP := o.IP
```
julsemaan: I think that since those options are required, we should perform validation that they are not empty since when they were in a URL we were 100% sure they would be there, while now they could be omitted by mistake

docs/api/spec/components/parameters/base.yaml
```diff
@@ -29,3 +29,11 @@ logic_operator:
   schema:
     type: string
   description: The logic operator to use for the search query. Can either be all or any.
+  
+local:
+  in: query
+  name: local
+  required: false
```
julsemaan: I think it would be safe to say it default to 0 here since when its omitted, the value in the service is an empty string and the code matches local == "1"

docs/api/spec/components/schemas/ipset.yaml
```diff
@@ -0,0 +1,48 @@
+IpsetNetwork:
+    type: object
+    properties:
+        network:
+            type: string
+            description: The IPv4 address of the network
```
julsemaan: I think it should be 'The IPv4 network identifier' since I'm not sure its technically an address

docs/api/spec/components/schemas/ipset.yaml
```diff
@@ -0,0 +1,48 @@
...
+        ip:
+            type: string
+            description: The IPv4 address of the client
+
+IpsetPort:
+    type: object
+    properties:
+        port:
+            type: string
+            description: The port type:integer example: tcp:65535
```
julsemaan: "The port to allow through. Can be prefixed by `tcp:` or `udp:`. When the protocol is omited, then the port will be opened for both protocols"

docs/api/spec/components/schemas/ipset.yaml
```diff
@@ -0,0 +1,48 @@
...
+        network:
+            type: string
+            description: The IPv4 address of the network
+
+IpsetRoleId:
+    type: object
+    properties:
+        role_id:
+            type: integer
+            description: The role of the clients node
```
extrafu: "client's node" I guess

docs/api/spec/components/schemas/ipset.yaml
```diff
@@ -0,0 +1,48 @@
...
+        mac:
+            type: string
+            description: The clients MAC address example: 01:23:45:67:89:AB
+
+IpsetType:
+    type: object
+    properties:
+        type:
+            type: string
+            description: The Ipset type values(Reg|Unreg)
```
julsemaan: I think you should be able to add an enum attribute here like this:
enum: [Reg, Unreg]

go/caddy/pfipset/utils.go
```diff
@@ -68,7 +68,7 @@ func updateClusterL2(ctx context.Context, Ip string, Mac string, Network string,
 	logger := log.LoggerWContext(ctx)
 
 	for _, member := range getClusterMembersIps(ctx) {
-		err := post(ctx, "https://"+member.String()+":22223/ipset/mark_layer2/"+Network+"/"+Type+"/"+Catid+"/"+Ip+"/"+Mac+"/1", body)
+		err := post(ctx, "https://"+member.String()+":22223/ipset/mark_layer2/"+Network+"/"+Type+"/"+Catid+"/"+Ip+"/"+Mac+"/?local=1", body)
```
julsemaan: This would need to be adjusted for the new request format since it calls the other pfipset daemons

julsemaan: And apply the comment above to the other calls in this file

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -107,7 +107,7 @@ func (pf pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 					switch ansb := ans.(type) {
 					case *dns.A:
 						for _, valeur := range v {
-							err := pf.apiClient.Call(ctx, "POST", "/api/v1/ipset/passthrough_isolation/"+ansb.A.String()+"/"+valeur+"/1", &unifiedapiclient.DummyReply{})
+							err := pf.apiClient.Call(ctx, "POST", "/api/v1/ipset/passthrough_isolation/"+ansb.A.String()+"/"+valeur+"/?local=1", &unifiedapiclient.DummyReply{})
```
julsemaan: This needs to be adjusted for the new request format

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -129,7 +129,7 @@ func (pf pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 				switch ansb := ans.(type) {
 				case *dns.A:
 					for _, valeur := range v {
-						err := pf.apiClient.Call(ctx, "POST", "/api/v1/ipset/passthrough/"+ansb.A.String()+"/"+valeur+"/1", &unifiedapiclient.DummyReply{})
+						err := pf.apiClient.Call(ctx, "POST", "/api/v1/ipset/passthrough/"+ansb.A.String()+"/"+valeur+"/?local=1", &unifiedapiclient.DummyReply{})
```
julsemaan: This needs to be adjusted for the new request format

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -149,7 +149,7 @@ func (pf pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
```
julsemaan: This needs to be adjusted for the new request format

go/dhcp/main.go
```diff
@@ -136,17 +136,16 @@ func main() {
...
 	router.HandleFunc("/api/v1/dhcp/stats/{int:.*}", handleStats).Methods("GET")
 	router.HandleFunc("/api/v1/dhcp/debug/{int:.*}/{role:(?:[^/]*)}", handleDebug).Methods("GET")
 	router.HandleFunc("/api/v1/dhcp/initialease/{int:.*}", handleInitiaLease).Methods("GET")
-	router.HandleFunc("/api/v1/dhcp/options/add/network/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{options:.*}", handleOverrideNetworkOptions).Methods("POST")
-	router.HandleFunc("/api/v1/dhcp/options/add/mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}/{options:.*}", handleOverrideOptions).Methods("POST")
-	router.HandleFunc("/api/v1/dhcp/options/del/network/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleRemoveNetworkOptions).Methods("GET")
-	router.HandleFunc("/api/v1/dhcp/options/del/mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", handleRemoveOptions).Methods("GET")
-	router.HandleFunc("/api/v1/dhcp/releaseip/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}", handleReleaseIP).Methods("POST")
-
+	router.HandleFunc("/api/v1/dhcp/options/network/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}/{options:.*}", handleOverrideNetworkOptions).Methods("POST")
```
julsemaan: The options would be better fit in the body

go/dhcp/main.go
```diff
@@ -136,17 +136,16 @@ func main() {
...
+	router.HandleFunc("/api/v1/dhcp/options/network/{network:(?:[0-9]{1,3}.){3}(?:[0-9]{1,3})}", handleRemoveNetworkOptions).Methods("DELETE")
+	router.HandleFunc("/api/v1/dhcp/options/mac/{mac:(?:[0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}}/{options:.*}", handleOverrideOptions).Methods("POST")
```
julsemaan: Same comment as above for the options

lib/pf/ipset.pm
```diff
@@ -302,9 +302,9 @@ sub iptables_mark_node {
             if ($net_addr->contains($ip)) {
 
                 if ($ConfigNetworks{$network}{'type'} =~ /^$NET_TYPE_INLINE_L3$/i) {
-                    call_ipsetd("/ipset/mark_layer3/".$network."/".$mark_type_to_str{$mark}."/".$role_id."/".$iplog."/0",{});
+                    call_ipsetd("/ipset/mark_layer3/".$network."/".$mark_type_to_str{$mark}."/".$role_id."/".$iplog."/?local=0",{});
```
julsemaan: Needs to be switched into the body

lib/pf/ipset.pm
```diff
@@ -302,9 +302,9 @@ sub iptables_mark_node {
...
                 } else {
-                    call_ipsetd("/ipset/mark_layer2/".$network."/".$mark_type_to_str{$mark}."/".$role_id."/".$iplog."/".$mac."/0",{});
+                    call_ipsetd("/ipset/mark_layer2/".$network."/".$mark_type_to_str{$mark}."/".$role_id."/".$iplog."/".$mac."/?local=0",{});
```
julsemaan: body here too

docs/api/spec/components/parameters/base.yaml
satkunas: You mean that it should default to 1, since 'default' has to be implicitly set to '0' for it to chain the req to the cluster?

lib/pf/parking.pm
```diff
@@ -48,7 +48,7 @@ sub park {
     my ($mac,$ip) = @_;
     get_logger->debug("Setting client in parking");
     if(isenabled($Config{parking}{place_in_dhcp_parking_group})){
-        pf::api::unifiedapiclient->default_client->call("POST", "/api/v1/dhcp/options/add/mac/".$mac."/", [{
+        pf::api::unifiedapiclient->default_client->call("POST", "/api/v1/dhcp/options/mac/".$mac."/", [{
```
satkunas: @fdurand This doesn't appear to be correct in the current context of the Go DHCP handler. More specifically, the post body has 2 issues that I'm unable to grasp:
1) How/Where is the "option", "value" and "type" Unmarshal'd into etcdInsert()
2) Why is this wrapped into an array?

Is this code-block outdated, or am I missing something? I'm attempting to move the /path/to/{options:*} into the POST body, but am having trouble wrapping my head around how the previous code works, specifically this code block.

extrafu:
@julsemaan Beside two conflicting files, are you ok to merge?

go/caddy/pfipset/handlers.go
```diff
@@ -34,18 +34,26 @@ func handlePassthrough(res http.ResponseWriter, req *http.Request) {
...
 	}
-	var o Options
+	var o PostOptions
 	err = json.Unmarshal(body, &o)
 	if err != nil {
 		panic(err)
 	}	
-	IP := o.IP
-	Port := o.Port
+	IP, valid_ipv4 := validate_ipv4(o.IP)
```
julsemaan: convention in our golang codebase is to use camel case everywhere

go/dhcp/main.go
julsemaan: This would still require to be adjusted unless there would be a good reason to leave this in the URL instead of the body

go/dhcp/main.go
julsemaan: Same as above

docs/api/spec/components/parameters/base.yaml
fdurand: I think it should be define to 0 since the code will detect if there ar other members in the cluster.

go/caddy/pfipset/handlers.go
```diff
@@ -70,20 +113,49 @@ func handleIsolationPassthrough(res http.ResponseWriter, req *http.Request) {
...
+	if !valid_type {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
```
fdurand: CamelCase

go/caddy/pfipset/handlers.go
```diff
@@ -116,60 +188,111 @@ func (IPSET *pfIPSET) IPSEThandleLayer2(ctx context.Context, IP string, Mac stri
...
+	if !valid_ipv4 {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
```
fdurand: CamelCase

go/caddy/pfipset/handlers.go
```diff
@@ -116,60 +188,111 @@ func (IPSET *pfIPSET) IPSEThandleLayer2(ctx context.Context, IP string, Mac stri
...
```
fdurand: CamelCase

go/caddy/pfipset/handlers.go
```diff
@@ -180,32 +303,57 @@ func handleMarkIpL3(res http.ResponseWriter, req *http.Request) {
...
+	if !valid_type {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Network, valid_network := validate_ipv4(o.Network)
+	if !valid_network {
+		handleError(res, http.StatusBadRequest)
+		return
+	}
+	Roleid, valid_roleid := validate_roleid(o.RoleID)
```
fdurand: CamelCase

go/caddy/pfipset/utils.go
```diff
@@ -211,10 +211,10 @@ func (IPSET *pfIPSET) initIPSet(ctx context.Context, db *sql.DB) {
 	var (
 		ipstr string
 		mac   string
-		catID string
+		nodeID string
```
fdurand: NodeId ?

go/caddy/pfipset/utils.go
```diff
@@ -288,3 +288,43 @@ func (IPSET *pfIPSET) detectType(ctx context.Context) error {
 	}
 	return nil
 }
+
+func validate_ipv4 (ipv4 string) (string, bool) {
```
fdurand: net.ParseIP() instead ?

go/dhcp/README.md
```diff
@@ -38,20 +45,20 @@ curl http://127.0.0.1:22222/api/v1/dhcp/stats/eth1.137 | python -m json.tool
...
-        "Options": {
-            "OptionDomainName": "inlinel2.fabianfence",
-            "OptionDomainNameServer": "10.10.0.1",
-            "OptionIPAddressLeaseTime": "123",
-            "OptionNetBIOSOverTCPIPNameServer": "172.20.135.2",
-            "OptionRouter": "192.168.0.1",
-            "OptionSubnetMask": "255.255.255.0"
+        "network": "192.168.0.0/24",
+        "options": {
+            "optionDomainName": "inlinel2.fabianfence",
```
fdurand, line 56: I think it will return OptionDomainName because we use https://github.com/krolaw/dhcp4/blob/master/packet.go

go/dhcp/main.go
```diff
@@ -196,9 +195,11 @@ func (h *Interface) run(jobs chan job) {
 					statistics = v.dhcpHandler.available.Stats()
 					var Options map[string]string
 					Options = make(map[string]string)
-					Options["OptionIPAddressLeaseTime"] = v.dhcpHandler.leaseDuration.String()
+					Options["optionIPAddressLeaseTime"] = v.dhcpHandler.leaseDuration.String()
```
fdurand, line 198: https://github.com/krolaw/dhcp4/blob/master/packet.go#L301

fdurand (changes requested):

go/dhcp/README.md
satkunas: See commit https://github.com/inverse-inc/packetfence/pull/2959/commits/e46b342e83d3f5989febcc3d69922d9276363e89

Tested that it returns first-char as lowercase.

extrafu:
@fdurand and @julsemaan are we all good on this PR?

julsemaan:
Other than the necessary rebase, this looks good to me

julsemaan:
I've performed the rebase.

@fdurand can perform the final check and merge if its all good

PR #2961, opened 2018-02-20
Feature/api frontend token info

julsemaan:
# Description
Allow to obtain the information about the current token via an API call

# Impacts
The API frontend

# Delete branch after merge
YES

### DIFF
--- a/docs/api/spec/paths/token_info.yaml
+++ b/docs/api/spec/paths/token_info.yaml
@@ -1,0 +1,23 @@
+/token_info:
+  post:
+    summary: Get information on the current token
+    responses:
+      '200':
+        description: Successfully found information for the token 
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                tenant_id:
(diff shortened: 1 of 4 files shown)


docs/api/spec/paths/token_info.yaml
```diff
@@ -0,0 +1,23 @@
...
+    summary: Get information on the current token
+    responses:
+      '200':
+        description: Successfully found information for the token 
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                tenant_id:
```
cgx, line 12: Isn't this information redundant? Calling `/tenants` should already return the list of accessible tenants for the specified token.

julsemaan: We haven't fully stated on it yet, but I believe the tenants API endpoint will be limited to the admin tokens.

So in that case, that would be the only way to get the current tenant ID

And also, this provides the special '0' tenant ID in the reply which indicates you can modify data on any tenant which the /tenants won't provide

PR #2966, opened 2018-02-21
Feature/api authentication pf

julsemaan:
# Description
Adds the ability to authenticate API users based on the internal sources of PacketFence

# Impacts
The API login

# Delete branch after merge
YES


julsemaan:
This is ready for review

extrafu:
@fdurand Please review and merge

lib/pf/UnifiedApi/Controller/Authentication.pm
```diff
@@ -0,0 +1,68 @@
...
+    
+    my ($status, $json) = $self->parse_json;
+    if (is_error($status)) {
+        $self->render(status => $status, json => $json);
+    }
+
+    my ($result, $roles, $tenant_id) = pf::authentication::adminAuthentication($json->{username}, $json->{password});
+
+    if($result == $LOGIN_SUCCESS) {
+        $self->render(status => 200, json => { result => $result, roles => $roles, tenant_id => ($tenant_id + 0) });
```
fdurand: "+ 0" ?

lib/pf/authentication.pm
```diff
@@ -442,6 +442,48 @@ sub getAuthenticationClassByType {
...
+            'password' => $password, 
+            'rule_class' => $Rules::ADMIN,
+            'context' => $pf::constants::realm::ADMIN_CONTEXT,
+        }, $source);
+    if ($result) {
+        if ($result == $LOGIN_CHALLENGE ) {
+          return $LOGIN_CHALLENGE;
+        }
+
+        #my $value = pf::authentication::match($source_id, { username => $user, 'rule_class' => $Rules::ADMIN , 'context' => $pf::constants::realm::ADMIN_CONTEXT}, $Actions::SET_ACCESS_LEVEL, undef, $extra);
```
fdurand: remove

lib/pf/UnifiedApi/Controller/Authentication.pm
julsemaan: Unfortunately yes, otherwise there is no guarantee that this will be an integer when passed to the JSON library for encoding. Then if its a string containing a number, the JSON decode will not work properly in Golang.

cgx: You could use the `int` function for clarity.

julsemaan:
Adjusted based on comments + rebased

Please give it another review

extrafu:
@fdurand Please do one more pass and if all is fine, please merge.

PR #2970, opened 2018-02-26
fixed typo /ect to /etc

atran-inverse:
# Description
fixed typo /ect to /etc on line 120

# Delete branch after merge
YES


extrafu:
No need to do a PR for such trivial change. In the future, please just push it to devel.

PR #2983, opened 2018-03-01
feature/pfmon-worker-limits

### LINKED ISSUE #2925: pfmon: Workers leak memory
When dealing with a system that has a high uptime, I've just seen the pfmon workers using a few GB worth of RAM.

Since the workers are never "rotated", then any minor leak leads up to a lot of RAM wasted after a few months.

Instead of tracking down all the little leaks that can occur, I'd suggest we do the same think as with pfqueue and restart the pfmon workers after they've ran X times their job.

If possible, we should also try to backport this to the 7.x maintenance branches if the fix isn't too big

jrouzierinverse:
Description
===========
Make the pfmon workers die off after 1000 +- 100 runs to avoid memory leaks.

Impacts
=======
pfmon

Issue
=====
Fixes #2925

NEWS file entries
=================

Enhancements
------------
* To avoid memory leaks have pfmon worker die after 1000 +- 100 runs

Bug Fixes
------------

* Workers leak memory (#2925)

### DIFF
--- a/sbin/pfmon
+++ b/sbin/pfmon
@@ -73,6 +73,8 @@ our $IS_CHILD = 0;
 our %CHILDREN;
 our @TASKS_RUN;
 our $ALARM_RECV = 0;
+our $DEFAULT_TASK_COUNT = 100;
+our $DEFAULT_TASK_JITTER = 10;
 
 my $logger = get_logger( $PROGRAM_NAME );
 my $old_child_sigaction = POSIX::SigAction->new;
@@ -137,6 +139,7 @@ flock($fh, LOCK_EX | LOCK_NB) or die "cannot lock $pidfile another pfmon is runn
 $HAS_LOCK = 1;
 
 our $running = 1;
+our $TASKS   = 0;
 our $process = 0;
 
 # standard signals and daemonize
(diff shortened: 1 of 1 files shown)


sbin/pfmon
```diff
@@ -73,6 +73,8 @@ our $IS_CHILD = 0;
 our %CHILDREN;
 our @TASKS_RUN;
 our $ALARM_RECV = 0;
+our $DEFAULT_TASK_COUNT = 1000;
```
julsemaan: IMO, this might be better at 100 for tasks that don't run frequently.

sbin/pfmon
```diff
@@ -73,6 +73,8 @@ our $IS_CHILD = 0;
...
+our $DEFAULT_TASK_JITTER = 100;
```
julsemaan: And then this would be 10 if the above became 100

jrouzierinverse: We could configure this easily per task.
The question is do we need that type of granularity?

julsemaan: IMO, this would be good enough since it fixes the issue and the leak seemed to happen in the workers that were performing frequent tasks, but if you feel like it, I see no issue in having a per task limit

jrouzierinverse: I will add it when we need it.

jrouzierinverse:
This has been fixed.
Let me know if you want me to put it into maintenance.

julsemaan:
IMO, that would be a good fit for the maintenance since its pretty self-contained

I'll merge now and perform the cherry-pick since you kept it to a single commit that I can put there without stealing your work

https://www.youtube.com/watch?v=toL1tXrLA1c

PR #2992, opened 2018-03-08
feature/api-search

jrouzierinverse:
Description
===========
Search API

Impacts
=======
pfunified_api

Delete branch after merge
=========================
NO


julsemaan:
If its ready for review, could you mark it as so with the tags.

Just so I don't comment on it for stuff you're already fixing

extrafu:
@julsemaan Looks good? The idea is to merge this quickly so we can start using it from the GUI.

We can fix any issue that arise directly in devel.

t/unittest/UnifiedApi/Controller/Locationlogs.t
```diff
@@ -73,16 +73,214 @@ $t->get_ok('/api/v1/locationlogs' => json => { })
...
+    '/api/v1/locationlogs/search' => json => { limit => 267 }
+  )
+  ->status_is(200)
+  ->json_hasnt('/nextCursor')
+  ->json_is('/prevCursor', 0)
+;
+$items = $t->tx->res->json->{items};
+
+is(@$items, 256, "Returns 256 items");
+
```
julsemaan, line 283: Maybe I'm blind but we're missing a test for using the cursor parameter for pagination

julsemaan (commented):
Just make sure that pagination is well tested in search and we're good for merge

t/unittest/UnifiedApi/Controller/Locationlogs.t
jrouzierinverse: https://github.com/inverse-inc/packetfence/pull/2992/files/4b0d2a630f5e269803c225d5dc740ca63d359413#diff-acbd078bd53801f2f2fc209e5c36a09fR230

PR #2999, opened 2018-03-09
OpenAPI Doc Updates

satkunas:
# Description
Various OpenAPI spec documentation updates/fixes

# Impacts
Changes
 * https://github.com/inverse-inc/packetfence/blob/devel/docs/api/spec/openapi.yaml
 * https://github.com/inverse-inc/packetfence/blob/devel/docs/api/spec/openapi.json

# Delete branch after merge
YES

### DIFF
--- a/docs/api/spec/components/schemas/dhcp.yaml
+++ b/docs/api/spec/components/schemas/dhcp.yaml
@@ -58,7 +58,7 @@ DhcpInterfaceStats:
                     description: A collection of MAC address and IPv4 address pairs
                     properties:
                         type: object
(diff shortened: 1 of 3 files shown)


julsemaan:
I think the changes in this PR are included in #3002 

If that is the case, either we close this PR and only focus on the other one, or the changes from this PR should be excluded from #3002. Simplest would be to focus on #3002 but its up to you.

satkunas:
@julsemaan yes, these changes also made it in #3002. I'm closing this PR.

PR #3001, opened 2018-03-09
Add PF support to Pip-Boy-3000

satkunas:
# Description
Add Packetfence support to Pip-Boy 3000.

# Delete branch after merge
YES

# NEWS file entries
## New Features
* The player character is equipped with the Pip-Boy 3000A; it is indicated that there are other models in the 3000 line, with A being one of the oldest. The Pip-Boy 3000A is worn on the character's left forearm. This multifunction device has been programmed by Pip-Boy developers to display the wearer's stats, area maps, inventory and item properties. It also includes a Geiger counter and a radio. Its screen doubles as a flashlight for fixing visibility issues in dark or obscured areas (enabled by holding down the Pip-Boy button). A user can set their color preference for the display. The four colors available are amber, blue, green and white. The display color must be changed through the pause menu rather than the Pip-Boy itself.


TomInverse:
Same for the T-61 armors detected as the T-51 if you don't have both shoulders shield.

PR #3002, opened 2018-03-11
Remap DHCP JSON response to accommodate draft-04 of the API specification

satkunas:
# Description
Remap the DHCP output JSON to avoid using dynamic keys. Only static keys are currently allowed in draft-04 of the API specification.

> In normal JSON Schema you could use patternProperties, but OpenAPI forbids it. If JSON Schema draft-06 or later were supported you could use propertyNames, but OpenAPI is based on draft-04
> Ref: [https://github.com/OAI/OpenAPI-Specification/issues/1505#issuecomment-372054402](url)

# Impacts
* Changes pfdhcp JSON response on `/dhcp/stats/{interface}` and `/dhcp/initialease/{interface}` endpoints.

# Delete branch after merge
YES

## Bug Fixes
* https://github.com/OAI/OpenAPI-Specification/issues/1505#issuecomment-372054402


docs/api/spec/components/schemas/dhcp.yaml
```diff
@@ -12,59 +12,62 @@ DhcpMacIp:
...
-                            description: Domain name
-                        optionDomainNameServer:
-                            type: string
-                            description: DNS IPv4 addresses
-                        optionIPAddressLeaseTime:
+                        mac:
                             type: string
-                            description: Lease time
-                        optionRouter:
+                            description: The IPv4 address of the dhcp member
```
julsemaan: I think the IP and MAC description are mixed up

julsemaan:
And actually, you were right, I was wrong for the openapi.json + openapi.yaml in the PRs, they shouldn't be included, when we talked I had devel in mind.

So to fix it:
```
# git checkout bugfix/api-dhcp
# git fetch origin
# git rebase origin/devel
# git checkout origin/devel docs/api/spec/openapi.json
# git checkout origin/devel docs/api/spec/openapi.yaml
# git commit -a -m "undo crap @julsemaan told me to do"
## The line above may say that there is nothing to commit, if that is the case, its fine, just ignore it
# git push origin bugfix/api-dhcp --force
```

julsemaan:
Also, very minor detail, the fix branches should be named `fix/prevent-zammitbug` and not `bugfix/prevent-zammitbug`

docs/api/spec/components/schemas/dhcp.yaml
```diff
@@ -12,59 +12,62 @@ DhcpMacIp:
...
-                    type: string
-                    description: Network address
-                free:
-                    type: integer
-                    description: Available number of IPv4 addresses in the dhcp scope
-                category:
-                    type: string
-                    description: Associated role of the dhcp scope
-                options:
+        description: Statistics of the dhcp scope
```
extrafu: dhcp -> DHCP everywhere

docs/api/spec/components/schemas/dhcp.yaml
satkunas: Looks good on my end

> # curl http://localhost:22222/api/v1/dhcp/stats/enp0s8
```
[{
    "interface": "enp0s8",
    "network": "10.10.0.0/16",
    "free": 65515,
    "category": "none",
    "options": {
        "optionDomainName": "inlinel2.satkunas.com",
        "optionDomainNameServer": "10.10.0.2",
        "optionIPAddressLeaseTime": "24h0m0s",
        "optionRouter": "10.10.0.2",
        "optionSubnetMask": "255.255.0.0"
    },
    "members": [{
        "mac": "ac:cf:85:26:a0:66",
        "ip": "10.10.0.17"
    }, {
        "mac": "b8:27:eb:91:4e:5b",
        "ip": "10.10.0.12"
    }],
    "status": "Normal"
}]
```

satkunas:
@julsemaan rebase commited. Also, `/s/bigfix/fix` noted.

PR #3013, opened 2018-03-15
feature/api-search-advanced

jrouzierinverse:
Description
===========
Provide a framework for creating joinable fields

### DIFF
--- a/lib/pf/UnifiedApi/SearchBuilder.pm
+++ b/lib/pf/UnifiedApi/SearchBuilder.pm
@@ -24,6 +24,35 @@ our %OP_HAS_SUBQUERIES = (
     'or' => 1,
 );
 
+
+=head2 $self->search($search_info)
+
+Search using search_info
+
+    my ($http_status, $search_result_or_error) = $self->search({
+        'dal' => 'pf::dal::node',
+        'cursor' => '1',
+        'limit' => 25,
+        'query' => {},
+        'fields' => [...],
+        'sort' => [...]
+    });
+
+dal    Class name of dal module. (Mandatory)
+
+cursor The cursor of the where start search. (Optional)
+
+limit  The maximum items to return
+
+query  The query object to use for searching
+
+fields An array of fields to return
+
+sort   An array of the sort order for the search
+
+=cut
+
+
 sub search {
     my ( $self, $search_info ) = @_;
     my ($status, $search_args) = $self->make_search_args($search_info);
@@ -53,6 +82,14 @@ sub search {
       }
       ;
 }
+
+=head2 $self->make_search_args($search_info)
+
+Make the search args from the search_info
+
+    my ($http_status, $search_args_or_error) = $self->make_search_args($search_info);
+
+=cut
 
 sub make_search_args {
     my ($self, $search_info) = @_;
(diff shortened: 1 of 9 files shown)


extrafu:
@julsemaan Please review and merge quickly, we can always adjust a bit after.

lib/pf/UnifiedApi/SearchBuilder.pm
```diff
@@ -0,0 +1,248 @@
...
+
+=head1 NAME
+
+pf::UnifiedApi::SearchBuilder -
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::UnifiedApi::SearchBuilder
```
julsemaan, line 11: Perldoc is required for the subroutines in this module

Given the complex nature of this module, I think it would be good to have solid perldoc in it if someone else than @jrouzierinverse has to play in it

jrouzierinverse:
When discussing the API/doc for search with @cgx 
We found parts of it confusing.
So here are some proposed changes.

## HTTP POST
* All the data is in the body even cursor.
* content-type is JSON.

## HTTP GET 
* All the data is in the query string (as there is no body).
* Using one of the following approaches 

### Flatten dot

 ```query.op=equals&query.field=mac&query.value=00:01:02:03:04:05```

#### Pros
* You can use it with html form 
* Easy to support with HTML::FormHandler
* The axios client supports it
* We are used to it.

#### Cons
* Not as wildly used as the bracket style

### Flatten bracket

```query[op]=equals&query[field]=mac&query[value]=00:01:02:03:04:05```

#### Pros
* You can use it with html form 
* Easy to support with HTML::FormHandler
* The axios client supports it
* More widely used

#### JSON encoded

* ```query=%7B%22op%22%3A%22and%22%2C%22values%22%3A%5B%7B%22field%22%3A%22mac%22%2C%22op%22%3A%22not_equals%22%2C%22value%22%3A%2200%3A03%3Aab%3A18%3A61%3Afc%22%7D%5D%7D```

#### Pros
* The JSON encoded is the easier to parse on the back end and handle.
* The axios client supports it
 
#### Cons
* You cannot use it with html forms.

**Comment by julsemaan** [comment]
IMO I don't think we need to limit ourselves to a syntax that is usable directly with HTML forms

And, not sure either we should have both a POST and a GET, picking one of them would be best IMO so we don't end up with multiple ways of doing the same thing.

Would it work to only use the POST option with everything in the body ?

**Comment by jrouzierinverse** [comment]
Supporting GET would make it easier to support sharable queries.
Supporting POST makes easier to write third party libraries.
Since this is just about data serialization and changing the lying behavior it does not change a lot.
In addition OpenAPI supports deep objects in queries using style=deepObject this style expect brackets.
We have all the pieces to support both formats with ease.

**Comment by julsemaan** [comment]
Even though we have all the pieces to support both, my concern is more on whether or not we should support both and have 2 ways of doing the same thing.

My personal opinion would be to pick one and stick to it (I prefer the POST personally)

**Comment by jrouzierinverse** [comment]
I documented the most non clear part of the SearchBuilder interface.
Which is how to add join fields.
Let me know if this is enough.

**Comment by julsemaan** [comment]
@jrouzierinverse 

I think it would be preferable to get all the subroutines documented at least with a description so we know what each subroutine job is

**Comment by jrouzierinverse** [comment]
The methods names seem pretty clear to on what they do.
Its feel like unnecessary repetition to just repeat exactly what the method name in the poddoc.

```
=head2 make_columns

Make columns array for search args

=cut
```
If a method name is not clear on what it does.
I would prefer to give it a better name.
Then to just repeat the same thing in the documentation.

julsemaan:
IMO we should document all subroutines. Something that is clear to me and you might not be for someone else. That is especially true when the code we're looking at has a lot of abstraction and dynamic variable assignment.

We can always agree to disagree on that one but there are many modules in our codebase that are like black magic to the team so I'm just trying to limit this with that comment.

jrouzierinverse:
If that is the case then I would prefer to document on how to use the function/method as that is more useful than just repeating information that can be easy understood by look at the method/function name.

jrouzierinverse:
Done

extrafu:
@julsemaan We merge?

PR #3014, opened 2018-03-15
Add support for filtering google oauth login for specific google-apps domainms

garci66:
# Description
Adds a "hosted domain" field to the Google OAuth authentication so that the user is presented with a login field with the domain of the organization and the returend token is verified to match the specific domain as well (via the HD attribute). More details here: [(https://developers.google.com/identity/protocols/OpenIDConnect#hd-param)]

This is very useful for enterprise or education uses where only employees / students should be able to login and not just anyone with a google account

# Impacts
Oauth workflow - callback is altered to verify the returned hosted domain attribute. Additionally, the google oauth class is extended to support the additional_client_attributes method so that the "hd" field is added on the outgoing oauth request (this allows google to present a login screen with only the accepted domain's accounts) 

# Code / PR Dependencies
The second part (of google presenting the login prompt with only the hosted domain) requires a minimal patch on the Net::OAuth2 library. This patch has not yet been pushed to the library but is available here:
[https://github.com/garci66/NetOauth2/commit/c4c97237e3870f3b606efa2c17407638f17d4532]
This patch is not mandatory. IF not applied, the login screen will be the typical google login screen and could contain multiple google accounts (@gmail, your own google apps account, etc). The returned oauth ticket will still be verified by packetfence to belong to the correct domain, but the user experience will be less than ideal as he could try - and fail- to login with his gmail account instead of his school  / work / whatever.
The proposed patch to Net::OAuth2 has been tested in production during the last few weeks without issues.

It was implemented identically to the "state" parameter used for linked-in OAuth logins. 

# NEW Package(s) required
Updated Net::OAuth2 library. (change has not yet been pushed upstream). Changes have been proposed to Mark Overmeer (maintainer of Oauth2) but I still need to hear from hi,

# Delete branch after merge
YES 

# NEWS file entries
## Enhancements
* Google OAuth now supports filtering for specific google apps domains

# UPGRADE file entries
Potentially upgrade Net::OAuth2 or require a newer version (0.64 proposed but not yet accepted)

# NOTE:
Not sure why the commits by @julsemaan got added and not 100% sure how to remove them. Sorry for the additional noise in the PR.

### DIFF
--- a/db/pf-schema-7.4.0.sql
+++ b/db/pf-schema-7.4.0.sql
@@ -3,8 +3,8 @@
 --
 
 SET @MAJOR_VERSION = 7;
-SET @MINOR_VERSION = 3;
-SET @SUBMINOR_VERSION = 9;
+SET @MINOR_VERSION = 4;
+SET @SUBMINOR_VERSION = 0;
 
 --
 -- The VERSION_INT to ensure proper ordering of the version in queries
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/OAuth/Google.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/OAuth/Google.pm
@@ -13,7 +13,61 @@ Google OAuth module
 use Moose;
 extends 'captiveportal::DynamicRouting::Module::Authentication::OAuth';
 
+use pf::log;
+use pf::auth_log;
+
 has '+source' => (isa => 'pf::Authentication::Source::GoogleSource');
+
+=head2 handle_callback
+
+Google override to handle the callback from the OAuth2 provider and fetch the protected resource
+
+=cut
+
+sub handle_callback {
+    my ($self) = @_;
+
+    my $token = $self->get_token();
+    return unless($token);
+
+    # request a JSON response
+    my $h = HTTP::Headers->new( 'x-li-format' => 'json' );
+    my $response = $token->get($self->source->{'protected_resource_url'}, $h ); 
+
+    if ($response->is_success) {
+        my $info = $self->_decode_response($response); 
+        my $pid = $self->_extract_username_from_response($info);
+        my $hd = $info->{hd};
+        my $hosted_domain = $self->source->hosted_domain;
+        # get_logger->info(sub { use Data::Dumper; "OAuth2  response : ".Dumper($info) });   
+        if ( defined $hosted_domain && $hosted_domain ne "" && (!defined $hd || $hosted_domain ne $hd)){
+            get_logger->info("OAuth2: google domain for user: ".$pid." did not match the allowed domain: ".$hosted_domain);
+            get_logger->debug(sub { use Data::Dumper; "OAuth2 failed response : ".Dumper($response) });
+            pf::auth_log::change_record_status($self->source->id, $self->current_mac, $pf::auth_log::FAILED);
+            $self->app->flash->{error} = "OAuth2: google domain did not match the allowed domain: ".$hosted_domain ;
+            $self->landing();
+            return;    
+        }
+        
+        $self->username($pid);
+
+        get_logger->info("OAuth2 successfull for username ".$self->username);
+        $self->source->lookup_from_provider_info($self->username, $info);
+        
+        pf::auth_log::record_completed_oauth($self->source->id, $self->current_mac, $pid, $pf::auth_log::COMPLETED);
+
+        $self->done();
+    }
+    else {
+        get_logger->info("OAuth2: failed to validate the token, redireting to login page.");
+        get_logger->debug(sub { use Data::Dumper; "OAuth2 failed response : ".Dumper($response) });
+        pf::auth_log::change_record_status($self->source->id, $self->current_mac, $pf::auth_log::FAILED);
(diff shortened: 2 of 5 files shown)


html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/OAuth/Google.pm
```diff
@@ -13,8 +13,62 @@ Google OAuth module
...
+    my $h = HTTP::Headers->new( 'x-li-format' => 'json' );
+    my $response = $token->get($self->source->{'protected_resource_url'}, $h ); 
+
+    if ($response->is_success) {
+        my $info = $self->_decode_response($response); 
+        my $pid = $self->_extract_username_from_response($info);
+        my $hd = $info->{hd};
+        my $hosted_domain = $self->source->hosted_domain;
+        # get_logger->info(sub { use Data::Dumper; "OAuth2  response : ".Dumper($info) });   
+        if ( defined $hosted_domain && $hosted_domain ne "" && (!defined $hd || $hosted_domain ne $hd)){
```
julsemaan: It would be better to modify the OAuth.pm authentication module so that it calls a filter_username function that would be overriden in this module to have this logic.

That way, this doesn't end up with duplicated code

ChangeLog
```diff
@@ -1,19 +1,35 @@
-commit 97d937cbca119b023d866871ccb697741e9dbbc3
+commit 6049670e742fa4983aa0473114f156d343f7c497
```
julsemaan, line 1: This here shouldn't be there.

This should be reverted to the devel version:
```
# git checkout origin/devel Changelog
```

db/pf-schema-7.4.0.sql
```diff
@@ -3,8 +3,8 @@
 --
 
 SET @MAJOR_VERSION = 7;
-SET @MINOR_VERSION = 3;
```
julsemaan, line 6: This here shouldn't be there.

This should be reverted to the devel version:
```
# git checkout origin/devel db/pf-schema-7.4.0.sql
```

lib/pf/Authentication/Source/GoogleSource.pm
```diff
@@ -45,6 +46,21 @@ sub lookup_from_provider_info {
...
 
+
+=head2 additional_client_attributes
+
+Supply a google-apps domain filter (hd parameter). See below
+https://developers.google.com/identity/protocols/OpenIDConnect#hd-param
+
+=cut
+
+sub additional_client_attributes {
```
julsemaan, line 57: This subroutine doesn't seem to be used anywhere, is this because I'm missing some of the magic from Net::OAuth2 ?

julsemaan (commented):
A few adjustments to do, but its a neat feature!

We'll need to hold-on to the new Net::OAuth2 version before this is merged as well

html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/OAuth/Google.pm
garci66: Sounds good. I'll try to refactor it in this way. Perl is far from my most proficient language thus why the copy-paste. I will try to refactor

julsemaan: Sounds good, I can always give it a shot if you struggle too much

lib/pf/Authentication/Source/GoogleSource.pm
garci66: Its part of the OAuthSource.pm and its used by the linkedin Oauth module to pass an additional parameter needed by it. I copied that same model to pass the HD attribute to google.

garci66:
Additionally, the required changes in Net::OAuth2 have been accepted by @markov2 in his code (which is now in Github). Link to the repo here:

https://github.com/markov2/perl5-Net-OAuth2

(not sure how to deal with making the new version of oauth2 a requirement or "recommended" install )

julsemaan:
@garci66 

We can build the package ourselves and put it in our public repo so that part isn't a problem.

garci66:
I am still struggling to get rid of the ChangeLog and schema commits. I did the checkoup them from upstream/devel but that doesn't seem to have reverted the commit. I'll keep digging.

julsemaan:
Doing the checkout should reset the file which you should then commit

Maybe that the only piece missing

extrafu:
No feedback, pushing to v8.1.

garci66:
Sorry for the delay. I will try to push this asap.

extrafu:
Still waiting for updates, pushing to v9.1.

extrafu:
Still waiting for updates, pushing to v9.2

extrafu:
@julsemaan Take what's best, scrap the rest.

julsemaan:
It seems this would require quite a bit of testing which we can't accomplish.

Also requires work in the new admin

All of this for something there isn't much interest for

@garci66 If you want to update the PR because its still relevant to you, please adjust it for v9 and reopen the PR

garci66:
Hi Julien

So far I've been keeping this as a private branch where I keep adapting as
needed. I will try to work on the PR when I have some cycles but it won't
be anytime soon. I will try to adapt this to version 9 to see what changes
have taken place.

cheers

PR #3015, opened 2018-03-16
Add Golang interval handler

satkunas:
# Description
Custom golang interval handler
* Concurrent
* Utilizes locking to prevent multiple instances of the same interval
* Adjusts interval offset to account for chained processing lag to ensure there's no incremental time drift
* Uses UTC to prevent system time/clock changes from affecting the interval timing
* Supports time.parseDuration with nanosecond precision

# Delete branch after merge
YES

### DIFF
--- a/go/interval/interval.go
+++ b/go/interval/interval.go
@@ -41,17 +41,17 @@ func (r *recurrent) nextRun() (time.Duration, error) {
 }
 
 func Every(duration string) *Job {
-    t, err := time.ParseDuration(duration)
-    if err != nil {
-        return &Job{err: errors.New("Could not parse duration: " + duration)}
-    }
-    r := new(recurrent)
-    r.started = time.Now().UTC()
-    r.count = 0
-    r.delay = t.Nanoseconds()
-    j := new(Job)
-    j.schedule = r
-    return j
+	t, err := time.ParseDuration(duration)
+	if err != nil {
+		return &Job{err: errors.New("Could not parse duration: " + duration)}
+	}
+	r := new(recurrent)
+	r.started = time.Now().UTC()
+	r.count = 0
+	r.delay = t.Nanoseconds()
+	j := new(Job)
+	j.schedule = r
+	return j
 }
 
 func (j *Job) Run(f func()) (*Job, error) {
@@ -89,7 +89,7 @@ func runJob(job *Job) {
 		job.err = errors.New("bad function chaining")
 		return
 	}
-    rj.count += 1
+	rj.count += 1
 	if job.IsRunning() {
 		return
 	}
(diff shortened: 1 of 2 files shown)


satkunas:
Please merge ASAP as this will be required for `pfstats`

extrafu:
@julsemaan Please review and merge quickly, we can always adjust a bit after.

go/interval/interval.go
```diff
@@ -0,0 +1,111 @@
...
+type recurrent struct {
+	delay   int64
+	started time.Time
+	count   int64
+	done    bool
+}
+
+func (r *recurrent) nextRun() (time.Duration, error) {
+	if r.delay == 0 {
+		return 0, errors.New("invlaid delay")
```
julsemaan: *invalid

julsemaan: Also wouldn't a negative delay be invalid as well ?

go/interval/interval.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	//adjust offset for processing time drift
+	offset := (r.count * r.delay) - time.Since(r.started).Nanoseconds()
+	return time.Duration(r.delay + offset), nil
+}
+
+func Every(duration string) *Job {
+	t, err := time.ParseDuration(duration)
+	if err != nil {
+		return &Job{err: errors.New("Could not parse duration: " + duration)}
```
julsemaan, line 46: The golang way would be to return (*Job, error) and not have errors in the job struct

Anyway the external caller can't even access err because its private

julsemaan (commented):
We also need unit tests for this new lib

julsemaan:
@satkunas 

You could always build this and pfstats into the same PR since we don't need this for anything else AFAIK

extrafu:
I agree with what @julsemaan just said - and it'll be easier to get gradual process on the whole thing and adjust as you go. Instead of dealing with tons of PR, merge issues, delays, etc.

satkunas:
@julsemaan @extrafu putting it here allowed me to include the `README.md` along with it. 

If you'd prefer to leave the `README.md` out, then I can build it in w/ `pfstats`

Another option could be for me to create a personal repo and include "github.com/satkunas/interval" instead.

go/interval/interval.go
satkunas: It's been adjusted. The error returns fine w/ the given example from the `README` as `err` is deferred (private scope) and picked up within `interval.Run()`

julsemaan:
@satkunas 

You could include the readme with your pfstats daemon that should be fine

Also a bit of unit testing for this would be good

go/interval/interval.go
julsemaan: Still not our standard way to do it though

And consumers from other packages won't be able to access err so it should definitely be put as a return argument instead

satkunas:
Closing, superseded by https://github.com/inverse-inc/packetfence/pull/3024

PR #3024, opened 2018-03-19
Add statsd metrics

### LINKED ISSUE #3017: pfstats runtime error
Details:
* Using most recent `devel` branch.
* Error occurs regardless of the contents of `conf/stats.conf`, or whether there's a `file=...` directive in the configuration file, or not.
* `bin/pfcmd pfconfig show config::Stats` successfully displays the contents of `conf/stats.conf`
* `bin/pfcmd pfconfig reload` has no effect.

The error appears to be coming from:

https://github.com/inverse-inc/packetfence/blob/642c0598b14d282378ec29018e8a54924b2d94a4/go/stats/main.go#L345

Tail from `logs/pfstats.log`, seen immediately after (re)starting pfstats:

> Mar 17 10:36:00 packetfence pfstats: panic: runtime error: invalid memory address or nil pointer dereference
> Mar 17 10:36:00 packetfence pfstats: [signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x575657]
> Mar 17 10:36:00 packetfence pfstats: goroutine 34 [running]:
> Mar 17 10:36:00 packetfence pfstats: github.com/inverse-inc/packetfence/go/log.LoggerWContext(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
> Mar 17 10:36:00 packetfence pfstats: /root/gospace/src/github.com/inverse-inc/packetfence/go/log/log.go:128 +0x37
> Mar 17 10:36:00 packetfence pfstats: main.tailFile(0xbea367d421eb625d, 0x6c81d0, 0x9254e0, 0x0, 0x0, 0xc4200ad490, 0x9, 0x0, 0x0, 0x0, ...)
> Mar 17 10:36:00 packetfence pfstats: /root/gospace/src/github.com/inverse-inc/packetfence/go/stats/main.go:407 +0x4e
> Mar 17 10:36:00 packetfence pfstats: created by main.main
> Mar 17 10:36:00 packetfence pfstats: /root/gospace/src/github.com/inverse-inc/packetfence/go/stats/main.go:388 +0x628

satkunas:
# Description
Adds statsd metrics

Basic Example `conf/`statsd.conf`:
```
[metric 'sample metric 1']
type=mysql_query
statsd_type=gauge
statsd_ns=source.statsd.test.1
# mysql_query must return a single column with a single row
mysql_query=select 0
interval=60s

[metric 'sample metric 2']
type=mysql_query
statsd_type=gauge
statsd_ns=source.statsd.test.2
# mysql_query must return a single column with a single row
mysql_query=select 1
interval=30s

[metric 'sample metric 3']
type=mysql_query
statsd_type=gauge
statsd_ns=source.statsd.test.3
# mysql_query must return a single column with a single row
mysql_query=select 3
interval=20s
```

# Impacts
Pushes metrics to statsd.
See:
* `/var/cache/netdata`
* http://packetfence:19999/netdata.conf
* http://packetfence:19999/api/v1/allmetrics
* http://packetfence:19999/api/v1/charts

# Replaces
https://github.com/inverse-inc/packetfence/pull/3015

# Issue
Fixes #3017 

# Delete branch after merge
YES 

# NEWS file entries
## New Features
* Add pfstats metric collections from mysql

## Bug Fixes
* Issue with pfstats runtime error (#3017)


go/stats/config.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
+)
+
+// readDBConfig read pfconfig database configuration
+func readDBConfig() pfconfigdriver.PfConfDatabase {
```
julsemaan: Use the database package instead to avoid code duplication

go/stats/main.go
```diff
@@ -334,19 +334,31 @@ func main() {
 	var keyConfStats pfconfigdriver.PfconfigKeys
 	keyConfStats.PfconfigNS = "config::Stats"
 	pfconfigdriver.FetchDecodeSocket(ctx, &keyConfStats)
+    re_metric := regexp.MustCompile("^metric .*")
```
julsemaan: Use camel casing in the Golang codebase

go/stats/main.go
```diff
@@ -334,19 +334,31 @@ func main() {
...
 
-		go tailFile(ConfStat, config, done)
+		if re_metric.MatchString(key) {
+			err = ProcessMetricConfig(ctx, ConfStat)
+			if err != nil {
+				log.LoggerWContext(ctx).Error(err.Error())
+			}
+		}
+
+		if ConfStat.File != "" {
```
julsemaan: I'd move this to ProcessMetricConfig with the type being `file`

go/stats/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+
+var MySQLdatabase *sql.DB
+
+func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error {
+
+	// Read DB config
+	configDatabase := readDBConfig()
+	connectDB(configDatabase, MySQLdatabase)
+	MySQLdatabase.SetMaxIdleConns(0)
+	MySQLdatabase.SetMaxOpenConns(500)
```
julsemaan: use the database package instead

go/stats/utils.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"database/sql"
+	"github.com/inverse-inc/packetfence/go/database"
+	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
+)
+
+// connectDB connect to the database
+func connectDB(configDatabase pfconfigdriver.PfConfDatabase, db *sql.DB) {
```
julsemaan: use the database package instead

julsemaan (changes requested):

go/stats/config.go
satkunas: Replicated from: https://github.com/inverse-inc/packetfence/blob/devel/go/dhcp/config.go#L57

I'm not sure what you mean. What replication are we trying to avoid? Do we want a separate struct from https://github.com/inverse-inc/packetfence/blob/devel/go/pfconfigdriver/structs.go#L110 within the stats package?

go/stats/main.go
satkunas: Conditional was added here to address to fix runtime error in #3017 and has nothing to do with [metric .*] pfconfigdriver keys from `conf/stats.conf`

go/stats/metrics.go
satkunas: Unclear, same response as https://github.com/inverse-inc/packetfence/pull/3024#discussion_r175586997

go/stats/utils.go
satkunas: Unclear, same response as https://github.com/inverse-inc/packetfence/pull/3024#discussion_r175586997

go/stats/main.go
julsemaan: Since we haven't started seriously to configure metrics and @fdurand left none in the example file, then I believe its safe to migrate this to the "mini-framework" you created

go/stats/config.go
julsemaan: Here is the proper way to connect to the DB, the code in dhcp/ was missed from my refactor.

```
import "github.com/inverse-inc/packetfence/go/database"
ctx := context.Background() // Or your current context
pfconfigdriver.PfconfigPool.AddStruct(ctx, &pfconfigdriver.Config.PfConf.Database)
database := database.ConnectFromConfig(pfconfigdriver.Config.PfConf.Database)
```

That way we don't duplicate calls to pfconfig and have a ton of readDBConfig methods everywhere, and it allows to configure the connection options in a single place

go/stats/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+
+var MySQLdatabase *sql.DB
+
+func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error {
+
+	// Read DB config
+	pfconfigdriver.PfconfigPool.AddStruct(ctx, &pfconfigdriver.Config.PfConf.Database)
+	configDatabase := pfconfigdriver.Config.PfConf.Database
+	connectDB(configDatabase)
+	MySQLdatabase.SetMaxIdleConns(0)
```
julsemaan: Lets remove this line since its already being configured in database/connection.go

go/stats/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	MySQLdatabase.SetMaxOpenConns(500)
```
julsemaan: Same as my comment above

go/stats/metrics.go
```diff
@@ -0,0 +1,82 @@
...
+	case "histogram":
+		StatsdClient.Histogram(conf.StatsdNS, metric)
+		break
+
+	case "unique":
+		StatsdClient.Unique(conf.StatsdNS, metric)
+		break
+
+	default:
+		log.LoggerWContext(ctx).Info("Unhandled statsd type: " + conf.StatsdType)
```
julsemaan: That would be a warn

go/stats/metrics.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+var MySQLdatabase *sql.DB
+
+func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error {
+
+	// Read DB config
+	pfconfigdriver.PfconfigPool.AddStruct(ctx, &pfconfigdriver.Config.PfConf.Database)
+	configDatabase := pfconfigdriver.Config.PfConf.Database
+	connectDB(configDatabase)
```
julsemaan: Seems you're connecting to the database for each metric defined in the file.

This should be moved into main.go and called outside the loop

go/stats/metrics.go
```diff
@@ -0,0 +1,82 @@
...
+					return
+				}
+				SendMetricConfig(ctx, conf, result)
+			}
+			return
+		}
+		break
+
+	default:
+		log.LoggerWContext(ctx).Info("Unhandled type: " + conf.Type)
```
julsemaan: This is more a warn

conf/stats.conf.example
```diff
@@ -0,0 +1,23 @@
+[metric 'sample metric count']
+type=mysql_query
+statsd_type=count
+statsd_ns=source.statsd.test.count
+# mysql_query must return a single column with a single row
```
julsemaan: Single column seems to be true but your code seems to handle multiple rows from what I can tell.

Lets make it consistent, whether its by limiting it to a single row or by adjusting the comments here

julsemaan (changes requested):

go/interval/README.md
```diff
@@ -0,0 +1,53 @@
...
+		return
+	}
+
+	// Keep the program from not exiting.
+	runtime.Goexit()
+}
+```
+
+## How it works?
+By specifying the chain of calls, a `Job` struct is instantiated and a goroutine is starts observing the `Job`.
```
extrafu: and a Goroutine is started

go/interval/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The goroutine will be on pause until:
```
extrafu: The Goroutine

extrafu:
@satkunas Let's make the small required adjustments and start pushing actual real tests in that PR.

Start with SQL-based ones:
- [x] total online devices
- [ ] online devices wired network
- [ ] online devices WiFi network
- [x] total offline devices
- [x] number of new registered devices during the past hour
- [x] number of new registered devices during the past day
- [x] number of new registered devices during the past week
- [x] number of new registered devices during the past month
- [x] number of new registered devices during the past year
- [x] number of devices currently online and in the registration network
- [x] number of devices currently registered
- [x] number of currently open violations

satkunas:
@extrafu can we get either @julsemaan or @jrouzierinverse to indicate to me which tables/columns are needed (faster) to query each of these, or would you prefer than I figure it out (slower)?

extrafu:
For the first four ones, have a look at what we have right now - we have from the Nodes tab an online/offline feature. I shouldn't be hard for you to grab that and adjust it.

julsemaan:
@satkunas 

For the "number of registered devices in the last .*"

You can look at the auth_log table and the completed_at column filtering by the httpd.portal process and we could have  a separate metric for httpd.aaa process so we can track the amount of 802.1x registration and portal registrations separately

satkunas:
I need some advice on the following 2 queries:
 * online devices wired network
 * online devices WiFi network

I noticed the `radacct`.`nasporttype` column. Do we want to add a `FULLTEXT` index here and utilize SQL `LIKE`, or is there a known better way to make a distinction between wired/wireless nodes?

julsemaan:
A full text index on the radacct table might shoot us in the foot because that table tends to have a lot of data.

I'd probably not differentiate both for now and stick to the amount of online devices unless @lzammit or @extrafu says otherwise

extrafu:
@julsemaan @satkunas @lzammit Let's KISS for now and we'll explore more metrics as we go. I agree that having a fulltext index in the accounting data will likely be bad.

extrafu:
I would also add:
- [x] number of successful authentications during the day
- [x] number of failed authentications during the day
- [x] number of free IP addresses in the registration network

extrafu:
Also for documentation purposes, we would like to add this badge:
- [ ] number of registered devices assigned to a role

So if we have 5 roles, we can have a badge for each of them (Employee: 15  Consultant: 2, etc..)

We should generate dynamically this when PacketFence's configuration changes (new role, delete role, change role name). We could get some inspiration from @fdurand automated ping checks when new authentication sources are added.

satkunas:
@extrafu re: "number of free IP addresses in the registration network"

For this we'll utilize the unifiedAPI to access the DHCP API, in order to access the available networks and the address pool size(s).

I'll build this in to `go/pfstats` statically since these are not exposed to mysql.

go/stats/main.go
```diff
@@ -376,27 +359,6 @@ func main() {
 	}
 }
 
-func tailFile(stats pfconfigdriver.PfStats, config tail.Config, done chan bool) {
```
julsemaan, line 362: I was more thinking of moving this logic to ProcessMetricConfig when the type=tail_file so that we can still keep doing this

julsemaan (changes requested):

go/stats/main.go
satkunas: I don't understand the purpose of this metric? What value will we pass to statsd with its use?

go/stats/metrics.go
```diff
@@ -0,0 +1,133 @@
...
+			var (
+				result string
+			)
+			for rows.Next() {
+				err := rows.Scan(&result)
+				if err != nil {
+					log.LoggerWContext(ctx).Error(err.Error())
+					return
+				}
+				switch conf.StatsdType {
```
julsemaan, line 42: Last detail before we merge is to get rid of that duplicated switch and extract that into a function so we can reuse it in all the metrics without copy/pasting it

go/stats/main.go
julsemaan: This actually tails a log file and checks for some specific patterns, when they do match, we increment a counter in statsd to know how many times that pattern matched.

The best example I could give would be to match on ERROR in a specific log to see how many errors there are in it.

Since we're not using it and I'm sure @extrafu would like to see that in ASAP, then lets scratch that and implement it when needed

go/stats/metrics.go
satkunas: @julsemaan `increment` and `unique` are useless in `icmp_ipv4`, while `timing` is only useful in `icmp_ipv4`. Since the switch differs between each of their use-cases, they can't be re-used. They were merged earlier, but I tore them apart when adding the `icmp_ipv4` type.

julsemaan: In that case, we could still create the function but add a validation through a map that ensures you only use valid statsd types for a specific metric type

satkunas: @julsemaan it doesn't seem counter-intuitive to always have at least one function parameter that isn't used? In order to create a function I'd need to utilize both a `metric interface{}` and a `client StatsdClient` parameter, and in every case at least one is not used due to:

```
	case "timing":
		client.Send(conf.StatsdNS)
```
So out function would be to be something like:

```
func SendMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats, metric interface{}, client StatsdClient) error {
...
}
```
Where we would either:
```
SendMetricConfig(ctx, conf, metric, nil)
```
or
```
SendMetricConfig(ctx, conf, nil, client)
```
or...

we use a single param for both, which IMO would be more confusing to someone looking at the function after-the-fact, since the single `interface{}` param can be either a `float64` or `int` or `Statsdclient`.

Please advise

julsemaan: Not sure I fully followed you on that one but here is the pseudo-code of what I had in mind:

```
func recordMetric(conf pfconfigdriver.PfStats, stat float64) error {
    // That should be a constant
    valid := map[string]map[string]bool{
        "mysql_query":map[string]bool{
              "count":true,
              "gauge":true,
        }
    }

    if validTypes, ok := valid[conf.Type]; !ok {
        return errors.New("Unrecognized metric type")
    } else {
        if _, ok := validTypes[conf.StatsdType]; !ok {
            return errors.New("Invalid statsd type")
        }
    }

    switch {
        .........
    }
}
```

With that above, we would just have to call the function and it would handle validating the statsd type works with the metric type and would centralize the whole logic of sending a metric to statsd.

satkunas: @julsemaan the above won't work. See line line 112 where the `timing` uses the `Statsdclient` type and not a ` float64`.  Also see line 63 where the `unique` uses a `string` type and not a `float64`.

julsemaan: Instead of using a wrapper that automatically sends the timing in that case, we could just track the start and the finish of the ping and then send it to StatsdClient.Timing("whatever", timing)

satkunas: @julsemaan this would prevent the optional use of the other chart types for a timing. I abstracted it this way so the user could decide to use either a `count`, `gauge`, `histogram` or the normal `timing` chart for `icmp`.

julsemaan: Ah, yes I do see this now.

IMO, I'd have kept ICMP to a simple timing because with every timing metric, you automatically get a count in statsd, but then, the code is there so we can leave it but it might be redundant a little bit for that part but to get a histogram we would need the implementation you did. We don't really use them but lets leave it as is in case we ever want to use them

satkunas: @julsemaan this was my thought as well, but with my tests I saw that the extra options were available with minimal changes, so I went for it.

Anyway, I'm glad to see this merged. Do we still want to move forward with the DHCP address pool size(s) as per @extrafu 's request?

julsemaan: @satkunas 

@extrafu can confirm but I believe we do want this.

So as discussed, I think the best route would be to code it instead of doing it via stats.conf

PR #3158, opened 2018-04-19
Feature/dashboard redis queue

### LINKED ISSUE #2981: Monitoring of the redis queues via monit
We should check if we could monitor the redis queues for any backlogging via monit or via our monitoring scripts

satkunas:
# Description
 * Advance the capability of `pfstats`.
   * Add plural (1x or 2x only) JSON Xpaths for appending metric namespace based on results (mirror `type=mysql_query` capability)
   * Add support for multiple JSON Xpath results/rows (mirror `type=mysql_query` capability)
   * Minor `pfstats` bugfixes

# Issue
* Fixes #2981

# Delete branch after merge
YES

# NEWS file entries
* Add API endpoint `/redis/queue/stats`

## Bug Fixes
* Issue with Monitoring of the redis queues via monit (#2981)

### DIFF
--- a/go/stats/metrics.go
+++ b/go/stats/metrics.go
@@ -273,7 +273,7 @@ func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error
  *         or 2x Xpaths separated with a comma (eg: $.items[0].somekey, $.item[0].somevalue)
  */
 func CompileJson(json interface{}, compile string) (interface{}, error) {
-	c := strings.Split(strings.Trim(compile, " "), ",")
+	c := strings.Split(compile, ",")
 	if len(c) > 1 {
 		// multiple XPath(s)
 		r1, err := CompileJson(json, c[0])
@@ -294,6 +294,7 @@ func CompileJson(json interface{}, compile string) (interface{}, error) {
 		return zipped, nil
 	}
 	//single Xpath
+	compile = strings.Trim(compile, " ")
 	p, err := jsonpath.Parse(compile)
 	if err != nil {
 		return nil, err
--- a/go/stats/metrics.go
+++ b/go/stats/metrics.go
@@ -206,6 +206,7 @@ func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error
 
 			case [][2]interface{}:
 				//double column
+                var namespace string
 				for _, row := range res.([][2]interface{}) {
 					namespace = CleanNameSpace(conf.StatsdNS + ";" + row[0].(string))
 					f64Result, _ := strconv.ParseFloat(row[1].(string), 64)
@@ -272,7 +273,7 @@ func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error
  *         or 2x Xpaths separated with a comma (eg: $.items[0].somekey, $.item[0].somevalue)
  */
 func CompileJson(json interface{}, compile string) (interface{}, error) {
-	c := strings.Split(strings.Trim(compile), ",")
+	c := strings.Split(strings.Trim(compile, " "), ",")
 	if len(c) > 1 {
 		// multiple XPath(s)
 		r1, err := CompileJson(json, c[0])
--- a/go/stats/metrics.go
+++ b/go/stats/metrics.go
@@ -270,7 +270,7 @@ func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error
  *
  *     Can be either:
  *         a single Xpath (eg: $.items[0].somevalue)
- *         or 2x Xpaths separated with a comma (eg: $.items[0].somekey, $.item[0].somevalue
+ *         or 2x Xpaths separated with a comma (eg: $.items[0].somekey, $.item[0].somevalue)
  */
 func CompileJson(json interface{}, compile string) (interface{}, error) {
 	c := strings.Split(compile, ",")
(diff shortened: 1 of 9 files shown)


julsemaan:
Make sure you get the milestone to '8.0' for your 8.0 PRs

'+1 patch release' is only for bug fixing that should go in the maintenance branch of the previous releases

go/stats/metrics.go
```diff
@@ -225,3 +256,75 @@ func ProcessMetricConfig(ctx context.Context, conf pfconfigdriver.PfStats) error
...
+		return ret, nil
+
+	default:
+		return nil, errors.New("Unhandled response type")
+	}
+
+	return nil, nil
+}
+
+func Zip(a, b []interface{}) ([][2]interface{}, error) {
```
julsemaan: This deserves a few comments as to what it does

lib/pf/UnifiedApi/Controller/Redis.pm
```diff
@@ -0,0 +1,78 @@
...
+        push $queue{$item->{queue}}{outstanding}, { name => $item->{name}, count => $item->{count} };
+    }
+    foreach my $item (@{ $stats->miss_counters }) {
+        push $queue{$item->{queue}}{expired}, { name => $item->{name}, count => $item->{count} };
+    }
+    #build json
+    my $json = [];
+    while( my( $key, $value ) = each %queue ){
+        # rebuild hash,
+        #   replace empty []'s w/ undef
```
julsemaan, line 39: Random question

Why is it better for these to be undef instead of empty arrays ?

satkunas: This is a personal habit for a couple reasons
 * Conditionals are more straight forward on an undef/null than on an empty array
 * It's visually more straight-forward, thus more readable
 * They're easier to struct in go

html/pfappserver/root/graph/dashboard.tt
```diff
@@ -711,5 +712,53 @@ END
...
+        <div class="tab-pane[% IF tab == "queue" %] active[% END %]" id="queue">
+            <div class="card">
+                <div class="card-title">
+                    <h2>[% l('Queue Counters') %]</h2>
+                </div>
+                <div class="card-block">
+                    <div class="row-fluid">
+                        <div class="span18 card">
+                            <div data-netdata="packetfence.redis.queue_stats_count"
+                            data-host="/netdata/127.0.0.1"
```
fdurand, line 988: We need to loop on all cluster members since redis is local to each one.

satkunas: Should we put in a "cluster" tab (2nd next to system) and put these charts there?

satkunas: I'm out until next week, unless an emergency is needed. This appears to be a simple front end adjustment in the dashboard template.

Can we merge and then either @cgx or I can update after?

julsemaan:
@satkunas 

I think we will have to postpone this to 8.1 since we can't really afford to have any more features in 8.0 that have a chance of breaking something.

lib/pf/UnifiedApi/Controller/Redis.pm
julsemaan: The thing is its not in our usual practices to do that so this makes it look different than the rest of our codebase

And in the spec we're "promising" an array and this isn't delivering it.

At least to stay consistent with the rest of our code, this should be an empty array

lib/pf/UnifiedApi/Controller/Redis.pm
```diff
@@ -0,0 +1,78 @@
...
+        push $json, { queue => $key, stats => {
+            count => $value->{count},
+            expired => ( $value->{expired}->[0] ? $value->{expired} : undef ),
```
julsemaan: Would be more obvious to check if there are elements in it by doing:
```
expired => ( @{$value->{expired}} ? $value->{expired} : undef ),
```

satkunas: I'll defer this change based on your earlier comment: https://github.com/inverse-inc/packetfence/pull/3158#discussion_r183021216

lib/pf/UnifiedApi/Controller/Redis.pm
satkunas: Changed in https://github.com/inverse-inc/packetfence/pull/3158/commits/06467b8f7d0f0d0d65ff21b17c4be2072faa16bb

julsemaan:
@satkunas 

Can you rebase this against devel to fix the conflicts, then we can merge that

jrouzierinverse (commented):
I think it would be better to name the endpoint /api/v1/queues
If we ever change the underlying technology behind we do not have to change name of the endpoint.

lib/pf/UnifiedApi.pm
```diff
@@ -203,6 +203,16 @@ our @API_V1_ROUTES = (
...
+    },
+    {
+        controller => 'Redis',
+        collection => {
+            subroutes    => {
+                'queue/stats' => {
+                    get => 'queue'
+                },
+            },
+        },
```
jrouzierinverse, line 253: To avoid the resource end points from being created add 

resource => undef,

fdurand:
Rebase please

satkunas:
@julsemaan @jrouzierinverse @fdurand rebased

lib/pf/UnifiedApi.pm
satkunas: added in https://github.com/inverse-inc/packetfence/pull/3158/commits/369898600038d89e2b35cc4e1e854b1fdab9b5ae

julsemaan (commented):
I'm happy with the code now, if @jrouzierinverse and @fdurand are happy too then we can merge this

extrafu:
@jrouzierinverse or @fdurand please merge.

PR #3236, opened 2018-05-14
Max node per role under autoreg ,moved unreg on accounting stop on the connection profile and vlan filters enhancement

fdurand:
# Description
Allow PacketFence to test MAx device allowed by role when autoregister
Moved unregister on accounting stop parameter on the connection profile
Allow a value return by the vlan filters to be dynamic.

# Impacts
Radius

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Allow max node per role on autoregister
* Moved unregister on accounting stop parameter on the connection profile
* Vlan filters can be set to ${node_info.category} and it will return the current category of the device

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
@@ -40,7 +40,7 @@ The main definition block
 
 has_block 'definition' =>
   (
-    render_list => [qw(id description root_module status preregistration autoregister reuse_dot1x_credentials dot1x_recompute_role_from_portal dpsk default_psk_key)],
+    render_list => [qw(id description root_module status preregistration autoregister reuse_dot1x_credentials dot1x_recompute_role_from_portal dpsk default_psk_key unreg_on_acct_stop)],
   );
 
 =head2 captive_portal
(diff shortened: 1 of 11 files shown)


julsemaan:
Need upgrade instructions for the unreg_on_accounting_stop parameter

html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
```diff
@@ -199,6 +199,24 @@ has_field 'autoregister' =>
...
+
+has_field 'unreg_on_acct_stop' =>
+  (
+   type => 'Toggle',
+   label => 'Automatically deregister devices on accounting stop',
+   checkbox_value => 'enabled',
+   unchecked_value => 'disabled',
+   default => 'disabled',
+   tags => { after_element => \&help,
+             help => 'This activates automatic deregistation of devices for the profile if PacketFence receive an radius accounting stop.' },
```
julsemaan: receives a RADIUS accounting stop for a device

lib/pf/Connection/Profile.pm
```diff
@@ -609,6 +609,17 @@ sub canAccessRegistrationWhenRegistered {
...
 
+=item unregOnAcctStop
+
+Deregister device on accounting stop
+
+=cut
+
+sub unregOnAcctStop {
+    my ($self) = @_;
+    return $self->{'_unreg_on_acct_stop'};
```
julsemaan: isenabled would be good here

lib/pf/role.pm
```diff
@@ -415,7 +415,7 @@ sub getRegisteredRole {
     # the role based on the rules defined in the different authentication sources.
     # FIRST HIT MATCH
     elsif ( defined $args->{'user_name'} && $args->{'connection_type'} && ($args->{'connection_type'} & $EAP) == $EAP ) {
-        if ( isdisabled($profile->dot1xRecomputeRoleFromPortal) ) {
+        if ( isdisabled($profile->dot1xRecomputeRoleFromPortal)  || $args->{'autoreg'} == 1) {
```
julsemaan: It seems autoreg can be set to 'yes' which is a string value and doesn't seem to be ever set to a number

fdurand:
Bump

extrafu:
@julsemaan Bump bump.

julsemaan:
Still needs upgrade instructions for the unreg_on_accounting_stop parameter

PR #3239, opened 2018-05-15
Feature/pfappserver nodes

satkunas:
# Description
Continue development on pfappserver 's nodes

# Impacts
None

# Delete branch after merge
YES

PR #3240, opened 2018-05-16
Feature/pfappserver nodes update

satkunas:
# Description
Continue development on pfappserver 's nodes

# Impacts
None

# Delete branch after merge
YES


satkunas:
Supersedes https://github.com/inverse-inc/packetfence/pull/3239

satkunas:
@jrouzierinverse I'll need the following API endpoints created, and  the ability to pass multiple node MAC's (preferably as a array) to perform all the following functions in bulk:

**setters**
- [x] clear violation
- [x] register node
- [x] deregister node
- [ ] reevalutate access
- [ ] restart switchport
- [ ] apply role
- [ ] apply bypass role
- [ ] apply violation

**getters**
- [x] node category(ies) - needed to map "nodes category_id" from `int` to `string`, therefore `id` must be an `integer`, not ` string` in order to perform mapping. [See comment](#issuecomment-389921761)

satkunas:
@cgx I'll need a (single) sample POST/PUT/DELETE in the state for the endpoints above as an example to work with as a template going-forward.

satkunas:
@jrouzierinverse please either update the OpenAPI spec, or respond inline with the documentation required to use the completed endpoints.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodeView.vue
```diff
@@ -185,6 +185,13 @@ export default {
   mounted () {
     this.$store.dispatch('$_nodes/getNode', this.mac)
     this.$store.dispatch('config/getViolations')
+    var vm = this
+    window.addEventListener('keyup', function (event) {
+      switch (event.keyCode) {
+        case 27: // escape
+          vm.close()
+      }
+    })
```
cgx: When adding an event listener, you must also remove it when the component is destroyed.

I've pushed a commit (dba8813) to handle the ESC key. You can remove this change.

satkunas:
@jrouzierinverse The API implementation for `/roles` is incorrect.

```/api/v1/config/roles
Response: {
	"items": [
		{
			"id": "default",
			"max_nodes_per_pid": "0",
			"notes": "Placeholder role\/category, feel free to edit"
		},
		{
			"id": "guest",
			"max_nodes_per_pid": "0",
			"notes": "Guests"
		},
		{
			"id": "gaming",
			"max_nodes_per_pid": "0",
			"notes": "Gaming devices"
		},
		{
			"id": "voice",
			"max_nodes_per_pid": "0",
			"notes": "VoIP devices"
		},
		{
			"id": "REJECT",
			"max_nodes_per_pid": "0",
			"notes": "Reject role (Used to block access)"
		}
	]
}
```
Here the `id` should be a integer from the `category_id` column, and a `name` column should be added to contain the string.

jrouzierinverse:
I am assuming you meant /config/roles not /roles

This is tricky since we have roles stored in two different places.

First we have roles stored in the roles.conf which is then synced to the 
database to the node_category table which as the name and id.

So what you are really looking for is /node_categories

I will expose the read portions of that for you.

||

satkunas:
@jrouzierinverse the `/nodes` API endpoint requires the following missing columns:
- [x] online/offline via radacct.online field

This can be received via the /node/:node_id/fingerbank_info this was changed to a GET instead of a POST
- [x] Fully Qualified Device Name (fingerbank)
- [x] Mobile (fingerbank) 

List may not be complete and is subject to more additions

julsemaan:
@satkunas 

You need to stop pulling if its not fast-forward only, otherwise you create useless "Merge branch" commits

I usually always rebase, never pull even if its fast-forward so that way I never do the mistake.

Otherwise, I think @cgx or @fdurand has a git alias that can pull only if its fast-forward

fdurand:
```
[alias]
        up = "!git remote update -p; git merge --ff-only @{u}"
        ready = rebase -i @{u}
```

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodeView.vue
```diff
@@ -21,27 +21,77 @@
...
               </b-form-group>
+              <pf-form-row id="pid" :label="$t('Owner')">
+                <b-form-input id="pid" v-model="node.pid"></b-form-input>
+              </pf-form-row>
+              <pf-form-row id="status" :label="$t('Status')">
+                <b-form-input v-model="node.status"></b-form-input>
+              </pf-form-row>
+              <pf-form-row id="category_id" :label="$t('Role')">
+                <b-form-input v-model="node.category_id"></b-form-input>
+              </pf-form-row>
```
cgx: Both *status* and *role* are already there.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodeView.vue
```diff
@@ -21,27 +21,77 @@
...
                 {{ node.last_seen }}
               </pf-form-row>
-              <pf-form-row label="IPv4 Address" v-if="node.ip4">
+              <pf-form-row :label="$t('IPv4 Address')" v-if="node.ip4">
                 {{ node.ip4.ip }}
-                  <b-badge variant="success" v-if="node.ip4.active">Since {{node.ip4.start_time}}</b-badge>
-                  <div v-else><icon class="text-warning" name="exclamation-triangle"></icon> Inactive since {{node.ip4.end_time}}</div>
+                  <b-badge variant="success" v-if="node.ip4.active">{{ $t('Since') }} {{node.ip4.start_time}}</b-badge>
+                  <b-badge variant="warning" v-else-if="node.ip4.end_time">{{ $t('Inactive since') }} {{node.ip4.end_time}}</b-badge>
+                  <b-badge variant="danger" v-else>{{ $t('Inactive') }}</b-badge>
```
cgx: I was testing the *warning triangle icon* instead of the badges, but I don't mind for now to keep the badges.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -9,12 +9,55 @@
...
+            <b-dropdown-item @click="applyReevaluateAccess()">
+              <icon class="position-absolute mt-1" name="sync"></icon>
+              <span class="ml-4">{{ $t('Revaluate Access') }}</span>
+            </b-dropdown-item>
+            <b-dropdown-item @click="applyRestartSwitchport()">
+              <icon class="position-absolute mt-1" name="retweet"></icon>
+              <span class="ml-4">{{ $t('Restart Switchport') }}</span>
+            </b-dropdown-item>
+            <b-dropdown-divider></b-dropdown-divider>
+            <b-dropdown-header id="header1">{{ $t('Apply Role') }}</b-dropdown-header>
```
cgx: Remove `id`.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -9,12 +9,55 @@
...
+            <b-dropdown-item v-for="role in roles" :key="role.id" @click="applyRole(role)">
+              <span>{{role.id}}</span>
+            </b-dropdown-item>
+            <b-dropdown-divider></b-dropdown-divider>
+            <b-dropdown-header id="header2">{{ $t('Apply Bypass Role') }}</b-dropdown-header>
```
cgx: Remove `id`.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -9,12 +9,55 @@
...
+            <b-dropdown-item v-for="role in roles" :key="role.id" @click="applyBypassRole(role)">
+              <span>{{role.id}}</span>
+            </b-dropdown-item>
+            <b-dropdown-divider></b-dropdown-divider>
+            <b-dropdown-header id="header2">{{ $t('Apply Violation') }}</b-dropdown-header>
```
cgx: Remove `id`.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -33,7 +76,24 @@
         </b-col>
       </b-row>
       <b-table stacked="sm" :items="items" :fields="visibleColumns" :sort-by="sortBy" :sort-desc="sortDesc"
-        @sort-changed="onSortingChanged" @row-clicked="onRowClick" hover no-local-sorting></b-table>
+        @sort-changed="onSortingChanged" @row-clicked="onRowClick" @head-clicked="clearChecked" hover no-local-sorting v-model="tableValues">
+        <template slot="HEAD_actions" slot-scope="head">
+          <input type="checkbox" v-model="checkedAll" @change="onCheckedAllChange" @click.stop>
+        </template>
+        <template slot="actions" slot-scope="data">
+          <input type="checkbox" :id="data.value" :value="data.item" v-model="checkedRows" @change="onCheckedRowsChange" @click.stop>
```
cgx: Are the `@click.stop` necessary? Just asking ..

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -119,6 +225,174 @@ export default {
           label: this.$i18n.t('Owner'),
           sortable: true,
           visible: true
+        },
+        {
+          key: 'last_ip',
```
cgx: I think we need to extend our API to support `last_ip`.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -119,6 +225,174 @@ export default {
...
+          label: this.$i18n.t('IP Address'),
+          sortable: true,
+          visible: true
+        },
+        {
+          key: 'tenant_name',
```
cgx: I think we need to extend our API to support `tenant_name`.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
```diff
@@ -119,6 +225,174 @@ export default {
...
+          visible: false
+        },
+        {
+          key: 'category_id',
+          label: this.$i18n.t('Role'),
+          sortable: true,
+          visible: true
+        },
+        {
+          key: 'switch_id',
```
cgx: I think none of the `switch` fields can currently be retrieved through the `/nodes/search` API call.

html/pfappserver/root/static.alt/src/views/Nodes/_store/index.js
```diff
@@ -17,7 +18,8 @@ const state = {
   searchSortBy: 'mac',
   searchSortDesc: false,
   searchMaxPageNumber: 1,
-  searchPageSize: localStorage.getItem(STORAGE_SEARCH_LIMIT_KEY) || 10
+  searchPageSize: localStorage.getItem(STORAGE_SEARCH_LIMIT_KEY) || 10,
+  visibleColumns: localStorage.getItem(STORAGE_VISIBLE_COLUMNS_KEY) || false
```
cgx: Since the localStorage will return a string, you'll need to wrap the value with `JSON.parse`.

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
jrouzierinverse: The name would change to ip4log.ip

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
jrouzierinverse: The name will change to locationlog.switch

html/pfappserver/root/static.alt/src/views/Nodes/_components/NodesSearch.vue
satkunas: Yes, otherwise it bubbles up an triggers a table-row click.

satkunas: Furthermore, For your info, the checkbox `checked` state is not available on the `click` event, only afterwards on the `change` event.

PR #3256, opened 2018-05-25
Feature/detect transport change

### LINKED ISSUE #2540: Perform wired -> wireless detection for MAC spoofing
In order to continue the battle against MAC spoofing, we could detect if a MAC is going from wired to wireless (or the other way around) and trigger a violation

julsemaan:
# Description
Allow to detect a change between wired and wireless connection for the same MAC

# Impacts
RADIUS

# Issue
fixes #2540 

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Added ability to trigger a violation when detecting a change of network transport (wired vs wireless) for a single MAC address

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -71,6 +71,14 @@ Will identify hostname changes and send an e-mail with these changes.
 This can help detect MAC spoofing.
 EOT
 
+[network.connection_type_change_detection]
+type=toggle
+options=enabled|disabled
+description=<<EOT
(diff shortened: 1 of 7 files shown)


jrouzierinverse:
The only issue I see is there is no translation for the new parameter.
Do you want to do that now or later?

julsemaan:
Might as well do it now, otherwise I'll forget for sure.

Just pushed it

julsemaan:
Ready for review/merge

PR #3257, opened 2018-05-28
Database encryption

fdurand:
# Description
Added support for database encryption

# Impacts
database

# NEW Package(s) required
Mariadb >= 10.2
Mariadb-backup >= 10.2

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Support for database encryption (aws and file key)


extrafu:
👍

julsemaan:
Not sure this should be 8.1 considering the feature freeze will begin Monday

What do @extrafu and @fdurand think?

extrafu:
I think we should consider it since it won't "break" anything unless enabled.

julsemaan:
In that case we need to get it in today so its part of our testing phase in case this causes unexpected side effects even when disabled

satkunas:
`MariaDB-aws-key-management` needed in .spec

julsemaan (commented):
code looks fine to me, if we merge this for 8.1, we'll need to test properly the new backups using the MariaDB backup tool

julsemaan:
Let me know when the packages are in devel and we'll merge

PR #3266, opened 2018-05-31
Added an alias for AeroHIVE web auth.

fdurand:
# Description
Added alias for Aerohive webauth 

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Added alias for Aerohive webauth, will work with /AeroHIVE::AP and /AeroHIVE as the portal url


extrafu:
Should we also alias /Aerohive ?

fdurand:
It's a fix because of the AP bug and i think it will be part of another PR (rename AeroHIVE to Aerohive).

PR #3285, opened 2018-06-12
Password of the day

### LINKED ISSUE #2779: Authentication sources > Password of the day
If would be nice to have an authentication source "Password of the day".

Forget for now how it can be done in PacketFence using just cronjobs and htpasswd files and just think about a simple interface for it:

- you specify when the password expires
- you specify who should receive the password of the day (list of email addresses, because there could be many persons)
- you could ask for additional info (full name, email address, etc.)
- you add rules/conditions if relevant
- etc.

fdurand:
# Description
Added a new source Password Of the Day.
The source Password of the day create a new user (the source pid) and rotate the password based on the period defined.
You can also define the length of the password.

# Impacts
Authentication on the portal, new pfmon task

# Issue
fixes #2779

# Delete branch after merge
YES

# NEWS file entries
## New Features
* New authentication source Password Of the Day


docs/PacketFence_Installation_Guide.asciidoc
```diff
@@ -479,6 +479,7 @@ PacketFence can authenticate users that register devices via the captive portal
 * Sponsored Email
 * Twitter (OAuth 2)
 * Windows Live (OAuth 2)
+* Password Of the Day
```
extrafu: Password of the day

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -52,4 +52,12 @@ call ValidateVersion;
 --
 ALTER TABLE pf_version engine = InnoDB;
 
+--
+-- Add potd column in person table
+--
+ALTER TABLE person
+    ADD `psk` varchar(255) NULL DEFAULT NULL,
```
nqb: `psk` is not related to that PR.

nqb (changes requested):
- HTML template should include password duration
- Email should be mandatory in source definition : if you have no email, you can't receive password
- Password length should be size limit. If pfmon takes more than `interval value` to generate new password, entries in `password` db are not updated.

lib/pf/password.pm
```diff
@@ -140,11 +140,16 @@ Generates the password
 =cut
 
 sub _generate_password {
+    my ($size) = @_;
+    my $min = 8;
+    my $max = 12;
 
-    my $password = word(8, 12);
+    $min = $max = $size if (defined $size);
```
nqb, line 147: This instruction doesn't limit size of `password` to 12 characters. With following one-liner:

```shell
perl -I/usr/local/pf/lib -Mpf::password -e "pf::password::generate("potd0",[{type => 'valid_from', value => '2018-08-27 09:39:05'},{type => 'expiration', value => '2018-08-27 09:50:05'}],undef, 0, {'password_length' => 64})"
```

`$size` equals to `64`

nqb (changes requested):

lib/pf/password.pm
nqb: Moreover, I didn't see `Password Length` limitation when configuring source.

fdurand: I am not sure to understand why do we want to limit the password length to 12 ?

nqb: On my side, I'm able to freeze PF (with a little config) when setting a password of 64 characters.

nqb:
As said with @jrouzierinverse, we need to create tests for that new authentication source.

See: https://github.com/inverse-inc/packetfence/blob/fa8ab3a9f28a503831809918919d60db0a31157a/t/data/authentication.conf

extrafu:
@nqb Please revisit your comments.

extrafu:
@jrouzierinverse Please help in adding the test case.

nqb:
Feedback on my new tests:
- At first creation of POTD source, I didn't receive POTD. In pfmon log:

```

Sep 30 20:05:34 potd pfmon: pfmon(12010) ERROR: [mac:unknown] Error running task password_of_the_day: Can't use string ("caubrott") as a HASH ref while "strict refs" in use at /usr/local/pf/lib/pf/pfmon/task/password_of_the_day.pm line 47.
```

- I can't see POTD sources on Portal (maybe due to first issue):

```
No authentication sources found in configuration. Select another option.
```
- As said before, the code of `_generate_password` function try to limit password length (between 8 and 12 characters). But currently, you can specify a password out of this limit.

jrouzierinverse:
Can you rebase this please?

lib/pf/password.pm
fdurand: It's because of Crypt::GeneratePassword::word who is taking time to generate a readable password.
I limited the password length from 1 to 15.

fdurand:
BUMP

nqb:
>     * At first creation of POTD source, I didn't receive POTD. In pfmon log:

I get same issue with new commits.

EDIT: I get same messages in `pfmon.log` but passwords are updated in DB. However, no emails.

nqb:
Still present too:

- I can't see POTD sources on Portal (maybe due to first issue):

```
No authentication sources found in configuration. Select another option.
```

I get another strange behavior: I see two "Password length" fields when creating a new Potd source.
screenshot from 2018-10-02 08-36-09

nqb:
Last commits fixed my previous issue.

@jrouzierinverse: is your unit test ready ?

jrouzierinverse:
Finished and found a bug in the process pushed the fix and the test.

nqb:
So I think we can merge ASAP.

nqb (approved):

PR #3286, opened 2018-06-12
Fixes #604 and add password length

### LINKED ISSUE #604: Local users should be created with the ability to select password encryption
Since we are able to do RADIUS authentication based on local users, but, by default, we shipped PacketFence with bcrypt encryption for local users, it would be great to do a 'per-user' selection of the encryption type rather than modify the global setting.

### LINKED ISSUE #3610: Create local account: Ability to specify password length
Right now, its preset to 10 characters or so which creates long passwords to remember.

We should allow the user to specify the length of the password in the source.

fdurand:
# Description
Fixed for #604 , choose password hash when create local user from a source

# Impacts
Create local user code

# Issue
fixes #604 
fixes #3610

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Chose hash algo when create a local user from a source
* Define the length of the password to generate


html/pfappserver/lib/pfappserver/Base/Form/Role/SourceLocalAccount.pm
```diff
@@ -40,8 +40,35 @@ has_field 'local_account_logins' => (
...
+    label => 'Database passwords hashing method',
+    options =>
+    [
+     { value => 'plaintext', label => 'Plaintext' },
+     { value => 'ntlm', label => 'NTLM' },
+     { value => 'bcrypt', label => 'Bcrypt' },
+    ],
+    required => 1,
+    element_class => ['input-small'],
+    default => 'ntlm',
```
nqb: `hash_passwords=bcrypt` in `pf.conf.defaults`, we need to do same thing for sources.

lib/pf/Authentication/CreateLocalAccountRole.pm
```diff
@@ -15,6 +15,8 @@ use pf::Authentication::constants qw($LOCAL_ACCOUNT_UNLIMITED_LOGINS);
 
 has 'create_local_account' => (isa => 'Str', is => 'rw', default => 'no');
 has 'local_account_logins' => (isa => 'Str', is => 'rw', default => $LOCAL_ACCOUNT_UNLIMITED_LOGINS);
+has 'hash_passwords'       => (isa => 'Maybe[Str]', is => 'rw', default => 'ntlm');
```
nqb: `hash_passwords=bcrypt` in `pf.conf.defaults`, we need to do same thing for sources.

nqb (changes requested):

nqb:
Another important point: if a user change its password from WebUI, hash will be change in DB to `$Config{'advanced'}{'hash_passwords'};`. Consequently, we lost "per-user" encryption mentioned in #604. If we want to keep original pref, we need to store it in DB.

@fdurand : what is the best way to test new `sql.conf` code, doing 802.1X authentications with local account created by sources ?

html/pfappserver/lib/pfappserver/Base/Form/Role/SourceLocalAccount.pm
julsemaan: Even better idea, why don't we take the value currently configure in `pf::ConfigStore::Pf->new->read("advanced")->{hash_passwords}`

julsemaan: Actually, lets scratch my previous comment.

This should point to the value in the source model metadata like the other sources

nqb:
Will try with `eapol_test`

lib/pf/Authentication/CreateLocalAccountRole.pm
julsemaan: @nqb's comment sounds legit

nqb: @julsemaan: not sure because `bcrypt` can't be use for 802.1X local auth.

julsemaan: Then we should see if we change our default in pf.conf.defaults, but IMO, the default for the sources should be the same default as for the system

nqb (changes requested):
We need to update Installation Guide:
- https://github.com/inverse-inc/packetfence/blob/12efe51b678b16f8d5e0362d4edfa433bb0f6605/docs/PacketFence_Installation_Guide.asciidoc#eap-guest-authentication-on-email-sponsor-and-sms-registration
- https://github.com/inverse-inc/packetfence/blob/12efe51b678b16f8d5e0362d4edfa433bb0f6605/docs/PacketFence_Installation_Guide.asciidoc#eap-local-user-authentication

I think this two sections can be merged.

fdurand:
BUMP

nqb:
Currently, I can't start `pfconfig`

nqb:
If I revert all the PR, I'm able to start `pfconfig`.

lib/pf/Authentication/CreateLocalAccountRole.pm
```diff
@@ -12,11 +12,10 @@ Role that defines the Authentication source behavior for creating local accounts
 
 use Moose::Role;
 use pf::Authentication::constants qw($LOCAL_ACCOUNT_UNLIMITED_LOGINS);
-use pf::ConfigStore::Pf;
 
 has 'create_local_account' => (isa => 'Str', is => 'rw', default => 'no');
 has 'local_account_logins' => (isa => 'Str', is => 'rw', default => $LOCAL_ACCOUNT_UNLIMITED_LOGINS);
-has 'hash_passwords'       => (isa => 'Maybe[Str]', is => 'rw', default => pf::ConfigStore::Pf->new->read("advanced")->{"hash_passwords"});
+has 'hash_passwords' => (isa => 'Maybe[Str]', is => 'rw', default => 'bcrypt');
```
nqb: @jrouzierinverse: could we use a var in place of 'bcrypt' hard-code value ?

jrouzierinverse: Of course we can

nqb:
I just rebased

nqb:
No more error with `pfconfig`. Well done @jrouzierinverse

nqb:
I get following message in audit log when using `eapol_test` to test RADIUS local auth with "created local account" using a sponsor source:

```
Module-Failure-Message = "pfsponsor: Failed to create the pair: Invalid vendor name in attribute name \"%hash_passwords%\""
Module-Failure-Message = "pfsponsor: Error parsing user data from database result" 
Module-Failure-Message = "pfsponsor: Error getting check attributes" 
User-Password = "******" 
Module-Failure-Message = "Failed retrieving values required to evaluate condition" 
```

PF has created a local user account like this:
* username: foo@foo.bar
* password: {ntlm}HASH_PASSWORD

I try with following peap.config file:
```
network={
ssid="peap-ssid"
key_mgmt=WPA-EAP
eap=PEAP
identity="foo@foo.bar"
anonymous_identity="foo@foo.bar"
password="secret"
phase2="autheap=MSCHAPV2"
# Uncomment the following to perform server certificate validation.
#ca_cert="/etc/raddb/certs/ca.der"
}
```

nqb:
Same issue is password stored in clear text.

nqb:
After doing a `raddebug`:

```
(101) Fri Oct  5 15:24:04 2018: Debug: pfsponsor: EXPAND %{User-Name}
(101) Fri Oct  5 15:24:04 2018: Debug: pfsponsor:    --> foo@foo.bar
(101) Fri Oct  5 15:24:04 2018: Debug: pfsponsor: SQL-User-Name set to 'foo@foo.bar'
(101) Fri Oct  5 15:24:04 2018: Debug: pfsponsor: EXPAND SELECT 1, pid, "%%hash_passwords%%", REPLACE(password,'{ntlm}',''), ":="           FROM password           JOIN activation using (pid)           WHERE pid = '%{SQL-User-Name}'             AND (SELECT type from activation WHERE pid='%{SQL-User-Name}' ORDER BY code_id DESC LIMIT 1) = "sponsor"             AND ( now() <= password.unregdate OR password.unregdate = '0000-00-00 00:00:00' )             AND password.tenant_id = '%{control:PacketFence-Tenant-Id}'             LIMIT 1 
(101) Fri Oct  5 15:24:04 2018: Debug: pfsponsor:    --> SELECT 1, pid, "%hash_passwords%", REPLACE(password,'{ntlm}',''), ":="           FROM password           JOIN activation using (pid)           WHERE pid = 'foo@foo.bar'             AND (SELECT type from activation WHERE pid='foo@foo.bar' ORDER BY code_id DESC LIMIT 1) = "sponsor"             AND ( now() <= password.unregdate OR password.unregdate = '0000-00-00 00:00:00' )             AND password.tenant_id = '1'             LIMIT 1 
(101) Fri Oct  5 15:24:04 2018: Debug: pfsponsor: Executing select query: SELECT 1, pid, "%hash_passwords%", REPLACE(password,'{ntlm}',''), ":="           FROM password           JOIN activation using (pid)           WHERE pid = 'foo@foo.bar'             AND (SELECT type from activation WHERE pid='foo@foo.bar' ORDER BY code_id DESC LIMIT 1) = "sponsor"             AND ( now() <= password.unregdate OR password.unregdate = '0000-00-00 00:00:00' )             AND password.tenant_id = '1'             LIMIT 1 
(101) Fri Oct  5 15:24:04 2018: ERROR: pfsponsor: Failed to create the pair: Invalid vendor name in attribute name "%hash_passwords%"
```

nqb:
With my last commit, I'm able to do RADIUS local auth with plain-text or NTLM passwords with `eapol_test`.

However, I get a new unexpected behavior related to that PR.

If I change `hash_passwords` settings from GUI: from `bcrypt` to `ntlm`, config files are updated but namespace `config::Pf` in `pfconfig` not. Consequently, passwords are still hash with bcrypt in DB.

I need to do a restart of `pfconfig` to fix issue (a `clear_backend` doesn't fix it). 

I didn't reproduce on other installations.

extrafu:
Bump.

extrafu:
@nqb What's the status on this one?

nqb:
@extrafu,

Feature is working but:
* I have an issue with `pfconfig` (see my last comment).
* If a user change its password from WebUI or an admin from WebAdmin, hash will be change in DB to $Config{'advanced'}{'hash_passwords'};

extrafu:
@julsemaan Can you please check both issues with @fdurand ?

nqb:
>     * I have an issue with `pfconfig` (see my last comment).

Maybe related to commits done by @jrouzierinverse

jrouzierinverse:
The issue is you must restart radiusd after changing the password hashing.

nqb:
> The issue is you must restart radiusd after changing the password hashing.

Not agree. On a fresh devel installation, I can change this parameter from webadmin and PF change password directly in DB.

jrouzierinverse:
The radius configuration is generated using values from pf.conf.
If radius is not restarted then the newer settings are not known to radius.

jrouzierinverse:
Can you test if the password hashing works after restarting radius.

nqb:
I speak with @jrouzierinverse about the issue. This is a `pfconfig` issue not a RADIUS issue.

extrafu:
@nqb So, what's the status on this? who takes care of the pfconfig issue?

nqb:
@extrafu,

There is always a bug that need to be fix. I don't know if @jrouzierinverse is working on it.

jrouzierinverse:
There seems to be two issues going on here.

- The first issue seems to be pfconfig has a problem refreshing values after updating them. 
   Need to look into why this happening.
   @julsemaan may have some insight since he was in that code recently

- The second issue is after changing advanced.hash_passwords RADIUS must be restarted to update the radius config based of pfconfig values.

My suggestion is that we we work around the pfconfig issue at the moment.
Since that was most likely caused by the recent changes in pfconfig.

Just retest with the proper config in place for RADIUS.

fdurand:
Just an update, the issue with radius is just because you need to rename conf/radiusd/sql.conf.example to conf/radiusd/sql.conf

nqb:
Any updates on this ?

fdurand:
Just test it again and i am able to start pfconfig and use 802.1x authentication with the local account created (of course you need to rename sql.conf.exemple to sql.conf).
So what next ?!

julsemaan:
@nqb 

Make sure you check this before tomorrow, we'll merge then, it works now

PR #3349, opened 2018-06-26
Fix/3128 (Graph for routed network DHCP)

### LINKED ISSUE #3128: pfstats DHCP stats with routed networks
Concern expressed by @satkunas in #3113:

I see a possible issue that we may need to address. We're generating DHCP leases based on `interfaces::listen_ints`. 

Is it possible to have more than one DHCP pool per `int`? If so, `pfstats`  only uses the first index from `/api/v1/dhcp/stats/[interface]` (`$[0].free`) and if this API endpoint returns more than one result for a single interface the subsequent results are ignored. Let me know if this has to be addressed and requires support for multiple DHCP servers for a single interface before merging. I'll need to add a second level of iteration within `foreach interface in listen_ints` both in the backend and in the dashboard template, or possibly switch to the `@dhcplistener_ints` - if you can confirm that this variable will work better.

fdurand:
# Description
DHCP Statistics for all networks.

# Impacts
Dashboard

# Issue
fixes #3128 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* DHCP Statistics for routed networks missing on the dashboard (#3128)


go/database/connection.go
```diff
@@ -2,14 +2,20 @@ package database
...
+	"os"
 
 	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
 )
 
 func Connect(user, pass, host, port, dbname string) *sql.DB {
 	var where string
 	if host == "localhost" {
-		where = "unix(/var/lib/mysql/mysql.sock)"
+		if _, err := os.Stat("/etc/debian_version"); err == nil {
```
julsemaan: It would be good to consolidate this function with the one in go/db/db.go to prevent duplication

julsemaan (commented):
Other for my DRYing up comment, this looks good and seems to have a small impact on the code so I'd be comfortable in merging this for 8.1 if nobody objects to it

go/database/connection.go
fdurand: Do we want to do it in this PR ?

julsemaan: Since we're already changing stuff in it, I'd say to put it in this PR

fdurand: Ok done

PR #3374, opened 2018-07-06
Feature/pfconfigpool locking timeout

### LINKED ISSUE #3360: Have pfconfigpool use the RW timeout lock from the Fingerbank collector
I recently created a mini library for a lock that can timeout in golang for the fingerbank collector

It would be good to use it for PacketFence as well since the current implementation has caused at least 1 issue of memory leak when locks do timeout

julsemaan:
# Description
Reworks the pfconfig pool locking strategy to use channels instead of sync.Mutex

# Impacts
pfconfig in golang

# Issue
fixes #3360 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Reworked the Golang pfconfig locking to prevent deadlocks while refreshing the configuration

### DIFF
--- a/go/timedlock/rw_lock.go
+++ b/go/timedlock/rw_lock.go
@@ -101,6 +101,7 @@ func (l *RWLock) Lock() (uint64, error) {
 				return id, nil
 			} else {
 				l.internalLock.Unlock()
+				time.Sleep(1 * time.Millisecond)
 			}
 		}
 
--- a/go/timedlock/rw_lock.go
+++ b/go/timedlock/rw_lock.go
@@ -1,6 +1,7 @@
 package timedlock
 
 import (
+	"errors"
 	"fmt"
 	"runtime"
 	"sync"
@@ -10,6 +11,8 @@ import (
 var timerPool = &sync.Pool{}
 
 const MAX_RLOCK = 1000
+
+var TIMEOUT_ERROR = errors.New("Timeout occured")
 
 type RWLock struct {
 	internalLock *sync.Mutex
@@ -79,7 +82,7 @@ func (l *RWLock) getNextId() uint64 {
 	return l.c
 }
 
-func (l *RWLock) Lock() uint64 {
+func (l *RWLock) Lock() (uint64, error) {
 	id := l.getNextId()
 	stopAt := time.Now().Add(l.Timeout)
 	timeoutTimer := l.getTimer(l.Timeout)
@@ -88,14 +91,14 @@ func (l *RWLock) Lock() uint64 {
 	select {
 	case <-timeoutTimer.C:
 		l.handleTimeout(id, "Timeout obtaining or releasing lock that came from: \n")
-		return 0
+		return 0, TIMEOUT_ERROR
 	case <-l.lockChan:
 		timeoutTimer.Stop()
 		for time.Now().Before(stopAt) {
 			l.internalLock.Lock()
 			if len(l.rlockChan) == MAX_RLOCK {
 				l.internalLock.Unlock()
-				return id
+				return id, nil
 			} else {
 				l.internalLock.Unlock()
 			}
@@ -104,7 +107,7 @@ func (l *RWLock) Lock() uint64 {
 		l.lockChan <- 1
 		l.handleTimeout(id, "Timeout obtaining or releasing lock that came from: \n")
 	}
-	return 0
+	return 0, TIMEOUT_ERROR
 }
 
 func (l *RWLock) Unlock(id uint64) {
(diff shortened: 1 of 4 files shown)


julsemaan:
This is a bit late for the party for 8.1 especially considering it touches the core of the golang services.

So I guess this can be reviewed and merged post 8.1 and put in the maintenance

julsemaan:
Like for 8.1, this seems to be late for the party, I think the best course of action would be to have this reviewed ASAP, then after 8.2 is released, merge this into devel and potentially put it in the maintenance after we validate it doesn't cause trouble in devel.

Or, since this is bugfixing, we rush this into devel but the risk seems to be fairly high although this locking strategy is working great in the fingerbank collector

extrafu:
👍

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,196 @@
...
+
+import (
+	"fmt"
+	"runtime"
+	"sync"
+	"time"
+)
+
+var timerPool = &sync.Pool{}
+var chanPool = &sync.Pool{
```
jrouzierinverse: The channel pool is not used

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,196 @@
...
+	New: func() interface{} {
+		return make(chan int)
+	},
+}
+
+type RWLock struct {
+	doneChans    map[uint64]chan int
```
jrouzierinverse: doneChans is not used

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,196 @@
...
+	l.c++
+	// 0 isn't an acceptable ID
+	if l.c == 0 {
+		l.c++
+	}
+	return l.c
+}
+
+func (l *RWLock) Lock() uint64 {
+	id := l.getNextId()
```
jrouzierinverse, line 86: It looks like id is not used.
Return an error instead.

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,196 @@
...
+	l.internalLock.Lock()
+	defer l.internalLock.Unlock()
+	if len(l.lockChan) != 0 {
+		panic("Unlock of unlocked mutex")
+	} else if id == 0 {
+		panic("Unlocking mutex with ID 0")
+	} else {
+		l.lockChan <- 1
+	}
+}
```
jrouzierinverse, line 124: Is id used?

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,196 @@
...
+		l.internalLock.Lock()
+		select {
+		case <-timeoutTimer.C:
+			l.lockChan <- 1
+			l.internalLock.Unlock()
+			l.handleTimeout(id, "Timeout obtaining or releasing read lock that came from: \n")
+			return 0
+		case <-l.rlockChan:
+			l.lockChan <- 1
+			timeoutTimer.Stop()
```
jrouzierinverse: Won't this happen when you release the timer?

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,196 @@
...
+			l.internalLock.Unlock()
+			return id
+		}
+	}
+}
+
+func (l *RWLock) RUnlock(id uint64) {
```
jrouzierinverse, line 182: It looks like id is not used.
Return an error instead.

jrouzierinverse (changes requested):
The public interface of timed rwlock does not match it's internal implementation.
Since returned id is not used for updating any internal state.
It would be better to simplify the return values to either just return a boolean or an error.

julsemaan:
Performed the cleanup and the signature changes as verbally discussed

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,191 @@
...
+	case <-l.lockChan:
+		timeoutTimer.Stop()
+		for time.Now().Before(stopAt) {
+			l.internalLock.Lock()
+			if len(l.rlockChan) == MAX_RLOCK {
+				l.internalLock.Unlock()
+				return id, nil
+			} else {
+				l.internalLock.Unlock()
+			}
```
jrouzierinverse, line 105: In this busy loop call time.Sleep() or runtime.Goscheld() after a few iterations do a avoid pegging the CPU to 100%

julsemaan:
Adjusted, ready for another review

julsemaan:
@jrouzierinverse 

Can you re-review this and merge if the changes are fine

PR #3385, opened 2018-07-12
feature/tenant-aware

jrouzierinverse:
Description
===========
Tenant Aware

Impacts
=======
pfconfig realm 

Delete branch after merge
=========================
NO

NEWS file entries
=================

New Features
------------
* pfconfig is tenant aware
* Realm are tenant scope

UPGRADE
=================
Upgrade realm.conf to be tenant aware:

perl  /usr/local/pf/addons/upgrade/to-8.2-realm-conf.pl

### DIFF
--- a/conf/realm.conf.defaults
+++ b/conf/realm.conf.defaults
@@ -1,14 +1,14 @@
-[DEFAULT]
+[1 DEFAULT]
 admin_strip_username=disabled
 radius_strip_username=disabled
 portal_strip_username=disabled
 
-[LOCAL]
+[1 LOCAL]
 admin_strip_username=disabled
 radius_strip_username=disabled
 portal_strip_username=disabled
 
-[NULL]
+[1 NULL]
 admin_strip_username=disabled
 radius_strip_username=disabled
 portal_strip_username=disabled
--- a/lib/pf/ConfigStore/Role/TenantID.pm
+++ b/lib/pf/ConfigStore/Role/TenantID.pm
@@ -1,0 +1,82 @@
+package pf::ConfigStore::Role::TenantID;
+
+=head1 NAME
+
+pf::ConfigStore::Role::TenantID -
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::ConfigStore::Role::TenantID
+
+=cut
+
+use strict;
+use warnings;
+use Moo::Role;
+use pf::config::tenant;
+
+=head2 _Section
+
+_Section
+
+=cut
+
+around _Sections => sub {
+    my $orig = shift;
+    my $self = shift;
+    my $tenant_id = pf::config::tenant::get_tenant();
+    return grep { s/^\Q$tenant_id\E // } $orig->($self, @_);
+};
+
+=head2 _formatSectionName
+
+=cut
+
+around _formatSectionName => sub {
+    my $orig = shift;
+    my $self = shift;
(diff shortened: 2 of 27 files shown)


extrafu:
@julsemaan Please review and test that PR.

conf/realm.conf.defaults
```diff
@@ -1,14 +1,14 @@
-[DEFAULT]
+[1 DEFAULT]
```
julsemaan, line 1: I think we discussed it but for the first tenant, the ID wouldn't be necessary (it would be implicit that its tenant 1 if there is no tenant ID there)

That would keep the configuration simpler and would allow for backward compatibility with our current admin

lib/pf/ConfigStore/Role/TenantID.pm
```diff
@@ -0,0 +1,82 @@
+package pf::ConfigStore::Role::TenantID;
```
julsemaan, line 1: Proper perldoc needed in this file

t/provisioner.t
```diff
@@ -32,7 +32,7 @@ our $ANDROID_OS = 'Android OS';
 our $TEST_OS = 'iOS';
 our $TEST_OS_FINGERBANK_ID = '33450';
 
-our $TEST_NODE_ATTRIBUTE = { category => $TEST_CATEGORY };
+our %TEST_NODE_ATTRIBUTE = ( category => $TEST_CATEGORY, device_type => $TEST_OS, device_name => $TEST_OS );
```
julsemaan: Not sure the changes in this file belong in this PR

julsemaan (changes requested):

julsemaan:
Also @jrouzierinverse 

Can you confirm that this has been fully tested for the case of an Active Directory 802.1x authentication and 802.1x local authentication both in multi-tenant and non multi-tenant contexts

conf/realm.conf.defaults
jrouzierinverse: I would prefer to keep it consistent and not have special cases.
With that said will need to add a script to handle the upgrade process.

jrouzierinverse:
Tested

extrafu:
@julsemaan Please merge if all is fine.

julsemaan:
We still need the upgrade script to be done since realm.conf isn't retro-compatible with this feature

jrouzierinverse:
Done

julsemaan:
Perhaps I'm blind, but there is no new commit for the perldoc and everything is as it was...

julsemaan (changes requested):
The admin interface for the realms hasn't been adjusted, so people have to actually figure out that they have to prefix the realm with the tenant ID so that it works. It should instead be an attribute of the realm configuration modal

julsemaan:
Also, the realms should be organized by tenant in the admin instead of being mixed together

jrouzierinverse:
No need at the moment we will only see the realm for tenant 1

jrouzierinverse:
Need to borrow my glasses?

https://github.com/inverse-inc/packetfence/pull/3385/commits/5c0c0a4a2a5118e0a318f49abbe6951bf11ef5b1

julsemaan:
The file on which I commented has the invalid perldoc was:  addons/upgrade/to-8.2-realm-conf.pl

jrouzierinverse:
done

PR #3395, opened 2018-07-17
Fix/person lookup strip

### LINKED ISSUE #3393: Person lookup in 802.1x doesn't use stripped username
The person lookup that is triggered via 802.1x requests isn't checking if the source requires the username to be stripped.

That causes objects to not be found and the lookup to fail

julsemaan:
# Description
Perform stripping of the username if necessary when performing a person lookup

# Impacts
person lookup

# Issue
fixes #3393 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed issue where username wasn't stripped during a person lookup (#3393)

### DIFF
--- a/lib/pf/lookup/person.pm
+++ b/lib/pf/lookup/person.pm
@@ -35,7 +35,8 @@ Lookup informations on a person
 =cut
 
 sub lookup_person {
-    my ($pid, $source_id) = @_;
+    my ($pid, $source_id, $context) = @_;
+    use Data::Dumper; get_logger->info(Dumper(\@_));
     my $logger = get_logger();
     unless (defined $source_id) {
         $logger->info("undefined source id provided");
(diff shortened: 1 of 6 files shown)


lib/pf/lookup/person.pm
```diff
@@ -35,7 +35,8 @@ Lookup informations on a person
 =cut
 
 sub lookup_person {
-    my ($pid, $source_id) = @_;
+    my ($pid, $source_id, $context) = @_;
+    use Data::Dumper; get_logger->info(Dumper(\@_));
```
fdurand: debug maybe

julsemaan: Removed it instead, it was just for my own debug

julsemaan:
@fdurand 

Can you perform a last check + merge if its all good

I'd like to have this in the maintenance for a deployment I'm finalizing

fdurand (approved):

PR #3429, opened 2018-07-30
Make userPrincipalName available for 802.1x authentication (ntlm_auth)

fdurand:
# Description
Permit the use of the userPrincipalName attribute in 802.1x.

# Impacts
RADIUS authentication workflow

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Make userPrincipalName for 802.1x


lib/pf/services/manager/radiusd_child.pm
```diff
@@ -408,6 +441,50 @@ sub generate_radiusd_sqlconf {
...
+   foreach my $ldap (keys %ConfigAuthenticationLdap) {
+      $tags{'servers'} .= <<"EOT";
+
+ldap $ldap {
+    server          = "$ConfigAuthenticationLdap{$ldap}->{host}"
+    port            = "$ConfigAuthenticationLdap{$ldap}->{port}"
+    identity        = "$ConfigAuthenticationLdap{$ldap}->{binddn}"
+    password        = "$ConfigAuthenticationLdap{$ldap}->{password}"
+    base_dn         = "$ConfigAuthenticationLdap{$ldap}->{basedn}"
+    filter          = "(userPrincipalName=%{User-Name})"
```
jrouzierinverse, line 519: We should quote the username to avoid ldap injection.
"(userPrincipalName=%{ldapquote:&User-Name})"

lib/pf/services/manager/radiusd_child.pm
```diff
@@ -408,6 +441,50 @@ sub generate_radiusd_sqlconf {
...
+    scope           = $ConfigAuthenticationLdap{$ldap}->{scope}
+    base_filter     = "(objectclass=user)"
+    rebind          = yes
+    chase_referrals = yes
+    update {
+        control:AD-Samaccountname := 'sAMAccountName'
+    }
+    user {
+        base_dn = "\${..base_dn}"
+        filter = "(|(userPrincipalName=%{User-Name})(sAMAccountName=%{%{Stripped-User-Name}:-%{User-Name}}))"
```
jrouzierinverse: We should use ldapquote to avoid ldap injection.

jrouzierinverse (commented):
We do not use ldapquote in the LDAP search filters

fdurand:
Question, do we want to be able to use another ldap attribute, like email ?

html/pfappserver/lib/pfappserver/Form/Config/Realm.pm
```diff
@@ -85,6 +86,29 @@ has_field 'admin_strip_username' =>
              help => 'Should the usernames matching this realm be stripped when used on the administration interface' },
   );
 
+has_field 'permit_userPrincipalName' =>
+  (
+   type => 'Toggle',
+   checkbox_value => "enabled",
+   unchecked_value => "disabled",
+   default => "enabled",
```
julsemaan: Should default to disabled actually, since when its empty in the config file it will not be enabled. And anyway I think we're better off keeping backward compatibility where the upn can only be used for machine auth

fdurand:
It miss the ldaps/ttls configuration generation in the Freeradius config.

extrafu:
@jrouzierinverse Please review and comment before tomorrow.

extrafu:
@jrouzierinverse What's the status on this?

jrouzierinverse:
rebase

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -605,6 +606,19 @@ sub preMatchProcessing {
...
 }
 
+=head2 _makefilter
+
+Create the filter to search for the dn
+
+=cut
+
+sub _makefilter {
+  my ($self,$username) = @_;
```
jrouzierinverse, line 616: Escape the LDAP user name with escape_filter_value

fdurand:
Done

lib/pf/services/manager/radiusd_child.pm
jrouzierinverse: ldapquote

lib/pf/services/manager/radiusd_child.pm
```diff
@@ -453,8 +490,76 @@ sub escape_freeradius_string {
...
+sub generate_radiusd_ldap {
+    my ($self, $tt) = @_;
+
+    my %tags;
+    $tags{'template'}    = "$conf_dir/radiusd/ldap_packetfence.conf";
+    $tags{'install_dir'} = $install_dir;
+    foreach my $ldap (keys %ConfigAuthenticationLdap) {
+        my $searchattributes;
+        foreach my $searchattribute (@{$ConfigAuthenticationLdap{$ldap}->{searchattributes}}) {
+            $searchattributes .= '('.$searchattribute.'=%{User-Name})';
```
jrouzierinverse, line 508: ldapquote

jrouzierinverse (changes requested):

lib/pf/services/manager/radiusd_child.pm
fdurand: I tested it but after that it doesn't work, so i revert.

jrouzierinverse: Ok

lib/pf/Authentication/Source/LDAPSource.pm
fdurand: done

PR #3437, opened 2018-08-02
support passthroughs for domains with dns load balance

### LINKED ISSUE #3417: Wrong IPSET entries are being created for a DNS query for some bypassed domains
### Packetfence 8.1.0

For a setup with vlan-enforcement, some DNS queries coming to pfdns from the registration vlan are somehow being proxied to the upstream DNS and back to the client without the response being added to the proper ipset.

I was looking at #3396 and added some spew debugging as well as merged the patches suggested in the thread and I came to the following conclusion:

Actually, a different response is being added to the IPSET. And its the response coming from the "LocalResolver" method. Its response is being used to populate the IPSET filter but its never sent upstream. Instead, the pfdns plugin is let through to the next plugin (proxy, as specified in the pfdns.conf) which then issues another request, for the same domain, but the response might be different.

In my case in particular, since the domains I'm trying to whitelist are the google OAuth ones, the response is pretty much ALWAYS different. As such, the client will probably receive a DNS response that doesn't match what was installed in the IPSET.

I believe that for all the cases where we got an answer from the LocalResolver, we should populate the response record and not continue in the plugin chain (like it happens, for example, for the portal address)

garci66:
# Description
Improved the handling of domains where DNS is performing load-balance so that the response back to the client and the whitelist added to pfilter matches. If the local resolver returns a valid response, this response is used to populate pffilter and also shortcuts pfdns proxy plugins sending the response directly back to the customer. In case of DNS errors, it also propagates the error back to the client (not returning the portal but rather an error)

# Impacts
PFDNS 

# Issue
fixes #3417 

# Delete branch after merge
YES 

# NEWS file entries
## Bug Fixes
* Wrong IPSET entries are being created for a DNS query for some bypassed domains (#3417)

### DIFF
--- a/go/coredns/plugin/pfdns/pfdns.go
+++ b/go/coredns/plugin/pfdns/pfdns.go
@@ -174,7 +174,7 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 			answer, err := pf.LocalResolver(state)
 			if err != nil {
 				fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
-				} else {
+			} else {
 				for _, ans := range append(answer.Answer, answer.Extra...) {
 					switch ansb := ans.(type) {
 					case *dns.A:
--- a/go/coredns/plugin/pfdns/pfdns.go
+++ b/go/coredns/plugin/pfdns/pfdns.go
@@ -122,21 +122,22 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		if k.MatchString(state.QName()) {
 			answer, err := pf.LocalResolver(state)
 			if err != nil {
-				fmt.Println("Local resolver error with the following error" + err.Error())
+				fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
 			} else {
 				for _, ans := range append(answer.Answer, answer.Extra...) {
 					switch ansb := ans.(type) {
 					case *dns.A:
 						for _, valeur := range v {
 							if err := pf.SetPassthrough(ctx, "passthrough", ansb.A.String(), valeur, true); err != nil {
-								fmt.Println("Not able to contact localhost", err)
+								fmt.Println("Not able to contact localhost for setPassthrough", err)
 							}
 						}
 					}
 				}
+				fmt.Println(srcIP + " : " + mac + " Domain bypass for fqdn " + state.QName())
+				w.WriteMsg(answer)
 			}
-			fmt.Println(srcIP + ":" + mac + " Domain bypass for fqdn " + state.QName())
-			return pf.Next.ServeDNS(ctx, w, r)
+			return 0, nil
 		}
 	}
 
@@ -147,7 +148,7 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 			if k.MatchString(state.QName()) {
 				answer, err := pf.LocalResolver(state)
 				if err != nil {
-					fmt.Println("Local resolver error with the following error" + err.Error())
+					fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
 				} else {
 					for _, ans := range append(answer.Answer, answer.Extra...) {
 						switch ansb := ans.(type) {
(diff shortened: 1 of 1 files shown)


go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -171,8 +173,8 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		if k.MatchString(state.QName()) {
 			answer, err := pf.LocalResolver(state)
 			if err != nil {
-				fmt.Println("Local resolver error with the following error" + err.Error())
-			} else {
+				fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
+				} else {
```
fdurand: Fix indentation

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -122,21 +122,22 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		if k.MatchString(state.QName()) {
 			answer, err := pf.LocalResolver(state)
 			if err != nil {
-				fmt.Println("Local resolver error with the following error" + err.Error())
+				fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
```
fdurand, line 125: return 0, nil

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -122,21 +122,22 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
 							}
 						}
 					}
 				}
+				fmt.Println(srcIP + " : " + mac + " Domain bypass for fqdn " + state.QName())
+				w.WriteMsg(answer)
 			}
-			fmt.Println(srcIP + ":" + mac + " Domain bypass for fqdn " + state.QName())
-			return pf.Next.ServeDNS(ctx, w, r)
+			return 0, nil
```
fdurand, line 140: It will never be executed

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -147,7 +148,7 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 			if k.MatchString(state.QName()) {
 				answer, err := pf.LocalResolver(state)
 				if err != nil {
-					fmt.Println("Local resolver error with the following error" + err.Error())
+					fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
```
fdurand, line 151: return 0, nil

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -159,9 +160,10 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 							}
 						}
 					}
+					fmt.Println(srcIP + " : " + mac + " isolation passthrough  for fqdn " + state.QName())
+					w.WriteMsg(answer)
 				}
-				fmt.Println(srcIP + " : " + mac + " isolation passthrough  for fqdn " + state.QName())
-				return pf.Next.ServeDNS(ctx, w, r)
+				return 0, nil
```
fdurand, line 166: It will never be executed

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -171,8 +173,8 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 		if k.MatchString(state.QName()) {
 			answer, err := pf.LocalResolver(state)
 			if err != nil {
-				fmt.Println("Local resolver error with the following error" + err.Error())
-			} else {
+				fmt.Println("Local resolver error for fqdn" + state.QName() + " with the following error" + err.Error())
```
fdurand: return 0, nil

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -182,10 +184,12 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
 						}
 					}
+					
 				}
+				w.WriteMsg(answer)
+				fmt.Println(srcIP + " : " + mac + " passthrough for fqdn " + state.QName())
 			}
-			fmt.Println(srcIP + " : " + mac + " passthrough for fqdn " + state.QName())
-			return pf.Next.ServeDNS(ctx, w, r)
+			return 0, nil
```
fdurand, line 192: remove

go/coredns/plugin/pfdns/pfdns.go
garci66: It is getting executed in my code `return 0, nil` on line 140, 166, 192 is shared between the `if` and `else` codes.  On my tests I am seeing being called on both the positive and negative cases in the if. 

The `w.WriteMsg(answer)` requires also a `return 0, nil` so my idea was to share it between the positive and negative case

fdurand: Yes make sense.

PR #3445, opened 2018-08-08
various improvements to maintenance script

julsemaan:
# Description
* Improves output of maintenance script
* Allows to patch Golang binaries even if no new patches are available
* Allows to patch Golang binaries without doing the text patches
* Adjusts the rsyslog configuration and restarts rsyslog after applying the maintenance.

# Impacts
Maintenance script

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Performed various improvements to the maintenance script

### DIFF
--- a/addons/pf-maint.pl
+++ b/addons/pf-maint.pl
@@ -293,7 +293,7 @@ sub accept_binary_patching {
 }
 
 sub install_binary_sign_key_if_needed {
-    my $rc = system("gpg --list-keys $BINARIES_SIGN_KEY_ID");
+    my $rc = system("gpg --list-keys $BINARIES_SIGN_KEY_ID > /dev/null");
     if($rc != 0) {
         $rc = system("gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys $BINARIES_SIGN_KEY_ID");
         die "Cannot install signing key\n" if $rc != 0;
--- a/addons/pf-maint.pl
+++ b/addons/pf-maint.pl
@@ -120,35 +120,64 @@ our $BINARIES_DIRECTORY = "/usr/local/pf/sbin";
 
 our $BINARIES_SIGN_KEY_ID = "A0030E2C";
 
+our $step = 0;
+
 my $base = $BASE_COMMIT || get_base();
 
 die "Cannot base commit\n" unless $base;
+
+$step++;
+print "=" x 110 . "\n";
+print "Step $step: Patching of text based codefiles\n";
 
 print "Currently at $base\n";
 
 my $head = $COMMIT || get_head();
 if ($base eq $head) {
-    print "Already up to date\n";
-    exit 0;
+    print "Already up to date for text based patches\n";
+} 
+else {
+    exit 1 if defined($test);
+
+    print "Latest maintenance version is $head\n";
+
+    my $patch_data = get_patch_data( $base, $head );
+    show_patch($patch_data);
+    accept_patch() unless $NO_ASK;
+    print "Downloading the patch........\n";
+    save_patch( $patch_data, $base, $head );
+    print "Applying the patch........\n";
+    apply_patch( $patch_data, $base, $head );
 }
-exit 1 if defined($test);
-
-print "Latest maintenance version is $head\n";
-
-my $patch_data = get_patch_data( $base, $head );
-show_patch($patch_data);
-accept_patch() unless $NO_ASK;
-print "Downloading the patch........\n";
-save_patch( $patch_data, $base, $head );
-print "Applying the patch........\n";
-apply_patch( $patch_data, $base, $head );
 
 if($BASE_BINARIES_URL) {
-    accept_binary_patching() unless $NO_ASK;
-    install_binary_sign_key_if_needed();
-    print "Downloading and replacing the binaries........\n";
-    download_and_install_binaries();
+    $step++;
+    print "=" x 110 . "\n";
+    print "Step $step: Patching of the Golang binaries\n";
+
+    my $should_patch = 0;
+    if($NO_ASK) {
+        $should_patch = 1;
+    }
+    elsif(accept_binary_patching()) {
+        $should_patch = 1;
+    }
+
+    if($should_patch) {
+        install_binary_sign_key_if_needed();
+        print "Downloading and replacing the binaries........\n";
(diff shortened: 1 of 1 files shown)


nqb:
@julsemaan: my 2cents.

addons/pf-maint.pl
```diff
@@ -252,12 +282,9 @@ sub print_dot {
 }
 
 sub accept_binary_patching {
-    print "=" x 110 . "\n";
     print "Should we patch the Golang binaries? ".join(",", @patchable_binaries)." Any custom code in them will be overwritten y/n [y]: ";
```
nqb: We need to make "Any custom code in them will be overwritten" more visible.

fdurand (approved):

nqb (changes requested):

addons/pf-maint.pl
julsemaan: Not sure how we can make it more visible other than by putting it just before the [y/n]
Unless we use some fancy bash coloring

nqb:
I think we can also hide `gpg` messages related to sig. They aren't really useful and print at each execution:
```console
Should we patch the Golang binaries? pfhttpd,pfdns,pfdhcp,pfstats,pfdetect Any custom code in them will be overwritten y/n [y]: y
pub   4096R/A0030E2C 2016-03-15
uid                  Inverse Support (RPM package signing) <support@inverse.ca>

Downloading and replacing the binaries........
Performing patching of pfhttpd.......
** GET https://inverse.ca/downloads/PacketFence/CentOS7/binaries/maintenance/8.1/pfhttpd.sig ==> 200 OK (1s)
gpg: Signature made Wed 08 Aug 2018 01:23:22 AM EDT using RSA key ID A0030E2C
gpg: Good signature from "Inverse Support (RPM package signing) <support@inverse.ca>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: B022 C48D 3D63 73D7 FC25  6A8C CB2D 3A2A A003 0E2C
```
could be:
```console
Should we patch the Golang binaries? pfhttpd,pfdns,pfdhcp,pfstats,pfdetect Any custom code in them will be overwritten y/n [y]: y
Downloading and replacing the binaries........
Performing patching of pfhttpd.......
```

addons/pf-maint.pl
nqb: Agree with these two options.

addons/pf-maint.pl
```diff
@@ -120,36 +120,66 @@ =head1 SYNOPSIS
...
 our $BINARIES_SIGN_KEY_ID = "A0030E2C";
 
+our $step = 0;
+
 my $base = $BASE_COMMIT || get_base();
 
 die "Cannot base commit\n" unless $base;
 
+$step++;
+print "=" x 110 . "\n";
```
jrouzierinverse: 110 should be a constant.

addons/pf-maint.pl
```diff
@@ -120,36 +120,66 @@ =head1 SYNOPSIS
...
 
 my $head = $COMMIT || get_head();
 if ($base eq $head) {
-    print "Already up to date\n";
-    exit 0;
+    print "Already up to date for text based patches\n";
+} 
+else {
+    print "Text based patche(s) available\n";
+    exit 1 if defined($test);
```
jrouzierinverse: exit 0 instead of 1 since it not a failure

julsemaan:
@nqb 

Let me know how to get this warning out and we can do it.

Has to be found quick though because we need this for 8.1.1 which we want to get out sooner than later

addons/pf-maint.pl
julsemaan: Its already just before the [y/n] prompt

jrouzierinverse: Verbiage suggestion.

```perl
print "Patch the Golang binaries? " . join(",", @patchable_binaries) . "\n";
print "Any custom code will be overwritten!!\n";
print "y/n [y]: ";
```

jrouzierinverse (changes requested):

addons/pf-maint.pl
julsemaan: It has to be "exit 1" since it has to be a failure for the monit check to send us an alert when a patch is available.

addons/pf-maint.pl
julsemaan: +1 lets do this

addons/pf-maint.pl
jrouzierinverse: Ok

nqb:
@julsemaan: I don't know how to hide `gpg` messages. Call to `system` function seems to send output to STDOUT. I didn't find any variable or parameter in [perldoc](https://perldoc.perl.org/functions/system.html) to change this behaviour. 

We can simply redirect STDOUT to `/dev/null`. Another option: use `qx` operator, see [perldoc perlop](https://perldoc.perl.org/perlop.html#Quote-Like-Operators).

@jrouzierinverse: What's your opinion on that point ?

jrouzierinverse:
Try the --quiet and --no-tty options
If those don't work then try
--logger-file /dev/null

```bash
gpg --quiet --no-tty --batch --yes --output $binary_path-maintenance-decrypted --decrypt $binary_path-maintenance-encrypted
```

nqb:
@jrouzierinverse: with suggested options, I get output on STDOUT.

@julsemaan: if we use `-t` option and there is no text based patches available (already up-to-date), we go at Step 2. Script should exit before.

julsemaan:
The thing is those get outputed on STDERR that's why.

Seems irresponsible to be to redirect error output to /dev/null for warnings but if everybody disagrees, then I'll follow the majority.

nqb:
Not agree for `--list-keys` option:
```shell
$ gpg --list-keys A0030E2C 2> /dev/null
pub   4096R/A0030E2C 2016-03-15
uid                  Inverse Support (RPM package signing) <support@inverse.ca>

$ gpg --list-keys A0030E2C > /dev/null
$ 
```

jrouzierinverse (approved):

nqb (approved):

julsemaan:
@nqb 

Want to have the honor of merging the PR ?

PR #3454, opened 2018-08-10
Feature/status page config

julsemaan:
# Description
Rework the device registration policy so its a self service policy controlling the status page configuration
At the same time, add the ability to only allow specific roles from being unregistered using the status page

# Impacts
Status page + device registration

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Can now prevent specific roles from unregistering devices on the self-service portal (status page)

# UPGRADE file entries

The device registration configuration file has been removed in favor of using a configuration file for all the self service portal features (status page + device registration).

In order to migrate your configuration, run the following script:
```
/usr/local/pf/addons/upgrade/to-8.2-selfservice-conf.pl
```


julsemaan:
Ready for review

jrouzierinverse (changes requested):
Some files were missed.

- t/unittest/UnifiedApi/Controller/Config/DeviceRegistrations.t
- go/api-frontend/aaa/authorization.go
- addons/packages/packetfence.spec
- debian/packetfence.conffiles
- conf/profiles.conf.defaults
- t/unittest/UnifiedApi/GenerateSpec.t
- html/captive-portal/lib/captiveportal/Controller/DeviceRegistration.pm

Would it be better to say 'The Self Service module is not enabled' here
https://github.com/inverse-inc/packetfence/pull/3454/files#diff-f4108bf635561a1df96019c6b71e7f38R36

julsemaan:
I suck at grep :/

Let me fix this

jrouzierinverse:
Don't use grep use https://geoff.greer.fm/ag/ it is just better

addons/upgrade/to-8.2-selfservice-conf.pl
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+to-8.2-selfservice-conf.pl
```
julsemaan: Upgrade script lacks update of profiles.conf

julsemaan:
@jrouzierinverse 

Pushed changes, as for changing the working in the user message, device registration is still device registration, it hasn't become self service, self service represents the self-service portal which includes the status page and device registration.

Same reason as above to keep html/captive-portal/lib/captiveportal/Controller/DeviceRegistration.pm
since the controller still exists for device registration html/captive-portal/lib/captiveportal/PacketFence/Controller/DeviceRegistration.pm

julsemaan:
reworked PR, ready for another review

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -76,13 +76,16 @@ sub is_lost_stolen {
 sub userIsAuthenticated : Private {
     my ( $self, $c ) = @_;
     my $pid   = $c->user_session->{"username"};
-    my @nodes = person_nodes($pid);
-    foreach my $node (@nodes) {
+    my @person_nodes = person_nodes($pid);
```
jrouzierinverse, line 79: To avoid doing two database calls.
Add the additional fields and joins [here](https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/person.pm#L264) instead.

html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
```diff
@@ -21,40 +21,62 @@ has_field 'id' =>
...
    element_class => ['chzn-deselect'],
    element_attr => {'data-placeholder' => 'Click to add an OS'},
    tags => { after_element => \&help,
-             help => 'List of OS which will be allowed to be register via the self service portal.' },
+             help => 'List of OS which will be allowed to be registered via the self service portal.' },
    fingerbank_model => "fingerbank::Model::Device",
   );
 
-has_block definition =>
+has_block global_definition =>
```
jrouzierinverse: Why change our current naming convention for the main block?

If you really feel a change is needed then to do a search and replace on all the forms and templates.

jrouzierinverse (changes requested):

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
julsemaan: From what I recall that specific join was removed from this query (even before DAL) because it was causing latency at some 3 letter telecom company which has a very large node table.

So that is why I didn't add it there since this is called in the RADIUS to check the max nodes

html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
julsemaan: Not sure how that got there, I'll revert the change

julsemaan: Actually, scratch my previous comment didn't look at the line.

I actually made it like this because there are multiple blocks which represents different "sections" of the self-service portal

There is one for device registration, one for status page, so the one that applied to both had to be called 'global_definition' to have it differentiate from the rest of the definition in the other forms which represent the whole config definition not sure a part of it

jrouzierinverse: Even then our current convention is to have a definition block then additional blocks.
This is just a matter of consistency.

If you want to change the convention I have no issue with that just change it everywhere.

```
html/pfappserver/lib/pfappserver/Form/Config/Domain.pm
html/pfappserver/lib/pfappserver/Form/Config/ProfileCommon.pm
html/pfappserver/lib/pfappserver/Form/Config/Source/Mirapay.pm
html/pfappserver/lib/pfappserver/Form/Config/Provisioning/opswat.pm
html/pfappserver/lib/pfappserver/Form/Config/Provisioning/mobileconfig.pm
html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
html/pfappserver/lib/pfappserver/Form/Config/Source.pm
```

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
jrouzierinverse: person_nodes is currently only used in the following
```
html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
html/pfappserver/lib/pfappserver/PacketFence/Model/User.pm
```
Unless my grepping is off I do not see how where is this called in RADIUS.

As for what I can tell we use pf::node::node_pid to check max nodes.

julsemaan: I'll change this in that case, my memory has failed me

I shall blame @lzammit anyway

extrafu:
@julsemaan Should we push this for v9.1?

julsemaan:
Would have been nice for 9.0 but its full of conflicts now.

I guess if I have time to get to it it will be in, otherwise, 9.1

jrouzierinverse:
rebase

julsemaan:
I think this will have to go to 9.1 given the amount of work that needs to go into this (including adjustments to the new GUI)

julsemaan:
I've refreshed this PR so that it supports the new v9 admin and rebased the code

I just didn't do that comment:
> To avoid doing two database calls.
> Add the additional fields and joins [here](https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/person.pm#L264) instead.

Because we would need to start joining to the node_category table to have the necessary fields and I'd have to replicate almost all the fields in node_view so that the user can have access to the full fields of node like he would expect in status.html

julsemaan:
It would be good to have this brought in for 9.1 so that this doesn't rot another time

Issue #3097, opened 2018-04-06
PF7.2 Guest Sponsor email CSS: Email is aligned to the right in Outlook


atran-inverse:
When sponsor receives email to register guest, Outlook aligns the email on the right:

julsemaan:
This was fixed in #3476

PR #3476, opened 2018-08-16
Email templates improvements

cgx:
# Description
Better responsive email templates using the [mjml framework](https://mjml.io/).

# Impacts
Email templates will be readable on all devices. Yeah!

# NEWS file entries

```
## Enhancements

* improved email templates
```

## Bug Fixes

* PF7.2 Guest Sponsor email CSS: Email is aligned to the right in Outlook (#3097)


jrouzierinverse (changes requested):
The packaging is not complete.

cgx:
I improved the Makefile and added package.json. Should be good to merge.

julsemaan (commented):
I'll throw the question we recently asked ourselves about generated templates/stylesheets/javascripties thingies

Should be have the generated HTML email templates not tracked in our git repo and build them in our packaging so that way we're not tracking generated files?

Possible answers I can think of:
- [ ] Yes
- [ ] No
- [ ] Semaan should mind his own business
- [ ] We should turn this into an @lzammit joke

cgx:
Most of these templates were not modified for the last two years. Generating the HTML files every time we package PacketFence didn't sound necessary to me.

julsemaan:
@cgx 

That sounds legit to me, its true that these aren't touched much, so on my side its good to merge

extrafu:
@jrouzierinverse Please merge and update the NEWS file.

PR #3484, opened 2018-08-20
Replace pfdhcp backend for options

fdurand:
# Description
Moved from etcd to mysql the pfdhcp backend to store the options.

# Impacts
DHCP

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* Use MySQL as backend for pfdhcp options

# UPGRADE file entries
* Remove packetfence-etcd systemd service


db/pf-schema-X.Y.Z.sql
```diff
@@ -1259,6 +1259,16 @@ CREATE TABLE `api_user` (
   PRIMARY KEY (`username`)
 ) ENGINE=InnoDB;
 
+--
+-- Table structure for table `keyval`
+--
+
+CREATE TABLE keyval (
```
extrafu: I think we should rename this table to something more meaningful. Are we planning to eventually use it for other key-value pairs? If not, it should have dhcp in its name.

julsemaan: I think that would be a good table for distributed caching that will be usable in Golang and in Perl since CHI only works in Perl AFAIK.

But then since this is persistent, perhaps not having cache in its name either would be good.

So I'd suggest: key_value_storage

Its pretty much what fab named it, just not abbreviated and a bit more self-explanatory.

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -52,4 +52,15 @@ call ValidateVersion;
...
 
+--
+-- Table structure for table `keyval`
+--
+
+CREATE TABLE keyval (
+  id VARCHAR(255),
+  value LONGBLOB,
+  PRIMARY KEY(id)
+) ENGINE=InnoDB;
```
satkunas: If `LONGBLOB` is going to hold a lot of data, we should try to use `ROW_FORMAT=COMPRESSED`

fdurand:
BUMP

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -91,4 +91,25 @@ BEGIN
...
+  id VARCHAR(255),
+  value BLOB,
+  PRIMARY KEY(id)
+) ENGINE=InnoDB;
+
+--
+-- Table structure for table `keyval`
+--
+
+CREATE TABLE keyval (
```
julsemaan: I think you forgot to remove that one

julsemaan:
Other than my cleanup comment, we're good on my side.

We'd have to see if we compress the long blobs like @satkunas suggested

fdurand:
It's not a long blobs anymore.

julsemaan:
Thats what she said

PR #3492, opened 2018-08-22
Feature/bandwidth balance increase

### LINKED ISSUE #3477: Max Bandwidth Balance
The maximum `bandwidth_balance` on a `node` is limited to 4GB - 1B using `UNSIGNED INT(10)`.

IMO this column should allow much larger values, as GB's of throughput can be common in many local and intranets.

satkunas:
...


satkunas:
Rebuild in https://github.com/inverse-inc/packetfence/pull/3493

PR #3502, opened 2018-08-23
add explicit output for database-backup-and-maintenance.sh in cluster env

### LINKED ISSUE #3428: database-backup-and-maintenance.sh: should be more explicit in cluster env
In Galera Cluster env, running `database-backup-and-maintenance.sh` script on a server which is not the first IP address of `wsrep_incoming_addresses`, doesn't return an explicit output:
```
# /usr/local/pf/addons/database-backup-and-maintenance.sh
/root/backup/ , folder already created. 

tar: Removing leading `/' from member names
tar: Removing leading `/' from hard link targets
packetfence-files-dump have been created in  /root/backup/ 

packetfence-files-dump older than 7 days have been removed. 
```

nqb:
# Description
Add explicit output for database-backup-and-maintenance.sh in cluster env

# Impacts
addons/database-backup-and-maintenance.sh

# Issue
fixes #3428 

# Delete branch after merge
YES

### DIFF
--- a/addons/database-backup-and-maintenance.sh
+++ b/addons/database-backup-and-maintenance.sh
@@ -83,6 +83,9 @@ if [ -f /var/lib/mysql/grastate.dat ]; then
     if ! ip a | grep $FIRST_SERVER > /dev/null; then
         SHOULD_BACKUP=0
         echo "Not the first server of the cluster: database backup canceled."
+        exit $BACKUPRC
+    else
+        echo -e "First server of the cluster : database backup will start.\n"
     fi
 fi
 
@@ -170,4 +173,3 @@ if [ $ACTIVATE_REPLICATION == 1 ]; then
 fi
 
 exit $BACKUPRC
-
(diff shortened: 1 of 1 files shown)


addons/database-backup-and-maintenance.sh
```diff
@@ -82,6 +82,7 @@ if [ -f /var/lib/mysql/grastate.dat ]; then
     FIRST_SERVER=`mysql -u$REP_USER -p$REP_PWD -e 'show status like "wsrep_incoming_addresses";' | tail -1 | awk '{ print $2 }' | awk -F "," '{ print $1 }' | awk -F ":" '{ print $1 }'`
     if ! ip a | grep $FIRST_SERVER > /dev/null; then
         SHOULD_BACKUP=0
+        echo "Not the first server of the cluster: database backup canceled."
```
nqb, line 85: I think I can add an `exit 0` statement after that line because processing replication block (at end of file) doesn't made sense in that situation.

What's your opinion ?

fdurand: exit 0 will bypass the replication

nqb: As discussed with @fdurand , we can do an `exit 0` at that point. If we are not the first server of the cluster, we don't have backups to replicate. Moreover, replication block is more related to active-passive clustering and not impact by following test:
```
if [ -f /var/lib/mysql/grastate.dat ]
```

nqb:
Please review

nqb:
@fdurand: ready to merge ?

addons/database-backup-and-maintenance.sh
mehdiroo: Hi

PR #3510, opened 2018-08-27
Replace MySQL user_name and computer_name column '=5C' with '\'

### LINKED ISSUE #3508: Backslash in usernames in Reports section is shown as "=5C"
In the following reports:
Top Authentication Failures
Top Authentication Successes

Authenticated wired 802.1x domain users (DOMAIN\user.name) and local computer users (.\administrator) are displayed with "=5C" instead of backslash in front of their username. Following the username link also doesn't work because the username in the link is also not correct (in Users sections they are correct - DOMAIN\user.name )

Attaching a screenshot of the issue: 

slash

satkunas:
# Description
Replaces '=5C'  with '\' in reports for top authentication successes and failures user_name and computer_name.

# Impacts
none

# Issue
fixes #3508 

# Delete branch after merge
YES

## Bug Fixes
* Backslash in usernames in Reports section is shown as "=5C" (#3508)

### DIFF
--- a/UPGRADE.asciidoc
+++ b/UPGRADE.asciidoc
@@ -19,7 +19,7 @@ Database backup
 Before making any changes to your database, ensure that you have a backup.
 A complete database backup can be taken using this command:
 
-  mysqldump --opt -u root -p pf | gzip > /root/packetfence_db.sql.gz
+  mysqldump --opt --routines -u root -p pf | gzip > /root/packetfence_db.sql.gz
 
 If your database is more than a few hundred megabytes, you may also want to consider using a tool such as Percona XtraBackup which makes for much faster restores than mysqldump.
 
@@ -427,7 +427,7 @@ Database schema update
 Before making any changes to your database, ensure that you have a backup.
 A complete database backup can be taken using this command:
 
-  mysqldump --opt -u root -p pf | gzip > /root/packetfence_db.sql.gz
+  mysqldump --opt --routines -u root -p pf | gzip > /root/packetfence_db.sql.gz
 
 If your database is more than a few hundred megabytes, you may also want to consider using a tool such as Percona XtraBackup which makes for much faster restores than mysqldump.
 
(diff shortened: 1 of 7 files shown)


satkunas:
@julsemaan not tested.

jrouzierinverse:
freeradius encodes the sql changing special characters to the following format '=6F' where 6F is the hexadecimal representation of the character.

So either this can be done in perl or MySQL
```perl
$value =~ s/=([a-fA-F0-9]{2})/chr(hex($1))/ge;
```

```sql
DELIMITER $$

DROP FUNCTION IF EXISTS FREERADIUS_DECODE $$

CREATE FUNCTION FREERADIUS_DECODE (str text) 
RETURNS text
DETERMINISTIC
BEGIN 
    DECLARE result text;
    DECLARE ind INT DEFAULT 0;

    SET result = str;
    WHILE ind <= 255 DO
       SET result = REPLACE(result, CONCAT('=', LPAD(LOWER(HEX(ind)), 2, 0)), CHAR(ind));
       SET result = REPLACE(result, CONCAT('=', LPAD(HEX(ind), 2, 0)), CHAR(ind));
       SET ind = ind + 1;
    END WHILE;

    RETURN result;
END$$

DELIMITER ;
```

```sql
select FREERADIUS_DECODE('=5C') as decoded_value;
+---------------+
| decoded_value |
+---------------+
| \             |
+---------------+
```

julsemaan:
I like the second option @jrouzierinverse has suggested since this allows us to integrate it easily in the custom reports

satkunas:
@julsemaan reworked

Note that `db/pf-schema-X.Y.Z.sql` does not currently contain any of our stored procedures. This is something we need to discuss before closing this PR.

julsemaan:
PR looks good to me

@satkunas were you able to test te new code or should this be tested prior to merge?

julsemaan:
We do put the stored procs in the schema file:
https://github.com/inverse-inc/packetfence/blob/devel/db/pf-schema-X.Y.Z.sql#L647

satkunas:
@julsemaan added to schema in https://github.com/inverse-inc/packetfence/pull/3510/commits/988d1b8d1918415471fe57039299fdc1088a3506

satkunas:
> @satkunas were you able to test te new code or should this be tested prior to merge?

@julsemaan not tested fully.

julsemaan:
Won't be cherry-picking because this contains schema changes

Issue #3390, opened 2018-07-13
pfdetect in 8.10 can't compile regex that validates OK in the UI


giannidaprile:
The following regex extracts MAC and IP from a syslog message using named captures, non-greedy matching:

.*\ ip\ (?<ip>.*)?\ .*\ mac\ (?<mac>.*)?\ -.*

pfdetect fails to start with:

panic: regexp: Compile(`.*\ ip\ (?<ip>.*)?\ .*\ mac\ (?<mac>.*)?\ -.*`): error parsing regexp: invalid or unsupported Perl syntax: `(?<`

Probably a side effect of pfdetect being written in golang vs perl the UI code .

giannidaprile:
Cound just be a documentation issue.

Using golang notation works in both the UI and pfdetect:

.*\ ip\ (?P<ip>.*)?\ .*\ mac\ (?P<mac>.*)?\ -.*

See:
https://github.com/google/re2/wiki/Syntax

giannidaprile:
My bad, Examples are being messed up by markup.  Resubmitting with with code blocks:

The following regex extracts MAC and IP from a syslog message using named captures, non-greedy matching:

`.*\ ip\ (?<ip>.*)?\ .*\ mac\ (?<mac>.*)?\ -.*`

pfdetect fails to start with:

panic: regexp: Compile(.*\ ip\ (?<ip>.*)?\ .*\ mac\ (?<mac>.*)?\ -.*): error parsing regexp: invalid or unsupported Perl syntax: (?<

Probably a side effect of pfdetect being written in golang vs perl the UI code .

Cound just be a documentation issue.

Using golang notation works in both the UI and pfdetect:

Cound just be a documentation issue.

Using golang notation works in both the UI and pfdetect:

`.*\ ip\ (?P<ip>.*)?\ .*\ mac\ (?P<mac>.*)?\ -.*`

See:
https://github.com/google/re2/wiki/Syntax

julsemaan:
@giannidaprile 

You're 100% correct, the re2 syntax should be used now

@jrouzierinverse will proceed to adjust our documentation and the upgrade guide

jrouzierinverse:
The real answer to this is to use re::engine::RE2 to test the regex.

PR #3597, opened 2018-09-05
feature/test-syslog-parser-via-api

### LINKED ISSUE #3562: New end point POST config/syslog_parser/:id/test

jrouzierinverse:
Description
===========
Expose the ability to perform a dry run for a regex parser

NEW Package(s) required
=======================
perl-re-engine-RE2

Issue
=====
fixes #3562 #3390

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
@@ -21,6 +21,12 @@ our $class_messages = {
     'regex' => 'It must be a valid RE2 regex',
 };
 
+=head2 get_class_messages
+
+Overridden get_class_messages
+
+=cut
+
 sub get_class_messages {
     my $self = shift;
     return {
--- a/html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
@@ -1,0 +1,73 @@
+package pfappserver::Form::Field::RE2;
+
+=head1 NAME
+
+pfappserver::Form::Field::RE2 - A RE2 regex field
+
+=head1 DESCRIPTION
+
+This field extends the default Text field and checks if the input value is an valid RE2 regex
+
+=cut
+
+use HTML::FormHandler::Moose;
+extends 'HTML::FormHandler::Field::Text';
+
+use pf::util;
+use namespace::autoclean;
+
+
+our $class_messages = {
+    'regex' => 'It must be a valid regex',
+};
+
+sub get_class_messages {
+    my $self = shift;
+    return {
+       %{ $self->next::method },
+       %$class_messages,
+    }
+}
+
+apply
+  (
+   [
+    {
+     check => sub {
+         use re::engine::RE2 -strict => 1;
+         my ( $value, $field ) = @_;
+         return eval { qr/$value/ };
+     },
+     message => sub {
+         my ( $value, $field ) = @_;
+         return $field->get_message('regex');
+     },
+    }
+   ]
+  );
+
+=head1 COPYRIGHT
+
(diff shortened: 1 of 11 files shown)


html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
```diff
@@ -0,0 +1,73 @@
...
+       %$class_messages,
+    }
+}
+
+apply
+  (
+   [
+    {
+     check => sub {
+         use re::engine::RE2 -strict => 1;
```
julsemaan, line 43: You probably know better than me, but I think this should go at the top of the file since when a module is used, then its imported at startup even if its there.

I may be wrong though, but just wondering why this isn't with the other imports

html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
```diff
@@ -0,0 +1,73 @@
...
+
+use HTML::FormHandler::Moose;
+extends 'HTML::FormHandler::Field::Text';
+
+use pf::util;
+use namespace::autoclean;
+
+
+our $class_messages = {
+    'regex' => 'It must be a valid regex',
```
julsemaan: Perhaps saying "valid RE2 regex" to give a clue to the user what spec to look at for his regex

lib/pf/UnifiedApi/Controller/Config/SyslogParsers.pm
```diff
@@ -49,7 +50,36 @@ our %TYPES_TO_FORMS = (
 sub type_lookup {
     return \%TYPES_TO_FORMS;
 }
- 
+
+sub dry_run {
```
julsemaan, line 60: perldoc

lib/pf/detect/parser/regex.pm
```diff
@@ -38,6 +38,7 @@ parse the Line using the rule
 
 sub parseLineFromRule {
     my ($self, $rule, $line) = @_;
+    use re::engine::RE2 -strict => 1;
```
julsemaan, line 41: use this at the top ?

html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
jrouzierinverse: That will turn on the RE2 engine in only within that scope to for all of the admin.
However we could do an empty import on the top level to detect if it was installed.

lib/pf/detect/parser/regex.pm
jrouzierinverse: See above

html/pfappserver/lib/pfappserver/Form/Field/RE2.pm
julsemaan: It would be so that a compile error fires if that is not installed like you said.

So the empty import would be good

jrouzierinverse: I just remembered it will fail no matter where it is.
The builds were breaking because I did not add it as a dependency.

jrouzierinverse:
Ready for rereview

julsemaan:
Does the spec need to be updated ?

jrouzierinverse:
No it is already updated in devel and I need to rebase

jrouzierinverse:
I thought you meant the packetfence.spec
I think you mean the openapi spec yes it does need to be updated.

julsemaan:
Yes sorry openapi spec

image

julsemaan:
Still missing spec update

julsemaan:
@jrouzierinverse 

Do you want to write the spec so we can merge this or should this be postponed to 8.3 ?

extrafu:
Bump.

PR #3617, opened 2018-09-14
Use several email addresses in alerting.emailaddr

### LINKED ISSUE #3578: alerting emailaddr : unexpected behavior with more than one address
The option `alerting.emailaddr` in `pf.conf` is designed for only one address but you can specify more than one, separated by comma without warnings. However, additional email addresses will never be notify.

In my opinion, we have two solutions:
1. disallow additional address in this field
1. permit additional address in this field

I'm not sure to have the big picture to choice between 1 and 2. What's your opinion ?

nqb:
# Description
Permit to use several email addresses in `alerting.emailaddr` (variable in `pf.conf`).
This variable is also used to generate `keepalived` configuration.

# Impacts
- `keepalived` config
- documentation
- `pfmailer()` function

# Issue
fixes #3578 

# Delete branch after merge
YES

# TODO

- [x] Correctly indent email addresses in `keepalived` generated config.

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -632,8 +632,8 @@ EOT
 [alerting.emailaddr]
 type=text
 description=<<EOT
-Email address to which notifications of rogue DHCP servers, violations 
-with an action of "email", or any other PacketFence-related message 
+Comma-delimited list of email addresses to which notifications of rogue DHCP servers, violations 
+with an action of "email", or any other PacketFence-related message.
 goes to.
 EOT
 
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -633,8 +633,7 @@ EOT
 type=text
 description=<<EOT
 Comma-delimited list of email addresses to which notifications of rogue DHCP servers, violations 
-with an action of "email", or any other PacketFence-related message.
-goes to.
+with an action of "email", or any other PacketFence-related message goes to.
 EOT
 
 [alerting.smtpserver]
--- a/lib/pf/services/manager/keepalived.pm
+++ b/lib/pf/services/manager/keepalived.pm
@@ -58,7 +58,8 @@ sub generateConfig {
 
     my %tags;
     $tags{'template'} = "$conf_dir/keepalived.conf";
-    $tags{'emailaddr'} = $Config{'alerting'}{'emailaddr'};
+    # split email addresses then rejoin them with line feed
+    $tags{'emailaddr'} = join("\n", split( /\s*,\s*/, $Config{'alerting'}{'emailaddr'}));
     $tags{'fromaddr'} = $Config{'alerting'}{'fromaddr'} || "keepalived\@$host_id";
     $tags{'smtpserver'} = $Config{'alerting'}{'smtpserver'};
     $tags{'router_id'} = "PacketFence-$host_id";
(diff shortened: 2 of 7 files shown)


lib/pf/services/manager/keepalived.pm
```diff
@@ -58,7 +58,8 @@ sub generateConfig {
 
     my %tags;
     $tags{'template'} = "$conf_dir/keepalived.conf";
-    $tags{'emailaddr'} = $Config{'alerting'}{'emailaddr'};
+    # split email addresses then rejoin them with line feed
+    $tags{'emailaddr'} = join("\n", split( /\s*,\s*/, $Config{'alerting'}{'emailaddr'}));
```
nqb: Guys,

I need your help to have a well indented `keepalived` generated config. With current code, I get:
```
global_defs {
   notification_email {
     foo@example.com
bar@example.com
}
```

jrouzierinverse: Try something like this
```perl
    $tags{'emailaddr'} = join("\n     ", split( /\s*,\s*/, $Config{'alerting'}{'emailaddr'}));
```

nqb:
I modified `lib/pf/config/util.pm` due to a [MIME::Lite bug](https://rt.cpan.org/Public/Bug/Display.html?id=65606#txn-1028353).

julsemaan (approved):
LGTM

conf/documentation.conf
```diff
@@ -632,8 +632,8 @@ EOT
 [alerting.emailaddr]
 type=text
 description=<<EOT
-Email address to which notifications of rogue DHCP servers, violations 
-with an action of "email", or any other PacketFence-related message 
+Comma-delimited list of email addresses to which notifications of rogue DHCP servers, violations
```
extrafu, line 635: Shouldn't we write "to which notifications of rogue DHCP servers ... should be sent, or or any other PacketFence-related message." ?

lib/pf/services/manager/keepalived.pm
nqb: @jrouzierinverse, @julsemaan: I will prefer using constants rather than a "blank space" and a piece of code like this:
```perl

# in pf/constants/{util,base}.pm ?
Readonly my $SPACE => q{ };
Readonly my $INDENT_SPACES => 4;

# in keepalived.pm
$tags{'emailaddr'} = join("\n" . ($SPACE x $INDENT_SPACES), split( /\s*,\s*/, $Config{'alerting'}{'emailaddr'}));
```
What's your opinion ?

conf/documentation.conf
nqb: I kept the end of text in a separated line

lib/pf/services/manager/keepalived.pm
jrouzierinverse: Since this is very specific to keep alive at the moment just create a constant in lib/pf/services/manager/keepalived.pm.

However if you feel that they should be global constant put it in pf/constants.pm

nqb:
You can review.

julsemaan:
Still LGTM, unless @jrouzierinverse opposes to the wedding of this branch and devel, he shall speak now or forever hold his peace.

PR #3641, opened 2018-09-25
Handle pure-text DHCP Option 82 suboptions Circuit-ID and Remote ID

ptrrkssn:
# Description
Adds a few simple lines of code for handling pure-text suboptions for Circuit-ID and Remote ID for the DHCP Option-82 code. The current code is Cisco-specific and assumes that the first byte for ascii options is 0x01 which isn't the case for Alcatel switches (and possibly others). 

The Alcatel switch configuration we're using looks like this:
ip helper dhcp-snooping option-82 format ascii circuit-id interface vlan delimiter " "
ip helper dhcp-snooping option-82 format ascii remote-id system-name base-mac delimiter " "

# Impacts
pfdhcplistener, if Option-82 support is enabled, and you have non-cisco switches configured for ASCII option-82 sub options.

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Adds support for pure ASCII Circuit-ID and Remote ID for the DHCP Option-82 support

### DIFF
--- a/lib/pf/ConfigStore/Switch.pm
+++ b/lib/pf/ConfigStore/Switch.pm
@@ -117,8 +117,8 @@ _normalizeUplink
 sub _normalizeUplink {
     my ($self, $switch) = @_;
     if ( $switch->{uplink_dynamic} ) {
-        $switch->{uplink}         = 'dynamic';
-        $switch->{uplink_dynamic} = undef;
+        $switch->{uplink_dynamic} = 'dynamic';
+        $switch->{uplink} = undef;
     }
 }
 
--- a/lib/pf/ConfigStore/Switch.pm
+++ b/lib/pf/ConfigStore/Switch.pm
@@ -65,7 +65,11 @@ sub cleanupAfterRead {
     # if the uplink attribute is set to dynamic or not set and the group we inherit from is dynamic
     if ( ($switch->{uplink} && $switch->{uplink} eq 'dynamic') ) {
         $switch->{uplink_dynamic} = 'dynamic';
+<<<<<<< HEAD
         $switch->{uplink} = undef;
+=======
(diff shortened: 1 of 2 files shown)


julsemaan:
That's very cool, thanks for the contribution.

Would you also be able to share the configuration you've put in the switch to enable DHCP option 82 so we can adjust our doc and officially support DHCP option 82 on alcatel

ptrrkssn:
Sorry about the added cruft to the pull request. It's another bugfix that wasn't supposed to go into this one...

(There seems to be a mixup with the uplink_dynamic and uplink fields in lib/pf/ConfigStore/Switch.pm which prevents API-updates of Switches and SwitchGroups).


Back to the Alcatel 6450 switches. It's pretty easy to enable DHCP-snooping (and Option-82):

ip helper dhcp-snooping enable
ip helper dhcp-snooping binding enable
ip helper dhcp-snooping linkagg 1 trust

! This isn't needed, it seems to be enabled by default:
! ip helper dhcp-snooping option-82 data-insertion enable

! By default the data sent is almost Cisco-compatible, but the Module and Port data is binary encoded
! as 2 bytes instead of one so the default code in PacketFence misinterprets the Module/Port values.
! With the patch above we can enable ASCII mode instead:
ip helper dhcp-snooping option-82 format ascii circuit-id interface vlan delimiter " "
ip helper dhcp-snooping option-82 format ascii remote-id system-name base-mac delimiter " "

(The stuff after "ascii circuit-id" and "ascii remote-id" can be changed - there are a lot of different options, but since we don't interpret the data we can use whatever we like).

fdurand:
The original repository doesn't exist anymore. 
We can fetch the code and create a new branch.

PR #3657, opened 2018-10-03
Give privileges to pf user to dump stored routines

### LINKED ISSUE #3652: backups: pf user can't backup DB
Due to following commit: 7fe7c61, pf user, in standalone setup without percona, is no longer able to dump DB:

```
# ./addons/database-backup-and-maintenance.sh 
[..]
mysqldump: pf has insufficent privileges to SHOW CREATE FUNCTION `FREERADIUS_DECODE`!
```
If I remove `--routines` option here: https://github.com/inverse-inc/packetfence/blob/7fe7c6111b6178b1442b7ca284bb13192099cee9/addons/database-backup-and-maintenance.sh#L136

DB backups as pf user work again.

nqb:
# Description
Allow `$pf::db::DB_Config->{user}` to dump stored routines.

# Impacts
- configurator
- upgrade

# Issue
fixes #3652 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Issue to backup PacketFence database as `$pf::db::DB_Config->{user}` with `mysqldump` (#3652)

# UPGRADE file entries
See commit on UPGRADE.asciidoc


UPGRADE.asciidoc
```diff
@@ -1781,15 +1781,24 @@ The api_user table has been deprecated
...
+
+An SQL upgrade script has been provided to upgrade the database from the X.X schema to X.Y.
+To upgrade the database schema and privileges, run the following commands:
 
-  mysql -u root -p pf -v < /usr/local/pf/db/upgrade-X.X.X-X.Y.Z.sql
+ PF_USER=$(perl -I/usr/local/pf/lib -Mpf::db -e 'print $pf::db::DB_Config->{user}')
+ mysql -u root -p -e "GRANT SELECT ON mysql.proc TO $PF_USER@'%';" \
+ -e "GRANT SELECT ON mysql.proc TO $PF_USER@'localhost';" \
+ -e "FLUSH PRIVILEGES;"
+ mysql -u root -p pf -v < /usr/local/pf/db/upgrade-X.X.X-X.Y.Z.sql
```
jrouzierinverse, line 1813: The instructions updating the privileges should be separate from the database instructions.
Provide a bash script to perform the upgrade.
Add a step to run the permission script across the entire cluster since the mysql system tables are not replicated.

nqb: > Add a step to run the permission script across the entire cluster since the mysql system tables are not replicated

I agree but is it really necessary ? In a cluster env, we use a dedicated account (`pfcluster`) to take dumps. This account is created by hand with correct rights.

julsemaan: @jrouzierinverse 

I think the user grants are replicated across a Galera cluster

But if they aren't, they I fully agree with you, they need to be adjusted on every server so its consistent.

jrouzierinverse: As long as that user is always created with the proper permissions you are right that step is not needed.
Just wrap those steps into a bash script then we are good.

jrouzierinverse: As long as that user is always created with the proper permissions.
Then you are right that step is not needed.
Just wrap those steps into a bash script then we are good.

nqb: I apply my one-liner on a host in a Galera Cluster: grants were replicated on all hosts.

jrouzierinverse: Then we are good

nqb: I will prepare a shell script and split instructions in two sections.

jrouzierinverse:
Rebase

nqb:
Done

nqb:
You can review !

PR #3682, opened 2018-10-16
fix/configurator-start-db

### LINKED ISSUE #3673: Configurator: Does not use the right command to start packetfence-mariadb in the configurator
screen shot 2018-10-11 at 4 10 39 pm

Apr 23 17:22:52 cluster1 httpd_admin: httpd.admin(3560) WARN: [mac:unknown] Problem trying to run command: LANG=C setsid sudo service mariadb start 2>&1 called from pfappserver::Model::Config::System::start_mysqld_service. Child exited with non-zero value 1 (pf::util::pf_run)

jrouzierinverse:
Description
===========
Configurator mysql starting

Impacts
=======
The admin Configurator

Issue
=====
Fix #3673

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------
* Start packetfence-mariadb instead of mysql when configuring PacketFence (#3673)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Config/System.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Config/System.pm
@@ -19,6 +19,8 @@ use pf::log;
 use pf::error qw(is_error is_success);
 use pf::util;
 
+our $DB_SERVICE_NAME = "packetfence-mariadb";
+
 extends 'Catalyst::Model';
 
 =head1 METHODS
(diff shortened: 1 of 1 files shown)


jrouzierinverse:
Looking into this further we already mask mariadb as service in packaging.
Maybe we should also mask firewalld on centos

extrafu:
@fdurand Looks good to you?

PR #3683, opened 2018-10-16
feature/rename-database-backup-and-maintenance

### LINKED ISSUE #3061: Rename database-backup-and-maintenance.sh to backup-and-maintenance.sh
Ideally we should rename that script, as it not only takes care of the database backup, but also the /usr/local/pf/ one for config files, etc. So the script name is a bit misleading.

jrouzierinverse:
Description
===========
Rename database-backup-and-maintenance.sh to backup-and-maintenance.sh

Issue
=====
Fixes #3061

Delete branch after merge
=========================
NO

NEWS file entries
=========================

Enhancements
------------
* Rename addons/database-backup-and-maintenance.sh to addons backup-and-maintenance.sh


nqb (changes requested):
I think we also need to change filename in UPGRADE.asciidoc.

jrouzierinverse:
Done

nqb:
We also need to upgrade this script: https://github.com/inverse-inc/packetfence-monitoring-scripts/blob/master/scripts/check-2388.sh#L8

I think a note in UPGRADE.asciidoc would be necessary for admins who use this script to backup PF from external tools.

jrouzierinverse:
We do not need to update that script since it is only applies for packetfence version 7.0 and 7.1

nqb:
Agree with you @jrouzierinverse

julsemaan:
I'm moving this to 8.3 since the merge window for features is long gone

fdurand:
Looks good to me.

PR #3695, opened 2018-10-21
fix/sms-pincode-reuse

### LINKED ISSUE #3436: Ran out of pin code
In the code even if the pin code has been used we never make it available again for another registration.
We need to find a way to "expire" the pin code.

jrouzierinverse:
Description
===========
Allow activation pin codes to be reused

Impacts
=======
SMS/Email activations.

Issue
=====
Fixes #3436

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------

* Allow activation pin codes to be reused (#3436)

### DIFF
--- a/lib/pf/activation.pm
+++ b/lib/pf/activation.pm
@@ -202,6 +202,31 @@ sub view_by_code {
     return ($item->to_hash);
 }
 
+=head2 is_code_in_use
+
+is_code_in_use
+
+=cut
+
+sub is_code_in_use {
+    my ($type, $code) = @_;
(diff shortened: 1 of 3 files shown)


extrafu:
@jrouzierinverse Please revisit this after the discussion we had. The SMS doesn't need to be unique per MAC address.

extrafu:
@jrouzierinverse Bump - what's the status on this?

jrouzierinverse:
Fixed

fdurand:
Just did some tests and i was able to reuse a pin code.

PR #3714, opened 2018-10-26
Reworked dhcp statistics

### LINKED ISSUE #3634: pfstats hammers pfdhcp and the API frontend with requests
I have a deployment where there are 549 routed networks.

The way the API call for leases statistics is designed, this results in 549 individual API calls every 24 seconds (https://github.com/inverse-inc/packetfence/blob/devel/lib/pfconfig/namespaces/config/Stats.pm#L51)

This results in high disk usage since httpd.api-frontend.access is ginormous and with high load on api-frontend, pfdhcp and rsyslog.

We should optimize this so that it performs a single API call to fetch all the leases information in one shot

fdurand:
# Description
Reduced the call to the api to have the dhcp statistics
Created 2 graphs with the usage on each networks and percent of usage
Created warning alarms that pop if the percent of used ip is greater than 80%, error if greater than 90%

# Impacts
pfstats and pfdhcp

# Issue
fixes #3634

# Delete branch after merge
YES

## Bug Fixes
* pfstats hammers pfdhcp and the API frontend with requests (#3634)

### DIFF
--- a/conf/monitoring/statsd.d/packetfence.conf.example
+++ b/conf/monitoring/statsd.d/packetfence.conf.example
@@ -83,7 +83,7 @@
     type = stacked
     dimension = pattern 'source.packetfence.redis.queue_stats_outstanding.*' '' last 1 1
 
-[dhcp.dhcp_free_leases]
+[dhcp.free_leases]
     name = dhcp.free_leases
     title = Dhcp free leases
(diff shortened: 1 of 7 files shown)


julsemaan:
will backport after releasing 8.2

PR #3716, opened 2018-10-26
Run some pfstats task only on the Master of the cluster or on the standalone

### LINKED ISSUE #3690: Some task from pfstat should only run on the server who have the vip
We have duplicate sql queries on te database coming from pfstats.
We will need to disable these queries from the servers who don't own the vip.

fdurand:
# Description
Get rid of sql queries made by pfstats on a cluster

# Impacts
Stats

# Issue
fixes #3690

# Delete branch after merge
YES

# NEWS file entries

## Enhancements
* pfstats optimization

## Bug Fixes
* Run some sql queries from pfstats from only one server (#3690)

### DIFF
--- a/conf/stats.conf.example
+++ b/conf/stats.conf.example
@@ -7,7 +7,7 @@ api_path=/api/v1/queues/stats
 api_compile=$.items[*].queue, $.items[*].stats.count
 interval=60s
 randomize=true
-management = false
+management=false
 
 [metric 'redis queue stats expired']
 type=api
@@ -18,7 +18,7 @@ api_path=/api/v1/queues/stats
 api_compile=$.items[*].stats.expired[*].name, $.items[*].stats.expired[*].count
 interval=60s
 randomize=true
-management = false
+management=false
 
 [metric 'redis queue stats outstanding']
 type=api
@@ -29,7 +29,7 @@ api_path=/api/v1/queues/stats
 api_compile=$.items[*].stats.outstanding[*].name, $.items[*].stats.outstanding[*].count
 interval=60s
 randomize=true
-management = false
+management=false
 
 [metric 'total number of registered devices per role']
 type=mysql_query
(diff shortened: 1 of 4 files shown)


go/stats/main.go
```diff
@@ -1764,3 +1781,38 @@ var Verb = TypeName{
 		},
 	},
 }
+
+// Detect the vip on management
+func detectVIP(management pfconfigdriver.ManagementNetwork) {
```
julsemaan, line 1786: This seems duplicated with the one in pfdhcp, could we centralize this in a package so we don't duplicate code

fdurand: It's not exactly the same that we have in pfdhcp.
Btw i can move this function in sharedutils.

conf/stats.conf.example
```diff
@@ -7,6 +7,7 @@ api_path=/api/v1/queues/stats
 api_compile=$.items[*].queue, $.items[*].stats.count
 interval=60s
 randomize=true
+management = false
```
satkunas: Use the same `key=value` spacing used elsewhere.

go/stats/main.go
```diff
@@ -1764,3 +1781,38 @@ var Verb = TypeName{
...
+	if keyConfCluster.Ip == "" {
+		VIP[management.Int] = true
+		return
+	}
+
+	eth, _ := net.InterfaceByName(management.Int)
+	adresses, _ := eth.Addrs()
+	var found bool
+	found = false
+	for _, adresse := range adresses {
```
satkunas, line 1801: This `for` loop appears to iterate over all `adresses` even after `found=true`. Is a `return` statement needed right after `found=true`?

Would it be faster to define `VIP[management.Int] = false` at the beginning of this function, and only set `VIP[management.Int] = true` and `return` immediately if `IP.Equal(VIPIp[management.Int])` ?

satkunas (changes requested):

fdurand:
BUMP

julsemaan:
@satkunas 

Can you check @fdurand's changes and remove your block if its all fine.

To me, we're all good to merge

satkunas:
@fdurand @julsemaan all good. Merge when ready.

PR #3718, opened 2018-10-29
fix/admin-styling

jrouzierinverse:
Description
===========
Fix styling of admin components 

Impacts
=======
Admin


cgx (approved):
👍

julsemaan:
Knowing we're pretty much done with our tests and that this is not a show stopper, should we aim to get it into 8.2 or leave it for 8.3.

IMO its not something that should absolutely get in

cgx:
I'll take care of testing.

PR #3722, opened 2018-10-29
fix/missing-saml-rework

### LINKED ISSUE #3720: Admin: Can't download SAML metadata in the admin
Seems the code to download the SP metadata from the admin to configure the SAML source is gone since the auth source rewrite

https://github.com/inverse-inc/packetfence/blob/maintenance/6.0/html/pfappserver/root/authentication/source/type/SAML.tt#L12
https://github.com/inverse-inc/packetfence/blob/maintenance/6.0/html/pfappserver/lib/pfappserver/PacketFence/Controller/Config/Authentication/SAML.pm#L33

jrouzierinverse:
Description
===========
Can't download SAML metadata in the admin

Impacts
=======
Admin/SAML

Issue
=====
Fixes #3720

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------

* Can't download SAML metadata in the admin (#3720)

### DIFF
--- a/lib/pf/Authentication/Source/SAMLSource.pm
+++ b/lib/pf/Authentication/Source/SAMLSource.pm
@@ -13,6 +13,8 @@ Model for a SAML source
 use pf::Authentication::constants;
 use pf::constants::authentication::messages;
 use pf::constants;
+use pf::constants::saml;
+use pf::config;
 use Template::AutoFilter;
 use File::Slurp qw(read_file write_file);
 use File::Temp qw(tempfile);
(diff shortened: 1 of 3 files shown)


julsemaan:
Please apply the proper tags and milestones so this doesn't get missed in the release

lib/pf/Authentication/Source/SAMLSource.pm
```diff
@@ -13,6 +13,8 @@ Model for a SAML source
 use pf::Authentication::constants;
 use pf::constants::authentication::messages;
 use pf::constants;
+use pf::constants::saml;
```
julsemaan, line 16: Are these 2 lines really necessary?

jrouzierinverse: I added those to stop the warnings on a compile test.
But as long as we guarantee that the libraries pf::constants::saml and pf::config are loaded before we use those constants and the config then no we do not need those two lines.

julsemaan: Lets leave them in then

extrafu:
Do we now support SMAL metadata? @lzammit says it's the future.

PR #3727, opened 2018-10-30
adjustments to pfdhcp for multi-zone cluster

### LINKED ISSUE #3726: Cluster zones: pfdhcp runs once in each zone
Since the logic of pfdhcp is to check if it has the VIP to start running and offering DHCP, then in the case of 2 zones, there are 2 pfdhcp that are online.

Although it initially seemed to not be causing any problem, when performing extensive tests, it seems the clients are losing their IP address from time to time, perhaps due to the fact there are concurrent ACK made by both servers

julsemaan:
# Description
Adjust pfdhcp so that it supports working with a peer in multi-zone cluster

# Impacts
pfdhcp, specifically the fact we won't be sending a NAK to a request for which we don't have a lease. Instead no reply at all will be made. In the case of a device attempting to do a request for a lease from another network, the lack of reply should start a discovery process

# Issue
fixes #3726 

# Delete branch after merge
YES

### DIFF
--- a/go/dhcp/main.go
+++ b/go/dhcp/main.go
@@ -292,7 +292,7 @@ func (h *Interface) ServeDHCP(ctx context.Context, p dhcp.Packet, msgType dhcp.M
 			GlobalTransactionLock.Unlock()
 			return answer
 		} else {
-			GlobalTransactionCache.Set(cacheKey, 1, time.Duration(5)*time.Second)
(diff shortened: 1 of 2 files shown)


julsemaan:
It seems I'm still having issues with Ubuntu during renewal but I think this is unrelated to cluster zones

PR #3782, opened 2018-11-13
fix/3771

### LINKED ISSUE #3771: configreload race condition
If you perform a configreload at the same time on more than one server, there is a race condition for the process of refreshing the radius_nas table.

The process inserts new clients with the starting timestamp as an identifier of the new batch (both servers have different timestamp for the identifier)

Assuming server 1 starts at 13:00:00 and server 2 starts at 13:00:02 and the process takes 5 seconds.

Server 1 will insert its clients normally, then 2 seconds later server 2 starts inserting clients.
Server 1 finishes 2 seconds before server 2 and deletes all the clients server 2 has inserted between 13:00:02 and 13:00:05
Then server 2 finishes the process and deletes all the clients server 1 inserted
We then end up with only the clients server 2 inserted between 13:00:05 and 13:00:07

jrouzierinverse:
Description
===========
Warn if radius_table is found in an inconsistent state

Issue
=====
Fixes #3771

Delete branch after merge
=========================
NO

NEWS file entries
=================

Bug Fixes
------------
* Warn if radius_table is found in an inconsistent state (#3771)


lib/pf/freeradius.pm
```diff
@@ -155,9 +154,57 @@ sub freeradius_populate_nas_config {
...
     }
     _delete_expired($timestamp);
+    validate_radius_nas_table($timestamp);
+}
+
+our $validate_radius_nas_table_sql = <<SQL;
+    SELECT
+        (SELECT COUNT(*) FROM radius_nas) = (SELECT COUNT(*) FROM radius_nas WHERE config_timestamp = ?) AS config_valid,
+        (SELECT COUNT(*) FROM radius_nas WHERE config_timestamp > ?) as count_greater_than,
+        (SELECT COUNT(*) FROM radius_nas WHERE config_timestamp < ?) as count_less_than;
```
satkunas: `SELECT COUNT(1) ...`

It's the same result as `COUNT(*)`, but faster.

lib/pf/freeradius.pm
```diff
@@ -155,9 +154,57 @@ sub freeradius_populate_nas_config {
         _insert_nas_bulk( @rows );
     }
     _delete_expired($timestamp);
+    validate_radius_nas_table($timestamp);
+}
+
+our $validate_radius_nas_table_sql = <<SQL;
+    SELECT
+        (SELECT COUNT(1) FROM radius_nas) = (SELECT COUNT(1) FROM radius_nas WHERE config_timestamp = ?) AS config_valid,
+        (SELECT COUNT(1) FROM radius_nas WHERE config_timestamp > ?) as count_greater_than,
```
julsemaan: This counter and the one below don't seem to be used. Any reason for keeping them?

PR #3786, opened 2018-11-14
add file to track fingerbank collector env variables

julsemaan:
# Description
Added ability to control the Fingerbank collector environment variables through the PacketFence config

# Impacts
Fingerbank collector

# NEW Package(s) required
fingerbank-collector-1.1.1

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Added ability to control the Fingerbank collector environment variables through the PacketFence config

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -1068,6 +1068,7 @@ fi
                         /usr/local/pf/conf/haproxy-db.conf.example
 %config(noreplace)      /usr/local/pf/conf/haproxy-portal.conf
                         /usr/local/pf/conf/haproxy-portal.conf.example
+%config                 /usr/local/pf/conf/fingerbank-collector.env.defaults
 %dir                    /usr/local/pf/conf/httpd.conf.d
 %config                 /usr/local/pf/conf/httpd.conf.d/captive-portal-common.tt
                         /usr/local/pf/conf/httpd.conf.d/captive-portal-common.tt.example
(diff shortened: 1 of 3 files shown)


julsemaan:
@jrouzierinverse 

Please check this and merge if its fine

Its pretty simple and we can get the collector 1.1.1 in after this gets merged.

If 1.1.1 is not in, it doesn't break either, it will just be an empty non-used file

jrouzierinverse (changes requested):
I found the following three issues
* Make conf/fingerbank-collector.env.defaults an example file
* Provide commented out examples of the allowed variables.
* Update packaging to mark conf/fingerbank-collector.env.defaults as a config file

julsemaan:
There is no point into making this an example file, its made only to push variables for the collector through the maintenance by us, not for usage by users

jrouzierinverse:
Ok no problem just update the packaging so it will not fail.

jrouzierinverse:
Would there ever be a scenario where a user could update the env file.
Then I would suggest the following pattern.

conf/fingerbank-collector.env.example 
conf/fingerbank-collector.env.defaults would have the following content.
```bash
PF_DIR=/usr/local/pf/
ENV="$PF_DIR/conf/fingerbank-collector.env"
if [ -e  "$ENV"  ];then
 . $ENV
fi
```

This pattern allows us to have custom variables for the collector while still allowing us to push variables for collector through the maintenance.

julsemaan:
There isn't really such scenario, if the user should configure stuff in the collector, then this should go through the collector section of fingerbank.conf which already sets up environment variables.

This is really just for maintenance purposes

jrouzierinverse:
Ok then the only thing to do is to update the packaging part then we are good.

julsemaan:
packaging fixed, ready for another review

PR #3255, opened 2018-05-25
Fix/pfconfig lock timeout

julsemaan:
# Description
Reworks the pfconfigpool to use the timedlock library that was made for the FB collector
Should be more resilient to failures and introduces read timeouts as well

# Impacts
All of pfconfig pool usage in Golang

# Issue
fixes #3148 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Reworked the pfconfig Golang driver to be more resilient to lock failures (#3148)

PR #3795, opened 2018-11-16
Avoid duplicates ip address sent by the dhcp

fdurand:
# Description
Adjust the code to avoid duplicate ip addresses given by the dhcp server

# Impacts
DHCP server

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Issue with the dhcp server that give sometimes a duplicate ip addresse

### DIFF
--- a/go/dhcp/main.go
+++ b/go/dhcp/main.go
@@ -5,7 +5,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"strings"
-	"sync"
 
 	"context"
 	_ "expvar"
@@ -15,14 +14,17 @@ import (
 	"time"
 
 	"github.com/coreos/go-systemd/daemon"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/fdurand/arp"
 	cache "github.com/fdurand/go-cache"
+	"github.com/go-errors/errors"
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/goji/httpauth"
 	"github.com/gorilla/mux"
 	"github.com/inverse-inc/packetfence/go/log"
 	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
 	"github.com/inverse-inc/packetfence/go/sharedutils"
+	"github.com/inverse-inc/packetfence/go/timedlock"
 	dhcp "github.com/krolaw/dhcp4"
 )
 
@@ -34,7 +36,7 @@ var GlobalIpCache *cache.Cache
 var GlobalMacCache *cache.Cache
 
 var GlobalTransactionCache *cache.Cache
-var GlobalTransactionLock *sync.Mutex
+var GlobalTransactionLock *timedlock.RWLock
 
 var RequestGlobalTransactionCache *cache.Cache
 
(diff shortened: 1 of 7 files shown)


go/dhcp/main.go
```diff
@@ -16,14 +15,17 @@ import (
 	"time"
 
 	"github.com/coreos/go-systemd/daemon"
+	"github.com/davecgh/go-spew/spew"
```
satkunas, line 17: Remove debugging

go/dhcp/main.go
```diff
@@ -608,3 +673,11 @@ func (h *Interface) ServeDHCP(ctx context.Context, p dhcp.Packet, msgType dhcp.M
 	return answer
 
 }
+
+func recoverName(options dhcp.Options) {
+	if r := recover(); r != nil {
+		fmt.Println("recovered from ", r)
+		fmt.Println(errors.Wrap(r, 2).ErrorStack())
+		spew.Dump(options)
```
satkunas, line 690: Is debugging needed here?

go/dhcp/pool/pool_test.go
```diff
@@ -1,62 +1,61 @@
...
 
 	// Try to reserve an IP outside the capacity
-	err = dp.ReserveIPIndex(cap)
+	err, _ = dp.ReserveIPIndex(cap, mac)
 
 	if err == nil {
 		t.Error("Didn't get an error when trying to reserve an IP outside the capacity")
 	}
 }
-func TestIsFreeIPAtIndex(t *testing.T) {
```
julsemaan, line 37: Why deleting the unit test ?

go/timedlock/rw_lock.go
```diff
@@ -0,0 +1,192 @@
+package timedlock
```
julsemaan: Lets see if we can get #3255 merged with this change and have you rebase this on devel instead

go/dhcp/main.go
fdurand: It's a way to see which dhcp packet cause the crash.

go/dhcp/pool/pool_test.go
fdurand: I have removed the function IsFreeIPAtIndex

julsemaan: It sounds like a pretty legit reason

fdurand:
Rebased since  #3255 has been merged

go/dhcp/main.go
```diff
@@ -154,9 +156,11 @@ func main() {
 	http.Handle("/", httpauth.SimpleBasicAuth(webservices.User, webservices.Pass)(router))
 
 	srv := &http.Server{
-		Addr:        "127.0.0.1:22222",
-		IdleTimeout: 5 * time.Second,
-		Handler:     router,
+		Addr:         "127.0.0.1:22222",
+		IdleTimeout:  5 * time.Second,
+		ReadTimeout:  5 * time.Second,
```
julsemaan: Do the new read and write timeout really belong in this PR ?

go/dhcp/pool/pool.go
```diff
@@ -6,6 +6,9 @@ import (
 	"sync"
 )
 
+const FreeMac = "00:00:00:00:00:00"
```
julsemaan, line 9: Create a package for constants and centralize them there so they aren't duplicated in each package

PR #3796, opened 2018-11-16
Reduce cache duration to 60s, cache mac + qname, return 127.0.0.1 for reg deviced in REJECT role

### LINKED ISSUE #3785: dns filters: mac filtered lookups are cached
When using filter = mac lookups are cached which means nodes not triggering the filter will still get the cached results

fdurand:
# Description
DNS Filter cache adjustement
Return 127.0.0.1 for devices with REJECT role (dns enforcement , Inline)

# Impacts
DNS Flow

# Issue
fixes #3785

# Delete branch after merge
YES

# NEWS file entries

## Bug Fixes
* Fixes mac filtered lookups are cached (#3785)

### DIFF
--- a/go/coredns/plugin/pfdns/pfdns.go
+++ b/go/coredns/plugin/pfdns/pfdns.go
@@ -247,7 +247,8 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 					return 0, nil
 				}
 			}
-			answer, found := pf.DNSFilter.Get(mac + category + strconv.FormatBool(violation) + state.QName())
+			cacheKey := pf.MakeKeyCache(mac, category, violation, state.QName())
+			answer, found := pf.DNSFilter.Get(cacheKey)
 			if found && answer != "null" {
 				log.LoggerWContext(ctx).Debug("Get answer from the cache for " + state.QName())
 				rr, _ = dns.NewRR(answer.(string))
@@ -259,7 +260,7 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 					"mac":      mac,
 				})
 				if err != nil {
-					pf.DNSFilter.Set(mac+category+strconv.FormatBool(violation)+state.QName(), "null", cache.DefaultExpiration)
+					pf.DNSFilter.Set(cacheKey, "null", cache.DefaultExpiration)
 					break
 				}
 				var answer string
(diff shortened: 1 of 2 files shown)


go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -219,17 +221,26 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
 				if err != nil {
 					log.LoggerWContext(ctx).Error(fmt.Sprintf("error getting node status %s %s\n", mac, err))
 				}
 				// Defer to the proxy middleware if the device is registered
-				if status == "reg" && !violation {
+				if status == "reg" && !violation && category != "REJECT" {
 					log.LoggerWContext(ctx).Debug(srcIP + " : " + mac + " serve dns " + state.QName())
 					return pf.Next.ServeDNS(ctx, w, r)
 				}
+				if status == "reg" && category == "REJECT" {
```
julsemaan: else if ?

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -219,17 +221,26 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
...
+					rr, _ = dns.NewRR("30 IN A 127.0.0.1")
+					a.Answer = []dns.RR{rr}
+					log.LoggerWContext(ctx).Debug("REJECT " + mac + " IP " + srcIP + " Query " + state.QName())
+					state.SizeAndDo(a)
+					w.WriteMsg(a)
+					return 0, nil
+				}
 			}
-			answer, found := pf.DNSFilter.Get(state.QName())
+			answer, found := pf.DNSFilter.Get(mac + state.QName())
```
julsemaan: I'd make the cache key, the mac, category, violation (as string) and the state.QName() and then the cache can be set to a high expiration since the cache key will accurately reflect the state of the endpoint.

Otherwise right now, if a MAC goes from rejected to allowed, its entries are cached for a minute while with the new cache key it misses the cache and returns the appropriate record

Also given you create the cache key in multiple places, I'd create a function to create the cache key from the parameters above.

go/coredns/plugin/pfdns/setup.go
```diff
@@ -73,7 +73,7 @@ func setuppfdns(c *caddy.Controller) error {
 	}
 
 	// Initialize dns filter cache
-	pf.DNSFilter = cache.New(300*time.Second, 10*time.Second)
+	pf.DNSFilter = cache.New(60*time.Second, 10*time.Second)
```
julsemaan: If we pick the strategy above, I think we can make this higher

fdurand:
BUMP

fdurand:
BUMP

go/coredns/plugin/pfdns/pfdns.go
```diff
@@ -263,7 +270,7 @@ func (pf *pfdns) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
 					}
 				}
 				log.LoggerWContext(ctx).Debug("Get answer from pffilter for " + state.QName())
-				pf.DNSFilter.Set(mac+state.QName(), answer, cache.DefaultExpiration)
+				pf.DNSFilter.Set(mac+category+strconv.FormatBool(violation)+state.QName(), answer, cache.DefaultExpiration)
```
julsemaan: Just need a helper function to create the cache key so that if we ever change the creation, we don't need to remember all the places to change

fdurand:
Done

PR #3802, opened 2018-11-20
Feature/no empty fields portal

julsemaan:
# Description
Do not prompt user to press continue after AUP when there are no fields to prompt (like null auth without email)

# Impacts
Portal authentication

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Do not prompt user to press continue after AUP when there are no fields to prompt

### DIFF
--- a/html/captive-portal/templates/signin.html
+++ b/html/captive-portal/templates/signin.html
@@ -10,7 +10,7 @@
     </div>
 
       [% IF aup_only %]
-      <script src="/common/aup-autosubmit.js">
+      <script src="/common/aup-autosubmit.js"></script>
       [% END %]
 
     [% END %]
--- a/html/captive-portal/templates/signin.html
+++ b/html/captive-portal/templates/signin.html
@@ -8,6 +8,11 @@
     <div class="c-card u-padding u-padding@tablet u-padding-top-none">
         [% form.get_field('aup').render | none %]
     </div>
+
+      [% IF aup_only %]
+      <script src="/common/aup-autosubmit.js">
+      [% END %]
+
     [% END %]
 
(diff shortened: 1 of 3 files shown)


html/captive-portal/templates/signin.html
```diff
@@ -8,6 +8,11 @@
     <div class="c-card u-padding u-padding@tablet u-padding-top-none">
         [% form.get_field('aup').render | none %]
     </div>
+
+      [% IF aup_only %]
+      <script src="/common/aup-autosubmit.js">
```
cgx: `<script>` is not an empty tag and should be closed with an end tag.

julsemaan: As they say in France: `fixaide`

html/common/aup-autosubmit.js
```diff
@@ -0,0 +1,6 @@
+$('label[for="aup"]').closest('div').click(function(e) {
+  e.preventDefault();
+  $('#aup').attr('checked', 'checked');
+  $('#button').click();
```
cgx, line 4: Instead of relying on the submit button, I would submit the form:

```javascript
$(this).closest('form').submit();
```

julsemaan: I had exactly that before but it didn't work when I tested it with a customer in his browser although it worked in mine.

This though has worked consistently so that's why I put it there although I agree it makes more sense to use the line you suggested

cgx: I don't like the fact that the Continue/submit button is disabled. I would expect a problem when programatically clicking this button more than submitting the form. If you feel confortable with your code, we can keep it like this.

julsemaan: I'd feel more comfortable to leave it as-is since the cleaner way has caused me trouble in the deployment where I tested this PR

PR #3808, opened 2018-11-22
Fix minor typos for Layer 3 clustering

nqb:
# Description
Minor typos for Layer 3 clustering

# Impacts
Clustering guide

# Delete branch after merge
YES

### DIFF
--- a/docs/PacketFence_Clustering_Guide.asciidoc
+++ b/docs/PacketFence_Clustering_Guide.asciidoc
@@ -734,7 +734,7 @@ ip=192.168.2.12
 
 Notes on the configuration:
 
- * The hostnames (pf1-dc1.example.com, pf2-dc1.example.com, etc) are not directly related to the cluster logic and the servers can have any hostname without impacting the cluster behavior. The assignment of a server to a cluster zone is made by the first part of the section name (ex: "DC1 pf.example.com" assigns server "pf.example.com" to cluster zone "DC1")
+ * The hostnames (pf1-dc1.example.com, pf2-dc1.example.com, etc) are not directly related to the cluster logic and the servers can have any hostname without impacting the cluster behavior. The assignment of a server to a cluster zone is made by the first part of the section name (ex: "DC1 pf-dc1.example.com" assigns server "pf-dc1.example.com" to cluster zone "DC1")
  * Each cluster zone needs to have its own "CLUSTER" definition that declares the virtual IPs to use for this cluster zone. This also declares the management IP on which the zone should be joined.
  * Given your zones aren't in the same layer 2 network, you cannot use the same virtual IP between your zones.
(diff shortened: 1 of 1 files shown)


docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -734,7 +731,7 @@ ip=192.168.2.12
 
 Notes on the configuration:
 
- * The hostnames (pf1-dc1.example.com, pf2-dc1.example.com, etc) are not directly related to the cluster logic and the servers can have any hostname without impacting the cluster behavior. The assignment of a server to a cluster zone is made by the first part of the section name (ex: "DC1 pf.example.com" assigns server "pf.example.com" to cluster zone "DC1")
+ * The hostnames (pf1-dc1.example.com, pf2-dc1.example.com, etc) are not directly related to the cluster logic and the servers can have any hostname without impacting the cluster behavior. The assignment of a server to a cluster zone is made by the first part of the section name (ex: "DC1 pf-dc1.example.com" assigns server "pf-dc1.example.com" to cluster zone "DC1")
```
julsemaan: I kept the DC name out of the hostname intentionally to show that it doesn't have anything to do with the cluster zone assignation

docs/PacketFence_Clustering_Guide.asciidoc
```diff
@@ -847,7 +847,7 @@ ip=192.168.22.12
...
```
julsemaan: Same as my comment above

nqb: I understand.

docs/PacketFence_Clustering_Guide.asciidoc
nqb: I understand.

nqb:
@julsemaan,

It could be nice to have list of network protocol(s) to open between DC1 servers and DC2 servers in the two configurations mentioned in the guide.

julsemaan:
@nqb 

Documenting the list of protocols in the guide directly might not be the best since new ports can be added/removed with new services.

Instead, we should aim to provide commands that allow the user to find the list of ports that we already maintain in iptables. Then its a case by case for what ports are needed for the specific environment. If they only want the replication traffic to be synced across DCs or if they have registration VLANs, etc.

The list of ports can be found using the following command:
```
iptables -S | grep ACCEPT | egrep -o '\-\-dport [0-9]+' | sort | uniq
```

But then as far as the specifics of what need to be opened is a case by case for each cluster.

Lets keep in mind that this technology is to be used (at least for now) by people with very strong linux and network skills that have a deep understanding of how a NAC solution works (thus the warnings in the docs) and the different pieces that are tied together or at least that will be able to figure out how the different pieces are connected together via advanced troubleshooting. Its currently not meant for that regular sysadmin that wants to toy around with a NAC.

PR #3814, opened 2018-11-27
Fix/openvas integration

julsemaan:
# Missing
* Precise requirements on what components are required on the OpenVAS server from @nqb 
* Necessary Debian dependencies (aren't in the official repos)

# Description
Fix the OpenVAS integration and update the documentation

# Impacts
OpenVAS scanning

# NEW Package(s) required
openvas-cli
openvas-libraries

# Delete branch after merge
(REQUIRED)
YES | NO

# Checklist
- [yes] Document the feature
- [n/a] Add unit tests
- [yes] Add acceptance tests (TestLink)

# NEWS file entries
## Bug Fixes
* Fixed the OpenVAS integration to work with OpenVAS Manager 7.0 (OpenVAS 9)


nqb:
I updated the NEWS section on this PR. We don't use OpenVAS 7 but OpenVAS 9 wi th OpenVAS Manager 7.

nqb:
Regarding usage of `omp` on Debian Jessie, I think we should point users to https://github.com/greenbone/gvm-tools in Installation Guide because `openvas-cli` packages are only available on Debian Stretch.

I'm able to install `gvm-tools` on Debian Jessie wit `pip`. However, `gvm-tools` needs at least Python 3.5, Jessie provides 3.4.2. But it works:
```console
vagrant@debian-8:~$ /home/vagrant/.local/bin/gvm-cli tls --hostname openvas --port 9390 --xml "<get_version/>"
<get_version_response status="200" status_text="OK"><version>7.0</version></get_version_response>
```

addons/packages/packetfence.spec
```diff
@@ -302,6 +302,9 @@ Requires: perl(DateTime::TimeZone)
 Requires: samba-winbind-clients, samba-winbind
 Requires: libdrm >= 2.4.74
 Requires: netdata, fping, MySQL-python
+#OpenVAS
+Requires: openvas-cli
+Requires: openvas-libraries
```
nqb, line 308: Is it really necessary to specify these softwares as `Requires` ? Only OpenVAS users will need it.

nqb: We could just add documentation.

nqb:
BUMP

extrafu:
@nqb Let's follow-up on this tomorrow during our R&D meeting.

extrafu:
@nqb Mention in the doc that Debian 8 is not supported for this feature but will be in v9.

addons/packages/packetfence.spec
nqb: @julsemaan: what's your opinion on this ? On top of that, I noticed that this two packages come from EPEL repo.

extrafu:
@nqb If all is fine, merge and update the NEWS file.

addons/packages/packetfence.spec
julsemaan: Its not in our convention to not install the necessary packages for making PF work.

Like we install WMI clients, samba clients, and countless other dependencies so things work as out of the box as possible

julsemaan:
We good to merge @nqb ?

addons/packages/packetfence.spec
nqb: @julsemaan, I know but packages are available in official repos not in EPEL.

julsemaan: samba and krb5-workstation are too and we're not asking people to install them manually so that the AD integration works... Same goes for all the other integrations we do have. Whether its in EPEL, base or zammrepo doesn't make a difference.

PR #3827, opened 2018-12-03
Change defaults conf files: portal_modules, profiles and switches

### LINKED ISSUE #3702: default connection profile: locale=en_US
I think we should have: `locale=` in `profiles.conf.defaults` to support all languages by default.

Currently we have `locale=en_US` which force english language for all web browsers.

nqb:
# Description
Change defaults in switches conf files to simplify tests.

# Impacts
- Network switches

# Delete branch after merge
YES

# Checklist
[n/a]

# NEWS file entries
## Enhancements
* Simplify tests with default switch

### DIFF
--- a/conf/switches.conf.example
+++ b/conf/switches.conf.example
@@ -8,7 +8,7 @@
 
 [192.168.0.1]
 description=Test Switch
-type = Cisco::Catalyst_2900XL
+type = Cisco::Catalyst_2960
 mode = production
 uplink = 23,24
 VoIPLLDPDetect=N


conf/portal_modules.conf.defaults
```diff
@@ -16,6 +16,7 @@ modules=default_login_policy,default_guest_policy,default_oauth_policy,default_b
 [default_login_policy]
 description=Username/password login
 type=Authentication::Login
+source_id=local
```
julsemaan: This makes it so that whatever "login type" sources (AD, LDAP, etc) you define on your connection profile ignored.

julsemaan (changes requested):

conf/portal_modules.conf.defaults
nqb: Tell me if I'm wrong but when you use a connection profile with a root portal module that use the default login policy, sources defined on your connection profiles are not used.

I make this change because:
* all others authentication portal module has pre-defined sources
* it was unclear in the doc that you need to add a source on this specific portal module

julsemaan: The sources on the connection profile are used if not sources are put in source_id

The Authentication::Login portal module encapsulates all the username/password sources, so that means you don't need to define all the source types or source IDs you want this to apply to unless you want to narrow it down to a specific source

nqb: @julsemaan,

> The sources on the connection profile are used if not sources are put in source_id

Ok, I understand why I get some strange behaviors. We need to enhance documentation on that topic.

nqb: > The sources on the connection profile are used if not sources are put in source_id

screenshot from 2018-12-07 13-53-29

There is no help text that mentioned that the source on the connection profiles will be used.

extrafu:
Bump - looks more work needs to be done by @nqb on this

nqb:
@julsemaan, 

I revert my change for `portal_modules.conf`.

I see this help message in PF sources:
```
             help => 'The sources to use in the module. If no sources are specified, all the sources on the Connection Profile will be used' },
```
here : https://github.com/inverse-inc/packetfence/blob/devel/html/pfappserver/lib/pfappserver/Base/Form/Role/WithSource.pm#L79

Is this what should be displayed ?

julsemaan:
@nqb 

Yes that is what should be displayed

cgx:
@nqb Make sure to squash your commits prior to merging.

nqb:
As said with @julsemaan, I will revert change for profiles.conf.defaults

For the help on source in portal modules, I will check in the new GUI.

nqb:
@cgx, commit squashed.

Useless changes have been reverted, ready to merge.

extrafu:
@jrouzierinverse If all is fine with this one-liner, let's merge it once for all.

PR #3848, opened 2018-12-14
iptables custom rules

becchett:
Iptables customization from web interface.

Dear all, I'd like to submit to Inverse developers this case because I think
that iptables settings is a great item for most of us.  In the networks of 
large companies there is always the need to make customizations to 
manage various types of Inernet and intranet traffic. So it would be very useful 
to have a section on iptables in the dashbord. 
PF after reading the rules entered by the user do some formal checks 
and then compile the file conf / iptables.conf.

# Delete branch after merge
NO

# Checklist
- [yes ] Document the feature
- [ yes] Add unit tests
- [ yes] Add acceptance tests (TestLink)

Thanks a lot !
Best Regards
Enrico


julsemaan:
This PR is attempting to merge maintenance/8.2 into devel which makes no sense at all.

Closing it

PR #3857, opened 2018-12-17
Make pf-maint.pl proxy ready

### LINKED ISSUE #3425: pf-maint.pl doesn't take *_proxy environment variables into account
As title says.

Steps to reproduce:
```console
# export http_proxy=http://1.1.1.1:3142
# export https_proxy=http://1.1.1.1:3142

# with use Data::Dumper; print Dumper($ua); at line 217
# ./pf-maint.pl 
[..]
'no_proxy' => [],
[..]
'proxy' => {
             'http' => 'http://1.1.1.1:3142',
             'https' => 'http://1.1.1.1:3142'
},
[..]

# yum info perl-libwww-perl.noarch 
[..]
Version     : 6.07
Release     : 4.2
[..]
From repo   : packetfence

```
I didn't see any traffic going to 1.1.1.1 on port 3142.

https://metacpan.org/pod/release/MSCHILLI/libwww-perl-6.07/lib/LWP/UserAgent.pm#Proxy-attributes

nqb:
# Description
`pf-maint.pl` can now retrieve proxies URL from environment variables on CentOS 7

# Impacts
`pf-maint.pl` script

# Issue
fixes #3425 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* `pf-maint.pl` doesn't take *_proxy environment variables into account (#3425)

### DIFF
--- a/addons/pf-maint.pl
+++ b/addons/pf-maint.pl
@@ -40,6 +40,8 @@ use Pod::Usage;
 use IO::Handle;
 use Term::ReadKey;
 use IO::Interactive qw(is_interactive);
+# use CONNECT method to deal with proxy, see #3425
+$ENV{PERL_NET_HTTPS_SSL_SOCKET_CLASS} = "Net::SSL";
 our $GITHUB_USER = 'inverse-inc';
 our $GITHUB_REPO = 'packetfence';
 our $PF_DIR      = $ENV{PF_DIR} || '/usr/local/pf';


extrafu:
@fdurand Please review, looks obvious to me and safe to merge.

addons/pf-maint.pl
```diff
@@ -40,6 +40,8 @@ =head1 SYNOPSIS
 use IO::Handle;
 use Term::ReadKey;
 use IO::Interactive qw(is_interactive);
+# use CONNECT method to deal with proxy, see #3425
+$ENV{PERL_NET_HTTPS_SSL_SOCKET_CLASS} = "Net::SSL";
```
fdurand, line 44: Make sure perl-Crypt-SSLeay is a dependency of PacketFence.

fdurand (changes requested):

addons/pf-maint.pl
nqb: On a CentOS 7.6 with PF installed:
```
# yum list installed|grep perl|grep -i ssl
[..]
perl-Crypt-SSLeay.x86_64                   0.64-5.el7               @base       
perl-IO-Socket-SSL.noarch                  2.049-1.1                @packetfence
[..]
perl-Net-SSLeay.x86_64                     1.55-6.el7               @base 
```

nqb: My code is wrong, with only `$ENV{PERL_NET_HTTPS_SSL_SOCKET_CLASS} = "Net::SSL";`

We need to add: `$ENV{PERL_LWP_SSL_VERIFY_HOSTNAME} = 0;` 

which is not a good idea for me, see [verify_hostname option](https://metacpan.org/pod/LWP::UserAgent#verify_hostname-=%3E-$bool) for details.

My two proposals:
- add documentation to explain how to use a proxy on CentOS 7 (with the `$ENV` variables)
- package `perl-libwww-perl` in PF repo for CentOS 7 at version 6.08 (same version as Debian)

fdurand: If the certificate match the fqdn then you don't need to set $ENV{PERL_LWP_SSL_VERIFY_HOSTNAME} = 0;
For the perl-Crypt-SSLeay you need to add it in the spec file and in the debian control file.

julsemaan:
Also, I couldn't get it to use the proxy until I downgraded perl-libwww-perl-6.05-2.el7, so maybe we need to adjust the packaging for that

addons/pf-maint.pl
nqb: > If the certificate match the fqdn then you don't need to set $ENV{PERL_LWP_SSL_VERIFY_HOSTNAME} = 0;

Didn't work with api.github.com, maybe due to wildcard.

julsemaan:
Lets move this to the next milestone since it still has pending items

extrafu:
I think we need to test that more with some customers before merging to make sure we cover all cases.

nqb:
Fixed by #5468

PR #3867, opened 2018-12-21
optimize voip detection for non-voip devices

### LINKED ISSUE #3486: Optimize the VoIP detection using SNMP
Currently, if a device isn't considered a phone, we will poll the switch using SNMP everytime that device connects until its detected as a phone (which will never happen for workstations).

This makes us extremely dependent to SNMP at all times.

An idea me and @fdurand have brainstormed is that for every successful lookup in the CDP/LLDP tables, we cache the result (success or failure) for 24 hours (or more) so that we do not constantly poll the switch for devices we haven't detected as a phone. The cache key would be "$switch_ip-$ifIndex-$mac" to take into account movement to another switch or switchport where CDP would be enabled (but not in the first port).

julsemaan:
# Description
optimize voip detection for non-voip devices

# Impacts
VoIP detection

# Issue
fixes #3486 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests

# NEWS file entries
## Enhancements
* Added caching of SNMP VoIP detection (#3486)

### DIFF
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -1589,9 +1589,11 @@ sub isPhoneAtIfIndex {
     }
 
     if (defined($ifIndex)) {
-        $logger->debug("determining if $mac is VoIP phone through discovery protocols");
-        my @phones = $self->getPhonesDPAtIfIndex($ifIndex);
-        return ( grep( { lc($_) eq lc($mac) } @phones ) != 0 );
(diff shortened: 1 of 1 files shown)


julsemaan:
This won't get much effect until this is fixed: #3868

Or if I set the expiration manually in the compute call

julsemaan:
@fdurand 

Could we get this merged in the codebase if its all good, and I'll fix #3868 in devel with the new timings

Issue #3612, opened 2018-09-12
[security] passwords : add more complexity


nqb:
Current generated passwords by PF are lowercase and 12 characters long. I think they could simply break by brute-force attacks. 

We need to add more complexity for passwords:
- upper
- digit
- special chars

Some help on the subject: https://ltb-project.org/documentation/self-service-password/1.3/config_ppolicy

julsemaan:
agree with upper and digits, we should be careful with special characters since some less techie users (like guests) may not be able to get it right, so we should keep it simple yet secure.

nqb:
Agree, the best will be to have a configurable password policy.

nqb:
First thing to have: mininum and maximum size of passwords generated by PF. Currently, they are hardcoded in `lib/pf/password.pm`

julsemaan:
@nqb 

Just as an FYI, the minimum and maximum in password.pm aren't enforced, they are defaults if no password length is defined

PR #3888, opened 2018-12-28
fix set min password length to 8

### LINKED ISSUE #3887: webadmin: default password length is 4 for created users
If you create user(s) without password, size of password generated is 4.

Related to #3879 and ec40e22

Having a password of 4 characters is not acceptable, need to be at least 8 (like before).

nqb:
# Description
Set minimum password length to 8

# Impacts
Password generation

# Issue
fixes #3887 

# Delete branch after merge
YES

### DIFF
--- a/lib/pf/password.pm
+++ b/lib/pf/password.pm
@@ -144,10 +144,9 @@ sub _generate_password {
     my $min = 8;
     my $max = 12;
 
-    my $absolute_min = 4;
-    if($size < $absolute_min) {
-        get_logger->warn("Password length less than $absolute_min, making the password $absolute_min long.");
-        $size = $absolute_min;
+    if($size < $min) {
+        get_logger->warn("Password length less than $min, making the password $min long.");
+        $size = $min;
     }
 
     $min = $max = $size if (defined $size);


julsemaan:
I disagree with this, some environments will want to have very simple passwords for guests

Its not our job to judge them, so 4 seems like a valid bottom limit

nqb:
> I disagree with this, some environments will want to have very simple passwords for guests

I disagree too: some would not.
Before you change, password was 8 characters long. I just revert to original value. 

As I said on #3612, this setting need to be configurable when you create accounts on webadmin not only on sources.

> so 4 seems like a valid bottom limit

I disagree. 8 is already weak for me.

julsemaan:
I don't disagree in making the default value 8 long or even larger, this is totally fine.

What I disagree with is forcing people to use passwords that are at least 8 characters and by reading your code, it is impossible to set a password length less than 8 characters even if I would want to.

The deployments that will want to have strong passwords should keep the default value (of 8 or whatever stronger you prefer), but deployments that wish to have simpler passwords should have the ability to select a weaker password length.

They are the administrators, not us, so we should ship with strong defaults and give the ability to the end user to decide what is best for him.

If we get to a deployment where they ask us to have a password length of 4 to make it simple for the guest users registering at the front desk, then we should be able to hack the code to provide it to them which doesn't make any sense.

extrafu:
I agree here with @julsemaan . Make it default to 8 but allow the admins to make it as low as 4. There are many deployments out there with a simple splash page portal, asking for a trivial PIN to get Internet access. It's not about strong security here, it's about limiting 99.9% of the lemmings out there having free WiFi. A simple 4-digit PIN code achieves that, while also eliminating support calls because of too-complex PINs.

nqb:
> What I disagree with is forcing people to use passwords that are at least 8 characters and by reading your code, it is impossible to set a password length less than 8 characters even if I would want to.

I agree. We didn't speak about the same thing. On my side, I was focus on password generation for account created directly from the web admin. For **these** passwords, there is, currently, nothing you can do to have auto-generated passwords with length greater than 4. This was the issue I tried to fix. Of course, an admin that want auto-generated passwords can do it itself by doing a user import.

On your side, you spoke about passwords created automatically from authentication sources (Potd, SMS, email, etc.) for which you have a "Password length" field since #3286.

julsemaan:
@nqb 

Although there is a field to control the length of the password, your change prohibits passwords that are less than 8 characters long when they are being generated. This is the problem that needs to be addressed before your PR can get merged.

julsemaan:
superseded by 350eacc26153e0e3e03dc1b7a9157320eee3947f

PR #3903, opened 2019-01-07
fix/i18n_translation_utf8

jrouzierinverse:
Description
===========
Fix utf8 encoding in templates and saving files

Impacts
=======
The portal, admin, email templates

Delete branch after merge
=========================
NO

UPGRADE file entries
====================

We had an issue where configuration we being saved as latin1 instead of utf8.

This script will convert all latin1 config file to utf8

addons/upgrade/to-8.3-conf-latin1-to-utf8.sh

### DIFF
--- a/addons/upgrade/to-8.3-conf-latin1-to-utf8.sh
+++ b/addons/upgrade/to-8.3-conf-latin1-to-utf8.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
-
-find /usr/local/pf/conf -type f -print0 | while read -d $'\0' file; do 
+PF_DIR="/usr/local/pf"
+find "$PF_DIR/html/captive-portal/profile-templates" "$PF_DIR/conf" -type f -print0 | while read -d $'\0' file; do
     if file --mime-encoding  "$file" | grep -q iso-8859-1  ;then
         echo "converting $file from iso-8859-1 to utf-8" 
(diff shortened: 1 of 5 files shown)


addons/upgrade/to-8.3-conf-latin1-to-utf8.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+find /usr/local/pf/conf -type f -print0 | while read -d $'\0' file; do 
+    if file --mime-encoding  "$file" | grep -q iso-8859-1  ;then
+        echo "converting $file from iso-8859-1 to utf-8" 
+        iconv --from-code=iso-8859-1 --to-code=utf-8 "$file" -o "$file"
+   fi
+done
```
julsemaan, line 8: Needs a message saying the job is done so there is output even though no file has been touched

julsemaan:
Also this PR is missing labels and milestone

addons/upgrade/to-8.3-conf-latin1-to-utf8.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+find /usr/local/pf/conf -type f -print0 | while read -d $'\0' file; do
```
julsemaan: Also, we need to do this for the templates in html/captive-portal/profile-templates

jrouzierinverse:
Done

PR #3922, opened 2019-01-10
fix/3920

### LINKED ISSUE #3920: pfcmd (debian): untainted variable
On Debian 8.11 with PF 8.3:
```console
# /usr/local/pf/bin/pfcmd
Can't run infocmp to get a termcap entry: Insecure $ENV{PATH} while running
with -T switch at /usr/share/perl/5.20/Term/Cap.pm line 297.
[..]
```

jrouzierinverse:
Description
===========
Set the PATH variable to avoid issues tainted

Impacts
=======
The path resolution in pfcmd

Issue
=====
fixes #3920

### DIFF
--- a/bin/pfcmd.pl
+++ b/bin/pfcmd.pl
@@ -22,6 +22,7 @@ $< = 0;
 
 # To ensure group permissions are properly added
 umask(0007);
+$ENV{PATH} = "/sbin:/usr/sbin:/bin:/usr/bin";
 
 use pf::cmd::pf;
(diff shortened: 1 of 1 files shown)


nqb:
@jrouzierinverse,

I applied your patch and warning disappear, anything to do before merge and push to maintenance ?

I didn't test on CentOS.

julsemaan:
I would say that before pushing this to the maintenance, testing this on CentOS would be imperative.

PR #3958, opened 2019-01-24
Update and rename Unifi.pm to UniFi.pm

rcmcdonald91:
# Description
(REQUIRED)
Changed all instances of 'Unifi' to be the proper style 'UniFi'

# Impacts
(REQUIRED)
Cosmetic changes to the GUI

# Delete branch after merge
(REQUIRED)
YES

### DIFF
--- a/lib/pf/Switch/Ubiquiti/UniFi.pm
+++ b/lib/pf/Switch/Ubiquiti/UniFi.pm
@@ -1,17 +1,17 @@
-package pf::Switch::Ubiquiti::Unifi;
+package pf::Switch::Ubiquiti::UniFi;
 
 =head1 NAME
 
-pf::Switch::Ubiquiti::Unifi
+pf::Switch::Ubiquiti::UniFi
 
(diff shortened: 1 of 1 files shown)


julsemaan:
Doing this will require an upgrade script to be written so that switches with the type Unifi are changed to UniFi

rcmcdonald91:
I assume the only file that would need to be upgraded is switches.conf?

julsemaan:
Yes that is correct

rcmcdonald91:
`sed -i -e 's/type=Ubiquiti::Unifi/type=Ubiquiti::UniFi/g' /usr/local/pf/conf/switches.conf`

Not sure where this would need to go. But this will replace all instances with the correct type.

Edit: Seems we need a script to put in the /usr/local/pf/addons/upgrade folder?

extrafu:
A script addons/upgrade/to-9.0-unifi-rename.sh should be provided to handle the rename. Pushing to v9.1 until that is provided.

PR #3983, opened 2019-02-04
feature/config-options

### LINKED ISSUE #3902: OPTIONS for schema for config items
OPTIONS method to obtain the schema for config items

OPTIONS -> ```config/<collection>``` returns schema for a new resource
  * Config items with types will only return allowed if types if type is not provided

OPTIONS -> ```config/<resource>/<id>``` returns schema for a resource

jrouzierinverse:
Description
===========
expose the options for the config entries

Issue
=====
fixes #3902

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/go/api-frontend/aaa/authorization.go
+++ b/go/api-frontend/aaa/authorization.go
@@ -70,12 +70,12 @@ var pathAdminRolesMap = []adminRoleMapping{
 }
 
 var methodSuffixMap = map[string]string{
-	"GET":    "_READ",
-	"OPTIONS":    "_READ",
-	"POST":   "_CREATE",
-	"PUT":    "_UPDATE",
-	"PATCH":  "_UPDATE",
-	"DELETE": "_DELETE",
+	"GET":     "_READ",
+	"OPTIONS": "_READ",
+	"POST":    "_CREATE",
+	"PUT":     "_UPDATE",
+	"PATCH":   "_UPDATE",
+	"DELETE":  "_DELETE",
 }
 
 type TokenAuthorizationMiddleware struct {
--- a/go/api-frontend/aaa/authorization.go
+++ b/go/api-frontend/aaa/authorization.go
@@ -71,6 +71,7 @@ var pathAdminRolesMap = []adminRoleMapping{
 
 var methodSuffixMap = map[string]string{
 	"GET":    "_READ",
+	"OPTIONS":    "_READ",
 	"POST":   "_CREATE",
 	"PUT":    "_UPDATE",
 	"PATCH":  "_UPDATE",
--- a/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
@@ -42,6 +42,7 @@ sub field_list {
 
     my $list = [];
     my $section = $self->section;
+    return [] if !defined $section;
     my $default_pf_config = pf::IniFiles->new(-file => $pf_default_file, -allowempty => 1);
     my @section_fields = $default_pf_config->Parameters($section);
     foreach my $name (@section_fields) {
--- a/lib/pf/ConfigStore.pm
+++ b/lib/pf/ConfigStore.pm
@@ -267,13 +267,24 @@ reads a section without doing post-read cleanup
 
 sub readRaw {
     my ($self, $id, $idKey ) = @_;
+    return $self->readAllFromSection($id, $idKey, $self->cachedConfig);
+}
+
+=head2 readAllFromSection
+
+readAllFromSection
+
+=cut
+
+sub readAllFromSection {
+    my ($self, $id, $idKey, $config) = @_;
     my $data;
-    my $config = $self->cachedConfig;
     $id = $self->_formatSectionName($id);
     if ( $config->SectionExists($id) ) {
         $data = {};
-        my @default_params = $config->Parameters($self->default_section)
-            if (defined $self->default_section && length($self->default_section));
+        my $default_section = $self->default_section;
+        my @default_params = $config->Parameters($default_section)
+            if (defined $default_section && length($default_section));
         $data->{$idKey} = $self->_cleanupId($id) if defined $idKey;
         foreach my $param (uniq $config->Parameters($id), @default_params) {
             my $val;
@@ -287,6 +298,22 @@ sub readRaw {
         }
     }
     return $data;
+}
+
+=head2 readFromImported
+
+readFromImported
+
+=cut
+
+sub readFromImported {
+    my ($self, $id, $idKey) = @_;
+    my $config = $self->cachedConfig();
+    if ($config->{imported}) {
+        return $self->readAllFromSection($id, $idKey, $config->{imported});
+    }
+
+    return {};
 }
 
 =head2 update
--- a/lib/pf/ConfigStore.pm
+++ b/lib/pf/ConfigStore.pm
@@ -272,7 +272,7 @@ sub readRaw {
 
 =head2 readAllFromSection
 
-readAllFromSection
+read all parameters from a section
 
 =cut
 
@@ -302,7 +302,7 @@ sub readAllFromSection {
 
 =head2 readFromImported
 
-readFromImported
+read from the parameters from a section
 
 =cut
 
--- a/lib/pf/UnifiedApi/Controller/Config.pm
+++ b/lib/pf/UnifiedApi/Controller/Config.pm
@@ -348,7 +348,6 @@ sub options {
 
 =head2 options_from_form
 
-options_from_form
 
 =cut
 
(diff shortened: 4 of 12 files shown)


satkunas:
Missing Internal|External|Exclusive|Billing categorization in OPTIONS /api/v1/config/sources > allowed_values > type
This will be added as a group like with switches.

satkunas:
Defaults/Placeholders all null in config/switches using:

/usr/local/pf/sbin/pfperl-api get -M OPTIONS /api/v1/config/switch/default | jq
/usr/local/pf/sbin/pfperl-api get -M OPTIONS -c '{"group":null}' /api/v1/config/switch/default | jq
/usr/local/pf/sbin/pfperl-api get -M OPTIONS -c '{"group":"default"}' /api/v1/config/switch/default | jq

Endpoint output does not match the defaults/placeholders seen in "Admin > Switches > Add Switch > default".

satkunas:
Defaults/Placeholders all null in config/switch_groups using:

/usr/local/pf/sbin/pfperl-api get -M OPTIONS /api/v1/config/switch_groups | jq

Endpoint output does not match the defaults/placeholders seen in "Admin > Switch Groups > Add Switch Group".

julsemaan:
I think we would need to be able to dictate the validation rules from the server side as well so they aren't duplicated since @satkunas wants to prevent anything that doesn't validate in the frontend from flying to the server. To reach this goal, it would be good to have absolute consistency of the validation between the backend and frontend for common validations (length, allowed characters, format, size of zayme's sandwich)

satkunas:
@julsemaan we will include meta for `min_value` and `max_value` to handle positive ints.

julsemaan:
@satkunas 

Works for me as long as @jrouzierinverse names the field "validation" instead of "meta" ;)

go/api-frontend/aaa/authorization.go
```diff
@@ -71,6 +71,7 @@ var pathAdminRolesMap = []adminRoleMapping{
 
 var methodSuffixMap = map[string]string{
 	"GET":    "_READ",
+	"OPTIONS":    "_READ",
```
julsemaan: Please run "go fmt" on the files you modify in Golang to keep the formatting consistent and to ensure other folks running it don't start modifying lines they didn't directly touch in future commits

lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -92,6 +93,12 @@ sub build_list_search_info {
     return 200, $info;
 }
 
+=head2 items
+
+items
```
julsemaan, line 98: In all the file the perldoc description is just the name of the sub

I know some of them are self-explanatory with their name but a few words to describe the less obvious ones would be welcome

lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -327,6 +334,278 @@ sub sort_items {
...
+    my %output = (
+        defaults => \%defaults,
+        placeholders => \%placeholders,
+        allowed => \%allowed,
+        meta => \%meta,
+    );
+
+    for my $field ($form->fields) {
+        my $name = $field->name;
+        next if $name eq 'id';
```
julsemaan: Shouldn't the ID be pushed as well for stuff like required or its formatting ?

lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -327,6 +334,278 @@ sub sort_items {
...
+    my %output = (
+        defaults => \%defaults,
+        placeholders => \%placeholders,
+        allowed => \%allowed,
+        meta => \%meta,
+    );
+    my $inherited_values = $self->inherited_values;
+    for my $field ($form->fields) {
+        my $name = $field->name;
+        next if $name eq 'id';
```
julsemaan: Same as my comment above for the ID

lib/pf/UnifiedApi/Controller/Config/Sources.pm
```diff
@@ -185,6 +189,46 @@ sub form_process_parameters_for_cleanup {
...
+    my ($self, $type) = @_;
+    my $class = "pf::Authentication::Source::${type}Source";
+    return {
+        value => $type,
+        label => $type,
+        class => $class->meta->find_attribute_by_name('class')->default
+    };
+}
+
+sub options_with_no_type {
```
julsemaan, line 208: perldoc

satkunas:
Endpoint @ OPTIONS `config/sources?type=Email`

`allowed.email_activation_timeout` is empty, expecting object `allowed.email_activation_timeout.unit` with an Array of { label: ..., value: ... } durations.

----

Endpoint @ OPTIONS `config/sources?type=SMS`

`allowed.sms_activation_timeout` is empty, expecting object `allowed.sms_activation_timeout.unit` with an Array of { label: ..., value: ... } durations.

satkunas:
~~422 Response Code on endpoint OPTIONS `config/sources?type=POTD`~~

jrouzierinverse:
> 422 Response Code on endpoint OPTIONS `config/sources?type=POTD`

It should `type=Potd` not `type=POTD`

extrafu:
@julsemaan Can you do one more review so we can merge? @jrouzierinverse told me it was ready.

html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
```diff
@@ -42,6 +42,7 @@ sub field_list {
 
     my $list = [];
     my $section = $self->section;
+    return [] if !defined $section;
```
julsemaan, line 45: Worth logging something here, even in debug?

lib/pf/ConfigStore.pm
```diff
@@ -267,13 +267,24 @@ reads a section without doing post-read cleanup
 
 sub readRaw {
     my ($self, $id, $idKey ) = @_;
+    return $self->readAllFromSection($id, $idKey, $self->cachedConfig);
+}
+
+=head2 readAllFromSection
+
+readAllFromSection
```
julsemaan: perldoc that isn't just the sub name would be better here since that sub isn't what we can call simple

lib/pf/ConfigStore.pm
```diff
@@ -289,6 +300,22 @@ sub readRaw {
     return $data;
 }
 
+=head2 readFromImported
+
+readFromImported
```
julsemaan: same as my comment above

lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -327,6 +334,351 @@ sub sort_items {
...
+        $extra->{min_value} = $min;
+    } elsif ($field->isa("HTML::FormHandler::Field::PosInteger")) {
+        $extra->{min_value} = 0;
+    }
+
+    if (defined $max) {
+        $extra->{max_value} = $max;
+    }
+
+    return ;
```
julsemaan, line 458: Might as well just remove this if we're not returning anything
Unless there is some perl magic this does that I don't get

lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -327,6 +334,351 @@ sub sort_items {
...
+    my $max = $field->maxlength;
+    if ($min) {
+        $extra->{min_length} = $min;
+    }
+
+    if (defined $max) {
+        $extra->{max_length} = $max;
+    }
+
+    return ;
```
julsemaan, line 479: Same as my comment above

lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -327,6 +334,351 @@ sub sort_items {
...
+    my $name = $field->name;
+    my $cs = $self->config_store;
+    my $default_section = $cs->default_section;
+    my $value;
+    if ($default_section) {
+        my $item = $self->cleanup_item($cs->read($default_section, 'id'));
+        $value = $item->{$name};
+    }
+
+    return $value // do {
```
julsemaan: That isn't really in our usual practice.

To keep things consistent, I'd say we use an if/else like everywhere else in the code

lib/pf/UnifiedApi/Controller/Config/Subtype.pm
```diff
@@ -43,6 +43,60 @@ sub type_lookup {
     return {}
 }
 
+=head2 options
+
+options
```
julsemaan: Same comment about the perldoc here and in the other subs below

julsemaan (changes requested):

julsemaan:
Also requires unit tests for a least a few resources (with/without subtype, base, etc)

html/pfappserver/lib/pfappserver/Form/Config/Pf.pm
jrouzierinverse: That is a valid state.

jrouzierinverse: So no

lib/pf/UnifiedApi/Controller/Config.pm
jrouzierinverse: That is to avoid perl from returning the value of the last statement

jrouzierinverse: That is to avoid perl from returning the value of the last statement

jrouzierinverse:
This is done

PR #3988, opened 2019-02-07
Feature/ssl certs api

julsemaan:
# Description
API endpoints to manage HTTP and RADIUS certificates

# Impacts
Nothing if not used, built-in certificates when used

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [x] Document the feature
- [x] Add unit tests

# NEWS file entries
## New Features
* Added API endpoint to manage certificates


lib/pf/UnifiedApi/Controller/Config/Certificates.pm
```diff
@@ -0,0 +1,380 @@
...
+
+=cut
+
+sub resource_config {
+    my ($self, $certificate_id) = @_;
+    $certificate_id //= $self->stash->{certificate_id};
+    return $self->certs_map->{$certificate_id};
+}
+
+=head2 read_from_disk
```
extrafu: I think we should s/disk/file/ everywhere. You're reading a file afterall - it can be on a "disk", or not.

lib/pf/UnifiedApi/Controller/Config/Certificates.pm
```diff
@@ -0,0 +1,380 @@
...
+    }
+}
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2018 Inverse inc.
```
extrafu: 2018 -> 2019 everywhere

lib/pf/UnifiedApi/Controller/Config/Certificates.pm
julsemaan: I took your advice but renamed this to `read_from_tape_backup`

Made much more sense to me.

Also thought of `read_from_zaymes_sandwich` but base64 was too complex to be stored on that given @lzammit considers base64 as encryption

extrafu: I would have preferred read_from_floppy_disk as it's likely the most secured medium to store data now because no-one can read them anymore.

julsemaan: We could also store this on an 8-track and use an Oldsmobile as the interface to read the certs from

satkunas: We should also encrypt/decrypt using ROT26 so the files are not readable across the wire.

julsemaan:
@cgx 

I've put a mockup of the interface here as discussed yesterday:
https://support.inverse.ca/~jsemaan/cert-mockup.html

nqb:
My comments:
- would be nice to have the serial of the certificates
- would be nice to see if a certificate is self-signed (not sure it's easy to do)
- on the CSR tab, I didn't see the `OU=<section of the organization>`

julsemaan:
@nqb 

* Not sure the serial of the cert is really used by common admins
* Yes pretty easy, the subject of the cert is the same as the issuer so the frontend guys can take care of that with the available info in the API
* OU isn't mandatory so I've left it out to keep the required info as small as possible. Technically we only care about the CN anyway

nqb:
>     * Not sure the serial of the cert is really used by common admins

I know at least one guy who use the serial : me ;-) Pretty useful when you are doing a renew: certs have same issuer, subject, cn. The serial like the dates could be use to distinguish the new and the old.

julsemaan:
Adjusted the code to add the serial

Could we get the final review done ASAP so this can be merged so that I can base the new lets encrypt stuff off of this

docs/api/spec/paths/config/certificate.yaml
```diff
@@ -0,0 +1,82 @@
...
+      '422':
+        $ref: '#/components/responses/UnprocessableEntity'
+  put:
+    description: Replace an item
+    operationId: api.v1.Config::Certificates.replace
+    parameters:
+    - in: path
+      name: certificate_id
+      schema:
+        type: string
```
jrouzierinverse, line 52: Add a check_chain as a query parameter.

jrouzierinverse (changes requested):

docs/api/spec/paths/config/certificate.yaml
julsemaan: Done

julsemaan:
Can we get this merged today ?

PR #3993, opened 2019-02-08
use a fresh struct everytime we fetch pfconfig

### LINKED ISSUE #3992: pfconfig golang: Issue when removing items from configuration arrays
Due to how the golang library encoding/json merges the parsed data with the existing data in the struct, removing an element of an array in a resource that is managed by the pfconfig pool (hot-reload) will leave that element there since the same struct is being reused.

For example, if you would remove one of the passthroughs, then it would still be in the golang structs even if they reload.

Didn't see a flag to disable this behavior, very likely a new struct will be needed every time a fetch is made

julsemaan:
# Description
Ensure reloading a pfconfig object will not leave stale data in it

# Impacts
pfconfig golang library

# Issue
fixes #3992

# Delete branch after merge
YES

# Checklist
## Bug Fixes
* Fixed stale data that was left in the memory configuration when reloading Golang processes (#3992)

### DIFF
--- a/go/pfconfigdriver/fetch.go
+++ b/go/pfconfigdriver/fetch.go
@@ -35,6 +35,9 @@ var clusterSummary *ClusterSummary
 
 var nsHasOverlayRe = regexp.MustCompile(`.*\(.*\)$`)
 
+var isNamespaceConfiguration = regexp.MustCompile(`^Pfconfig`)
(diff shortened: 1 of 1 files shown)


julsemaan:
@fdurand 

Could you take a look at this and merge if it looks good.
That's tested though a separate PR so it should be safe for the maintenance

PR #4010, opened 2019-02-16
Fix sql request to detect the tenant id

fdurand:
# Description
Use the NAS-IP-Address attribute instead of the source ip address.
In case of a cluster the sql request try to find the tenant id of the management vip address, because it never exist then for each radius request we do 2 sql request for nothing.

# Impacts
- RADIUS flow

# Issue
Too many request in the db that never return a result.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Reduce the stress on the db from radius

### DIFF
--- a/raddb/policy.d/packetfence
+++ b/raddb/policy.d/packetfence
@@ -35,16 +35,21 @@ packetfence-local-auth {
 
 #Update the PacketFence-Tenant-Id defaulting to the default tenant_id
 packetfence-set-tenant-id {
+    if(!NAS-IP-Address || NAS-IP-Address == "0.0.0.0"){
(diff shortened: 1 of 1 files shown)


fdurand:
When i applied the fix::
tenant_id_fix

julsemaan:
> When i applied the fix::
> tenant_id_fix

I'm sure its a coincidence and its @lzammit stopping his OTBM mirror on that server while you patched it

fdurand:
Did you cherry-pick in the maintenance branch ?

PR #4017, opened 2019-02-21
Add support for Let's Encrypt through API

julsemaan:
# Description
Add API support for Let's Encrypt + a cronjob to renew the certificates

# Impacts
SSL certificates management

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [later when UI is done] Document the feature
- [n/a] Add unit tests

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -506,7 +506,6 @@ done
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/radiusd
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge
-%{__install} -D -m0644 conf/ssl/acme-challenge/success.txt $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge/success.txt
 %{__install} -d -m2775 $RPM_BUILD_ROOT%logdir
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/raddb/sites-enabled
 %{__install} -d -m2775 $RPM_BUILD_ROOT/usr/local/pf/var
@@ -1065,7 +1064,6 @@ fi
                         /usr/local/pf/conf/scan.conf.example
 %dir                    /usr/local/pf/conf/ssl
 %dir                    /usr/local/pf/conf/ssl/acme-challenge
-%config                 /usr/local/pf/conf/ssl/acme-challenge/success.txt
 %dir                    /usr/local/pf/conf/systemd
 %config                 /usr/local/pf/conf/systemd/*
 %config(noreplace)      /usr/local/pf/conf/switches.conf
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -505,6 +505,7 @@ done
 %{__install} -d -m2770 $RPM_BUILD_ROOT/usr/local/pf/conf
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/radiusd
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl
+%{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge
 %{__install} -d -m2775 $RPM_BUILD_ROOT%logdir
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/raddb/sites-enabled
 %{__install} -d -m2775 $RPM_BUILD_ROOT/usr/local/pf/var
@@ -1062,6 +1063,7 @@ fi
 %config(noreplace)      /usr/local/pf/conf/scan.conf
                         /usr/local/pf/conf/scan.conf.example
 %dir                    /usr/local/pf/conf/ssl
+%dir                    /usr/local/pf/conf/ssl/acme-challenge
 %dir                    /usr/local/pf/conf/systemd
 %config                 /usr/local/pf/conf/systemd/*
 %config(noreplace)      /usr/local/pf/conf/switches.conf
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -506,6 +506,7 @@ done
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/radiusd
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge
+%{__install} -D -m0644 conf/ssl/acme-challenge/success.txt $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge/success.txt
 %{__install} -d -m2775 $RPM_BUILD_ROOT%logdir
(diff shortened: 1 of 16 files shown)


conf/documentation.conf
```diff
@@ -1576,3 +1576,31 @@ Which changes (changing from a device class to another) should trigger internal:
 This setting is independant from trigger_on_device_class_change and allows to specify exactly which transitions should trigger internal::fingerbank_device_change. 
 Comma delimited transitions using the following format: $PREVIOUS_DEVICE_CLASS_ID->$NEW_DEVICE_CLASS_ID where $PREVIOUS_DEVICE_CLASS_ID and $NEW_DEVICE_CLASS_ID are the IDs in the Fingerbank database.
 EOT
+
+[lets_encrypt.radius]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Whether or not Let's Encrypt should be enabled for RADIUS certificate management
```
extrafu, line 1611: certificate -> certificates ? Same below.

conf/documentation.conf
```diff
@@ -1576,3 +1576,31 @@ Which changes (changing from a device class to another) should trigger internal:
...
+options=enabled|disabled
+description=<<EOT
+Whether or not Let's Encrypt should be enabled for HTTP certificate management
+EOT
+
+[lets_encrypt.live]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Whether or not Let's Encrypt should be using the live ACME API. Put to disabled only for testing since the certificates that will be issued will not be valid.
```
extrafu: Put -> Set

conf/documentation.conf
julsemaan: There is just one certificate for RADIUS unless I'm getting my grammar wrong

lib/pf/cmd/pf/renew_lets_encrypt.pm
```diff
@@ -0,0 +1,170 @@
...
+
+pf::cmd::pf::renew_lets_encrypt add documentation
+
+=head1 SYNOPSIS
+
+ pfcmd renew_lets_encrypt [http|radius|all]
+
+Renews Let's Encrypt certificates
+
+  http   | Renews the certificate for the HTTP resource if its enabled
```
extrafu: its -> it's or it is, same below

packetfence.cron.d
```diff
@@ -6,6 +6,9 @@ PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 # Database backup and maintenance script
 30 00 * * * root /usr/local/pf/addons/database-backup-and-maintenance.sh
 
+# Renew any Let's Encrypt certificates the first of the month
+1 0 1 * * root /usr/local/pf/bin/pfcmd renew_lets_encrypt
```
extrafu, line 10: Is root necessary or could it run under pf?

julsemaan: pfcmd usually always runs as root but the code will "downgrade" to PF rights when performing the update:
https://github.com/inverse-inc/packetfence/pull/4017/files#diff-83d8609c0217f8d78d579e0b28b2295aR78

julsemaan:
Adjusted the mockup for LE
https://support.inverse.ca/~jsemaan/cert-mockup-le.html

conf/documentation.conf
julsemaan: So, final verdict, should it be certificate or certificates ?

addons/packages/packetfence.spec
```diff
@@ -505,6 +505,8 @@ done
 %{__install} -d -m2770 $RPM_BUILD_ROOT/usr/local/pf/conf
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/radiusd
 %{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl
+%{__install} -d $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge
+%{__install} -D -m0644 conf/ssl/acme-challenge/success.txt $RPM_BUILD_ROOT/usr/local/pf/conf/ssl/acme-challenge/success.txt
```
fdurand: It looks that success.txt is missing in the code.

julsemaan: Actually forgot to remove it from the packaging since it uses a dynamic filename now

lib/pf/ssl/lets_encrypt.pm
```diff
@@ -0,0 +1,177 @@
...
+
+sub process_challenge {
+   my ($challenge) = @_;
+
+   my $logger = get_logger;
+
+   my $fname = $challenge->{token};
+   my $content = "$challenge->{token}.$challenge->{fingerprint}";
+   $logger->info("A file '$fname' with the text: $challenge->{token}.$challenge->{fingerprint} will be created for the Let's Encrypt challenge");
+   write_file("/usr/local/pf/conf/ssl/acme-challenge/$fname", $content);
```
fdurand: Use constant.

lib/pf/ssl/lets_encrypt.pm
```diff
@@ -0,0 +1,177 @@
...
+use strict;
+use warnings;
+
+use pf::file_paths qw($acme_challenge_dir);
+use pf::config qw(%Config);
+use pf::ConfigStore::Pf;
+use pf::constants qw($TRUE $FALSE);
+use pf::error qw(is_error);
+use Crypt::LE;
+use Data::Dumper;
```
fdurand: remove

lib/pf/ssl/lets_encrypt.pm
```diff
@@ -0,0 +1,177 @@
+package pf::ssl::lets_encrypt;
+
+=head1 NAME
+
+pf::ssl
```
fdurand: pf::ssl::lets_encrypt

fdurand (commented):
Just details.

fdurand:
Rebase please

docs/api/spec/components/schemas/configcertificate.yaml
```diff
@@ -53,6 +59,9 @@ ConfigCertificateReturnValue:
 
 ConfigCertificateDetails:
   properties:
+    common_name:
+      type: string
+      description: "The common name (CN) of the certificate"
```
cgx: Please use *The common name of the certificate* as you did elsewhere. Make sure to use the same string everywhere.

docs/api/spec/paths/config/certificate.yaml
```diff
@@ -21,6 +21,50 @@
...
+              $ref: '#/components/schemas/ConfigCertificateReturnValue'
+      '404':
+        $ref: '#/components/responses/NotFound'
+      '422':
+        $ref: '#/components/responses/UnprocessableEntity'
+
+/config/certificate/{certificate_id}/lets_encrypt:
+  description: pf::UnifiedApi::Controller::Config::Certificates
+  put:
+    description: Generate a CSR for a certificate resource
```
cgx: Bad copy-paste.

julsemaan:
Additional issues @cgx reported:
* perl-Crypt-LE RPM package doesn't compile when installed
* The spec declares we return whether or not let's encrypt is enabled but its not implemented

julsemaan:
Issues have been addressed except for the package which I need to fix.

Was able to replicate and its because of an autoload for windows that occurs on Linux for a reason that is still to be discovered

julsemaan:
Package perl-Crypt-LE has been fixed and is being pushed in devel right now and should be available for usage in about 15 minutes

PR #4029, opened 2019-03-01
Fix realm attribute on machine authentication

fdurand:
# Description
In the case of a machine authentication PacketFence detect the realm in multidomain but never set the realm in the radius request.
Used unlang do detect machine authentication and set the realm

# Impacts
- RADIUS authorize workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Bug Fixes
* Set the realm in the radius request when it's machine authentication

### DIFF
--- a/conf/radiusd/packetfence-tunnel.example
+++ b/conf/radiusd/packetfence-tunnel.example
@@ -20,13 +20,14 @@ server packetfence-tunnel {
 #  Make *sure* that 'preprocess' comes before any realm if you
 #  need to setup hints for the remote radius server
 authorize {
-    # TTLS does not send an EAP-Message to be parsed so the eap module
-    # cannot assign the EAP-Type
-    if ( outer.EAP-Type == TTLS) {
-        update request {
-            &EAP-Type := TTLS
-        }
+  # TTLS does not send an EAP-Message to be parsed so the eap module
+  # cannot assign the EAP-Type
+  if ( outer.EAP-Type == TTLS) {
+    update request {
+      &EAP-Type := TTLS
     }
+  }
+  packetfence-set-realm-if-machine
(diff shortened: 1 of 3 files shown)


conf/radiusd/packetfence.example
```diff
@@ -25,6 +25,7 @@ authorize {
 		&control:PacketFence-Request-Time := 0
 		&control:Load-Balance-Key := "%{Calling-Station-ID} %{User-Name}"
 	}
+	packetfence-set-realm-if-machine
```
julsemaan, line 28: Do we need to do this in the inner as well since technically the username in the inner could be different

fdurand: It will cover really rare cases but we can add it.

fdurand:
Bump

extrafu:
@julsemaan Review and merge

conf/radiusd/packetfence-tunnel.example
```diff
@@ -20,13 +20,14 @@ server packetfence-tunnel {
 #  Make *sure* that 'preprocess' comes before any realm if you
 #  need to setup hints for the remote radius server
 authorize {
-    # TTLS does not send an EAP-Message to be parsed so the eap module
-    # cannot assign the EAP-Type
-    if ( outer.EAP-Type == TTLS) {
-        update request {
-            &EAP-Type := TTLS
-        }
+  # TTLS does not send an EAP-Message to be parsed so the eap module
```
julsemaan: Just fix the indentation here so it matches the rest of the authorize section and I'll merge

fdurand: Fixaide

Issue #4049, opened 2019-03-12
Use Unicode characters like Chinese/Arabic not are not correctly shown in web admin


amiirrmm:
Hi

I am using ZEN vm version. I get my users and authentication from my active directory and in the AD my users have Unicode characters.   It is not correctly shown in web admin and maybe in the database. Is there any solution to this problem now or in the future? 

![Capture](https://user-images.githubusercontent.com/45257636/54189460-18658d80-44c7-11e9-8de7-1eb48a671e04.JPG)

julsemaan:
@jrouzierinverse 

Do you recall working on supporting different encodings when fetching users via LDAP?

I thought it was already addressed but that may just be my memory failing me

jrouzierinverse:
Bad memory
![](https://media.giphy.com/media/mYttD81qQ6vLi/giphy.gif)

amiirrmm:
So this is a bug, or at least a feature request, please.

Is there any temporary solution to this until an official fix ?

julsemaan:
That would be a feature request if it has never been part of PF

Will mark it for 9.1 and we'll see about addressing this in a future R&D meeting

PR #4075, opened 2019-03-17
BugFixes&&NewFeatures

simbazad:
# Description
BugFixes && New Features

# Impacts

# Delete branch after merge

YES

## Bug Fixes

* Issue with Unicode  (#4049)
* Issue with Nessus (api not available https://www.tenable.com/blog/a-clarification-about-nessus-professional) Nessus > 7 disabled create_scan api --- In REST_M.pm i've added routine "get_token" to extract the X-API UUID so we can use it to access the create_Scan api.
* Issue with the wmi tab never working - Fixed - check notes on the relevant commits.
* Issue with multiple scan engines never executed -- Fixed -  check notes on the relevant commits.
* Issue with " scan on registration " never starts -- Fixed
* More minor issues ...

### DIFF
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -52,7 +52,7 @@ CREATE TABLE class (
   delay_by int(11) NOT NULL default 0,
   external_command varchar(255) DEFAULT NULL,
   PRIMARY KEY (security_event_id)
-) ENGINE=InnoDB;
+) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
 --
 -- Table structure for table `person`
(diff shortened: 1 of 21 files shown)


jrouzierinverse:
To make easier please separate these into distinct pull requests per feature and fix.

simbazad:
@jrouzierinverse Rearranging those commits to separates pull request is a bit of an hassle , just tag me here again if it's necessary nonetheless

db/pf-schema-X.Y.Z.sql
```diff
@@ -52,7 +52,7 @@ CREATE TABLE class (
   delay_by int(11) NOT NULL default 0,
   external_command varchar(255) DEFAULT NULL,
   PRIMARY KEY (security_event_id)
-) ENGINE=InnoDB;
+) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
julsemaan, line 55: To anybody from @inverse-inc thinking of merging this PR, this will require internal discussions on the modifcation of the encoding of the database since this has very large concequences (external DBs, upgrade of existing databases, code that relies on a specific encoding)

julsemaan (changes requested):

jrouzierinverse:
As @julsemaan mentioned changing the database encoding could cause other issues and should be tested thoroughly.

This is why I would prefer to have the features/fixes in different branches.
So we can test each change before merging to ensure the changes that work get in quickly.

extrafu:
I agree with @jrouzierinverse . There's NO WAY your PR will ever get merged like this with a bunch of non-related fixes and features wrapped in a single PR.

simbazad:
@jrouzierinverse @extrafu  Done

PR #4079, opened 2019-03-18
Randomize the key balanced 

fdurand:
# Description
Randomize the key balanced calculation algorithm.
Let's say you have 1 internal server that proxy to 2 radius server then proxy to 2 others then with the modulo calculation the first proxy servers will always send the request to an only one server.
This code will allow to randomize the KeyBalanced attribute to load-balance (50/50) to the next servers.

# Impacts
Radius Proxy flow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Randomize KeyBalanced to randomize the load balance in Freeradius Proxy.

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1607,6 +1607,13 @@ description=<<EOT
 Treat OCSP response errors as 'soft' failures and still accept the certificate.
 EOT
 
+[radius_configuration.forward_key_balanced]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Send in the proxy request the attribute PacketFence-KeyBalanced to randomize the load balance in the next PacketFence RADIUS server
+EOT
+
 [pfdhcp.ip2mac_lookup]
 type=toggle
 options=enabled|disabled
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1611,7 +1611,8 @@ EOT
 type=toggle
 options=enabled|disabled
 description=<<EOT
-Send in the proxy request the attribute PacketFence-KeyBalanced to randomize the load balance in the next PacketFence RADIUS server
+Send in the proxy request the attribute PacketFence-KeyBalanced to randomize the load balance in the next PacketFence RADIUS server.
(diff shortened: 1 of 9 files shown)


conf/pf.conf.defaults
```diff
@@ -1155,6 +1155,11 @@ normalize_radius_machine_auth_username=disabled
 # Which attributes to use to get the username from a RADIUS request
 # The order of the attributes are listed in this configuration parameter is followed while performing the lookup
 username_attributes=PacketFence-UserNameAttribute,TLS-Client-Cert-Subject-Alt-Name-Upn,TLS-Client-Cert-Common-Name,User-Name
+#
+# radius_configuration.forward_key_balanced
+#
+# Send in the proxy request the attribute PacketFence-KeyBalanced to randomize the load balance in the next PacketFence RADIUS server
+PacketFence-KeyBalanced=disabled
```
julsemaan: That parameter name doesn't follow the convention in pf.conf.defaults

And also, not the same name as in documentation.conf (which is fine there)

jrouzierinverse:
Was this tested?

fdurand:
Not in production

jrouzierinverse:
Let's move to 9.1?

extrafu:
I think we can move it to v9.1 and assess later if we have time to get it into v9.

fdurand:
Bump

conf/documentation.conf
```diff
@@ -1607,6 +1607,13 @@ description=<<EOT
 Treat OCSP response errors as 'soft' failures and still accept the certificate.
 EOT
 
+[radius_configuration.forward_key_balanced]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Send in the proxy request the attribute PacketFence-KeyBalanced to randomize the load balance in the next PacketFence RADIUS server
```
julsemaan: Can you just add a brief description to explain the benefit of using that attribute like:
"Used to ensure requests are balanced when they come from @lzammit's sandwich"

PR #4080, opened 2019-03-18
New Nessus Support

simbazad:
# Description
Support for Nessus > 7

# Issue
Issue with Nessus (api not available https://www.tenable.com/blog/a-clarification-about-nessus-professional) Nessus > 7 disabled create_scan api --- In REST_M.pm i've added routine "get_token" to extract the X-API UUID so we can use it to access the create_Scan api.
# Delete branch after merge
(REQUIRED)
YES | NO

### DIFF
--- a/lib/pf/scan/nessus6.pm
+++ b/lib/pf/scan/nessus6.pm
@@ -26,7 +26,7 @@ use pf::util;
 use pf::node;
 use pf::constants qw($TRUE $FALSE);
 use pf::constants::scan qw($SCAN_SECURITY_EVENT_ID $PRE_SCAN_SECURITY_EVENT_ID $POST_SCAN_SECURITY_EVENT_ID $STATUS_STARTED);
-use Net::Nessus::REST;
+use Net::Nessus::REST_M;
 
 sub description { 'Nessus6 Scanner' }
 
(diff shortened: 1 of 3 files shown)


lib/pf/scan/nessus6.pm
```diff
@@ -26,7 +26,7 @@ use pf::util;
 use pf::node;
 use pf::constants qw($TRUE $FALSE);
 use pf::constants::scan qw($SCAN_SECURITY_EVENT_ID $PRE_SCAN_SECURITY_EVENT_ID $POST_SCAN_SECURITY_EVENT_ID $STATUS_STARTED);
-use Net::Nessus::REST;
+use Net::Nessus::REST_M;
```
julsemaan, line 29: Will this module still work with Nessus 6 ?

simbazad: @julsemaan  Yep , I've just added a routine (In the nessus rest implementation  - check out REST_M)  to extract the x-api token so we can call the create_Scan api

simbazad:
@julsemaan It will be great if you can also merge this .

julsemaan:
IMO this would better belong (the REST_M part) in https://github.com/guillomovitch/Net-Nessus-REST so that the package we use gets fixed and then we can update our dependency

That also has the advantage of fixing the module for non-PF users

If they fail to integrate this or take too long we can always merge as-is but lets try to make this as clean as possible

PR #4082, opened 2019-03-18
unicode support

simbazad:
# Description
Unicode support

# Delete branch after merge
(REQUIRED)
YES

### DIFF
--- a/conf/mariadb/mariadb.conf.tt.example
+++ b/conf/mariadb/mariadb.conf.tt.example
@@ -4,12 +4,18 @@
 #
 # See the examples of server my.cnf files in /usr/share/mysql/
 #
-
+[client]
+default-character-set = utf8mb4
+[mysql]
+default-character-set = utf8mb4
 # this is read by the standalone daemon and embedded servers
 [server]
 
 # this is only for the mysqld standalone daemon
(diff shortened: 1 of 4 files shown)


julsemaan (changes requested):
As I commented on #4075, modifying the character will require a lot of testing and precise upgrade instructions. So although the change here is simple, the team will need to work to make the transition smooth for all users

simbazad:
@julsemaan There are probably edge cases i didn't consider so testing this properly is obviously a good idea ,  And also you can take a look at this article (https://mathiasbynens.be/notes/mysql-utf8mb4) which contains useful information including queries to convert tables/columns character set.

satkunas:
Character sets affect query performance. It's not a big difference, but adds-up when the table size gets BIG. Since utf8 > ascii column size, we can increase the number of rows per page in the b-tree by switching from utf8 to ascii or binary.

We will eventually deprecate `latin`/`iso8859-1`/`utf8` in tables that only require `ascii`/`bin` character sets and we will split high-use tables (such as `node` and `rad*` into 2-or-more separate tables leaving the main table with INTs only (fixed row size), and the sibling tables to contain the verbiage that requires `utf8`.

PR #4083, opened 2019-03-18
Fix scan engines && wmi tab

simbazad:
## Bug Fixes

*Issue with the wmi tab never working - Fixed - check notes on the relevant commits.

*Issue with multiple scan engines never executed -- Fixed - check notes on the relevant commits.

Issue with " scan on registration " never starts -- Fixed

More minor issues ...

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
@@ -30,12 +30,12 @@ sub process_view {
     my $mac = $c->stash->{mac};
     my ($status, @scans);
     ($status, my $node) = $c->model("Node")->view($mac);
-    my $device_class = $node->{device_class};
+    my $device_class = $node->{fingerbank_info}->{device_fq};
     my $host_ip = $node->{iplog}->{ip};
     my $scan_model = $c->model("Config::Scan");
     eval {
         my $profile = pf::Connection::ProfileFactory->instantiate($mac);
-        foreach my $scan (split(/\s*,\s*/, $profile->{_scans})) {
+        foreach my $scan (@{ $profile->{_scans} }) {
             ($status, my $item) = $scan_model->read($scan);
             if (is_success($status)) {
                 push @scans, $item;
--- a/html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
@@ -108,7 +108,7 @@ sub parseWmi {
     my $scan_result = $scan->runWmi($scan_config, $rule_config);
     if ($scan_result =~ /ACCESS_DENIED/) {
         $rule_config->{item_exist} = "Access denied";
-    }elsif ($scan_result =~ /$WMI_NS_ERR/ || !@$scan_result) {
+    }elsif ($scan_result =~ /$WMI_NS_ERR/ || !$scan_result) {
         $rule_config->{item_exist} = 'No';
     }elsif ($scan_result =~ /TIMEOUT/ || $scan_result =~ /UNREACHABLE/) {
         $rule_config->{item_exist} = 'Request failed';
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -52,7 +52,7 @@ use pf::pfqueue::stats();
 use pf::pfqueue::producer::redis();
 
 use List::MoreUtils qw(uniq);
-use List::Util qw(pairmap);
+use List::Util qw(pairmap any);
 use File::Copy::Recursive qw(dircopy);
 use NetAddr::IP;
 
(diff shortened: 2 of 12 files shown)


html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
```diff
@@ -108,7 +108,7 @@ sub parseWmi {
     my $scan_result = $scan->runWmi($scan_config, $rule_config);
     if ($scan_result =~ /ACCESS_DENIED/) {
         $rule_config->{item_exist} = "Access denied";
-    }elsif ($scan_result =~ /$WMI_NS_ERR/ || !@$scan_result) {
+    }elsif ($scan_result =~ /$WMI_NS_ERR/ || !$scan_result) {
```
simbazad, line 111: When error occurs the $scan_result is of type string which yields this error:
Can't use string ("NTSTATUS: NT_STATUS_HOST_UNREACH"...) as an ARRAY ref while "strict refs"

lib/pf/scan.pm
```diff
@@ -145,8 +147,9 @@ sub parse_scan_report {
                'mac' => $mac,
                'reason' => 'manage_vclose',
             );
+        
             $apiclient->notify('close_security_event', %data );
-            $apiclient->notify('reevaluate_access', %data );
+
```
simbazad, line 152: No need to reevaluate access after parsing of the report , reevaluate should be in the event actions .

html/pfappserver/lib/pfappserver/Model/Node/Tab/WMI.pm
```diff
@@ -30,12 +30,12 @@ sub process_view {
     my $mac = $c->stash->{mac};
```
julsemaan, line 30: All of this will need to be adjusted in the API as well so the new admin can get these improvements

lib/pf/api.pm
```diff
@@ -981,30 +981,99 @@ sub trigger_scan :Public :Fork :AllowedAsAction($ip, mac, $mac, net_type, TYPE)
...
+
+        if( any { pf::util::isenabled($_->{'_registration'}) } @scanners ) {
+            $added = pf::security_event::security_event_add( $postdata{'mac'}, $pf::constants::scan::SCAN_SECURITY_EVENT_ID );
+        }
+
+        return if (!defined($added) || $added == 0 || $added == -1);
+
+        return if(pf::security_event::security_event_scaning($postdata{'mac'},$pf::constants::scan::SCAN_SECURITY_EVENT_ID));
+
+        iter_block:{
```
julsemaan: That's not really in our usual practice, use a while loop instead

lib/pf/api.pm
```diff
@@ -981,30 +981,99 @@ sub trigger_scan :Public :Fork :AllowedAsAction($ip, mac, $mac, net_type, TYPE)
...
+        
+            my $current_scan = pop @scanners;
+            
+            return if (!defined($current_scan));
+
+            if(pf::util::isenabled($current_scan->{'_registration'})){
+
+                $logger->info("On Regestration Scan - Current Scan Engine is : $current_scan");
```
julsemaan: *registration

lib/pf/api.pm
```diff
@@ -981,30 +981,99 @@ sub trigger_scan :Public :Fork :AllowedAsAction($ip, mac, $mac, net_type, TYPE)
...
-        # pre_registration
-        if (defined($scanner) && pf::util::isenabled($scanner->{'_pre_registration'})) {
-            $added = pf::security_event::security_event_add( $postdata{'mac'}, $pf::constants::scan::PRE_SCAN_SECURITY_EVENT_ID );
+
+        return if (!defined($added) || $added == 0 || $added == -1);
+
+        return if(pf::security_event::security_event_scaning($postdata{'mac'},$pf::constants::scan::POST_SCAN_SECURITY_EVENT_ID));
+
+
+        iter_block:{
```
julsemaan: Same comment about the while loop

lib/pf/api.pm
```diff
@@ -981,30 +981,99 @@ sub trigger_scan :Public :Fork :AllowedAsAction($ip, mac, $mac, net_type, TYPE)
...
+        if (any { pf::util::isenabled($_->{'_pre_registration'}) } @scanners ) {
+                $added = pf::security_event::security_event_add( $postdata{'mac'}, $pf::constants::scan::PRE_SCAN_SECURITY_EVENT_ID );
+            }
+
+        return if (!defined($added) || $added == 0 || $added == -1);
+
+        return if(pf::security_event::security_event_scaning($postdata{'mac'},$pf::constants::scan::PRE_SCAN_SECURITY_EVENT_ID));
+
+
+        iter_block:{
```
julsemaan: Here again

lib/pf/cmd/pf/schedule.pm
```diff
@@ -98,8 +98,23 @@ sub action_now {
...
+    my $host_mac = $mac || pf::ip4log::ip2mac($hostaddr);
+
+    my $profile = pf::Connection::ProfileFactory->instantiate($host_mac);
+    my @scanners = $profile->findScans($host_mac);
+    
+
     require pf::scan;
-    pf::scan::run_scan($hostaddr);
+
+    iter_block:{
```
julsemaan: And again a while loop

lib/pf/scan.pm
```diff
@@ -188,9 +191,11 @@ Prepare the scan attributes, call the engine instantiation and start the scan
 =cut
 
 sub run_scan {
-    my ( $host_ip, $mac ) = @_;
+    my ( $host_ip, $mac , $C_scanner ) = @_;
```
julsemaan: $C_scanner doesn't really follow our naming convention and isn't self-explanatory as a variable name

lib/pf/scan.pm
```diff
@@ -334,8 +341,8 @@ Check if the category matches the configuration of the scanner
 sub matchCategory {
     my ($self, $node_attributes) = @_;
     my $category = [split(/\s*,\s*/, $self->{_categories})];
-    my $node_cat = $node_attributes->{'category'};
-
+    my $node_cat = $node_attributes->{'category'} || 'default';
```
julsemaan: No role *is not* the default role. These are 2 completely different things in the PacketFence logic currently

julsemaan (changes requested):

julsemaan:
BTW, thanks a lot for the contribution and congrats for understanding that complex piece of code of the scan engines!

lib/pf/api.pm
simbazad: Done

lib/pf/api.pm
simbazad: Done

lib/pf/api.pm
simbazad: Done

lib/pf/cmd/pf/schedule.pm
simbazad: Done

lib/pf/scan.pm
simbazad: Done

lib/pf/scan.pm
simbazad: Ok good to know , check my changes because we still have to return before "any { $_ eq $node_cat } @$category" as it will cause an error when $node_cat not defined

simbazad:
> BTW, thanks a lot for the contribution and congrats for understanding that complex piece of code of the scan engines!
You welcome :)

@julsemaan  I've changed everything according to your notes , let me know if you need anything else

extrafu:
@julsemaan Should we merge this in v9?

julsemaan:
Lets merge since v9 will have us test everything

PR #4101, opened 2019-03-21
 Pfappserver Updates

satkunas:
# Description
Continue work on the pfappserver

# Impacts
Increased PR count :100: 

# Delete branch after merge
YES


html/pfappserver/root/static.alt/src/views/Configuration/_components/AccessDurationView.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <pf-config-view
     :isLoading="isLoading"
+    :percentLoading="percentLoading"
```
cgx: Please remove this parameter if it's not (yet) useful.

html/pfappserver/root/static.alt/src/store/index.js
```diff
@@ -4,6 +4,7 @@ import Vuex from 'vuex'
 import notification from './modules/notification'
 import session from './modules/session'
 import config from './modules/config'
+import performance from './modules/performance'
```
cgx: Please remove this store if it isn't used (yet).

html/pfappserver/root/static.alt/src/views/Configuration/_components/AdminRoleView.vue
```diff
@@ -144,22 +148,29 @@ export default {
...
-      this.$store.dispatch(`${this.storeName}/deleteAdminRole`, this.id).then(response => {
-        this.close()
-      })
     }
   },
   created () {
     this.init()
+  },
+  watch: {
+    isClone: {
```
cgx, line 169: Could we do this check in `mounted` instead? I don't think a watcher is necessary since `isClone` won't changed.

cgx (changes requested):

html/pfappserver/root/static.alt/src/utils/api.js
```diff
@@ -37,16 +38,154 @@ Object.assign(apiCall, {
...
+            })
+            if (num > 0 && time > 0) {
+              return { time, num }
+            }
+          }
+          // TODO search siblings, parent
+          // method not available, inspect parent for same method
+          urlParts.pop()
+          if (urlParts.length > 0) {
+            return stat(this.benchmarks, method, `/${urlParts.join('/')}`)
```
cgx: No need to prefix your url with a slash since `stat` will ignore it anyway.

cgx (approved):

PR #4104, opened 2019-03-22
Feature/api interfaces

julsemaan:
# Description
Add API endpoints to manage network interfaces

# Impacts
Unified API

# Delete branch after merge
YES

### DIFF
--- a/docs/api/spec/components/schemas/configinterface.yaml
+++ b/docs/api/spec/components/schemas/configinterface.yaml
@@ -1,4 +1,5 @@
-ConfigInterface:
+ConfigInterfaceBase:
+  type: object
   properties:
     dhcpd_enabled:
       description: Whether or not DHCP is enabled for this interface
@@ -15,23 +16,8 @@ ConfigInterface:
       enum:
         - 1
         - 0
-    hwaddr:
-      description: The MAC address of the interface
-      type: string
-    ifindex:
-      description: The ifindex
-      type: string
     ipaddress:
       description: The IP address of the interface
-      type: string
-    is_running:
-      description: Whether or not the interface is running (up)
-      type: boolean
-    master:
-      description: The physical interface associated to a VLAN interface when applicable
-      type: string
-    name:
-      description: The name of the interface
       type: string
     nat_enabled:
       description: Whether or not NAT is enabled for this interface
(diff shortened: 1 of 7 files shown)


julsemaan:
To be able to fully replicate the interfaces page of the current admin, API endpoints for networks will be needed (for the "ADD ROUTED NETWORK" button). Until then, everything should be available to replicate the page except for that feature. I'll very likely open another PR for the networks API

extrafu:
💪

satkunas:
`type` should be an array (and accept an array): "type": ["management","portal"]
```json
{
  "items": [
    {
      "enp0s3": { ... },
      "enp0s8": { ... },
      "enp0s9": {
        ...
        "type": "management,portal",
      },
      "is_running": false,
      "network_iseditable": false
    }
  ],
  "status": 200
}
```

Move the interface name from the object key into the value object as `id`:
```json
{
  "items": [
    { "id": "enp0s3", ... },
    { "id": "enp0s8", ... },
    { "id": "enp0s9", ... },
  ]
}
```

satkunas:
Is `"type": "none"` a valid value, or should we use `"type": null` instead?
```json
{
  "item": {
    ...
    "type": "none"
  },
  "status": 200
}
```

julsemaan:
@satkunas 

None is a value accepted by the current pfappserver model and a valid "enforcement" method so I believe it should stay since changing it could have some side effects.

I'll take care of translating to an array but I think type will have to stay a string and I'll push the additional stuff in additional_daemons which is used during the POST/PUT

satkunas:
Errors have 1 too many nested levels of "message".

```json
{
  "errors": [],
  "message": {
    "message": "Interface lo doesn't exist"
  },
  "status": 404
}
```

julsemaan:
@satkunas 

I've performed adjustments on the API endpoints, please check them again and let me know if I messed something else

julsemaan:
@satkunas 

Can you check this again so we can get this ASAP and it doesn't rot here

extrafu:
@satkunas Please confirm and merge so the GUI work around this can get going.

satkunas:
@julsemaan looks good. Please rebase so I can merge.

PR #4107, opened 2019-03-22
feature/fingerbank_api

jrouzierinverse:
Description
===========
Fingerbank API

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


extrafu:
@julsemaan Please review/comment ASAP in order to get it in.

lib/pf/UnifiedApi/Controller/Fingerbank.pm
```diff
@@ -0,0 +1,281 @@
...
+
+sub remove {
+    my ($self) = @_;
+    my $id = $self->id;
+    my ($status, $msg) = $self->fingerbank_model->delete($id);
+    if (is_error($status)) {
+        return $self->render_error($status, $msg);
+    }
+
+    return $self->render(json => {});;
```
julsemaan, line 131: Not sure its a plain 200 the other deletion calls return

lib/pf/UnifiedApi/Controller/Fingerbank.pm
```diff
@@ -0,0 +1,281 @@
...
+    my ($self) = @_;
+    url_unescape($self->stash->{$self->url_param_name});
+}
+
+sub build_list_search_info {
+    my ($self) = @_;
+    my $params = $self->req->query_params->to_hash;
+    my $scope = $self->stash->{scope};
+    my $model = $self->fingerbank_model;
+    my $db = fingerbank::DB_Factory->instantiate(schema => 'Local');
```
julsemaan, line 150: We can't search in the upstream schema or is this just for metadata ?

lib/pf/UnifiedApi/Controller/Fingerbank/UserAgent.pm
```diff
@@ -0,0 +1,48 @@
+package pf::UnifiedApi::Controller::Fingerbank::UserAgent;
```
julsemaan: Why duplicate controllers for files that contain exactly the same thing ?

Anything different here than from the other resources ?

lib/pf/UnifiedApi/Search/Builder/Fingerbank.pm
```diff
@@ -0,0 +1,526 @@
...
+    my ($self, $search_info) = @_;
+    my ($status, $search_args) = $self->make_search_args($search_info);
+    if ( is_error($status) ) {
+        return $status, $search_args;
+    }
+
+    my $model = $search_info->{model};
+    my $limit = $search_args->{'-limit'};
+    my $offset = $search_args->{'-offset'};
+    #use Data::Dumper;print Dumper($search_args);
```
julsemaan: remove

t/unittest/UnifiedApi/Plugin/RestCrud.t
```diff
@@ -1,490 +0,0 @@
-#!/usr/bin/perl
```
julsemaan, line 1: Looks like this is gone but not replaced by anything else.

Anyway we can keep the same level of testing ?

julsemaan (changes requested):

julsemaan:
We're also missing the spec for the new API endpoints

lib/pf/UnifiedApi/Controller/Fingerbank.pm
jrouzierinverse: It is

lib/pf/UnifiedApi/Controller/Fingerbank.pm
jrouzierinverse: This is for metadata

lib/pf/UnifiedApi/Controller/Fingerbank.pm
jrouzierinverse: It is

lib/pf/UnifiedApi/Controller/Fingerbank.pm
jrouzierinverse: Just for meta

lib/pf/UnifiedApi/Controller/Fingerbank/UserAgent.pm
jrouzierinverse: This is a mistake will fix

t/unittest/UnifiedApi/Plugin/RestCrud.t
jrouzierinverse: Same level of testing is not needed.

The plugin only provides very simple shortcuts to the route class.
If the shortcuts did not work then routing would not work.

PR #4119, opened 2019-03-25
Feature/accounting triggers history

### LINKED ISSUE #4066: Allow suricata trigger + accounting to be in the same security event trigger
Right now accounting + suricata (any event) triggers cannot be mixed in the same event trigger.

We should treat accounting triggers as non-event triggers since they can be computed from DB data

This is necessary for the new security events v9 GUI

julsemaan:
# Description
Allow to search current accounting triggers when triggering another security event.
Makes the accounting triggers non-event (can be mixed with Suricata, nexpose, etc)

# Impacts
Security event engine
Accounting maintenance

# Fixes issue
fixes #4066

# Delete branch after merge
YES

### DIFF
--- a/html/pfappserver/root/security_event/view.tt
+++ b/html/pfappserver/root/security_event/view.tt
@@ -86,6 +86,7 @@
               <select id="trigger_type" class="input-medium">
                 <option value=""></option>
                 [% FOREACH type IN trigger_types -%]
+                [% NEXT IF type == "recorded_accounting" %]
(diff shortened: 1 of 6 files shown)


julsemaan:
@fdurand 

Can you review + merge the PR ?
Its necessary to have it in 9.0 since the new GUI supports accounting triggers mixed with events triggers like suricata

extrafu:
@fdurand Please test and merge ASAP. We want no open PR for v9 after tomorrow.

PR #4155, opened 2019-03-28
Feature/domain queue join

### LINKED ISSUE #3561: v9-gui: Add the ability to join rejoin and unjoin domains via the api

julsemaan:
# Description
Allow to start a pfqueue task asynchronously and poll its status
Use the above for long-running domain operations

# Impacts
pfqueue plumbing
Unified API

# Issue
fixes #3561 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

### DIFF
--- a/go/caddy/job-status/job-status.go
+++ b/go/caddy/job-status/job-status.go
@@ -23,6 +23,7 @@ const REDIS_PORT = 6380
 
 const STATUS_PENDING = "Pending"
 const STATUS_COMPLETED = "Completed"
+const STATUS_FAILED = "Failed"
 
 const POLL_TIMEOUT = 15
 
@@ -80,10 +81,24 @@ func buildJobStatusHandler(ctx context.Context) (JobStatusHandler, error) {
 func (h JobStatusHandler) handleStatusPoll(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
 	ctx := r.Context()
 	jobId := p.ByName("job_id")
+
+	data, err := h.redis.HGetAll(h.jobStatusKey(jobId)).Result()
+	if err != nil {
+		msg := "Unable to get job status from redis database"
+		w.WriteHeader(http.StatusInternalServerError)
+		h.writeMessage(ctx, msg, w)
+		log.LoggerWContext(ctx).Error(msg + ": " + err.Error())
+	} else if status := data["status"]; status == STATUS_COMPLETED || status == STATUS_FAILED {
+		res, _ := json.Marshal(data)
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprintf(w, string(res))
+		return
+	}
+
 	updatesKey := h.jobStatusUpdatesKey(jobId)
 
 	sub := h.redis.Subscribe(updatesKey)
-	_, err := sub.Receive()
+	_, err = sub.Receive()
 	if err != nil {
 		msg := "Unable to get job status from redis database"
 		w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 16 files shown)


julsemaan:
Example:
https://support.inverse.ca/~jsemaan/domain-join.html

lib/pf/pfqueue/status_updater/redis.pm
```diff
@@ -0,0 +1,239 @@
...
+
+sub set_in_status_hash {
+    my ($self, $key, $data) = @_;
+    if($self->finalized) {
+        get_logger->trace("Not reporting data for $key since status has been finalized");
+        return;
+    }
+    
+    $self->connection->multi();
+    $self->connection->expire($self->status_key, $self->status_ttl);
```
jrouzierinverse: The expire should be done after the hset.
Because if the hash does not exists the expire will fail.

jrouzierinverse (changes requested):
Redis PUB/SUB is unreliable

The current approach will cause some messages to be lost.
If the subscriber is not listening when the message is published that message will be lost.

The only way to avoid losing message is to use BRPOPLPUSH/RPOPLPUSH.

go/caddy/job-status/job-status.go
```diff
@@ -0,0 +1,195 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"git.inverse.ca/inverse/fingerbank-processor/sharedutils"
+
+	"github.com/go-redis/redis"
```
jrouzierinverse, line 10: We already have a redis connection pool
"github.com/inverse-inc/packetfence/go/redisclient"

This uses the config of pfqueue to connect to redis.
I know it is too late to rework at least use the config to connect to redis.

julsemaan:
We don't really care if a publication gets lost since it doesn't contain anything important as the status key holds all the data.

Receiving a publication on the channel just allows the polling to complete faster. The worst that can happen is that the user will wait 15 seconds because of a missed publication that occurred between his 2 requests. The result of the operation will be the same though.

To me that isn't a big issue but if that's a blocker to the team, I can adjust it but the publisher (pfqueue) shouldn't be blocking and ensuring his publication went through and was received by someone as there is no guarantee someone will be polling the status. Someone could just give up on polling the status of the job or some jobs could be started with intent of polling but the consumer just decides to let it do its job without polling.

julsemaan:
As verbally discussed, I made polling exit early if the job has already completed

go/caddy/job-status/job-status.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return jobStatus, nil
+}
+
+func (h JobStatusHandler) handleStatusPoll(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	ctx := r.Context()
+	jobId := p.ByName("job_id")
+	updatesKey := h.jobStatusUpdatesKey(jobId)
+
+	sub := h.redis.Subscribe(updatesKey)
```
jrouzierinverse, line 106: defer sub.Close() to avoid a resource leak

jrouzierinverse (changes requested):

julsemaan:
Performed the requested adjustments

jrouzierinverse:
rebase

julsemaan:
Adjustments performed for password handling + rebased

jrouzierinverse:
rebase then I will merge

julsemaan:
rebased

PR #4210, opened 2019-04-04
eduroam local realm proxy

fdurand:
# Description
Added the way to proxy eduroam local realm to another radius server (like nps)

# Impacts
- RADIUS eduroam workflow

# Delete branch after merge
(REQUIRED)
YES | NO

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Added a way to forward eduroam local realm to a define radius server instead of packetfence.

### DIFF
--- a/conf/radiusd/eduroam-cluster.example
+++ b/conf/radiusd/eduroam-cluster.example
@@ -29,8 +29,6 @@ server eduroam.cluster {
         authorize {
             suffix
             ntdomain
-            %%reject_realm%%
-            %%local_realm%%
             if(!NAS-IP-Address || NAS-IP-Address == "0.0.0.0"){
                     update request {
                             NAS-IP-Address := "%{Packet-Src-IP-Address}"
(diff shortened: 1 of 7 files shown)


fdurand:
Bump

julsemaan:
I'm not sure we should have both the eduroam source and these configuration options because it will be confusing as to what should be configured where and what settings apply in both cases.

If we plan on having this as the way to integrate eduroam, we should deprecate the eduroam source and provide an upgrade script + instructions so that we can migrate to that

fdurand:
The eduroam sources are still necessary if you want to deal with eduroam.
The goal of this PR is per example you don't want to join the PacketFence server to the domain but to authenticate incoming request (from eduroam server) to a 3rd party server (like the MS NPS server).

julsemaan:
So just to make sure, if we leave all of this as default all the existing inbound/outbound eduroam we already have will work as-is?

fdurand:
yes

PR #4275, opened 2019-04-22
Improve documentation conventions

nqb:
# Description
Add two new conventions

# Impacts
Docs

# Delete branch after merge
YES

### DIFF
--- a/docs/PacketFence_Developers_Guide.asciidoc
+++ b/docs/PacketFence_Developers_Guide.asciidoc
@@ -178,7 +178,7 @@ Navigate to menu:File[Import>Import csv] to import a csv file.
 
 a|Navigate to menu:File[Import>Import csv] to import a csv file.
 
-|GUI button and menu (non experimental)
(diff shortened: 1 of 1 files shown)


docs/PacketFence_Developers_Guide.asciidoc
```diff
@@ -178,6 +178,15 @@ Navigate to menu:File[Import>Import csv] to import a csv file.
 
 a|Navigate to menu:File[Import>Import csv] to import a csv file.
 
+|GUI button and menu (non experimental)
```
extrafu: non-experimental

extrafu: We might want to write "GUI buttons and menus...." - if we want to use the plural form everywhere.

nqb: I use singular form because we use previously:
```
GUI Button
[..]
GUI Menu
```

PR #4541, opened 2019-05-17
Pfappserver - append "-copy" to id on clone

### LINKED ISSUE #4468: Add suffix or prefix to cloned configuration
When cloning a piece of configuration (lets say a source), the name in the clone is the same as the source you're cloning which results in a warning right when you get onto the page

Let's say the initial source name is "@lzammit's sexy LDAP server", cloning it could make the ID of the new source "@lzammit's sexy LDAP server (copy)"

Not sure if its decided in the frontend or backend so assigning @satkunas and @jrouzierinverse

satkunas:
# Description
Automatically append "-copy" suffix to the form `id` on clone.

# Impacts
When cloning the id will have "-copy" appended to it in all configuration sections, expect interfaces and l2_networks (routed).

# Issue
Fixes #4468

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature
- [n/a] Add unit tests

## Bug Fixes
Fixes #4468

### DIFF
--- a/html/pfappserver/root/static.alt/src/views/Configuration/_components/AdminRoleView.vue
+++ b/html/pfappserver/root/static.alt/src/views/Configuration/_components/AdminRoleView.vue
@@ -123,6 +123,7 @@ export default {
         if (this.id) {
           // existing
           this.$store.dispatch(`${this.storeName}/getAdminRole`, this.id).then(form => {
+            if (this.isClone) form.id = `${form.id}-${this.$i18n.t('copy')}`
             this.form = form
           })
         } else {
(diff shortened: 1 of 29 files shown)


html/pfappserver/root/static.alt/src/views/Configuration/_components/AdminRoleView.vue
```diff
@@ -123,6 +123,7 @@ export default {
         if (this.id) {
           // existing
           this.$store.dispatch(`${this.storeName}/getAdminRole`, this.id).then(form => {
+            if (this.isClone) form.id = `${form.id}-${this.$i18n.t('copy')}`
```
julsemaan, line 126: Is it possible to isolate this logic in a function instead of a full copy/paste ?

satkunas: @julsemaan not currently. There are long-term plans to move the majority of the business logic out of the templates and in to the store modules. Until then, the individual templates are currently the only code-path that has access to both `isClone` and the API response.

PR #4558, opened 2019-05-21
fix/allowed_options

### LINKED ISSUE #4536: v9_gui: admin role issue.
Let's create a admin role hotel:
image

with the following user options:
image

If i assign a local user to this admin role, connect to the admin gui with that user and i try to create a user then the access duration list is empty and the role list contain all the role available.

### LINKED ISSUE #4854: Admin roles: Allowed node roles not taken into account
When a user has an admin role with a limited list of `Allowed node roles`, he is able to create nodes with all roles (not only `Allowed node roles`).

**To Reproduce**
1. Create two roles: `role1` and `role2`
1. Create an admin role called `Role-1`
   * Actions: Nodes - Create
   * Node options - Allowed node roles: `role1`
1. Create a user `foo` and assign `Role-1` as access level
1. Log in as `foo` user and try to create a node with `role2`

=> Node is created.

**Expected behavior**
User `foo` should be able to create only nodes with role `role1`.

jrouzierinverse:
Description
===========
Use the Admin Roles allowed_access_levels

Issue
=====
Fixes #4536
Fixes #4854

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/html/pfappserver/root/static.alt/src/components/pfCSVImport.vue
+++ b/html/pfappserver/root/static.alt/src/components/pfCSVImport.vue
@@ -146,8 +146,20 @@
               </b-col>
               <b-col>
 
-                <!-- BEGIN SUBSTRING -->
-                <pf-form-input v-if="isFieldType(substringValueType, staticMap)"
+                <pf-form-chosen v-if="isComponentType([componentType.SELECTMANY, componentType.SELECTONE], staticMap)"
+                  :value="staticMap.value"
+                  label="name"
+                  track-by="value"
+                  :ref="staticMap.key"
+                  :disabled="isDisabled"
+                  :multiple="isComponentType([componentType.SELECTMANY], staticMap)"
+                  :options="fieldTypeValues[staticMap.key](this)"
+                  :vuelidate="getStaticMappingVuelidateValue(index)"
+                  @input="staticMap.value = $event"
+                  collapse-object
+                ></pf-form-chosen>
+
+                <pf-form-input v-else-if="isComponentType([componentType.SUBSTRING], staticMap)"
                   v-model="staticMap.value"
                   :ref="staticMap.key"
                   :disabled="isDisabled"
@@ -155,8 +167,7 @@
                   :vuelidate="getStaticMappingVuelidateValue(index)"
                 ></pf-form-input>
 
-                <!-- BEGIN DATE -->
-                <pf-form-datetime v-else-if="isFieldType(dateValueType, staticMap)"
+                <pf-form-datetime v-else-if="isComponentType([componentType.DATE], staticMap)"
                   v-model="staticMap.value"
                   :ref="staticMap.key"
                   :config="{format: 'YYYY-MM-DD'}"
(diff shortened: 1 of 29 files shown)


fdurand:
Still missing the access duration.
Also the user need to have the "Nodes - Read" action to be able to see the roles.
Add all the roles appear in the list and not only the one allowed in the admin role.

julsemaan:
@jrouzierinverse 

Can you fix this ASAP so that it can get into 9.0.1

julsemaan:
Pushing to 9.0.2 so we can get a full fix for the issue

nqb:
We will need this PR merged for 9.2.

julsemaan:
@jrouzierinverse 

Could you work on a full fix based on @fdurand's comments. Its indeed overdue as the current admin lacks some advanced access control due to that missing feature

extrafu:
@jrouzierinverse 

Make sure we handle roles and access duration validation.

extrafu:
@julsemaan If all good please merge.

julsemaan:
Given @fdurand is the assignee and probably knows the problem with more details than me I'll just let him double check the PR to ensure it meets what he expects for the fix

fdurand:
Still doesn't work, from a user i created with node_manager role, i am able to change the role of a device to any role and same for user and role/access duration.
Also my own user is able to change his own admin role and grant access to ALL.

jrouzierinverse:
Let me see

fdurand:
Nodes options looks good.
User options no.
I am not able to set any role/access duration/access level

fdurand:
After added:
Security event  - READ
Nodes - READ
Configuration Main - READ
System - READ

I am able to set the role of the device (only the allowed roles) but i can set any access duration.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfiguration.js
```diff
@@ -514,6 +543,38 @@ export const pfConfigurationActions = {
       }
     }
   },
+  set_role_by_acl_node: {
```
cgx: @satkunas This action doesn't seem to be used anywhere. Forgot to commit your changes to the Node components?

html/pfappserver/root/static.alt/src/components/pfCSVImport.vue
```diff
@@ -741,17 +686,33 @@ export default {
...
-        const field = this.fields[index]
-        if ('types' in field && field.types.includes(type)) {
-          return true
+    isComponentType (componentTypes, { key }) {
+      if (key) {
+        const index = this.fields.findIndex(field => field.value === key)
+        if (index >= 0) {
+          const field = this.fields[index]
+          for (let t = 0; t < componentTypes.length; t++) {
+            if (field.types.map(type => fieldTypeComponent[type]).includes(componentTypes[t])) return true
```
cgx: @satkunas Put the `map` outside the *for* loop for better performance.

cgx (changes requested):

cgx (approved):

Issue #4579, opened 2019-05-29
v9 Fingerbank empty set


WallaPF:
**bug description**
Fingerbank "Combination" and "User_Agent" tables return an empty value

**To Reproduce**
Get these tables from the API

**packetfence.log errors**
May 29 13:04:31 PacketFence-ZEN packetfence: INFO pfperl-api(24632): Searching for 'Combination' entries in schema(s) returned an empty set (fingerbank::Base::CRUD::search)
May 29 13:04:31 PacketFence-ZEN packetfence: INFO pfperl-api(24633): Searching for 'Combination' entries in schema(s) returned an empty set (fingerbank::Base::CRUD::search)
May 29 13:04:31 PacketFence-ZEN packetfence: INFO pfperl-api(24633): Searching for 'User_Agent' entries in schema(s) returned an empty set (fingerbank::Base::CRUD::search)
May 29 13:04:32 PacketFence-ZEN packetfence: INFO pfperl-api(24639): Fetched user account information successfully (fingerbank::API::account_info)

**Expected behavior**
Get Fingerbank's response from these tables

julsemaan:
These tables aren't shipped anymore from PF 8.0 and above as the database is only used for creating local combinations but not for discovery anymore as it goes through the API at 100% now

WallaPF:
@julsemaan Note that it did work in ver 8.3

julsemaan:
Not sure how that ever worked unless you had an outdated database version or pointing to the wrong API server since its stripped from the DB since the launch of Fingerbank v2

WallaPF:
**Refining the case**
When a node is identified, Its fingerprint won't populate in the in the node tab.
These are the relevant packetfence.log errors:

Jun  2 15:44:44 PF-9 packetfence: WARN pfperl-api(13318): Can't find scan engine for 68:f7:28:bd:95:17 since we don't have it's OS (pf::Connection::Profile::findScan)
Jun  2 15:44:50 PF-9 packetfence: INFO pfperl-api(13319): Could not find any 'Device' entry with ID '31' in schema 'Upstream' (fingerbank::Base::CRUD::read)
Jun  2 15:44:50 PF-9 packetfence: ERROR pfperl-api(13319): Unable to read device 31 (Class::MOP::Class:::after)

Manual "Refresh Fingerbank" in the node tab will return the correct values.
So, in the scanning a node process (especially WMI) the fingerprinting is broken.

julsemaan:
[...]

PR #4591, opened 2019-06-06
fix

simbazad:
# Description
Fix Issue 4579

## Bug Fixes
Issue 4579

### DIFF
--- a/lib/pf/dhcp/processor.pm
+++ b/lib/pf/dhcp/processor.pm
@@ -196,7 +196,7 @@ sub processFingerbank {
     # DHCP filters (Fingerbank scope)
     # If there is a match, we override Fingerbank call
     my $dhcp_filter_rule = $self->filterEngine->filter('Fingerbank', $fingerbank_args);
-    unless ( $dhcp_filter_rule ) {
+    unless ( (keys $dhcp_filter_rule) > 0 ) {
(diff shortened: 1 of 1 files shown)


julsemaan:
@simbazad 

Can you fix the branch so that it doesn't contain the merge commit so that it keeps the history clean

Otherwise I can cherry-pick the fix in devel

Also, can you confirm how this fixes #4579?

Thanks

simbazad:
> @simbazad
> 
> Can you fix the branch so that it doesn't contain the merge commit so that it keeps the history clean
> 
> Otherwise I can cherry-pick the fix in devel
> 
> Also, can you confirm how this fixes #4579?
> 
> Thanks

@julsemaan I've opened a new pull request here:https://github.com/inverse-inc/packetfence/pull/4614

And regarding the original issue 4579 - The person who opened this issue is the CISO in the company that i also work for, so i'm a developer and on my free time i'm helping him by fixing the bugs that he encounters and the issue was that the fingerbank process logic wasn't initiated by the app only manually from the gui , the DEVICE 31 unknown error was already fixed by one of your commits (it was in a legacy default security event) and this pull requests fixes the problem that the fingerbank process call never executes.
 "unless ( $dhcp_filter_rule ) {" will never reach the block even if "$dhcp_filter_rule" is empty

PR #4600, opened 2019-06-10
fix hardcode os

simbazad:
# Description
Fix incorrect hard coded os name


julsemaan:
Merged in devel + cherry-picked in maintenance

Thanks for the fix

PR #4606, opened 2019-06-11
Family Zone firewall sso support

fdurand:
# Description
Added Family Zone firewall sso support.

# Impacts
pfsso

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Family Zone Firewall SSO support


go/firewallsso/familyzone.go
```diff
@@ -0,0 +1,67 @@
...
+func (fw *FamilyZone) startHttp(ctx context.Context, info map[string]string, timeout int) (bool, error) {
+	id, err := uuid.NewUUID()
+	if err != nil {
+		log.LoggerWContext(ctx).Info("Oups too bad")
+	}
+	s := fw.Password + "__" + info["username"] + "_PacketFence_" + id.String()
+	h := sha1.New()
+	h.Write([]byte(s))
+	sha1_hash := hex.EncodeToString(h.Sum(nil))
+	resp, err := fw.getHttpClient(ctx).Get("https://login." + fw.Region + ".linewize.net/login/agent?deviceid=" + fw.DeviceID + "&mac=" + info["mac"] + "&ip=" + info["ip"] + "&username=" + strings.ToLower(info["username"]) + "&agent=PacketFence&hash=" + sha1_hash + "&salt=" + id.String())
```
julsemaan: Would be better to be able to set the full host instead of just the region so that it supports other domains

go/firewallsso/familyzone.go
```diff
@@ -0,0 +1,67 @@
...
+	log.LoggerWContext(ctx).Info("Sending SSO to FamilyZone using HTTP")
+	return fw.startHttp(ctx, info, timeout)
+}
+
+// Send a start to the FamilyZone using the HTTP transport
+// Will return an error if it fails to get a valid reply from it
+func (fw *FamilyZone) startHttp(ctx context.Context, info map[string]string, timeout int) (bool, error) {
+	id, err := uuid.NewUUID()
+	if err != nil {
+		log.LoggerWContext(ctx).Info("Oups too bad")
```
julsemaan: That's an error and the message isn't precise enough nor does it contain the actual error

html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/FamilyZone.pm
```diff
@@ -0,0 +1,85 @@
...
+use pf::config;
+use pf::util;
+use File::Find qw(find);
+
+has_field '+password' =>
+  (
+   type => 'ObfuscatedText',
+   label => 'Secret or Key',
+   tags => { after_element => \&help,
+             help => 'If using the HTTP transport, specify the password for the Palo Alto API' },
```
julsemaan: Palo Alto -> Family Zone

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationFirewalls.js
```diff
@@ -200,6 +200,32 @@ export const pfConfigurationFirewallViewFields = (context) => {
             }
           ]
         },
+        {
+          if: ['FamilyZone'].includes(firewallType),
+          label: i18n.t('Region'),
+          text: i18n.t('Please define the Virtual System number. This only has an effect when used with the HTTP transport.'),
```
julsemaan: virtual system number ?

go/firewallsso/familyzone.go
jrouzierinverse: Should we do a ```go return false, err ``` after logging the error.
Is the UUID return after an error is valid.

jrouzierinverse (changes requested):

go/firewallsso/familyzone.go
fdurand: it shouldn't exist one more domain because linewize.net is the cloud api

PR #4607, opened 2019-06-11
Enable CoA in inline

fdurand:
# Description
Let's say you have an ssid where you configured mac-auth but the target vlan where the device will stand is a inline vlan.
So in this case the location log will contain a mac-auth and an inline entry for the same device.
When packetfence will try to reevaluate the access it will change the ipset set and will not try to deauth the device from the network.

This PR add the ability to send a CoA to the network equipment and to update the ipset.

# Impacts
Inline setup

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Make CoA available for inline setup when there is a mix between radius and inline


extrafu:
@jrouzierinverse Review and merge ASAP.

PR #4611, opened 2019-06-13
BugFixes

simbazad:
# Description
This pull requests fixes multiple bugs:
* Added to the interfaces config in the admin UI the ability to select the "dhcp-listener" Daemon and by that allowing the iplog to continuously update

* Remove the same checks again & again for duplicates types to a single routine .

* Added a safety check ("next if(!defined($network) or !defined($netmask));") to avoid error when calling  "new Net::Netmask($network, $netmask);" when the interface is missing netmask || network like the case of a physical interface with multiple vlans.

* Removed the "return if ( $type =~ /^other$/i );" to solve the following scenario:
    * add an extra listening daemon for an interface with type = Other
    * If we will try to remove the extra listening daemon it will fail because of that irrelevant return statment

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Interface.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Interface.pm
@@ -140,7 +140,7 @@ sub options_additional_listening_daemons {
     my $self = shift;
 
     return map { { value => $_, label => $_ } }
-        qw(portal radius dhcp dns);
+        qw(portal radius dhcp dns dhcp-listener);
 }
 
 =head2 validate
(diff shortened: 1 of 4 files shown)


html/pfappserver/lib/pfappserver/Form/Interface.pm
```diff
@@ -152,52 +152,33 @@ Force DNS to be defined when the 'inline' type is selected
...
-        }
-    }
-    # Remove double dhcp type if exist
-    @types = qw(dhcp);
-    if ( defined $self->value->{type} && any { $_ eq $self->value->{type} } @types ) {
-        my %daemons = map { $_ => 1 } @{$self->value->{additional_listening_daemons}};
-        if ( exists($daemons{'dhcp'}) ) {
-            my $index = firstidx { $_ eq 'dhcp' } @{$self->value->{additional_listening_daemons}};
-            splice @{$self->value->{additional_listening_daemons}}, $index, 1;
+    sub deDup {
```
julsemaan: Move this out of the validate subroutine so that its truly part of the module

simbazad: @julsemaan Done

simbazad:
@julsemaan closing this pull request because of unnecessary merge commit. new one at:https://github.com/inverse-inc/packetfence/pull/4615

PR #4612, opened 2019-06-13
bugfix

simbazad:
# Description

* Fixes the bug that was introduced by this commit: f04c27763d0c53a265b4904cdce6d6be9a1e736e
by removing the filter method the "actions" key was added to the "fields" param for this api endpoint : 
"/api/v1/nodes" which resulted in an error and the entire node search ui breakage .


satkunas:
This bug was fixed with https://github.com/inverse-inc/packetfence/commit/90dc265e76b8a6918182acd7b429209813793b6c

Closing this PR

PR #4614, opened 2019-06-14
Fix Issue 4591

simbazad:
# Description
Fix Issue 4591


julsemaan:
Thanks for the bug fix + explanation in #4591

PR #4615, opened 2019-06-14
Interfaces&dhcpD

simbazad:
# Description

This pull requests fixes multiple bugs:

Added to the interfaces config in the admin UI the ability to select the "dhcp-listener" Daemon and by that allowing the iplog to continuously update

Remove the same checks again & again for duplicates types to a single routine .

Added a safety check ("next if(!defined($network) or !defined($netmask));") to avoid error when calling "new Net::Netmask($network, $netmask);" when the interface is missing netmask || network like the case of a physical interface with multiple vlans.

Removed the "return if ( $type =~ /^other$/i );" to solve the following scenario:

add an extra listening daemon for an interface with type = Other
If we will try to remove the extra listening daemon it will fail because of that irrelevant return statment

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Interface.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Interface.pm
@@ -140,7 +140,7 @@ sub options_additional_listening_daemons {
     my $self = shift;
 
     return map { { value => $_, label => $_ } }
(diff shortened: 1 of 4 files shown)


julsemaan:
tested and it works fine, will not put in maintenance as the change is relatively large

PR #4663, opened 2019-07-12
Development environment can be set up automatically with Ansible and Vagrant

nqb:
# Description
- Adjust minor things in the build process to automate it
- Create a Vagrant environment with:
  * an Ansible inventory to:
    * start virtual machines with specific parameters
    * provision virtual machines based on their Ansible inventory
  * Ansible provisionners to:
    * automate installation of PacketFence dependencies
    * prepare development environment
    * reconfigure PacketFence any time 
  * shell provisionner to build PacketFence from source code
- Documentation of this development environment

# Impacts
Build process

# Delete branch after merge
YES

# Checklist
- [X] Document the feature

# NEWS file entries
## Enhancements
Development environment can be set up automatically with Ansible and Vagrant


addons/vagrant/inventory/hosts
```diff
@@ -0,0 +1,80 @@
...
+              ssh_port: 2600
+              mgmt_ip: 172.17.1.14
+              reg_ip: 192.168.103.10
+              iso_ip: 192.168.104.10
+              ansible_host: 127.0.0.1
+              ansible_port: "{{ ssh_port }}"
+              cpus: 1
+              memory: 8192
+  vars:
+    tz: Europe/Paris
```
julsemaan: default timezone should either be UTC or America/New_York not Europe/:baguette_bread:

addons/vagrant/setup-dev-env.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
julsemaan: I think this would be better in the dev-helpers so that we can use this for other things than for vagrant

nqb:
@julsemaan, fixed. Branch rebased.

PR #4686, opened 2019-07-25
Packer images with Golang environment

nqb:
# Description
- First Packer template to build Docker images that will be use in the pipeline to build PacketFence packages
- Refactor of `setup-go-env.sh` to use it to install Golang environment in Packer images
- Install Golang vendor dependencies in Packer images
- Update `vendor.json` file with latest checksums
- Fix a duplicate dependencies in `vendor.json`

# Impacts
Golang vendor dependencies and `setup-go-env.sh` script

# Delete branch after merge
YES


nqb:
@julsemaan, if possible, could you review and merge quickly ?

Thanks.

nqb:
`github.com/go-sql-driver/mysql` was present twice in `vendor.json`, I remove the oldest one.

nqb:
As discussed with @julsemaan, an improvement could be:

# Packer build
1. copy only `go/vendor/vendor.json` in `$GOPATH/src/github.com/inverse-inc/packetfence` inside Docker image (in place of the whole `go` directory)
2. run `govendor sync`
=> you got all dependencies in `go/vendor/` dir in Docker image

# RPM/Debian builds
1. sync `go` directory (in the build directory) with `go` dir (in `$GOPATH`). We could replace `vendor.json` because files should be the same.
2. change directory to `$GOPATH/src/github.com/inverse-inc/packetfence/go` to run the `go build`
3. move Golang binaries to the build directory

With this approach, `go build` is done in a `$GOPATH` with a vendor directory but we leave the build directory.

EDIT:
1. sync `go` directory (in the build directory) with `go` dir (in `$GOPATH`). We could replace `vendor.json` because files should be the same.
2. run a `make` target from the build directory to put Golang binaries into `sbin` dir

nqb:
I tested carefully changes made by this PR in a fresh environment to be sure this will not break current build system.

nqb:
@fdurand: could you review ?

fdurand:
It looks good, the only thing is why do you update go/vendor/vendor.json ?

nqb:
> It looks good, the only thing is why do you update go/vendor/vendor.json ?

For two reasons:
- I removed useless occurrence of `github.com/go-sql-driver/mysql` (already present)
- Each time I made a `govendor sync`, `vendor.json` has been updated and it should not be the case. According to @julsemaan, this is certainly because `vendor.json` has not been correctly updated in our Git repo.

PR #4734, opened 2019-08-22
Use access duration choices for password rotation period

### LINKED ISSUE #3761: POTD: Rotate on a precise day instead of duration
Instead of rotating on a duration, it would be useful to be able to rotate every sunday or every 1st of the month, etc for better control of the password change

fdurand:
# Description
Use the access duration configuration in the password of the day rotation field.
It add the way to choose a rotation based on the beginning  of the period + a duration (like each day at 6am).

# Impacts
Password of the day source

# Issue
fixes #3761

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Rotate on a precise day instead of duration (#3761)

# UPGRADE file entries
Run addons/upgrade/to-9.2-update-potd.pl

### DIFF
--- a/addons/upgrade/to-9.2-update-potd.pl
+++ b/addons/upgrade/to-9.2-update-potd.pl
@@ -1,0 +1,86 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+Move password_rotation value from Potd source to guests_admin_registration.access_duration_choices 
+
+=cut
+
+=head1 DESCRIPTION
+
+Moved password_rotation value to guests_admin_registration.access_duration_choices
+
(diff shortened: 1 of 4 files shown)


jrouzierinverse:
If someone has an already has a password rotation that does not exist in the 
guests_admin_registration.access_duration_choices updating the source will fail.
My suggestion is to create a script that adds all POTD password rotations to guests_admin_registration.access_duration_choices.

extrafu:
@jrouzierinverse Review one last time and merge.

PR #4748, opened 2019-08-27
feature/config_raw

jrouzierinverse:
Description
===========
Return the raw config entry to avoid the over head of cleaning up items

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


extrafu:
I don't know if it's too optimistic to have that in v9.1.

julsemaan:
Performance doesn't seem to have improved with this.

@jrouzierinverse will investigate

jrouzierinverse:
The performance has improved the issue was the raw option was not passed properly.

Issue #4776, opened 2019-09-04
ipset upgrade cause issues when installing PacketFence


nqb:
When you try to install latest builds of PacketFence under CentOS 7:
```
    Error: Package: ipset-symlink-1.0.1-3.1.x86_64 (packetfence-devel)                                                                         
               Requires: ipset >= 7.1                                                                                            
               Installed: ipset-6.38-3.el7_6.x86_64 (@koji-override-1)                                                        
                   ipset = 6.38-3.el7_6                                                                                                
               Available: ipset-6.38-2.el7.x86_64 (base)                                                                             
                   ipset = 6.38-2.el7
```

**Expected behavior**
It should be possible to install latest builds with ipset-6.38-3

nqb:
Related to #4732, same issue if you try to install PF 9.0.1 on CentOS 7.

nqb:
@fdurand is building a new package. Will be uploaded in a few minutes.

nqb:
I still got an issue with new package:
```
Transaction check error:
  file /usr/lib64/libipset.so.11 from install of ipset-symlink-1.0.1-4.1.x86_64 conflicts with file from package ipset-libs-6.38-3.el7_6.x86_64

Error Summary
```

nqb:
CentOS 7.7 has been released today with new version of ipset. Something to have in mind for our 9.1 release.

julsemaan:
I'm tired of this ipset bull***t, I'm adding ipset to our repos and pinning it

shanehull:
Yep, I'm getting:

```
Error: Package: packetfence-9.1.0-1.el7.noarch (packetfence)
           Requires: ipset = 6.38
           Installed: ipset-7.1-1.el7.x86_64 (@anaconda)
               ipset = 7.1-1.el7
           Available: ipset-6.38-3.el7_6.x86_64 (packetfence)
               ipset = 6.38-3.el7_6
```

And 6.38 doesn't appear to be available.

nqb:
Issue presents in nightly builds too.

nqb:
### Current state

#### CentOS 7.6 with PF 9.1
Only `ipset-symlink` has been installed from packetfence repo:
```console
# yum list ipset*
[..]
Installed Packages
ipset.x86_64                                                                                         6.38-3.el7_6                                                                                  @koji-override-1
ipset-libs.x86_64                                                                                    6.38-3.el7_6                                                                                  @koji-override-1
[...]

PR #4821, opened 2019-09-19
Adjust dependencies for CentOS 7.7

### LINKED ISSUE #4820: Issues with CentOS 7.7
See #4776

nqb:
# Description
Identical to #3831

# Impacts
- Golang binaries
- RPM packages

# NEW Package(s) required
RPM packages

# Issue
fixes #4820 #4776 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
Unable to install PacketFence 9.1 under CentOS 7.7

### DIFF
--- a/addons/packages/packetfence.spec
+++ b/addons/packages/packetfence.spec
@@ -140,7 +140,7 @@ requires: perl(Const::Fast)
 # Perl core modules but still explicitly defined just in case distro's core perl get stripped
 Requires: perl(Time::HiRes)
 # Required for inline mode.
-Requires: ipset = 6.38, ipset-symlink
+Requires: ipset >= 7.1, ipset-symlink
 Requires: sudo
 Requires: perl(File::Which), perl(NetAddr::IP)
 Requires: perl(Net::LDAP)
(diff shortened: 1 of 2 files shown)


nqb:
Not able to build `pfhttpd` with ipset 7.1, error message on `pfbuilder`:
```
# github.com/inverse-inc/packetfence/go/vendor/github.com/inverse-inc/go-ipset
../../../vendor/github.com/inverse-inc/go-ipset/ipset.go:161:2: could not determine kind of name for C.ipset_envopt_parse
```
We certainly need to adapt https://github.com/inverse-inc/go-ipset to support this new version like we did in the past:
* https://github.com/inverse-inc/go-ipset/commit/0a0972e5af167061ebab363a87b643dc508d9b43
* https://github.com/inverse-inc/go-ipset/commit/3b787d33ec19ef4e074fed7fb7eacb86c4744261

or take a look at https://github.com/digineo/go-ipset/tree/v2 as suggested by @salted in https://github.com/digineo/go-ipset/pull/3

addons/packages/packetfence.spec
```diff
@@ -140,7 +140,7 @@ requires: perl(Const::Fast)
 # Perl core modules but still explicitly defined just in case distro's core perl get stripped
 Requires: perl(Time::HiRes)
 # Required for inline mode.
-Requires: ipset = 6.38, ipset-symlink
+Requires: ipset >= 7.1, ipset-symlink
```
julsemaan, line 143: As discussed in the internal email, this is a bad idea because:

- It prevents us from releasing maintenance binaries for previous PF versions
- It needs to be pinned `= 7.1` if we decide to upgrade it so that an OS update will not break the binaries on the system

julsemaan (changes requested):

PR #4832, opened 2019-09-26
Migrate to go mod for Golang binaries

julsemaan:
# Description
This migrates to `go mod` + Golang 1.13 for all our Golang binaries.
Ignore the branch name, I'm not upgrading caddy in this PR to keep it !@lzammit (lean)

# Impacts
All our golang binaries (pfdns, pfdetect, pfdhcp, pfhttpd, pfdns)

# Delete branch after merge
NO

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -69,7 +69,7 @@ all: pfhttpd pfdhcp pfdns pfstats pfdetect
 
 .PHONY: copy
 copy:
-	cp -f pfhttpd pfdhcp pfdns pfstats pfdetect $(SBINDIR)
+	cp -f pfhttpd pfdhcp pfdns pfstats pfdetect $(DESTDIR)$(SBINDIR)
 
 .PHONY: clean-coredns-src
 clean-coredns-src:


julsemaan:
I was able to build on pfbuilder with this branch (I put golang 1.13 temporarily on it and then removed it)

We'll need to see if @nqb's new build system will work correctly with this or if it will need adjustments

nqb:
`GOVERSION` need to be adjust in `go/Makefile`

nqb:
@julsemaan, in order to make adjustements, I will need some answers:

1. Could you provide Golang documentations you follow to make this PR ?
2. If I understand correclty this PR, download of Golang dependencies can be done outside a `GOPATH` (not the case with `govendor` approach). In which place Golang dependencies should be put if we want to download them **before** running the build and use it during the build ?

Thanks.

julsemaan:
1. https://blog.golang.org/using-go-modules
2. https://stackoverflow.com/questions/52082783/how-do-i-find-the-go-module-source-cache

jrouzierinverse (changes requested):
Tested it and it works!
The only thing I could see is to add golang-1.13 as a build dependency.

nqb:
>  The only thing I could see is to add golang-1.13 as a build dependency.

@jrouzierinverse, I will prefer to keep golang out of SPEC files for now because Debian Buster and Stretch doesn't provide golang-1.13 as packages. Consequently, we need to install it from upstream for Debian. Using the same installation method on CentOS is more clear from my point of view.

jrouzierinverse:
I would prefer to have one single place to define the build dependencies for our packages.
Instead of having it split between a README file and the spec files.

This does force us to build non-standard packages.
But it does makes it easier to have reproducible builds and reduce those but it worked on my machine scenarios.

For example to set up a machine for PacketFence development or to build a package.
To could be boiled down to the following commands more or less.

Debian
```bash
apt install devscripts
mk-build-deps --install --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes' debian/control
```
Centos
```bash
yum install yum-utils
yum-builddep --define="el$EL_VERSION 1" --define="builddoc 0" addons/packages/packetfence.spec
```

nqb:
> I would prefer to have one single place to define the build dependencies for our packages.
> Instead of having it split between a README file and the spec files.

I completely agree. For new build system, I rely on `setup-go-env.sh` script to set same Golang versions on CentOS and Debian: https://github.com/inverse-inc/packetfence/blob/bf2c8b848dbfc26f6162e135af6af7933cacfcaa/ci/packer/buildpkg.json#L99-L104

Regarding auto installation of build dependencies for each distro, I do exactly the same with `ci-build-pkg`: https://github.com/inverse-inc/packetfence/blob/bf2c8b848dbfc26f6162e135af6af7933cacfcaa/ci/packer/buildpkg.json#L90-L96

And it works. You can get Docker images for each distro: https://hub.docker.com/u/inverseinc

When you want to have **same** software versions on all distros, you don't really have the choice to install from upstream (like we currently do for Golang) and having false SPEC files. But if you provide ready-to-use images to run build with these dependencies, I don't see the problem. I do the same for asciidoctor version 2 (see https://github.com/inverse-inc/packetfence/blob/feature/packer-build-deps/ci/packer/provisionners/playbooks/rubygems.yml).

addons/dev-helpers/setup-go-env.sh
```diff
@@ -50,19 +50,6 @@ else
...
-    setup
-    declare -p GOPATH GO_REPO
-    mkdir -v -p $GOPATH/src/$GO_REPO
-    if [ -d $GOPATH/src/$GO_REPO/go ]; then
-        die "Directory $GOPATH/src/$GO_REPO/go already exists, cannot symlink it to /usr/local/pf/go"
-    else
-        ln -s -v /usr/local/pf/go $GOPATH/src/$GO_REPO/go
-    fi
-fi
+cd /usr/local/pf/go
```
nqb: - You expect to be on a system where PF is installed. When building images, PF is not installed and I ẁill only copy `go.mod` file in temp location. We should use an environment variable here with a default value that can be override. We could still use `GO_REPO` and adapt its value.
- Some parts of the script can be drop because they will not be used.

nqb:
I made some changes to this PR.

Another thing I would like to do is to convert `go/README.md` to AsciiDoc and put this file under `docs` with a symlink in `go` directory. That's way we can integrate this README directly in our Developer's Guide and it is still visible in GitHub when you browse `go` directory.

nqb (approved):

nqb:
I finished my work on this PR. Let me know if there is something to adjust.

I'm working on setup to build maintenance binaries before 9.2 with go1.9.3 and futur maintenance binaries with go1.13.1.

nqb:
@fdurand: could you review and if all is fine merge this PR ? Thanks.

julsemaan:
Just as a side note, we will ABSOLUTELY need the new build system to do the release of 9.2 and the new build system to publish the signed maintenance binaries.

@nqb, could you confirm we are in that state right now?

nqb:
> @nqb, could you confirm we are in that state right now?

Yes !

nqb:
> Just as a side note, we will ABSOLUTELY need the new build system to do the release of 9.2 and the new build system to publish the signed maintenance binaries.

Because of the several PRs involved to accomplish this goal, I will need your help in order to make quick reviews.

extrafu:
@fdurand and @jrouzierinverse please review quickly - we need that in ASAP for the new build system.

jrouzierinverse:
rebase

julsemaan:
I think this isn't needed anymore since @nqb took the work of this PR and put it in #4841 

He can confirm and close this PR

jrouzierinverse:
I am waiting in Julien's response to https://github.com/inverse-inc/packetfence/pull/4832#issuecomment-535975733 .
Edit nevermind

nqb:
@julsemaan,

I expect this PR was merged **before** #4841 in order to rebase #4841 on devel after merge to avoid such situation. History will be hard to understand.

I rebased this one on devel and I will merge it now.

PR #4834, opened 2019-09-30
feature/admin_api_audit_log

jrouzierinverse:
Description
===========
Admin API Audit Log

Impacts
=======
pfperl after dispatching

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests
- [x] Add acceptance tests (TestLink)


lib/pf/UnifiedApi/Controller/Config.pm
```diff
@@ -1075,6 +1078,77 @@ sub fix_permissions {
...
+    my ($self) = @_;
+    my $stash = $self->stash;
+    my $status =  $self->res->code;
+    my $req = $self->req;
+    my $method = $req->method;
+    my $path = $req->url->path;
+    my $request = $self->make_audit_record_request();
+
+    my $current_user = $stash->{current_user};
+    return {
```
julsemaan, line 1111: You should also get the current tenant in this payload

lib/pf/UnifiedApi/Controller/Config/Bases.pm
```diff
@@ -81,6 +81,14 @@ sub test_smtp {
     return $self->render(json => { message => 'Testing SMTP success' });
 }
 
+=head2 fields_to_mask
+
+fields_to_mask
+
+=cut
+
+sub fields_to_mask { qw(smtp_password pass galera_replication_password) }
```
julsemaan, line 90: You should also go through the fields of the scan engines and the firewall sso entries

lib/pf/UnifiedApi/Controller/Config.pm
jrouzierinverse: It is automatically added.

lib/pf/UnifiedApi/Controller/Config/Bases.pm
jrouzierinverse: By default, any field named password will get masked.
Will verify if the scan engines have other fields that need to be masked.

jrouzierinverse: They are good.

jrouzierinverse:
I reviewed and answered issues brought up by @julsemaan.

extrafu:
@julsemaan Please update the NEWS file

julsemaan:
@jrouzierinverse, can you provide the NEWS entry in your PR description next time so that the merge procedure handles that

PR #4840, opened 2019-10-03
feature/switch_supports

jrouzierinverse:
Description
===========
Expose what each switch supports via the type attributes.

Impacts
=======
Switches

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
+++ b/html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
@@ -201,6 +201,7 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
     id = null,
     isNew = false,
     isClone = false,
+    advancedMode = false,
     options: {
       meta = {}
     },
@@ -211,6 +212,20 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
   const placeholder = (key = null) => {
     const { [key]: { placeholder = null } = {} } = meta
     return placeholder
+  }
+
+  const switchTypeSupport = (options) => {
+    const { type: { allowed = [] } = { } } = meta
+    return advancedMode || allowed.find(group => {
+      return group.options.find(switche => {
+        if (switche.value === form.type) {
+          return switche.supports.find(option => {
+            // Return true if the switch model supports *any* of the specified options
+            return options.includes(option)
+          })
+        }
+      })
+    })
   }
 
   return [
@@ -265,7 +280,8 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
                 ...{
                   groupLabel: 'group',
                   groupValues: 'options'
-                }
+                },
+                optionsLimit: 1000
               },
               validators: pfConfigurationValidatorsFromMeta(meta, 'type', i18n.t('Type'))
             }
@@ -328,6 +344,7 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
         {
           label: i18n.t('External Portal Enforcement'),
           text: i18n.t('Enable external portal enforcement when supported by network equipment.'),
+          if: switchTypeSupport(['ExternalPortal']),
           fields: [
             {
               key: 'ExternalPortalEnforcement',
@@ -359,6 +376,7 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
         {
           label: i18n.t('VoIPLLDPDetect'),
           text: i18n.t('Detect VoIP with a SNMP request in the LLDP MIB.'),
+          if: switchTypeSupport(['Lldp']),
           fields: [
             {
               key: 'VoIPLLDPDetect',
@@ -375,6 +393,7 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
         {
           label: i18n.t('VoIPCDPDetect'),
           text: i18n.t('Detect VoIP with a SNMP request in the CDP MIB.'),
+          if: switchTypeSupport(['Cdp']),
           fields: [
             {
               key: 'VoIPCDPDetect',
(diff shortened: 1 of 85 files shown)


julsemaan:
As discussed with @cgx, he will add a way to show all the fields for a switch entry if needed in case there was a mistake or a field was omitted

extrafu:
@jrouzierinverse Why assign it to me?

jrouzierinverse:
@extrafu because you asked for the feature ¯\_(ツ)_/¯

extrafu:
@jrouzierinverse Please rebase.

jrouzierinverse:
Done

jrouzierinverse:
done

julsemaan (approved):
I've checked the Perl code and it looks good to me.

Probably a good idea to have @satkunas review the JS code @cgx wrote just to make sure although it looks fairly straightforward to me

extrafu:
@satkunas Please review the JS code. Merge if all fine.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
```diff
@@ -213,6 +214,20 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
     return placeholder
   }
 
+  const switchTypeSupport = (options) => {
+    const { type: { allowed = [] } = { } } = meta
+    return advancedMode || allowed.find(group => {
+      return group.options.find(switche => {
+        if (switche.value === form.type) {
+          return switche.supports.find(option => {
```
satkunas, line 222: Javascript `.find()` only returns the first match. If you need to return all matches use `.filter()`. Additionally `find` will return `undefined` if there are no matches and `filter` will return an empty array if there are no matches.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
```diff
@@ -213,6 +214,20 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
...
+            // Return true if the switch model supports *any* of the specified options
+            return options.includes(option)
+          })
+        }
+      })
```
satkunas, line 227: Line 220 should return a boolean, so this line should compare length with `})` => `}).length > 0`

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
```diff
@@ -213,6 +214,20 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
...
+    })
```
satkunas, line 228: Line 219 should return a boolean, so this line should compare length with `})` => `}).length > 0`

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
```diff
@@ -213,6 +214,20 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
     return placeholder
   }
 
+  const switchTypeSupport = (options) => {
+    const { type: { allowed = [] } = { } } = meta
+    return advancedMode || allowed.find(group => {
```
satkunas, line 219: Javascript .find() only returns the first match. If you need to return all matches use .filter(). Additionally find will return undefined if there are no matches and filter will return an empty array if there are no matches.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
```diff
@@ -484,9 +507,10 @@ export const pfConfigurationSwitchGroupViewFields = (context = {}) => {
     {
       tab: i18n.t('Roles'),
       fields: [
-        { label: i18n.t('Role mapping by VLAN ID'), labelSize: 'lg' },
+        { label: i18n.t('Role mapping by VLAN ID'), labelSize: 'lg', if: switchTypeSupport(['RadiusDynamicVlanAssignment']) },
```
satkunas: Reorder so the if statement is first so conditions are more visible. Apply to all additions below and in other files.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitches.js
```diff
@@ -219,6 +220,20 @@ export const pfConfigurationSwitchViewFields = (context = {}) => {
     return placeholder
   }
 
+  const switchTypeSupport = (options) => {
```
satkunas, line 223: Refactor function's `find` and `length` same as above.

satkunas (commented):
Review added. Some JS modifications needed.

julsemaan:
@cgx, can you adjust the code based on @satkunas comments

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
cgx: As commented in the code, this function returns true if *any* of the `options` is supported by the selected switch, therefore `find` is correct and will make the function faster. `undefined` will correctly be evaluated as `false` in a boolean context while an an empty array would be `true`.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
cgx: Not relevant since we use `find`.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
cgx: Not relevant since we use `find`.

html/pfappserver/root/static.alt/src/globals/configuration/pfConfigurationSwitchGroups.js
cgx: `find` is what we want.

PR #4848, opened 2019-10-10
feature/switch_template

jrouzierinverse:
Description
===========
Allow switches to be defined via configuration

Impacts
=======
Switch Configuration

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/docs/developer/template_switches.asciidoc
+++ b/docs/developer/template_switches.asciidoc
@@ -60,20 +60,41 @@ Functions: ${f1("",$var, f2())}
 
 ==== Available variables for RADIUS scope
 
+  * acceptRole - (Same as acceptVlan)
+  * reject - (Same as acceptVlan)
   * acceptVlan
-    * vlan
+    * autoreg
+    * connection_sub_type
+    * connection_type
+    * eap_type
+    * fingerbank_info
+    * ifDesc
+    * ifIndex
+    * isPhone
     * mac
+    * nas_port_id
+    * nas_port_type
+    * node_info
+    * profile
+    * radius_request
+    * realm
+    * session_id
+    * source_ip
+    * ssid
+    * stripped_user_name
+    * switch
+    * switch_ip
+    * switch_mac
+    * user_name
+    * user_role
+    * wasInline
+    * vlan
   * coa
-    * vlan
     * mac
-    * role
-  * acceptRole
     * role
     * mac
   * disconnect
     * disconnectIp
-    * mac
-  * reject
     * mac
 
 ==== Available functions
--- a/docs/developer/template_switches.asciidoc
+++ b/docs/developer/template_switches.asciidoc
@@ -1,0 +1,118 @@
+// to display images directly on GitHub
+ifdef::env-github[]
+:imagesdir: ../images
+endif::[]
+
+Creating a new Switch via a Template
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To create a new Switch Template you must create a template file in the /usr/local/pf/lib/pf/Switch directory.
+The file must have the following pattern /usr/local/pf/lib/pf/Switch/<Vendor>/<SwitchName>.def
+
+For example to create a template for the vendor "Cyberdyne" and the switch "Switchinator 800".
+The file name is called /usr/local/pf/lib/pf/Switch/Cyberdyne/Switchinator800.def
+The file name must must only consist of alphanumeric characters and under scores and must begin with a letter.
+
+The file consist of parameter names and their values.
+
+Parameters required:
+
+* description      - The description of the switch.
+* radiusDisconnect - The RADIUS disconnect methods to use. Must be one of the following values coa|disconnect|coaOrdisconnect .
+
+Parameters for set of RADIUS attributes
+
+* acceptVlan - Attributes for setting up the vlan.
+* acceptRole - Attributes for setting up the role.
+* reject     - Attributes sent on rejection.
+* disconnect - Attributes sent on disconnect (required if radiusDisconnect is 'disconnect' or 'coaOrdisconnect')
+* coa        - Attributes sent on CoA (required if radiusDisconnect is 'coa' or 'coaOrdisconnect')
+
+Comments
+The line of a comment must begin with a '#'
+For example:
+# This is a comment
+
+Defining RADIUS Attributes.
+
+name = <<EOT
+Attribute-Name1 = value1
+Attribute-Name2 = value2
+EOT
+
+RADIUS Vendor Attributes for CoA
+name = <<EOT
+VendorName:Attribute-Name1 = value1
+VendorName:Attribute-Name2 = value2
+EOT
+
+Dynamic RADIUS Attribute Value Syntax
+
+Some values depends on the context of the current request.
+So here is mini templating language to format values.
+
+text replacement:  $name
+embedded text replacement:  start-${name}-end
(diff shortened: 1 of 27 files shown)


docs/developer/template_switches.asciidoc
```diff
@@ -0,0 +1,118 @@
+// to display images directly on GitHub
+ifdef::env-github[]
+:imagesdir: ../images
+endif::[]
+
+Creating a new Switch via a Template
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nqb: Please use "=" syntax for section titles, see https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-titles

docs/developer/template_switches.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+
+coa = <<EOT
+Calling-Station-Id= ${macToEUI48($mac)}
+NAS-IP-Address = $disconnectIp
+Cisco:Cisco-AVPair = subscriber:command=bounce-host-port
+EOT
+....
+
+Contributing
+~~~~~~~~~~~~
```
nqb: Please use "=" syntax for section titles, see https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-titles

nqb (changes requested):
`docs/developer/template_switches.asciidoc` need to follow documentation conventions, see https://packetfence.org/doc/PacketFence_Developers_Guide.html#_documentation_conventions.

julsemaan:
@jrouzierinverse 

I added a fix in the PR to have the form consider the template switches as valid types since I wasn't able to save a switch with a templated type in the old admin

lib/pf/Switch/Template.pm
```diff
@@ -0,0 +1,364 @@
...
+use pf::constants;
+use pf::util qw(isenabled);
+use pf::constants::role qw($REJECT_ROLE);
+use pf::access_filter::radius;
+use pf::roles::custom;
+use pf::mini_template;
+
+our $DISCONNECT_TYPE_COA = 'coa';
+our $DISCONNECT_TYPE_DISCONNECT = 'disconnect';
+our $DISCONNECT_TYPE_BOTH = 'coaOrdisconnect';
```
julsemaan: Should be coaOrDisconnect

docs/developer/template_switches.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+=== Dynamic RADIUS Attribute Value Syntax
+
+Some values depends on the context of the current request.
+So here is mini templating language to format values.
+
+Text replacement:  $name
+Embedded text replacement:  start-${name}-end
+Functions: ${f1("",$var, f2())}
+
+==== Available variables for RADIUS scope
```
julsemaan, line 65: It would be useful to have access to the switch, node_info and other args part of the AAA context in here so we can do fancy stuff if we'd like

docs/developer/template_switches.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+    * vlan
+    * mac
+  * coa
+    * vlan
+    * mac
+    * role
+  * acceptRole
+    * role
+    * mac
+  * disconnect
```
julsemaan: In this scope we will need to have the last accounting data fetched through pf::accounting::node_accounting_dynauth_attr

jrouzierinverse: What name do you want to give that variable?

julsemaan: `last_accounting` ?

docs/developer/template_switches.asciidoc
jrouzierinverse: done

docs/developer/template_switches.asciidoc
jrouzierinverse: done

lib/pf/Switch/Template.pm
jrouzierinverse: done

docs/developer/template_switches.asciidoc
jrouzierinverse: Done

jrouzierinverse:
Ready for review

nqb (changes requested):
@jrouzierinverse, you should review AsciiDoc markup in `template_switches.asciidoc`.

Try to generate HTML and see resulting document.

jrouzierinverse:
@nqb Which version of asciidoctor are you using?
I cannot even build the HTML docs on my setup.

nqb:
@jrouzierinverse, I'm using asciidoctor 2.0.10.

You need to install it using `gem install asciidoctor` because this version is not available as a package.

jrouzierinverse:
Fixed the formatting

docs/developer/template_switches.asciidoc
```diff
@@ -38,43 +38,70 @@ For example:
...
+   ** last_accounting
+   ** mac
+   ** role
   * disconnect
-    * disconnectIp
-    * mac
-  * reject
-    * mac
+   ** disconnectIp
+   ** last_accounting
```
julsemaan: Please add the available fields of last_accounting

docs/developer/template_switches.asciidoc
```diff
@@ -38,43 +38,70 @@ For example:
...
-    * mac
+   ** autoreg
+   ** connection_sub_type
+   ** connection_type
+   ** eap_type
+   ** fingerbank_info
+   ** ifDesc
+   ** ifIndex
+   ** isPhone
+   ** last_accounting
```
julsemaan: Add the sub fields here too

docs/developer/template_switches.asciidoc
```diff
@@ -38,43 +38,70 @@ For example:
...
+   ** mac
+   ** nas_port_id
+   ** nas_port_type
+   ** node_info
```
julsemaan: Add the sub fields here too

docs/developer/template_switches.asciidoc
julsemaan: Also, it would be useful if your example configuration contained a nested field so we can have an example on how to properly use them

nqb:
Regarding @julsemaan's comments, I wonder if AsciiDoc documentation is the right place to add such details. Maintaining a list like this would certainly be painful and error prone. I suggest to include, if possible, this list directly from PacketFence's sources. With this design, when we updated the code, we also updated documentation.

AsciiDoc allow to include any source file with following syntax:
```asciidoc
[source,ruby]
----
include::app.rb[]
----
```
@jrouzierinverse, what's your point of view ?

jrouzierinverse:
@nqb it makes sense we should share this information with the VLAN filters docs.
I would want to write a script to extract this from code.
I would create a separate issue for this.
So if we do not have time for this release.
We can do it for the next release.

jrouzierinverse:
@julsemaan Updated the doc can you review to see if I added too much or not enough.

julsemaan (approved):
@jrouzierinverse 

That looks good on me, @nqb still has a lock on that PR I'll let him remove it and we can merge afterwards

nqb (approved):

extrafu:
@julsemaan Please adjust the NEWS file

julsemaan:
@jrouzierinverse, can you provide the NEWS entry in your PR description next time so that the merge procedure handles that

PR #4860, opened 2019-10-16
Set a default timeout for ipset inline set

fdurand:
# Description
Added timeout value on ipset set to auto clean the set.

# Impacts
Inline setup

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Auto clean ipset with a timeout value based on the dhcp configuration


lzammit:
Works.

lzammit (commented):
All is ok!

PR #4864, opened 2019-10-17
CLI-Access connection profile filter.

fdurand:
# Description
Added connection profile filter for CLI-Access.
Allow switch and switch_group in authentication/administration rules.

# Impacts
CLI-Access

# Delete branch after merge
YES
 | NO

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* New connection profile filter for CLI-Access. Switch and switch_group in authentication/administration rules.

### DIFF
--- a/lib/pf/Authentication/Source.pm
+++ b/lib/pf/Authentication/Source.pm
@@ -97,7 +97,7 @@ sub common_attributes {
           { value => 'computer_name', type => $Conditions::SUBSTRING },
           { value => "mac", type => $Conditions::SUBSTRING },
           { value => "realm", type => $Conditions::SUBSTRING },
-          { value => "switch", type => $Conditions::SUBSTRING },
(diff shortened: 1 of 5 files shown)


lib/pf/Authentication/Source.pm
```diff
@@ -97,7 +97,9 @@ sub common_attributes {
           { value => 'computer_name', type => $Conditions::SUBSTRING },
           { value => "mac", type => $Conditions::SUBSTRING },
           { value => "realm", type => $Conditions::SUBSTRING },
-         ];
+          { value => "switch", type => $Conditions::SUBSTRING },
```
rouzier: The switch in $args of pf::radius::switch_access is an object and cannot be used for string operations.
I would suggest that you use switch_id or switch_ip for this and fill it in as you did with switch_group.

rouzier (changes requested):

PR #4869, opened 2019-10-22
feature/bulk_import

### LINKED ISSUE #4865: API: Import Bulk Nodes/Users endpoint needed
The front-end requires 2 new bulk endpoints from the perl back-end:
- nodes
- users (persons)

This endpoint will be responsible to handle bulk import from CSV.

Expected request payload:
```json
{
  "items": [
    {
      "mac": "00:00:00:00:00:00",
      "owner": "default",
      "computer_name": "foobar"
    },
    { ... },
    { ... }
  ],
  "stopOnFirstError": true|false,
  "ignoreUpdateIfExists": true|false,
  "ignoreInsertIfNotExists": true|false
}
```

The endpoint should:
- process the `items` in the same order given in the request payload.
  - when `stopOnFirstError: true` stop execution on the first error and auto-fill all subsequent items in the response with a custom status code to indicate that they were skipped. Alternatively when `stopOnFirstError: false` do not stop execution on the first error (or any error).
- determine if the node or user already exists/not-exists, and should update/insert respectively.
  - if `ignoreInsertIfNotExists' do NOT insert a new item if the item does not already exist.
  - if `ignoreUpdateIfExists` do NOT update an existing item if the item already exists.
  - include a `true`/`false` boolean in the response payload to indicate if the resource already existed, (SQL update) or is new (SQL insert). (eg: `isNew`: `true`|`false`)
- only insert/update the key/value pairs that are in the request payload
  - perform server-side validation on each key/value pair (if required)
- return a response in the exact same order as the request (Array)
  - include the request data in the response
  - always return the resource id (primary key) in the response data
  - include any server-side errors in the response
- ideally the request payload size should be unlimited, however, please indicate what payload size (10/100/1000) would be most ideal to avoid excessive memory use.

Expected response payload:
```json
{ 
  "status": 200,
  "items": [
    {
      "status": 400,
      "item": {
        "mac": "00:00:00:00:00:00",
        "owner": "default",
        "computer_name": "foobar"
      },
      "errors": [
         { "field":"mac", "message": "Mac address is invalid." },
         { "field":"owner", "message": "Owner does not exist."}
      ],
      "isNew": false,
      "isError": true
    },
    { ... },
    { ... }
  ]
}
```

jrouzierinverse:
Description
===========

Added new endpoints for bulk import for users/nodes

* /api/v1/nodes/bulk_import
* /api/v1/users/bulk_import

Issue
=====

Fixes #4865

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/lib/pf/UnifiedApi.pm
+++ b/lib/pf/UnifiedApi.pm
@@ -527,7 +527,7 @@ sub setup_api_v1_nodes_routes {
 
     $resource_route->register_sub_actions({
         method => 'POST',
-        actions => [ qw( register deregister restart_switchport reevaluate_access apply_security_event close_security_event fingerbank_refresh park unpark)],
+        actions => [ qw( register deregister restart_switchport reevaluate_access apply_security_event close_security_event fingerbank_refresh park unpark bulk_import) ],
         auditable => 1,
     });
 
--- a/lib/pf/UnifiedApi.pm
+++ b/lib/pf/UnifiedApi.pm
@@ -527,7 +527,7 @@ sub setup_api_v1_nodes_routes {
 
     $resource_route->register_sub_actions({
         method => 'POST',
-        actions => [ qw( register deregister restart_switchport reevaluate_access apply_security_event close_security_event fingerbank_refresh park unpark bulk_import) ],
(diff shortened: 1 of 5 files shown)


jrouzierinverse:
@satkunas review the pull request

PR #4877, opened 2019-10-29
Multiple Input element w/ variable separator

### LINKED ISSUE #2616: Multiple LDAP servers should be configurable from the GUI
We already support multiple LDAP servers in the LDAP source.
It would be nice if it was user discoverable, for example by having each server on one line with a + button.

satkunas:
# Description
New VueJS component that re-uses `Vue-multiselect` chosen element to allow multiple items in a single input element and proxies the Array with any separator (default: `,`) to provide CSV to the form.

# Issue
Fixes #2616

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests
- [n//a] Add acceptance tests (TestLink)

## Bug Fixes
* Multiple LDAP servers should be configurable from the GUI (#2616)


nqb:
It works as expected:
Capture d’écran de 2019-10-30 14-19-37

I got following content in `authentication.conf`:
```ini
[inverse]
write_timeout=5
usernameattribute=sAMAccountName
host=10.1.1.1,10.1.1.2
[..]
```
Content is also correct if I use API.

@satkunas, let me know if I need to check other things.

nqb (approved):

html/pfappserver/root/static.alt/src/components/pfFormInputMultiple.vue
```diff
@@ -0,0 +1,300 @@
...
+    }
+  },
+  methods: {
+    onFocus (event) {
+      this.focus = true
+    },
+    onBlur (event) {
+      this.focus = false
+    },
+    focus () {
```
cgx, line 130: It is confusing that have an attribute and a method with the same name.

html/pfappserver/root/static.alt/src/components/pfFormInputMultiple.vue
```diff
@@ -0,0 +1,300 @@
...
+      this.$refs.input.focus()
+    },
+    addTag (newTag) {
+      this.$set(this, 'inputValue', [ ...this.inputValue, ...[{ text: newTag, value: newTag }] ])
+    }
+  }
+}
+</script>
+
+<style lang="scss">
```
cgx, line 140: I propose to move common styling of components based on vue-multiselect to `src/styles/_multiselect.scss` to avoid style duplication in `pfFormInputMultiple.vue` and `pfFormChosen.vue`.

PR #4887, opened 2019-11-02
Juniper EX2300 CoA Port Bounce

lyubomirtraykov:
# Description
Perform CoA Port-Bounce on Juniper EX2300 Switch 

# Impacts
(REQUIRED)
Port Bounce method

# Delete branch after merge
YES

# Checklist
- [n/a ] Document the feature
- [ yes] Add unit tests
- [ n/a] Add acceptance tests (TestLink)

# NEWS file entries
## Enhancements
* Enhance Juniper EX2300


lib/pf/Switch/Juniper/EX2300.pm
```diff
@@ -105,6 +105,72 @@ sub radiusDisconnect {
...
+    }
+
+    if (!defined($self->{'_radiusSecret'})) {
+        $logger->warn(
+            "Unable to perform RADIUS CoA-Request on $self->{'_id'}: RADIUS Shared Secret not configured"
+        );
+        return;
+    }
+
+    $logger->info("boucing $mac using RADIUS CoA-Request method");
```
julsemaan: boucing -> bouncing

julsemaan (changes requested):
Just a minor logging detail to fix and then its ready to merge for 9.2

lyubomirtraykov:
Using _radiusBounceMac is better because we don't need to get the mac from the  ifIndex. Will the _radiusBounceMac be called from the GUI?

julsemaan:
From what I can tell _radiusBounce isn't being called anywhere so you need to use setAdminStatus like in the previous version of your PR or else it won't do anything

lib/pf/Switch/Juniper/EX2300.pm
julsemaan: You still have boucing instead of bouncing

PR #4906, opened 2019-11-08
[packer] support docker multi-tags

nqb:
# Description
- Multi-tag a Docker image if `DOCKER_EXTRA_TAG` is set *and* if `PACKER_EXTRA_ARG` set to null

Needed to tag a Docker image release with two tags: `maintenance/X.X` and `vX.X.X`

- Add `REGISTRY_USER` environment variable to simplify test outside CI

# Impacts
Build of Docker images

# Delete branch after merge
YES

- [Pipeline triggered](https://gitlab.com/inverse-inc/packetfence/-/jobs/345562346) without default values : work as expected
- [Pipeline triggered](https://gitlab.com/inverse-inc/packetfence/-/jobs/345586558) with PACKER_EXTRA_ARG as empty value: in progress

### DIFF
--- a/ci/packer/Makefile
+++ b/ci/packer/Makefile
@@ -4,24 +4,44 @@
 # - command variables (make VAR=VALUE target)
 
 include ../../config.mk
+
+#==============================================================================
+# Packer command line variables
+#==============================================================================
 ACTIVE_BUILDS = 'pfbuild-centos-7,pfbuild-stretch'
+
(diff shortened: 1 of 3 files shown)


nqb:
Second use case is not working as expected, `packer` push only latest tag to Docker registry. I certainly need to use docker-push post-processor before using docker-tag post-processor a second time.

PR #4929, opened 2019-11-13
Realm match default

fdurand:
# Description
If the realm is not defined in the configuration then packetfence use the default realm to match the sources associated on the portal.

# Impacts
Source selection

# Issue
Return the authentication sources where the default realm has been associated if the realm used by the connection contain a realm that is not defined in the configuration.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Return the authentication sources where the default realm has been associated if the realm used by the connection contain a realm that is not defined in the configuration.

### DIFF
--- a/lib/pf/config/util.pm
+++ b/lib/pf/config/util.pm
@@ -415,10 +415,14 @@ Find sources for a specific realm
 
 sub get_realm_authentication_source {
     my ( $username, $realm, $sources ) = @_;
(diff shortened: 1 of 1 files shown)


julsemaan:
What are the side effects of this ? Can it have side effects to behaviors we assumed were correct in previous versions ?

PR #4950, opened 2019-11-21
Feature/inline detection mecanism bypass passthrough

fdurand:
# Description
If the device is reg and it try to hit a captive portal url detection then the dispatcher will proxy the request and the device will think that it have internet access.
It happen in inline setup when the device cache the dns request and still think that the captive portal url detection match with the ip of the portal

# Impacts
Portal detection

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Proxy the captive portal detection url when the device is reg

### DIFF
--- a/go/httpdispatcher/proxy.go
+++ b/go/httpdispatcher/proxy.go
@@ -545,7 +545,7 @@ func (p *Proxy) nodeStatus(ctx context.Context, mac string) bool {
 	status := false
 	var Status string
 	err := p.Nodedb.QueryRow(mac).Scan(&Status)
-	if err != nil {
+	if err == nil {
 		if Status == "reg" {
 			status = true
 		}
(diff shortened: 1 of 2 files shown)


fdurand:
BUMP

go/httpdispatcher/proxy.go
```diff
@@ -527,6 +524,54 @@ func (p *Proxy) detectPortalURL(r *http.Request) (bool, url.URL) {
...
+			if p.nodeStatus(ctx, MAC) && passThrough.checkDetectionMechanisms(ctx, fqdn) {
+				log.LoggerWContext(ctx).Info("Device register and match the portal detection mechanism for " + MAC)
+				p.reverse(ctx, w, r, r.Host)
+			}
+		}
+	}
+}
+
+// nodeStatus search for status of the device
+func (p *Proxy) nodeStatus(ctx context.Context, mac string) bool {
```
julsemaan: Should be called nodeIsReg since you're not returning the status column value but whether or not the status is reg

julsemaan:
Also, should this be in the maintenance ?

fdurand:
> Also, should this be in the maintenance ?

Not sure if we need to put in maintenance, it's not really a bug, it's just a way to still have the portal with the fqdn.

extrafu:
@julsemaan and @fdurand - do we keep that for v10.1? If so we need to get it in today.

PR #4951, opened 2019-11-21
Lower ttl to 5 seconds if the fqdn match with a captive portal detection host

fdurand:
# Description
Force the ttl to 5 seconds for captive portal detection host

# Impacts
dns

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Set TTL to 5 seconds when the host match with a captive portal detection host


jrouzierinverse (approved):
Looks good

PR #4972, opened 2019-11-27
Feature/pfpki

fdurand:
# Description
PacketFence PKI in golang

# Impacts
Provisioning + certificates

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* PacketFence pki service in golang

# UPGRADE file entries

PacketFence-pki is deprecated in favour of the new packetfence pki written in golang.
If you previously ran the PacketFence-pki you will need to migrate from the sqlite db to mysql.
To migrate be sure that the database is running and the new pki too and do the following:

`/usr/local/pf/addons/upgrade/to-10.0-packetfence-pki-migrate.pl`

Next edit the pki providers (Configuration -> PKI Providers) and redefined the profile to use. 
And if you use oscp then change the url to use this one:
`http://127.0.0.1:22225/api/v1/pki/ocsp`


satkunas:
Notes on TODO:
 * Mail header/footer disabled
 * No error is triggered if invalid ID is passed (@ revoke Cert).
 * SMTP settings need to be pulled from pfconfig (currently static).

Nice to have's:
 * Allow admin to set a filename (prefix) when cert is emailed and mime-attached, fallback to 'cert.p12'

satkunas:
Since we're going to include the schema externally, we'll need to avoid using automigrate

https://github.com/inverse-inc/packetfence/blob/e2e59e2f4760450ed6600a23ea18cd7b6ed01bd8/go/caddy/pfpki/models.go#L251

satkunas:
TODO:

Manually add foreign keys with `ON DELETE CASCADE` w/ `cas -> profiles`, `profiles -> certs` and `profiles -> revoked_certs`.

conf/caddy-services/api.conf.example
```diff
@@ -39,6 +39,11 @@
   proxy /api/v1/firewall_sso/ 127.0.0.1:8777 {
     transparent
   }
+
+  # pfpki API access
+  proxy /api/v1/pki/ 127.0.0.1:22225 {
```
julsemaan, line 44: We need admin roles for these API endpoints

conf/pf.conf.defaults
```diff
@@ -1183,7 +1188,7 @@ ocsp_override_cert_url = yes
 # radius_configuration.ocsp_url
 #
 # The overridden OCSP Responder URL.
-ocsp_url = http://127.0.0.1/ocsp/
+ocsp_url = http://127.0.0.1:22225/api/v1/pki/ocsp
```
julsemaan, line 1191: I think it would be time to allow configuration of the various settings of OCSP through the admin

db/pf-schema-X.Y.Z.sql
```diff
@@ -1363,6 +1363,154 @@ CREATE TABLE `admin_api_audit_log` (
    KEY `created_at` (`created_at`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=COMPRESSED;
 
+
+--
+-- Table structure for table `cas`
+--
+
+CREATE TABLE `cas` (
```
julsemaan: I think the PKI tables should be prefixed since some can cause confusion with PF features (like the `profiles` table)

go/caddy/pfpki/handlers.go
```diff
@@ -0,0 +1,581 @@
...
+
+	// Set the default Content-Type
+	Information.ContentType = "application/json; charset=UTF-8"
+
+	switch v := object.(type) {
+	case CA:
+
+		switch {
+
+		case len(regexp.MustCompile(`/pki/cas/search$`).FindStringIndex(req.URL.Path)) > 0:
```
julsemaan: Why not use an actual HTTP router instead of an enormous switch case for all the URLs this supports.

It creates an enormous function that will be hard to maintain.

Golang practices usually have a function per URL+verb to handle.

go/caddy/pfpki/models.go
```diff
@@ -0,0 +1,755 @@
...
+	if CaDB := pfpki.DB.First(&ca, p.CaID).Find(&ca); CaDB.Error != nil {
+		Information.Error = err.Error()
+		return Information, CaDB.Error
+	}
+
+	if err := pfpki.DB.Create(&Profile{Name: p.Name, Ca: ca, CaID: p.CaID, CaName: ca.Cn, Validity: p.Validity, KeyType: p.KeyType, KeySize: p.KeySize, Digest: p.Digest, KeyUsage: p.KeyUsage, ExtendedKeyUsage: p.ExtendedKeyUsage, P12MailPassword: p.P12MailPassword, P12MailSubject: p.P12MailSubject, P12MailFrom: p.P12MailFrom, P12MailHeader: p.P12MailHeader, P12MailFooter: p.P12MailFooter}).Error; err != nil {
+		Information.Error = err.Error()
+		return Information, errors.New("A database error occured. See log for details.")
+	}
+	pfpki.DB.Select("id, name, ca_id, ca_name, validity, key_type, key_size, digest, key_usage, extended_key_usage, p12_mail_password, p12_mail_subject, p12_mail_from, p12_mail_header, p12_mail_footer").Where("name = ?", p.Name).First(&profiledb)
```
julsemaan, line 399: I don't think you have to specify all the columns in gorm when performing a select

go/caddy/pfpki/models.go
```diff
@@ -0,0 +1,755 @@
...
+
+	return Information, nil
+}
+
+func (p Profile) search(pfpki *Handler, vars Vars) (Info, error) {
+	Information := Info{}
+	sql, err := vars.Sql(p)
+	if err != nil {
+		Information.Error = err.Error()
+		return Information, errors.New("A database error occured. See log for details.")
```
julsemaan, line 456: It doesn't look like the error is actually logged unless I'm missing some magic

go/caddy/pfpki/ocsp/oscp.go
```diff
@@ -0,0 +1,747 @@
+// Copyright 2013 The Go Authors. All rights reserved.
```
julsemaan: Why are we adding this code to our own repo ?

go/db/db.go
```diff
@@ -37,11 +41,37 @@ func ConnectDb(ctx context.Context, user, pass, host, dbName string) (*sql.DB, e
...
+	db.SetConnMaxLifetime(time.Minute * 5)
 	if err != nil {
 		log.LoggerWContext(ctx).Error(fmt.Sprintf("Error while connecting to DB: %s", err))
 		return nil, err
 	} else {
 		return db, nil
 	}
 }
+
+func ReturnURI(ctx context.Context, dbName ...string) string {
```
julsemaan, line 42: Use this in ConnectDb to prevent code duplication

lib/pf/services/manager/pfpki.pm
```diff
@@ -0,0 +1,60 @@
...
+use pf::config qw(
+    %Config
+);
+use pf::util;
+
+extends 'pf::services::manager';
+
+has '+name' => ( default => sub { 'pfpki' } );
+
+sub isManaged {
```
julsemaan: This is useless

julsemaan (changes requested):

julsemaan:
On top of what I reviewed we're missing:
* Upgrade notes detailing the changes this brings and the necessary upgrade steps to migrate the config from the "old" PKI to this
* Upgrade script to migrate the certificates and keys from an "old" PKI to this PKI

conf/caddy-services/api.conf.example
fdurand: done

go/caddy/pfpki/models.go
fdurand: It's to get rid of some columns

go/caddy/pfpki/models.go
fdurand: gorm logs the sql error in the logs, it's just for the frontend.
Btw i need to work on the loglevel.

go/caddy/pfpki/ocsp/oscp.go
fdurand: It's a custom version (parseRequest)

fdurand:
BUMP

addons/upgrade/to-10.0-packetfence-pki-migrate.pl
```diff
@@ -0,0 +1,173 @@
...
+BEGIN {
+    use constant INSTALL_DIR => '/usr/local/pf';
+    use lib INSTALL_DIR . "/lib";
+}
+
+use pf::db;
+use WWW::Curl::Easy;
+
+my $driver   = "SQLite";
+my $database = "db.sqlite3";
```
julsemaan: Are you sure this is the right path for the DB ?

To me it sounds like it would take db.sqlite3 in my cwd which may not be where the PKI DB is

go/caddy/pfpki/pfpki.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Default http timeout
+	http.DefaultClient.Timeout = 10 * time.Second
+
+	pfpki.router = mux.NewRouter()
+	PFPki := &pfpki
+	api := pfpki.router.PathPrefix("/api/v1").Subrouter()
+
+	// CAs (GET: list, POST: create)
+	api.Handle("/pki/cas", getSetCA(PFPki)).Methods("GET", "POST")
```
julsemaan, line 89: Is it possible you're missing a file in the PR ?

I don't see getSetCA defined in any file here.

Same for a few other functions I tested below

julsemaan:
Also, are there any necessary steps to migrate configuration from an old PKI ?

Or all the config is contained in the DB that will already be migrated ?

go/caddy/pfpki/pfpki.go
fdurand: https://github.com/inverse-inc/packetfence/blob/feature/pfpki/go/caddy/pfpki/handlers.go#L83

julsemaan: #semaanisblind

fdurand:
> Also, are there any necessary steps to migrate configuration from an old PKI ?
> 
> Or all the config is contained in the DB that will already be migrated ?

There is no need to migrate the configuration from the old PKI , only the ca/profiles/certificates are important.
Also i added a note in the upgrade section to modify the pki provider and the oscp url.

julsemaan (approved):

julsemaan:
On my side all is good, I see a comment of @satkunas that hasn't seen any answer.

Is this done ? If so then we're ready to merge

> TODO:
> 
> Manually add foreign keys with `ON DELETE CASCADE` w/ `cas -> profiles`, `profiles -> certs` and `profiles -> revoked_certs`.

fdurand:
> On my side all is good, I see a comment of @satkunas that hasn't seen any answer.
> 
> Is this done ? If so then we're ready to merge
> 
> > TODO:
> > Manually add foreign keys with `ON DELETE CASCADE` w/ `cas -> profiles`, `profiles -> certs` and `profiles -> revoked_certs`.

Not yet

fdurand:
It can be merged.

nqb:
A file is called oscp.go in place of ocsp.go,  fixed by @fdurand

nqb:
To avoid any confusion, we should put packetfence-pki repository in read-only with an updated README to mention that packetfence-pki is not integrated into PF.

PR #5007, opened 2019-12-20
Update path to route binary

nqb:
# Description
On Debian stretch, `route` binary is located in `/sbin` which is not part of the `PATH` for `pf` user

# Impacts
Detection of running state for `routes` service on Debian.

# NEW Package(s) required
YES

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
*

### DIFF
--- a/lib/pf/services/manager/routes.pm
+++ b/lib/pf/services/manager/routes.pm
@@ -131,7 +131,7 @@ sub isAlive {
         }
     }
     my $routes_applied = $FALSE;
-    $routes_applied = defined(pf_run("route | grep ".$route_exist)) if ($route_exist);
+    $routes_applied = defined(pf_run("/sbin/route | grep ".$route_exist)) if ($route_exist);
     $routes_applied = $TRUE if (-f "$install_dir/var/routes_applied");
     return (defined($pid) && $routes_applied);
 }


lib/pf/services/manager/routes.pm
```diff
@@ -131,7 +131,7 @@ sub isAlive {
         }
     }
     my $routes_applied = $FALSE;
-    $routes_applied = defined(pf_run("route | grep ".$route_exist)) if ($route_exist);
+    $routes_applied = defined(pf_run("/sbin/route | grep ".$route_exist)) if ($route_exist);
```
fdurand, line 134: Shouldn't be in the debianize.patch ?, on centos it's /usr/sbin/route

nqb: On CentOS:

```shell
# sudo -u pf -i

-sh-4.2$ ls -rtld /sbin
lrwxrwxrwx. 1 root root 8  1 juin   2019 /sbin -> usr/sbin
```

Behavior will be the same for Debian Buster: https://www.debian.org/releases/buster/amd64/release-notes/ch-whats-new.en.html#merged-usr

nqb:
@fdurand, 

Can this PR be merge ?

PR #5009, opened 2019-12-20
security_events: split example and defaults in correct files

nqb:
# Description
### `security_events.conf.defaults`
- All defaults and shipped security events are located in this file
- Set `max_enables` for Parking security event to `0` 
=> If a user reached `max_enables` for Parking security event, user can't leave parking state without an action from an admin. Admin should delete security events by hand in DB.

###  `security_events.conf.example`
- Remove all security events from this file in favor of  `security_events.conf.defaults`

# Impacts
Security events

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Parking security can now be trigger without limitation


julsemaan:
IMO the default violations (built-in and shipped) should be in the .defaults and we should remove what is in the .example instead

nqb:
@julsemaan,

I exactly did the opposite because I was sure it will cause issues when you want to delete them.

julsemaan:
I don't think we should allow users to delete built-in security events. Otherwise that might create unexpected errors in the code paths that expect these to be there.

Users can disable these built-in security events if they want but I don't think we should let them delete them

nqb:
@julsemaan, I will do the change you requested unless your opinion changed.

julsemaan:
Still the same

nqb:
@julsemaan, I did change you requested. I'm building package to check behavior and also deletion of security event (see #5010).

nqb:
It works as expected:
- unable to delete all default security events (it's a change, see #5010)
- change to default security events are written in security_events.conf
- new security events are created in security_events.conf

nqb:
@julsemaan, do you anything else before merging ?

conf/security_events.conf.defaults
```diff
@@ -71,7 +72,20 @@ enabled=N
 [1100004]
 desc=Ancient OS isolation example
 template=banned_os
-trigger=DEVICE::28,DEVICE::29,DEVICE::30
+trigger=DEVICE::28,DEVICE::29,DEVICE::30,DEVICE::31,DEVICE::32
```
julsemaan: Remove DEVICE::31 and DEVICE::32, they don't exist anymore, the configuration in the .example was faulty

conf/security_events.conf.defaults
```diff
@@ -71,7 +72,20 @@ enabled=N
...
+#
+# Example config to block a specific Browser User Agent
+# Trigger format: The number is the ID of the user-agent from the 'User Agent' list in Fingerbank (either 'upstream' or 'local' or both)
+#
+# The below example blocks user-agents IDs 101 and 102.
+#
+[1100005]
+desc=Browser isolation example
+template=banned_devices
+trigger=USER_AGENT::101,USER_AGENT::102
```
julsemaan: The USER_AGENT triggers don't exist anymore, remove this whole security event (.example was faulty on this too)

conf/security_events.conf.defaults
```diff
@@ -222,7 +242,7 @@ actions=enforce_provisioning,log
 template=enforce_provisioning
 trigger=Provisioner::check
 enabled=Y
-grace=15m
+grace=0s
```
julsemaan: Its not a good idea to have a grace of 0s on this, lets put the 15m back or at least 1m.

Otherwise if it gets triggered twice in a connection, it results in 2 checks

nqb:
I've addressed all your comment @julsemaan, do you think we need to add some notes in Upgrade guide related to removal of devices 31/32 and user agent security event.

PR #5057, opened 2020-01-15
Lazy non-blocking toggle

satkunas:
# Description
Refactor the toggles on Config Lists so they are lazy and non-blocking. Allows faster synchronous control. 

# Impacts
n/a

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature
- [n/a] Add unit tests

### DIFF
--- a/html/pfappserver/root/static.alt/src/components/pfTableSortable.vue
+++ b/html/pfappserver/root/static.alt/src/components/pfTableSortable.vue
@@ -14,7 +14,7 @@
       class="pf-table-sortable-empty justify-content-md-center"
     >
       <b-col cols="12" md="auto">
-        <slot name="empty" v-bind="{ isLoading }">
+        <slot name="empty" v-bind:isLoading="isLoading">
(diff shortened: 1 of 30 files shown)


html/pfappserver/root/static.alt/src/components/pfTableSortable.vue
```diff
@@ -14,7 +14,7 @@
       class="pf-table-sortable-empty justify-content-md-center"
     >
       <b-col cols="12" md="auto">
-        <slot name="empty" v-bind="{ isLoading }">
+        <slot name="empty" v-bind:isLoading="isLoading">
```
cgx: Use kebab-case and the `:` shortcut.

```html
<slot name="empty" :is-loading="isLoading">
```

julsemaan: I personally prefer the shish taouk case with a bit of extra hummus

PR #5078, opened 2020-01-21
Adjust backup-and-maintenance.sh for locationlog_history table

### LINKED ISSUE #5076: locationlog1to1: backup-and-maintenance.sh to adjust
**Describe the bug**
With #4995:
1. We keep only one entry per node in `locationlog` but pfmon `locationlog_cleanup` job will remove this entry after 1 week by default. I'm not sure it's a problem.

1. `backup-and-maintenance.sh` script has several references to `locationlog_archive` table which didn't exist anymore:
- https://github.com/inverse-inc/packetfence/blob/15e5365172113258c7206bc7d492cac11e15d9af/addons/backup-and-maintenance.sh#L116
- https://github.com/inverse-inc/packetfence/blob/15e5365172113258c7206bc7d492cac11e15d9af/addons/backup-and-maintenance.sh#L162

These references need to be replace by `locationlog_history` I guess.

nqb:
# Description
- Exclude `locationlog_history` from `mysqldump`
- Remove call to `database-cleaner.pl` because `pfmon` already do the job
- Update docs

# Impacts
- `backup-and-maintenance.sh` script

# Issue
fixes #5076 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Adjust backup-and-maintenance.sh for `locationlog_history` table (#5076)

### DIFF
--- a/addons/backup-and-maintenance.sh
+++ b/addons/backup-and-maintenance.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 #
-# Backup of PF tree and database
+# Backup of $PF_DIRECTORY and $DB_NAME
 #
-# - compressed PF tree to $BACKUP_DIRECTORY, rotate and clean
+# - compressed $PF_DIRECTORY to $BACKUP_DIRECTORY, rotate and clean
 # - compressed mysqldump to $BACKUP_DIRECTORY, rotate and clean
 #
 # Copyright (C) 2005-2019 Inverse inc.
--- a/addons/backup-and-maintenance.sh
+++ b/addons/backup-and-maintenance.sh
@@ -1,11 +1,9 @@
 #!/bin/bash
 #
-# Database maintenance and backup
+# Backup of PF tree and database
 #
-# - Move entries older than a month from locationlog to locationlog_archive
-# - Optimize tables on sunday
+# - compressed PF tree to $BACKUP_DIRECTORY, rotate and clean
 # - compressed mysqldump to $BACKUP_DIRECTORY, rotate and clean
-# - archive locationlog_archive entries older than a year the first day of each month
 #
 # Copyright (C) 2005-2019 Inverse inc.
 #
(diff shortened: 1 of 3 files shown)


addons/backup-and-maintenance.sh
```diff
@@ -113,7 +109,7 @@ should_backup(){
 }
 
 backup_db(){
-    /usr/local/pf/addons/database-cleaner.pl --table=locationlog_archive --date-field=end_time --older-than="1 MONTH"
+    /usr/local/pf/addons/database-cleaner.pl --table=locationlog_history --date-field=end_time --older-than="1 MONTH"
```
julsemaan: This is already cleaned by the locationlog_cleanup job so there is no point on even keeping that line

nqb: Done

docs/PacketFence_Installation_Guide.asciidoc
```diff
@@ -4307,7 +4307,7 @@ Keeping Tables Small
 
 Over time, some of the tables will grow large and this will drag down performance (this is especially true on a wireless setup).
 
-One such table is the `locationlog` table. We recommend that closed entries in this table be moved to the archive table `locationlog_archive` after some time. A closed record is one where the `end_time` field is set to a date (strictly speaking it is when `end_time` is not null and not equals to 0).
+One such table is the `locationlog` table. We recommend that closed entries in this table be moved to the archive table `locationlog_history` after some time. A closed record is one where the `end_time` field is set to a date (strictly speaking it is when `end_time` is not null and not equals to 0).
```
julsemaan: This line isn't true anymore, this happens auto-magically so we can just remove this

nqb: You're right, I missed it. I will adjust.

nqb:
@julsemaan, it should be good now.

addons/backup-and-maintenance.sh
```diff
@@ -1,19 +1,16 @@
 #!/bin/bash
 #
-# Database maintenance and backup
+# Backup of PF tree and database
```
julsemaan: PF tree ?

PR #5086, opened 2020-01-23
Feature/change captive portal detection

fdurand:
# Description
Change the captive portal behaviour to return a 200 on captive portal detection (wispr desactivated by default).

On a layer 2 network return the ip address 192.0.2.1 for the captive portal to be able to have the popup detection on samsung devices. 

Faster portal detection from apple devices.

Added the ip address 192.0.2.1 on the loopback interface and enabled haproxy-portal on this ip.
 
# Impacts
Captive portal detection workflow.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Better captive portal detection for Samsung devices
* Faster detection for Apple devices


conf/systemd/packetfence-keepalived.service
```diff
@@ -1,8 +1,9 @@
 # Copyright (C) Inverse inc.
 [Unit]
 Description=PacketFence LVS and VRRP High Availability Monitor
-After=packetfence.target packetfence-haproxy-db.service
-Wants=packetfence.target packetfence-iptables.service
+Before=packetfence-haproxy-portal.service
```
julsemaan, line 4: Shouldn't it be after haproxy-portal is started as in a cluster it would start before the server is ready to offer service

conf/systemd/packetfence-keepalived.service
```diff
@@ -13,6 +14,7 @@ ExecStartPre=/bin/perl -I/usr/local/pf/lib '-Mpf::services::manager::keepalived'
 ExecStart=/usr/sbin/keepalived -f /usr/local/pf/var/conf/keepalived.conf --pid=/usr/local/pf/var/run/keepalived.pid
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=on-failure
+Slice=packetfence.slice
 
 [Install]
-WantedBy=packetfence-cluster.target
+WantedBy=packetfence.target
```
julsemaan, line 20: Not sure it will behave correctly in a cluster where it will start before radiusd load balancer is ready

go/httpdispatcher/proxy.go
```diff
@@ -150,42 +151,73 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			destURL.Host = host
+		}
 
 		if parking {
 			PortalURL.Path = ""
 			PortalURL.RawQuery = ""
 		}
 
+		// Detect wispr user agent but exclude istuff
```
julsemaan: lets call them iOS devices or "Apple stuff" if it applies to all their devices

istuff looks like stuff but you added an i at the beginning because you forgot you were already in insert mode

julsemaan:
As a solution to the couple of startup order issues I found, we could fix the issue by improving the keepalived node health checks in order to ensure the VIP will only go up if all the necessary services are ready. That could be good also to migrate the VIP if any load balancing service dies. 

As the french would describe this:
- "1 croissant, 2 baguettes"
- "1 rock, 2 coups"
- "2 oiseaux, une roche"

conf/systemd/packetfence-keepalived.service
fdurand: It's before because in standalone i won't be able to bind if the ip address is not here.

conf/systemd/packetfence-keepalived.service
fdurand: It's because i need it in a standalone mode, so i need to change the target.

go/httpdispatcher/proxy.go
```diff
@@ -150,42 +151,73 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+		// Detect wispr user agent but exclude Apple stuff
+		wispr := regexp.MustCompile(`(?i)wispr`)
+		CaptiveNetworkSupport := regexp.MustCompile(`(?i)CaptiveNetworkSupport`)
+
+		PortalURL.RawQuery = "destination_url=" + destURL.String()
+
 		w.Header().Set("Location", PortalURL.String())
-		w.WriteHeader(http.StatusFound)
+		t := template.New("foo")
```
extrafu: Might want to give a more relevant temporary name...

fdurand:
Bump

rpm/packetfence.spec
```diff
@@ -286,7 +286,7 @@ Requires: perl(Test::NoWarnings), perl(Test::ParallelSubtest)
 # required for the fake CoA server
 Requires: perl(Net::UDP)
 # For managing the number of connections per device
-Requires: haproxy >= 1.8.9, keepalived >= 1.4.3
+Requires: haproxy >= 1.8.9, keepalived >= 2.0.0
```
julsemaan, line 289: The package doesn't seem to be in the devel repo yet

fdurand: Yes it's on the OBS , will push it on devel repo

fdurand: Done http://inverse.ca/downloads/PacketFence/CentOS7/devel/x86_64/RPMS/keepalived-2.0.20-1.1.x86_64.rpm

PR #5087, opened 2020-01-24
Fix EX2300 CoA Port Bounce

lyubomirtraykov:
# Description
Override bouncePort, because  setAdminStatus is called two times in Switch.pm.
Replaced locationlog_view_open_switchport and added a check for VoIP nodes.

# Impacts
EX2300 CoA Port Bounce

# Delete branch after merge
YES

# Checklist
- [n/a ] Document the feature
- [n/a ] Add unit tests
- [ n/a] Add acceptance tests (TestLink)

### DIFF
--- a/lib/pf/Switch/Juniper/EX2300.pm
+++ b/lib/pf/Switch/Juniper/EX2300.pm
@@ -116,8 +116,17 @@ sub setAdminStatus {
     my $logger = $self->logger;
 
     #We need to fetch the MAC on the ifIndex in order to bounce switch port with CoA.
-    my @locationlog = locationlog_view_open_switchport( $self->{_ip}, $ifIndex );
+    my @locationlog = locationlog_view_open_switchport_no_VoIP( $self->{_ip}, $ifIndex );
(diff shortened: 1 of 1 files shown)


lib/pf/Switch/Juniper/EX2300.pm
```diff
@@ -116,8 +116,17 @@ sub setAdminStatus {
...
+    my @locationlog = locationlog_view_open_switchport_no_VoIP( $self->{_ip}, $ifIndex );
     my $mac = $locationlog[0]->{'mac'};
+    if (!$mac) {
+        @locationlog = locationlog_view_open_switchport_only_VoIP( $self->{_ip}, $ifIndex );
+        $mac = $locationlog[0]->{'mac'};
+    }
+    
+    if (!$mac) {
+        $logger->info("Can't find MAC address in the locationlog... we won't perform port bounce");
+        return 1;
```
julsemaan: Add "use pf::constants" at the top of the file and use '$TRUE' instead of the hardcoded '1'

PR #5088, opened 2020-01-24
fix/5075

### LINKED ISSUE #5075: API: delete of a node is possible even with a locationlog entry opened
**Describe the bug**
From web admin, it's possible to delete a node even if this node has a locationlog entry opened. That's not the case with `pfcmd node delete XX:XX:XX:XX:XX:XX`.

From my point of view, we should have same behavior between two tools and I prefer `pfcmd` approach.

**To Reproduce**
1. Generate a RADIUS request (to create a locationlog entry):
``` 
# cat >> /opt/tls.config << EOF
network={
ssid="CORPORATE SSID"
key_mgmt=WPA-EAP
eap=TLS
identity="inverse"
ca_cert="/usr/local/pf/raddb/certs/ca.pem"
client_cert="/usr/local/pf/raddb/certs/client.pem"
private_key="/usr/local/pf/raddb/certs/client.key"
private_key_passwd="whatever"
eapol_flags=3
}
EOF

# eapol_test -c /opt/tls.config -s testing123 -a MGMT_IP -A 127.0.0.1 -N 61:d:15 -M 11:11:11:11:11:11 -N 4:x:0xc0a80101
```
2. Click on node added
3. Click on Delete

=> Your node is removed from `node` table but its location stays in place in `locationlog` table.

**Expected behavior**
It should not possible to delete a node if a locationlog entry is opened.

jrouzierinverse:
Description
===========
Node deletion in consistent between the the API and pf::node::node_delete

Impacts
=======
API

NEWS file entries
=======================
Enhancements
-------------
* Node deletion in consistent between the the API and pf::node::node_delete

Issue
=====
fixes #5075

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/lib/pf/node.pm
+++ b/lib/pf/node.pm
@@ -182,7 +182,6 @@ sub node_delete {
     }
 
     require pf::locationlog;
-    # TODO that limitation is arbitrary at best, we need to resolve that.
     my ($can_delete, $msg) = _can_delete($mac);
     if (!$can_delete) {
         $logger->warn($msg);
--- a/lib/pf/node.pm
+++ b/lib/pf/node.pm
@@ -177,15 +177,14 @@ sub node_delete {
     $mac = clean_mac($mac);
 
     if ( !node_exist($mac) ) {
-        $logger->error("delete of non-existent node '$mac' failed");
-        return ($FALSE);
+        return ($FALSE, "delete of non-existent node '$mac' failed");
     }
 
     require pf::locationlog;
     my ($can_delete, $msg) = _can_delete($mac);
     if (!$can_delete) {
         $logger->warn($msg);
-        return ($FALSE);
+        return ($FALSE, $msg);
     }
 
     my %options = (
@@ -200,11 +199,11 @@ sub node_delete {
 
     my ($status, $count) = pf::dal::node->remove_items(%options);
     if (is_error($status)) {
-        return ($FALSE);
+        return ($FALSE, "Unable to delete '$mac' internal error");
     }
 
     $logger->info("node $mac deleted");
-    return ($TRUE);
+    return ($TRUE, '');
 }
 
 our %DEFAULT_NODE_VALUES = (
--- a/lib/pf/node.pm
+++ b/lib/pf/node.pm
@@ -156,6 +156,16 @@ sub node_view_reg_pid {
     return;
 }
 
+sub _can_delete {
+    my ($mac) = @_;
+    if ( defined( pf::locationlog::locationlog_view_open_mac($mac) ) ) {
+        my $msg = "$mac has an open locationlog entry. Node deletion prohibited";
+        return (0, $msg);
+    }
+
+    return (1,'');
+}
+
 #
 # delete and return 1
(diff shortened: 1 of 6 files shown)


lib/pf/node.pm
```diff
@@ -156,6 +156,16 @@ sub node_view_reg_pid {
...
+sub _can_delete {
+    my ($mac) = @_;
+    if ( defined( pf::locationlog::locationlog_view_open_mac($mac) ) ) {
+        my $msg = "$mac has an open locationlog entry. Node deletion prohibited";
+        return (0, $msg);
+    }
+
+    return (1,'');
+}
+
```
nqb, line 168: Should we not replace `0` and `1` by `$FALSE` and `$TRUE` ?

lib/pf/node.pm
```diff
@@ -173,8 +183,9 @@ sub node_delete {
 
     require pf::locationlog;
     # TODO that limitation is arbitrary at best, we need to resolve that.
-    if ( defined( pf::locationlog::locationlog_view_open_mac($mac) ) ) {
-        $logger->warn("$mac has an open locationlog entry. Node deletion prohibited");
+    my ($results, $msg) = _can_delete($mac);
+    if (!$results) {
+        $logger->warn($msg);
```
nqb, line 186: 1. Is TODO comment still necessary ?
1. I will prefer `$result` in place `$results` like in other files.

lib/pf/node.pm
```diff
@@ -173,8 +183,9 @@ sub node_delete {
...
         return (0);
```
nqb: Should we not replace `0` by `$FALSE` ?

nqb (changes requested):

nqb:
@jrouzierinverse,

Another thing I noticed. Now when you remove a node using `pfcmd`, you got two lines in logs:
```
Feb  3 09:50:06 pfcen7stable packetfence: WARN pfcmd.pl(29687): 11:11:11:11:11:11 has an open locationlog entry. Node deletion prohibited (pf::node::node_delete)
Feb  3 09:50:06 pfcen7stable packetfence: ERROR pfcmd.pl(29687): Cannot delete node 11:11:11:11:11:11 since there are some records in locationlog table indicating that this node might still be connected and active on the network  (pf::cmd::pf::node::action_delete)
```
and you get "Cannot delete node 11:11:11:11:11:11 .." on STDERR.

IMO, we should adjust `action_delete` function in `pf::cmd::pf::node` to:
- use message returns by API in place of this different message
- display message returns by API on STDERR

lib/pf/node.pm
nqb: @jrouzierinverse, what about the TODO item ?

jrouzierinverse: I find $can_delete is more clear on the value is for.

nqb:
@jrouzierinverse, what about my last comment ?

nqb:
It works as expected.

PR #5104, opened 2020-02-03
New switch template for Aruba switches

nqb:
# Description
New switch template to handle these cases:
* Aruba OS 16.10 require an additional attribute to handle RADIUS disconnect messages
* `getIfIndexByNasPortId` not needed anymore 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Support Aruba switchs using Aruba OS 16.10


lib/pf/Switch/Aruba/ArubaSwitchNG.def
```diff
@@ -0,0 +1,24 @@
...
+EOT
+
+disconnect = <<EOT
+Calling-Station-Id = ${macToEUI48($mac)}
+NAS-IP-Address = $disconnectIp
+EOT
+
+coa = <<EOT
+Calling-Station-Id = ${macToEUI48($mac)}
+NAS-IP-Address = $disconnectIp
```
julsemaan, line 24: I don't think its necessary to add this as its implicit from what I know

Unless that overrides the default behavior?

nqb: It's a WIP. I'm waiting a feedback from a customer to add extra attributes because Aruba Switches complain about "Missing attribute" when sending RADIUS disconnect messages like this.

I'm working on a separate PR on improvements to Template guide, it will be interesting to know the default behavior to add it in documentation.

nqb:
@jrouzierinverse,

When creating a new switch template, if we don't define some scope like "reject" or "acceptVlan" should we consider [Standard.def](https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/Switch/PacketFence/Standard.def) as default ?

Based on your answer, I will update docs.

nqb:
> When creating a new switch template, if we don't define some scope like "reject" or "acceptVlan" should we consider [Standard.def](https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/Switch/PacketFence/Standard.def) as default ?

I did a test without a `reject` scope on a new switch module, Reply message defined in Standard.def is not returned, is it the expected behavior @jrouzierinverse ?

jrouzierinverse:
That is expected behavior.
Maybe a reject message should be required.
That makes sense to inherit from PacketFence::Standard.
The only problem is if there are any new scopes we add in the future PacketFence::Standard everyone will have them.
That means all switch will then support that automatically that feature.
This may cause misbehaviors since the definition says it supports it but in reality, the switch does not.
If we do add support for inheriting from a switch then we should consider a separate definition for inheriting from instead PacketFence::Standard to avoid those issues.

nqb:
Thanks a lot for this explanation, I will update docs and PR.

nqb:
@julsemaan, you can review.

PR #5130, opened 2020-02-07
Return the portal ip if the fqdn match with one of the portal fqdn

fdurand:
# Description
Return the portal ip if the qname match a portal fqdn defined in the configuration. (for inline and reg devices). It help when you do email registration and you need to hit the portal once registered.

# Impacts
pfdns

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Inline setup, reg device, return the portal ip if the qname match one of the portal fqdn

### DIFF
--- a/conf/pfdns.conf.example
+++ b/conf/pfdns.conf.example
@@ -21,6 +21,13 @@ proxy . /etc/resolv.conf
     }
     # Anything not handled by pfdns will be resolved normally
 [% domain %]
+
+    # This configuration will check in the /etc/hosts and if the fqdn match it will return the corresponding ip.
+    # If it doesn't match then it will continue
+    #hosts {
+    #    fallthrough
+    #}
+
 [% inline %]
 
     # Default to system resolv.conf file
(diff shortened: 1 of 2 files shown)


julsemaan:
Should this be a +1 patch release ?

fdurand:
Not sure we need to put it in the maintenance, it's not really a bug.

extrafu:
@julsemaan  Review and merge if all is fine. Don't forget to update the NEWS file.

conf/pfdns.conf.example
```diff
@@ -21,6 +21,13 @@ proxy . /etc/resolv.conf
     }
     # Anything not handled by pfdns will be resolved normally
 [% domain %]
+
+    # This configuration will check in the /etc/hosts and if the fqdn match it will return the corresponding ip.
+    # If it doesn't match then it will continue
+    #hosts {
```
julsemaan: Shouldn't that be enabled by default so that this works out of the box ?

Otherwise we would need to point to this somewhere in the doc for inline

fdurand: Yes, we can enable it by default. I can't see any issue with that.
I mean that if you are at this step in pfdns it means that your device is registered and you want to add custom dns entries (like internal).

julsemaan: In that case lets go with that so that it works out of the box

PR #5131, opened 2020-02-10
feature/optimize_handle_accounting_metadata

jrouzierinverse:
Description
===========
Optimizations found after profiling handle accounting metadata

Enhancements
-------------
* Optimized node lookup
* Optimized API client creation
* Optimized logging usage
* Optimized mac clean

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -1455,13 +1455,13 @@ sub radius_rest_accounting :Public :RestPath(/radius/rest/accounting) {
     my $timer = pf::StatsD::Timer->new();
     my $logger = pf::log::get_logger();
 
-    my %remapped_radius_request = %{pf::radius::rest::format_request($radius_request)};
+    my $remapped_radius_request = pf::radius::rest::format_request($radius_request);
 
-    my $return = $class->handle_accounting_metadata(%remapped_radius_request);
+    my $return = $class->handle_accounting_metadata($remapped_radius_request);
 
     my $radius = new pf::radius::custom();
     eval {
-        $return = $radius->accounting(\%remapped_radius_request, $headers);
+        $return = $radius->accounting($remapped_radius_request, $headers);
     };
     if ($@) {
         $logger->error("radius accounting failed with error: $@");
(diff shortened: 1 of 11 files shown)


julsemaan:
I checked it out and it looks good.

Have you also performed the profiling using the local API client so that we can tune the stuff in radius.pm ?

jrouzierinverse:
Yes, nothing that I could optimize easily.

julsemaan:
Can you post performance before and after (using local client)

Its not because it won't be optimizable easily that we shouldn't do it. 
The goal is to have extremely scalable accounting for v10 and if we need to rework some of the core of the accounting, we'll have to do so.

lib/pf/api.pm
```diff
@@ -1475,51 +1443,54 @@ sub radius_rest_accounting :Public :RestPath(/radius/rest/accounting) {
...
-    if ($RAD_REQUEST{'Acct-Status-Type'} != $ACCOUNTING::STOP){
+    if ($acct_status_type != $ACCOUNTING::STOP){
+        my $advanced = $pf::config::Config{advanced};
         # Tracking IP address.
-        if(pf::util::isenabled($pf::config::Config{advanced}{update_iplog_with_accounting})){
+        my $framed_ip = $RAD_REQUEST->{"Framed-IP-Address"};
+        if ($framed_ip && pf::util::isenabled($advanced->{update_iplog_with_accounting})) {
             $logger->info("Updating iplog from accounting request");
-            $client->notify("update_ip4log", mac => $mac, ip => $RAD_REQUEST{'Framed-IP-Address'}) if ($RAD_REQUEST{'Framed-IP-Address'} );
+            $client->notify("update_ip4log", mac => $mac, ip => $framed_ip);
```
fdurand, line 1467: To get rid of a call , can we test if the $framed_ip is a valid ip ?

lib/pf/api.pm
```diff
@@ -1475,51 +1443,54 @@ sub radius_rest_accounting :Public :RestPath(/radius/rest/accounting) {
...
         }
         else {
-            pf::log::get_logger->debug("Not handling iplog update because we're not configured to do so on accounting packets.");
+            $logger->debug("Not handling iplog update because we're not configured to do so on accounting packets.");
         }
         
-        if(pf::util::isenabled($pf::config::Config{advanced}{scan_on_accounting}) && $RAD_REQUEST{"Framed-IP-Address"}) {
+        if ($framed_ip && pf::util::isenabled($advanced->{scan_on_accounting})) {
```
fdurand, line 1473: if (valid_ip($framed_ip) && ...

PR #5143, opened 2020-02-14
Filter Engines GUI

satkunas:
# Description
Complete rework/refactor of Graphical User Interface (Admin) for Filter Engines 

# Impacts
Changes the Filter Engine workflow.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* TDB

## Enhancements
* TBD


julsemaan:
Initial comments on this:

- [ ] Actions should have stub values for arguments like the custom syslog parser engine. Ideally, they would have a builder but that could be for the next iteration
- [ ] Fields and values don't automatically populate for the filter, we have to type them manually
- [ ] My conditions are getting simplified where if I created a condition `OR (AND condition1, condition2)`, it simplifies it to `AND condition1, condition2`. From what I recall we said we wouldn't try to "optimize" the input of the user to prevent them from thinking the system is broken because what they save isn't the same as what they see after the save
- [x] I'm unable to select text in the input boxes that are part of the conditions
- [x] Should be able to add a description to the filters
- [x] Would be nice to move sub-conditions anywhere within the main condition. For example I have a condition node_info.machine_account starts_with host/ and I want to move it into a nested OR condition I just added afterwards. It works if that nested OR already contains a condition, not if its empty
- [x] Having difficulty dragging and dropping sub-conditions where we need to drag it into the sub-condition and then highlight the new position of what we're dragging so that it goes at the right place 

Pretty sure some of the above was already thought of by @satkunas and @jrouzierinverse in their own checklist but I wanted to at least get them here so that it can be tracked.

I'll update this comment with more feedback/items today if needed

julsemaan:
I'm adding this separately as its pretty major but it looks like none of the VLAN filter rules are loaded in the engine so its impossible to test the actual rules this produces

Log of the request:
```
Mar  6 11:40:07 pf-julien packetfence_httpd.aaa: httpd.aaa(22136) DEBUG: [mac:02:00:00:00:00:01] No engine found for AutoRegister (pf::access_filter::test)
```

My configuration:
```
[myass]
scopes=NodeInfoForAutoReg,RegisteredRole,AutoRegister
actions=modify_node: mac = $mac, notes = UNA MACHINA
action.0=modify_node: mac = $mac, notes = UNA MACHINA
status=enabled
role=default
condition=((starts_with(node_info.machine_account, "jse")&&connection_type == "Wireless-802.11-EAP")||node_info.role == "machines")
```

satkunas:
@julsemaan drag/drop refactored

jrouzierinverse:
> Initial comments on this:
> 
> * [x]  Should be able to add a description to the filters
> * [ ]  Actions should have stub values for arguments like the custom syslog parser engine. Ideally, they would have a builder but that could be for the next iteration
I will provide this in the OPTIONS
> * [ ]  Fields and values don't automatically populate for the filter, we have to type them manually
I will provide this in the OPTIONS.

However, this will be something we cannot automatically generate we must update this dictionary manually.

> * [x]  I'm unable to select text in the input boxes that are part of the conditions
> * [ ]  My conditions are getting simplified where if I created a condition `OR (AND condition1, condition2)`, it simplifies it to `AND condition1, condition2`. From what I recall we said we wouldn't try to "optimize" the input of the user to prevent them from thinking the system is broken because what they save isn't the same as what they see after the save

That is not an optimization I cannot represent that as condition string.
OR (AND condition1, condition2)  = (condition1 && condition2)
AND (condition1) = condition1
OR (condition1) = condition1

> * [x]  Would be nice to move sub-conditions anywhere within the main condition. For example I have a condition node_info.machine_account starts_with host/ and I want to move it into a nested OR condition I just added afterwards. It works if that nested OR already contains a condition, not if its empty

> * [x]  Having difficulty dragging and dropping sub-conditions where we need to drag it into the sub-condition and then highlight the new position of what we're dragging so that it goes at the right place
> 
> Pretty sure some of the above was already thought of by @satkunas and @jrouzierinverse in their own checklist but I wanted to at least get them here so that it can be tracked.
> 
> I'll update this comment with more feedback/items today if needed

jrouzierinverse:
The stubs are for the action parameters are now available in the OPTIONS

satkunas:
Adding 2 basic conditions results in missing end brackets in string-ified condition

Screenshot_20200313_142514

`dhcp_filters.conf`
```
[DHCP Filter]
scopes=
status=enabled
description=Test
condition=contains(foo, "bar" && contains(bar, "foo"
```

resulting condition should be:
```
[snipped]
condition=contains(foo, "bar") && contains(bar, "foo")
```

julsemaan:
Stuff that in my opinion will be necessary:

- [ ]  Fields and values don't automatically populate for the filter, we have to type them manually
- [ ] Getting a warning for when the condition that is saved in the backend isn't the same as the condition that was created by the user; issue:  

> My conditions are getting simplified where if I created a condition OR (AND condition1, condition2), it simplifies it to AND condition1, condition2. From what I recall we said we wouldn't try to "optimize" the input of the user to prevent them from thinking the system is broken because what they save isn't the same as what they see after the save

Various issues:

- [x] I'm unable to create a condition with only one sub-condition. It wants me to create a condition with 2 sub-condition so I can't create a condition like `node_info.category contains bussi`
- [x] I can't disable the `pf_deauth_from_wireless_secure` VLAN filter from the index (list) page. It just stays enabled

satkunas:
[resolved]

julsemaan:
Comments from new tests

- [x] The RADIUS filter merge_answer should be a toggle between `yes` and `no`
- [x] When typing in the attribute fields of my condition, the focus keeps being stolen and I can't type anymore in the field that was selected
- [ ] I'm still unable to select text in the fields of a condition (which was supposed to have been fixed)

- [x] Trying to negate a condition using NOT ALL creates a bad condition:
GUI:
image
Configuration:
```
[test1]
status=enabled
answer.0=3Com-Product-ID = hello
description=test1
scopes=
condition=not_and(, "")
```

- [ ] The defaults of the DNS filters have not been converted

- [ ] We're missing an upgrade script to convert existing configuration from the old format to the new format

- [x] It would be useful in the DHCP filters answers to have the option ID next to the name of the option to make it clearer and less prone to errors

Reported previously and still active:

- [ ]  Fields and values don't automatically populate for the filter, we have to type them manually
- [x] Getting a warning for when the condition that is saved in the backend isn't the same as the condition that was created by the user; issue:  

> My conditions are getting simplified where if I created a condition OR (AND condition1, condition2), it simplifies it to AND condition1, condition2. From what I recall we said we wouldn't try to "optimize" the input of the user to prevent them from thinking the system is broken because what they save isn't the same as what they see after the save

- [ ] The backend is still not doing anything so we can't test these actually do work

satkunas:
>  When typing in the attribute fields of my condition, the focus keeps being stolen and I can't type anymore in the field that was selected

This is fixed, but one side effect my be some unexpected behavior after drag-drop. Keep an eye out.

satkunas:
> I'm still unable to select text in the fields of a condition (which was supposed to have been fixed)

Unable to fix (Mozilla only): 
https://bugzilla.mozilla.org/show_bug.cgi?id=800050

Hack is unreasonable: https://stackoverflow.com/a/21682033

jrouzierinverse:
Fixed the NOT_AND serialization

julsemaan:
> Fixed the NOT_AND serialization

Doesn't work for me, I'm getting a different result but still not something that makes sense:
```
[test1]
status=enabled 
answer.0=3Com-Product-ID = hello
description=test1
scopes=
condition=!() 
```

julsemaan:
>  The RADIUS filter merge_answer should be a toggle between yes and no

When I save a new RADIUS filter without touching the merge_answer toggle, it doesn't save the field to `merge_answer=no` and simply omits it from the request payload. IMO, it should explicitly set it to `merge_answer=no`

This is the request payload:
```
{"answers":null,"condition":{"field":null,"op":"and","value":null,"values":[]},"description":"test3","id":"test3","merge_answer":null,"radius_status":null,"scopes":null,"status":"enabled"}
```

julsemaan:
I have unchecked the elements that are still buggy from the list above to help tracking the progress, just re-check them once its fixed

satkunas:
> > The RADIUS filter merge_answer should be a toggle between yes and no
> 
> When I save a new RADIUS filter without touching the merge_answer toggle, it doesn't save the field to `merge_answer=no` and simply omits it from the request payload. IMO, it should explicitly set it to `merge_answer=no`
> 
> This is the request payload:
> 
> ```
> {"answers":null,"condition":{"field":null,"op":"and","value":null,"values":[]},"description":"test3","id":"test3","merge_answer":null,"radius_status":null,"scopes":null,"status":"enabled"}
> ```

@jrouzierinverse Please set the meta default for `merge_answer` to 'no', it is currently `null`.
```
/usr/local/pf/sbin/pfperl-api get -M OPTIONS /api/v1/config/filter_engines/radius_filters| jq '.meta.merge_answers.default'
null
```

julsemaan:
>  Trying to negate a condition using NOT ALL creates a bad condition:

and 

> When I save a new RADIUS filter without touching the merge_answer toggle, it doesn't save the field to merge_answer=no and simply omits it from the request payload. IMO, it should explicitly set it to merge_answer=no

are fixed, I've updated the list above

jrouzierinverse:
Actions for the VlanFilters are not required.

jrouzierinverse:
Make Scopes required for all filters.

extrafu:
TODO @jrouzierinverse 

- [ ] Fields and values don't automatically populate for the filter, we have to type them manually (backend part)
- [x] Warning for the simplification of a condition (check if it should be in the frontend)
- [x]  The backend is still not doing anything so we can't test these actually do work
- [ ] The defaults of the DNS filters have not been converted
- [ ] We're missing an upgrade script to convert existing configuration from the old format to the new format

TODO @satkunas 
- [ ] issue of actions being required when they are not
- [x] scopes are now required (at least one needs to be defined)
- [ ] Fields and values don't automatically populate for the filter, we have to type them manually

julsemaan:
>  My conditions are getting simplified where if I created a condition OR (AND condition1, condition2), it simplifies it to AND condition1, condition2. From what I recall we said we wouldn't try to "optimize" the input of the user to prevent them from thinking the system is broken because what they save isn't the same as what they see after the save

@jrouzierinverse, this is still happening

What I am saving:
image

What is displayed after I create it, then refresh the page:
image

The user should be warned in this case that what he saved isn't what is saved in the configuration.

@jrouzierinverse, if its @satkunas that needs to handle this, please let us know

satkunas:
@extrafu actions are only required when "Run actions" is enabled.

> issue of actions being required when they are not

satkunas:
> Actions for the VlanFilters are not required.

done

> Make Scopes required for all filters.

done

jrouzierinverse:
> > My conditions are getting simplified where if I created a condition OR (AND condition1, condition2), it simplifies it to AND condition1, condition2. From what I recall we said we wouldn't try to "optimize" the input of the user to prevent them from thinking the system is broken because what they save isn't the same as what they see after the save
> 
> @jrouzierinverse, this is still happening
> 
> What I am saving:
> image
> 
> What is displayed after I create it, then refresh the page:
> image
> 
> The user should be warned in this case that what he saved isn't what is saved in the configuration.
> 
> @jrouzierinverse, if its @satkunas that needs to handle this, please let us know

This should be fixed for all cases not just the simple ones.

html/pfappserver/lib/pfappserver/Form/Field/DHCPAnswer.pm
```diff
@@ -0,0 +1,211 @@
...
+}
+
+=head2 options_type
+
+Provide a list DHCP option types
+
+=cut
+
+my %options = (
+    OptionSubnetMask                                 => 1,
```
julsemaan, line 87: It would greatly help to have the option ID in the label to recognize them better

lib/pf/UnifiedApi/Controller/Config/FilterEngines/DHCPFilters.pm
```diff
@@ -0,0 +1,59 @@
...
+use Mojo::Base qw(pf::UnifiedApi::Controller::Config);
+
+has 'config_store_class' => 'pf::ConfigStore::DhcpFilters';
+has 'form_class' => 'pfappserver::Form::Config::FilterEngines::DHCPFilter';
+has 'primary_key' => 'dhcp_filter_id';
+
+use pf::ConfigStore::DhcpFilters;
+use pfappserver::Form::Config::FilterEngines::DHCPFilter;
+
+sub is_sortable {
```
julsemaan, line 27: This code is being repeated in a lot of controllers, a role would be better (or a util function function you reuse)

julsemaan (changes requested):

html/pfappserver/lib/pfappserver/Form/Field/DHCPAnswer.pm
jrouzierinverse: It already is. Look at the options_type method below

julsemaan:
Republishing TODO with new bugs

TODO @jrouzierinverse 

- [x] Fields and values don't automatically populate for the filter, we have to type them manually (backend part)
- [x] Warning for the simplification of a condition (check if it should be in the frontend)
- [x]  The backend is still not doing anything so we can't test these actually do work
- [x] The defaults of the DNS filters have not been converted
- [x] We're missing an upgrade script to convert existing configuration from the old format to the new format
- [x] Remove the rcode from the DNS Filter
- [x] Broken RADIUS filters backend engine

TODO @satkunas 
- [x] issue of actions being required when they are not
- [x] scopes are now required (at least one needs to be defined)
- [x] Fields and values don't automatically populate for the filter, we have to type them manually
- [ ] Unable to configure actions in VLAN and RADIUS filters

BUGS:
- [x] DNS filters scopes are wrong

PR #5148, opened 2020-02-20
Vagrant: add an Active Directory server in environment

nqb:
# Description
Provision an AD server with:
- a domain
- a DNS server
- AD CS services with auto-enrollment
- a packetfence domain account

# Impacts
CI tests

# Delete branch after merge
YES

# Checklist
- [x] Document the feature


lzammit:
I did not test it out but it looks good to me.

nqb:
I tested it several times and it works so you can merge.

PR #5172, opened 2020-02-28
Ensure sponsor field is updated when using forced_sponsor

### LINKED ISSUE #5171: Sponsor field is missing on sponsored users when using forced sponsor
**Describe the bug**
When you use "Forced sponsor" field on a Sponsor portal module, sponsor field of sponsored users is not filled with email address of sponsor. That's not the case if you don't use "Forced sponsor" field.

**To Reproduce**
1. Create a user with Mark a sponsor + email address
1. Configure a portal module with Forced Sponsor using email address of your sponsor user
2. Try to register `foo` user using portal module defined at last step

Sponsor registration works but `foo` has an empty sponsor field in DB.

**Expected behavior**
Sponsor field of sponsored users should contain email address of their sponsor.

**Additional context**
Issue is caused by the fact that when we called `update_person_fields` function, we didn't pass `sponsor` variable (defined above) as argument: https://github.com/inverse-inc/packetfence/blob/ad5f313760fcb480cdcf45fc34df00e615f65151/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm#L212

It works for users that don't use "Forced Sponsor" field because  `update_person_fields` function called `person_modify` function with `$self->request_fields` and  `sponsor` email is stored in this hash.
https://github.com/inverse-inc/packetfence/blob/ad5f313760fcb480cdcf45fc34df00e615f65151/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm#L173

nqb:
# Description
Ensure sponsor field is updated when using forced_sponsor

# Impacts
Sponsor registration

# Issue
fixes #5171

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Sponsor field is missing on sponsored users when using forced sponsor (#5171)

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
@@ -209,7 +209,13 @@ sub do_sponsor_registration {
     $self->app->session->{email} = $email;
     $self->username($email);
 
-    $self->update_person_from_fields(additionnal_fields => {notes => $note, sponsor => $sponsor});
+    # update sponsor field with forced_sponsor value
+    if (!defined($self->request_fields->{sponsor})) {
+        $self->update_person_from_fields(additionnal_fields => {notes => $note, sponsor => $sponsor});
+    }
+    else {
+        $self->update_person_from_fields(additionnal_fields => {notes => $note});
+    }
 
     $self->waiting_room();
 }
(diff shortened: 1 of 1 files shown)


nqb:
I'm not very proud of this fix because it will certainly override `$self->request_fields->{sponsor}; ` when doing basic sponsor.

What about doing a test before `update_person_fields` call on `$self->request_fields->{sponsor}; ` ?

nqb:
New solution works.

PR #5177, opened 2020-02-28
Allow vlan greater than 1023

fdurand:
# Description
Allow vlan number greater than 1023 for floating devices

# Impacts
Floating devices

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Allow vlan number greater than 1023 for floating devices.

### DIFF
--- a/lib/pf/Switch/Cisco/Catalyst_2950.pm
+++ b/lib/pf/Switch/Cisco/Catalyst_2950.pm
@@ -840,8 +840,7 @@ sub setTaggedVlans {
     my @bits = split //, ("0" x 4096);
     foreach my $t (@vlans) {
         if ($t > 4096) {
-            $logger->warn("We do not support Tagged Vlans > 4096 for now on Cisco switches. Sorry... but we could! " .
-                      "interested in sponsoring the feature?");
+            $logger->warn("Tagged VLAN can't be greater than > 4096");
         } else {
             $bits[$t] = "1";
         }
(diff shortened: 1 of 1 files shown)


extrafu:
That PR has some serious licking impact!

lib/pf/Switch/Cisco/Catalyst_2950.pm
```diff
@@ -837,25 +837,33 @@ sub setTaggedVlans {
     my $OID_vlanTrunkPortVlansEnabled3k = '1.3.6.1.4.1.9.9.46.1.6.1.1.18';
     my $OID_vlanTrunkPortVlansEnabled4k = '1.3.6.1.4.1.9.9.46.1.6.1.1.19';
 
-    my @bits = split //, ("0" x 1024);
+    my @bits = split //, ("0" x 4096);
     foreach my $t (@vlans) {
-        if ($t > 1024) {
-            $logger->warn("We do not support Tagged Vlans > 1024 for now on Cisco switches. Sorry... but we could! " .
+        if ($t > 4096) {
```
extrafu, line 842: I think we can remove that statement... otherwise it's switch vendors that will contact us to enhance that capability in their product ;-)

julsemaan: Agree with @extrafu, that statement can never be true unless we f*cked up or Cisco starts supporting VLANs above 4096 which will likely require much more work than adjusting this sub

fdurand: https://github.com/inverse-inc/packetfence/commit/0b9e82b33139eb2669817883c86cb7261dc3e222

PR #5181, opened 2020-03-02
feature/pfacct

jrouzierinverse:
Description
===========
NOT READY
A rewrite of accounting to handle in go.

Impacts
=======
Radius Accounting.
Bandwidth Accounting

NEWS file entries
=======================
Enhancements
-------------
* Remove pfbandwidthd
* Faster radius accounting

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


db/pf-schema-X.Y.Z.sql
```diff
@@ -616,8 +616,10 @@ CREATE TABLE radius_nas (
   start_ip INT UNSIGNED DEFAULT 0,
   end_ip INT UNSIGNED DEFAULT 0,
   range_length INT DEFAULT 0,
+  unique_session_attributes varchar(255),
   PRIMARY KEY nasname (nasname),
-  KEY id (id)
+  KEY id (id),
+  INDEX radius_nas_start_ip_end_time (start_ip, end_ip)
```
fdurand: radius_nas_start_ip_end_ip ?

go/jsonrpc2/client.go
```diff
@@ -0,0 +1,169 @@
+package jsonrpc2
+
+import (
+	"context"
+    "io/ioutil"
```
fdurand: go fmt

extrafu:
Based on a discussion with @jrouzierinverse, what needs to be done:

1. Track session using a 24 hour idle time in case we don't receive a stop
1. Enabling/disabling netflow per network .

Both will be completed tomorrow based on @jrouzierinverse.

@fdurand and @julsemaan from there review and test. We are weeks behind on this work.

julsemaan:
Is this ready?

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -204,6 +204,23 @@ DELIMITER ;
...
+    out_bytes BIGINT UNSIGNED NOT NULL,
+    total_bytes BIGINT UNSIGNED AS (in_bytes + out_bytes) PERSISTENT,
+    PRIMARY KEY (tenant_id, unique_session_id, mac, time_bucket),
+    KEY bandwidth_accounting_time_bucket (time_bucket)
+);
+
+\! echo "Alter table radius_nas";
+ALTER TABLE radius_nas
+  ADD COLUMN IF NOT EXISTS unique_session_attributes varchar(255),
+  ADD INDEX IF NOT EXISTS radius_nas_start_ip_end_ip (start_ip, end_ip)
```
fdurand: missing ; at the end

julsemaan (changes requested):
- [ ] There is an issue where radiusd-acct doesn't get disabled properly on update since pfcmd service pf updatesystemd doesn't disable it properly

julsemaan (changes requested):
Missing from the PR:

- [x] Log lines when its an unknown RADIUS client or the shared secret is invalid

go/acct/radius.go
```diff
@@ -0,0 +1,334 @@
...
+	return hex.EncodeToString(sum)
+}
+
+func (h *PfAcct) sendRadiusAccounting(r *radius.Request) {
+	attr := packetToMap(r.Context(), r.Packet)
+	attr["PF_HEADERS"] = map[string]string{
+		"X-FreeRADIUS-Server":  "packetfence",
+		"X-FreeRADIUS-Section": "accounting",
+	}
+	client := jsonrpc2.NewAAAClientFromConfig(r.Context())
```
julsemaan: This creates tons of useless calls to pfconfig, you should create the client once and reuse it

julsemaan (changes requested):

go/jsonrpc2/client.go
```diff
@@ -0,0 +1,168 @@
...
+type JsonRPC2Response struct {
+	JsonRPC string         `json:"jsonrpc"`
+	Result  interface{}    `json:"result,omitempty"`
+	Error   *JsonRPC2Error `json:"error,omitempty"`
+	Id      uint           `json:"id"`
+}
+
+func NewClientFromConfig(ctx context.Context) *Client {
+	var webservices pfconfigdriver.PfConfWebservices
+	pfconfigdriver.FetchDecodeSocket(ctx, &webservices)
```
julsemaan, line 69: You should use a global variable for webservices and use FetchDecodeSocketCache here. It will prevent calls to pfconfig when the data hasn't changed

go/jsonrpc2/client.go
```diff
@@ -0,0 +1,168 @@
...
+		Host:     webservices.Host,
+		Port:     webservices.Port,
+	}
+}
+
+func NewAAAClientFromConfig(ctx context.Context) *Client {
+	var webservices pfconfigdriver.PfConfWebservices
+	var ports pfconfigdriver.PfConfPorts
+	pfconfigdriver.FetchDecodeSocket(ctx, &webservices)
+	pfconfigdriver.FetchDecodeSocket(ctx, &ports)
```
julsemaan, line 83: You should use a global variable for webservices and use FetchDecodeSocketCache here. It will prevent calls to pfconfig when the data hasn't changed

julsemaan (changes requested):
The bandwidth accounting security event triggers can't currently work since that still uses the radacct tables from what I can tell.

Also, the time buckets have a timestamp of 0000-00-00 00:00:00 and also don't seem to rotate and create new buckets.

go/acct/radius.go
```diff
@@ -0,0 +1,360 @@
...
+
+	callingStation := rfc2865.CallingStationID_GetString(r.Packet)
+	mac, _ := mac.NewFromString(callingStation)
+	in_bytes := int64(rfc2866.AcctInputOctets_Get(r.Packet))
+	out_bytes := int64(rfc2866.AcctOutputOctets_Get(r.Packet))
+	giga_in_bytes := int64(rfc2869.AcctInputGigawords_Get(r.Packet))
+	giga_out_bytes := int64(rfc2869.AcctOutputGigawords_Get(r.Packet))
+	out_bytes += giga_out_bytes << 32
+	in_bytes += giga_in_bytes << 32
+	timestamp := rfc2869.EventTimestamp_Get(r.Packet)
```
julsemaan, line 67: I think this may be why the time buckets on my deployment are at 0

If this is a zero value, you should default to now for the timestamp

go/acct/pfacct.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"database/sql"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/inverse-inc/packetfence/go/db"
+	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
+	"net"
+	"time"
+)
+
+const DefaultTimeDuration = 5 * time.Minute
```
julsemaan, line 15: This should be configurable

julsemaan:
I got it to switch time buckets but it doesn't update the new bucket:
```
MariaDB [pf]> select * from bandwidth_accounting;
+-----------+-------------------+----------------------------------+---------------------+----------+-----------+-------------+
| tenant_id | mac               | unique_session_id                | time_bucket         | in_bytes | out_bytes | total_bytes |
+-----------+-------------------+----------------------------------+---------------------+----------+-----------+-------------+
|         1 | a0:00:00:00:00:01 | 40ae44694e00db971e168111e318421c | 2020-03-10 07:55:00 |     7000 |     14000 |       21000 |
|         1 | a0:00:00:00:00:01 | 40ae44694e00db971e168111e318421c | 2020-03-10 08:00:00 |        0 |         0 |           0 |
+-----------+-------------------+----------------------------------+---------------------+----------+-----------+-------------+
```

I sent the same series of RADIUS requests at an interval of 5 minutes to get this result

NOTE: What I did was to send the packets for the same session ID twice in an interval of 5 minutes. We'd have to see what we expect of this as adding it to the 2nd bucket as well would result in doubled accounting

julsemaan:
I'm getting an insane amount of the following message when stress testing the service:
```
Mar 10 08:24:06 localhost pfacct[15384]: t=2020-03-10T08:24:05-0400 lvl=warn msg="Not handled Event-Timestamp\n" pid=15384
```

go/acct/main.go
```diff
@@ -0,0 +1,60 @@
...
+	if processor != nil {
+		processor.Start()
+	}
+	w.Wait()
+}
+
+func NotifySystemd(msg string) {
+	_, err := daemon.SdNotify(false, msg)
+	if err != nil {
+		log.LoggerWContext(context.Background(), fmt.Sprintf("Error sending systemd ready notification: %s", err.Error()))
```
julsemaan, line 64: You should use a global context with a logger in it, otherwise you'll be creating a new logger everytime

go/acct/net_flow.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (h *PfAcct) netflowProcessor() (*processor.Processor, error) {
+	addr := "127.0.0.1:" + h.NetFlowPort
+	conn, err := net.ListenPacket("udp", addr)
+	if err != nil {
+		return nil, err
+	}
+
+	log.LoggerWContext(context.Background()).Info("Starting listening to netflow at '" + addr + "'")
```
julsemaan, line 146: You should use a global context with a logger in it, otherwise you'll be creating a new logger everytime

go/acct/radius.go
```diff
@@ -0,0 +1,360 @@
...
+		} else {
+			attributes[da.Name] = item
+		}
+	} else {
+		logWarn(fmt.Sprintf("Not handled %s\n", da.Name))
+	}
+}
+
+func logError(msg string) {
+	log.LoggerWContext(context.Background()).Error(msg)
```
julsemaan: You should use a global context with a logger in it, otherwise you'll be creating a new logger everytime

go/acct/radius.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func logWarn(msg string) {
+	log.LoggerWContext(context.Background()).Warn(msg)
```
julsemaan: You should use a global context with a logger in it, otherwise you'll be creating a new logger everytime

go/acct/radius.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func logInfo(msg string) {
+	log.LoggerWContext(context.Background()).Info(msg)
```
julsemaan: You should use a global context with a logger in it, otherwise you'll be creating a new logger everytime

julsemaan (changes requested):
The logging strategy causes issues of syslog delivery errors when there are a lot of concurrent requests. Likely because a new logger is created everytime which opens a new file descriptor

go/jsonrpc2/client.go
jrouzierinverse: Done

go/acct/radius.go
jrouzierinverse: Fixed

go/jsonrpc2/client.go
jrouzierinverse: I only create one client. So need to do it here for now.

julsemaan (changes requested):
This PR doesn't seem to rework/optimize the core code of the RADIUS accounting in Perl.

I still see a lot of stress being put on the DB for RADIUS accounting packets and the httpd.aaa is slow to respond to pfacct + puts a lot of stress on the DB.

@jrouzierinverse, can you confirm if you completed the profiling of the *whole* code that is called in the httpd.aaa as discussed previously. We'll definitely need to get this to perform better.

julsemaan:
When sending 5 concurrent requests to the daemon I'm getting this:
```
Mar 11 08:43:38 localhost pfacct: panic: Unix syslog delivery error
Mar 11 08:43:38 localhost pfacct: goroutine 4418 [running]:
Mar 11 08:43:38 localhost pfacct: github.com/inverse-inc/packetfence/go/sharedutils.CheckError(...)
Mar 11 08:43:38 localhost pfacct: /root/rpmbuild/centos-7/BUILD/packetfence-9.3.9/go/sharedutils/util.go:119
Mar 11 08:43:38 localhost pfacct: github.com/inverse-inc/packetfence/go/log.initContextLogger(0x8e33c0, 0xc00144b5f0, 0x822540, 0x448b01)
Mar 11 08:43:38 localhost pfacct: /root/rpmbuild/centos-7/BUILD/packetfence-9.3.9/go/log/log.go:113 +0x4a0
Mar 11 08:43:38 localhost pfacct: github.com/inverse-inc/packetfence/go/log.LoggerNewContext(0x8e33c0, 0xc00144b5f0, 0x7b6a20, 0x8d2100)
Mar 11 08:43:38 localhost pfacct: /root/rpmbuild/centos-7/BUILD/packetfence-9.3.9/go/log/log.go:182 +0x39
Mar 11 08:43:38 localhost pfacct: main.(*PfAcct).RADIUSSecret(0xc0003479e0, 0x8e3300, 0xc0001e0e40, 0x8df4e0, 0xc00144a060, 0xc000a43e60, 0x8e, 0x90, 0xc000b56640, 0x406dac, ...)
Mar 11 08:43:38 localhost pfacct: /root/rpmbuild/centos-7/BUILD/packetfence-9.3.9/go/acct/radius.go:179 +0x375
Mar 11 08:43:38 localhost pfacct: github.com/inverse-inc/go-radius.(*PacketServer).Serve.func2(0xc000449a00, 0xc0002a0960, 0xc0002b2630, 0x8e5220, 0xc0000922f8, 0xc000a43e60, 0x8e, 0x90, 0x8df4e0, 0xc00144a060)
Mar 11 08:43:38 localhost pfacct: /root/go/pkg/mod/github.com/inverse-inc/go-radius@v0.0.0-20200310093817-85565fe93aa5/server-packet.go:136 +0xe2
Mar 11 08:43:38 localhost pfacct: created by github.com/inverse-inc/go-radius.(*PacketServer).Serve
Mar 11 08:43:38 localhost pfacct: /root/go/pkg/mod/github.com/inverse-inc/go-radius@v0.0.0-20200310093817-85565fe93aa5/server-packet.go:133 +0x3e5
Mar 11 08:43:38 localhost pfacct[27868]: t=2020-03-11T08:43:37-0400 lvl=eror msg="Post http://127.0.0.1:7070: dial tcp 127.0.0.1:7070: i/o timeout" pid=27868
Mar 11 08:43:38 localhost pfacct[27868]: t=2020-03-11T08:43:37-0400 lvl=eror msg="Post http://127.0.0.1:7070: dial tcp 127.0.0.1:7070: i/o timeout" pid=27868
Mar 11 08:43:38 localhost pfacct[27868]: t=2020-03-11T08:43:37-0400 lvl=eror msg="Post http://127.0.0.1:7070: dial tcp 127.0.0.1:7070: i/o timeout" pid=27868
Mar 11 08:43:38 localhost pfacct[27868]: t=2020-03-11T08:43:37-0400 lvl=eror msg="Post http://127.0.0.1:7070: dial tcp 127.0.0.1:7070: i/o timeout" pid=27868
```

So errors in both calling the httpd.aaa and the logging

You can replicate the issue using:
```
wget https://support.inverse.ca/~jsemaan/acct-tester
chmod +x acct-tester
./acct-tester -host=192.168.1.5 -secret=`cat /usr/local/pf/conf/local_secret` -lt-nodes-count=500 -lt-concurrency=5
```

Replacing 192.168.1.5 by your mgmt IP

lib/pf/api.pm
```diff
@@ -1487,53 +1455,55 @@ sub radius_rest_accounting :Public :RestPath(/radius/rest/accounting) {
...
-    if ($RAD_REQUEST{'Acct-Status-Type'} != $ACCOUNTING::STOP){
+    if ($acct_status_type != $ACCOUNTING::STOP) {
+        my $advanced = $pf::config::Config{advanced};
         # Tracking IP address.
-        if(pf::util::isenabled($pf::config::Config{advanced}{update_iplog_with_accounting})){
+        my $framed_ip = $RAD_REQUEST->{"Framed-IP-Address"};
+        if ($framed_ip && pf::util::isenabled($advanced->{update_iplog_with_accounting})) {
             $logger->info("Updating iplog from accounting request");
-            $client->notify("update_ip4log", mac => $mac, ip => $RAD_REQUEST{'Framed-IP-Address'}) if ($RAD_REQUEST{'Framed-IP-Address'} );
+            $client->notify("update_ip4log", mac => $mac, ip => $framed_ip);
```
fdurand, line 1480: To get rid of a call , can we test if the $framed_ip is a valid ip ?

lib/pf/api.pm
```diff
@@ -1487,53 +1455,55 @@ sub radius_rest_accounting :Public :RestPath(/radius/rest/accounting) {
...
         }
         else {
-            pf::log::get_logger->debug("Not handling iplog update because we're not configured to do so on accounting packets.");
+            $logger->debug("Not handling iplog update because we're not configured to do so on accounting packets.");
         }
         
-        if(pf::util::isenabled($pf::config::Config{advanced}{scan_on_accounting}) && $RAD_REQUEST{"Framed-IP-Address"}) {
+        if ($framed_ip && pf::util::isenabled($advanced->{scan_on_accounting})) {
```
fdurand, line 1486: if (valid_ip($framed_ip) && ...

go/acct/radius.go
```diff
@@ -141,7 +141,7 @@ func (h *PfAcct) sendRadiusAccounting(r *radius.Request) {
 		"X-FreeRADIUS-Section": "accounting",
 	}
 
-	if err := h.AAAClient.Notify("radius_accounting", attr, 1); err != nil {
+	if _, err := h.AAAClient.Call("radius_accounting", attr, 1); err != nil {
```
fdurand: Notify or Call ?

jrouzierinverse: Call for now

julsemaan: Better to call for now since otherwise pfacct accepts too many concurrent requests which can bring it to a halt

julsemaan:
When I perform sequential testing of pfacct vs the previous implementation on 9.3, it performs with the same timings

Could be because of the following comment I made earlier today:

> This PR doesn't seem to rework/optimize the core code of the RADIUS accounting in Perl.
> 
> I still see a lot of stress being put on the DB for RADIUS accounting packets and the httpd.aaa is slow to respond to pfacct + puts a lot of stress on the DB.
> 
> @jrouzierinverse, can you confirm if you completed the profiling of the _whole_ code that is called in the httpd.aaa as discussed previously. We'll definitely need to get this to perform better.

julsemaan:
Results of a basic performance test (single threaded with empty DB)

This uses the acct-tester in go/acct-tester
```
time ./acct-tester -host=172.20.155.47 -secret=`cat /usr/local/pf/conf/local_secret` -lt-nodes-count=500 -lt-concurrency=1
```

Results:
 * `packetfence-9.3.0`: 1m30s
 * `packetfence-9.3.9-20200210103550.125841666.0007.feature~optimize~handle~accounting~metadata`: 1m
 * `packetfence-9.3.9-20200312193611.125840269.0007.feature~pfacct`: 45s

julsemaan (changes requested):
One thing that was a design requirement that is either missing or just not working:

Initially we discussed that the buckets would be 5 minutes but as they get older they get summarized into buckets of 1 hour, then buckets of 1 day, etc

Right now I only see buckets of 5 minutes so that means that for our 70K daily endpoints goal we would get 288 buckets per day which is around 2 million rows for a single day of accounting.

We will need to get these summarized buckets in place and potentially partitioning to help with table size

go/jsonrpc2/client.go
```diff
@@ -0,0 +1,178 @@
...
+
+	r, err := c.buildRequest(&request)
+	if err != nil {
+		return nil, err
+	}
+
+	start := time.Now()
+	resp, err := httpClient.Do(r)
+	elapsed := time.Now().Sub(start)
+	log.LoggerWContext(ctx).Info(fmt.Sprintf("jsonrpc2.Call took %v for %s", elapsed, method))
```
julsemaan: This is way too verbose, probably debug is better

julsemaan (changes requested):
I'm getting rows with a MAC of 00:00:00:00:00:00:
```
|         1 | 00:00:00:00:00:00 | ff95122251ef985a475e46d522f91ca8 | 2020-03-13 09:00:00 |     2302 |      4302 |        6604 |         0 |
```

Can be replicated using:
```
./acct-tester -host=172.20.155.47 -secret=`cat /usr/local/pf/conf/local_secret` -lt-nodes-count=500 -lt-concurrency=1 -lt-max-interim-per-node=1
```

I did a radsniff and I don't see a Calling-Station-Id with this value and no warnings/errors in the pfacct.log

jrouzierinverse:
I fixed the parsing issue with calling station id.

db/pf-schema-X.Y.Z.sql
```diff
@@ -1528,6 +1530,23 @@ CREATE TABLE `pki_revoked_certs` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
+--
+-- Table structure for table `bandwidth_accounting`
+--
+
+CREATE TABLE bandwidth_accounting (
+    tenant_id INT,
+    mac char(17) NOT NULL,
+    unique_session_id char(32),
```
satkunas: `mac` and `unique_session_id` should be INT columns to drastically increase index performance.

Re `mac`: strip chars ![a-f0-9] and store as either BINARY OR BIGINT
Re `unique_session_id`: BINARY(16) === CHAR(32) (as hex)
Re: `tenant_id`: Ideally UNSIGNED SMALLINT or UNSIGNED MEDIUMINT since we don't need negative ids. INT increases the index sizes beyond what's needed, thus unnecessarily increasing INDEX lookup performance.

db/upgrade-X.X.X-X.Y.Z.sql
```diff
@@ -221,6 +221,25 @@ CREATE TABLE IF NOT EXISTS dhcppool (
   KEY released (released)
 ) ENGINE=INNODB;
 
+\! echo "Adding new table bandwidth_accounting";
+CREATE TABLE IF NOT EXISTS bandwidth_accounting (
+    tenant_id INT,
+    mac char(17) NOT NULL,
+    unique_session_id char(32) NOT NULL,
```
satkunas, line 228: Same as above.

julsemaan:
> One thing that was a design requirement that is either missing or just not working:
> 
> Initially we discussed that the buckets would be 5 minutes but as they get older they get summarized into buckets of 1 hour, then buckets of 1 day, etc
> 
> Right now I only see buckets of 5 minutes so that means that for our 70K daily endpoints goal we would get 288 buckets per day which is around 2 million rows for a single day of accounting.
> 
> We will need to get these summarized buckets in place and potentially partitioning to help with table size

@jrouzierinverse, we will need to get this adjusted, I have tested with 2.6 million rows to get the daily accounting of a device and it takes 2.77 seconds.

I think the following bucket aggregation would be good:
- last hour should have buckets of 5 minutes
- last day should have buckets of 1 hour
- last month should have buckets of 1 day
- last year should have buckets of 1 month

This means that if we keep accounting for 24 hours like we currently do we'll have at most 35 buckets (12 for the last hour, 23 for the 23 other hours)

Once you get the acct_maintenance job adjusted as well, we'll have to check the performance of that query as it will be used a lot when having accounting triggers. Also, we'll have to see your adjustments for the accounting report and ensure that the bucket size will also work

julsemaan (changes requested):
We need to have the cleanup jobs of pfmon adjusted for the new table structure

satkunas:
>  I have tested with 2.6 million rows to get the daily accounting of a device and it takes 2.77 seconds.

@julsemaan I would love to see this same bench with INT columns

jrouzierinverse:
I updated the index the speed is much faster on my table of 2 million records.

julsemaan:
@jrouzierinverse, we still need to have the aggregation of buckets since my initial calculation for 70K endpoints was wrong, if we keep it at 5 minutes, its 20 million records that we'll have, with my suggested aggregation we'll have 2.4 million

Also, please expand on the much faster, is it 1 second instead of 2 which would still be slow

jrouzierinverse:
1.5329 ms

jrouzierinverse:
Just tested the query on a 19872000 record table.
The total time was 2.2128 ms

julsemaan:
We still need the aggregation as we discussed in the design meeting, the 70K endpoints for 1 day is the basic requirement.

We should be able to keep accounting data for days/months with our new design. 
Without aggregation this will lead to enormous tables and indexes on disk and in memory.

Also, the query we looked at is the simplest query we have, we haven't got into reports and acct_maintenance yet which contain much more complex queries.

jrouzierinverse:
Fixing the test query it is now 0.6283 ms

julsemaan:
As discussed with @jrouzierinverse yesterday, this should be feature complete by the end of today.

I've summarized the missing points of the PR here, feel free to add other stuff that might be missing:

- [x] Summarize/aggregate buckets to reduce row count
- [ ] Adjust the acct_maintenance job to count accounting data usage in the new table
- [ ] Adjust the existing accounting reports to use the new table
- [x] Adjust the existing cleanup job to use the new table
- [ ] Go through all the other accounting queries that use the radacct and radacct_log tables and adjust them
- [ ] Maintenance task form
- [ ] Add cleanup to bandwidth_accounting table
- [x] The size of the buckets needs to be configurable

The points above are necessary for the feature completeness of this.

Stuff we could do post merge:
- [ ] Evaluate the usage of partitioning with @satkunas and optimize some columns (like using an int for the MAC, etc)

satkunas:
@julsemaan will admin > reporting be affected as well?

julsemaan:
> @julsemaan will admin > reporting be affected as well?

At **least** the following reports in the admin are impacted:
image

julsemaan:
While running some tests I found that pfacct is unable to insert rows in the bandwidth_accounting table with 20 million rows in it.

To replicate:
```
./db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=mysupersecurepassouardh -endpoints=70000 -start-date="2000-01-01 00:00:00" -rows=20160000
./acct-tester -host=172.20.155.47 -secret=`cat /usr/local/pf/conf/local_secret` -lt-nodes-count=500
```

html/pfappserver/lib/pfappserver/Model/DB.pm
```diff
@@ -45,7 +45,7 @@ sub assign {
 
     # Create global PF user
     foreach my $host ("'%'","localhost") {
-        my $sql_query = "GRANT SELECT,INSERT,UPDATE,DELETE,EXECUTE,LOCK TABLES ON $db.* TO ?\@${host} IDENTIFIED BY ?";
+        my $sql_query = "GRANT SELECT,INSERT,UPDATE,DELETE,EXECUTE,LOCK TABLES, CREATE TEMPORARY TABLE ON $db.* TO ?\@${host} IDENTIFIED BY ?";
```
julsemaan, line 48: Do we really need this grant? If so then it should be in the upgrade script for v10

jrouzierinverse: That was part of some possible changes needed for the aggregation.
I committed it by mistake. If in the final code it is not needed I will remove it.

julsemaan:
> While running some tests I found that pfacct is unable to insert rows in the bandwidth_accounting table with 20 million rows in it.
> 
> To replicate:
> 
> ```
> ./db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=mysupersecurepassouardh -endpoints=70000 -start-date="2000-01-01 00:00:00" -rows=20160000
> ./acct-tester -host=172.20.155.47 -secret=`cat /usr/local/pf/conf/local_secret` -lt-nodes-count=500
> ```

I retested this morning and I'm obtaining timings that are close to when the table is empty so the fix @jrouzierinverse did works fine

julsemaan:
I ran my tests again to validate the maths work for the accounting and all is fine as before.

I also ran the test with 500 nodes and the perf is on par with what I had a few days ago on this branch

julsemaan:
I have a bunch of errors during aggregation:
```
Mar 20 08:42:09 localhost pfmon: pfmon(20880) ERROR: [mac:unknown] Database query failed with non retryable error: PROCEDURE pf.process_bandwidth_accounting does not exist (errno: 1305) [CALL process_bandwidth_accounting(?);]{100} (pf::dal::db_execute)
Mar 20 08:42:09 localhost pfmon: pfmon(20880) ERROR: [mac:unknown] Error calling process_bandwidth_accounting (pf::bandwidth_accounting::call_process_bandwidth_accounting)
Mar 20 08:42:09 localhost pfmon: pfmon(20880) INFO: [mac:unknown] deleted 0 for process_bandwidth_accounting (1584708129 1584708129)  (pf::bandwidth_accounting::process_bandwidth_accounting)
Mar 20 08:42:09 localhost pfmon: pfmon(20880) ERROR: [mac:unknown] Error running task bandwidth_maintenance: Undefined subroutine &pf::bandwidth_accounting::is_success called at /usr/local/pf/lib/pf/bandwidth_accounting.pm line 44.

```

julsemaan:
Re-commenting the TODO since it was now lost in a pile of commits and was hidden on Github:

- [x] Summarize/aggregate buckets to reduce row count
- [x] Adjust the acct_maintenance job to count accounting data usage in the new table
- [x] Adjust the existing accounting reports to use the new table
- [x] Adjust the existing cleanup job to use the new table
- [ ] Go through all the other accounting queries that use the radacct and radacct_log tables and adjust them
- [x] Maintenance task form
- [x] Add cleanup to bandwidth_accounting table
- [x] The size of the buckets needs to be configurable

The points above are necessary for the feature completeness of this.

Stuff we could do post merge:
- [ ] Evaluate the usage of partitioning with @satkunas and optimize some columns (like using an int for the MAC, etc)

julsemaan:
Still not able to get a build working, getting the following error now to start pfmon:

```
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: pfmon(18954) FATAL: [mac:[undef]] Attribute (history_timeout) does not pass the type constraint because: Validation failed for 'PfInterval' with value 10s
                                                    2D at /usr/lib64/perl5/vendor_perl/Moose/Meta/Attribute.pm line 1279.
                                                            Moose::Meta::Attribute::verify_against_type_constraint('Moose::Meta::Attribute=HASH(0xa1dbcd8)', '10s\x{a}2D', 'instance', 'pf::pfmon::task::bandwidth_maintenance=HASH(0xada9b20)') called at /usr/lib64/perl5/vendor_perl/Moose/Meta/Attribute.pm line 1266
                                                            Moose::Meta::Attribute::_coerce_and_verify('Moose::Meta::Attribute=HASH(0xa1dbcd8)', '10s\x{a}2D', 'pf::pfmon::task::bandwidth_maintenance=HASH(0xada9b20)') called at /usr/lib64/perl5/vendor_perl/Moose/Meta/Attribute.pm line 536
                                                            Moose::Meta::Attribute::initialize_instance_slot('Moose::Meta::Attribute=HASH(0xa1dbcd8)', 'Moose::Meta::Instance=HASH(0xa1db4b0)', 'pf::pfmon::task::bandwidth_maintenance=HASH(0xada9b20)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Class/MOP/Class.pm line 525
                                                            Class::MOP::Class::_construct_instance('Moose::Meta::Class=HASH(0xa1d69f0)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Class/MOP/Class.pm line 498
                                                            Class::MOP::Class::new_object('Moose::Meta::Class=HASH(0xa1d69f0)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Moose/Meta/Class.pm line 284
                                                            Moose::Meta::Class::new_object('Moose::Meta::Class=HASH(0xa1d69f0)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Moose/Object.pm line 28
                                                            Moose::Object::new('pf::pfmon::task::bandwidth_maintenance', 'HASH(0xada9a18)') called at /usr/local/pf/lib/pf/factory/pfmon/task.pm line 47
                                                            pf::factory::pfmon::task::new('pf::factory::pfmon::task', 'bandwidth_maintenance') called at /usr/local/pf/sbin/pfmon line 156
                                                            main::registertasks() called at /usr/local/pf/sbin/pfmon line 126
                                                            main::start() called at /usr/local/pf/sbin/pfmon line 131
                                                     (Moose::Meta::Class::raise_error)
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Attribute (history_timeout) does not pass the type constraint because: Validation failed for 'PfInterval' with value 10s
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: 2D at /usr/lib64/perl5/vendor_perl/Moose/Meta/Attribute.pm line 1279.
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Moose::Meta::Attribute::verify_against_type_constraint('Moose::Meta::Attribute=HASH(0xa1dbcd8)', '10s\x{a}2D', 'instance', 'pf::pfmon::task::bandwidth_maintenance=HASH(0xada9b20)') called at /usr/lib64/perl5/vendor_perl/Moose/Meta/Attribute.pm line 1266
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Moose::Meta::Attribute::_coerce_and_verify('Moose::Meta::Attribute=HASH(0xa1dbcd8)', '10s\x{a}2D', 'pf::pfmon::task::bandwidth_maintenance=HASH(0xada9b20)') called at /usr/lib64/perl5/vendor_perl/Moose/Meta/Attribute.pm line 536
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Moose::Meta::Attribute::initialize_instance_slot('Moose::Meta::Attribute=HASH(0xa1dbcd8)', 'Moose::Meta::Instance=HASH(0xa1db4b0)', 'pf::pfmon::task::bandwidth_maintenance=HASH(0xada9b20)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Class/MOP/Class.pm line 525
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Class::MOP::Class::_construct_instance('Moose::Meta::Class=HASH(0xa1d69f0)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Class/MOP/Class.pm line 498
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Class::MOP::Class::new_object('Moose::Meta::Class=HASH(0xa1d69f0)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Moose/Meta/Class.pm line 284
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Moose::Meta::Class::new_object('Moose::Meta::Class=HASH(0xa1d69f0)', 'HASH(0xad9d710)') called at /usr/lib64/perl5/vendor_perl/Moose/Object.pm line 28
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: Moose::Object::new('pf::pfmon::task::bandwidth_maintenance', 'HASH(0xada9a18)') called at /usr/local/pf/lib/pf/factory/pfmon/task.pm line 47
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: pf::factory::pfmon::task::new('pf::factory::pfmon::task', 'bandwidth_maintenance') called at /usr/local/pf/sbin/pfmon line 156
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: main::registertasks() called at /usr/local/pf/sbin/pfmon line 126
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: main::start() called at /usr/local/pf/sbin/pfmon line 131
Mar 20 11:52:23 localhost.localdomain pfmon[18954]: pfmon(18954) INFO: [mac:[undef]] stopping pfmon (main::)
Mar 20 11:52:23 localhost.localdomain systemd[1]: packetfence-pfmon.service: main process exited, code=exited, status=2/INVALIDARGUMENT
Mar 20 11:52:23 localhost.localdomain systemd[1]: Failed to start PacketFence pfmon Service.
Mar 20 11:52:23 localhost.localdomain systemd[1]: Unit packetfence-pfmon.service entered failed state.
Mar 20 11:52:23 localhost.localdomain systemd[1]: packetfence-pfmon.service failed.
Mar 20 11:52:23 localhost.localdomain systemd[1]: packetfence-pfmon.service holdoff time over, scheduling restart.
Mar 20 11:52:23 localhost.localdomain systemd[1]: Stopped PacketFence pfmon Service.
```

julsemaan:
New error in aggregation:
```
Mar 20 12:49:06 localhost pfmon: pfmon(27475) ERROR: [mac:unknown] Database query failed with non retryable error: Unknown column 'to_process.mac' in 'where clause' (errno: 1054) [CALL process_bandwidth_accounting(?);]{100} (pf::dal::db_execute)
Mar 20 12:49:06 localhost pfmon: pfmon(27475) ERROR: [mac:unknown] Error calling process_bandwidth_accounting (pf::bandwidth_accounting::call_process_bandwidth_accounting)
```

julsemaan:
The data seems to be hourly aggregated but not daily aggregated (haven't tried monthly but my guess is it might not work either)

To replicate:
```
/root/db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=caca -endpoints=1 -start-date="2020-03-19 00:00:00" -rows=288
/root/db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=caca -endpoints=1 -start-date="2020-03-20 00:00:00" -rows=144
```

I'd expect the data of 2020-03-19 to be daily aggregated

db/pf-schema-X.Y.Z.sql
```diff
@@ -1528,6 +1530,127 @@ CREATE TABLE `pki_revoked_certs` (
...
+             mac,
+             new_time_bucket,
+             sum(in_bytes) AS in_bytes,
+             sum(out_bytes) AS out_bytes
+            FROM to_delete
+            GROUP BY tenant_id, mac, new_time_bucket
+            ON DUPLICATE KEY UPDATE
+                in_bytes = in_bytes + VALUES(in_bytes),
+                out_bytes = out_bytes + VALUES(out_bytes)
+            ;
```
satkunas, line 1599: `total_bytes` also needs to be aggregated to `bandwidth_accounting_history`.

jrouzierinverse: It is automatically since it is a generated field.
https://github.com/inverse-inc/packetfence/pull/5181/files/3c544a9f3fb69659dd5950f934ae9ec1cc501744#diff-9a3daa21c2b9052d5f85954a88f0592cR1560

https://mariadb.com/kb/en/generated-columns/

julsemaan:
Preliminary definition: interruption means the service takes more than 1 second to respond to at least 1 packet

I first loaded the DB with some data for the past day:
```
 /root/db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=caca -endpoints=70000 -start-date="2020-03-22 00:00:00" -rows=70000
 /root/db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=caca -endpoints=70000 -start-date="2020-03-23 00:00:00" -rows=980000 -bucket-size=3600
 /root/db-tester -db-host=172.20.155.47 -db-user=pf -db-pass=caca -endpoints=70000 -start-date="2020-03-23 14:00:00" -rows=840000
```

With 70K endpoints with data that spans across a day, I was able to run continuous service without interruption using:
```
/root/acct-tester -host=172.20.155.47 -secret=`cat /usr/local/pf/conf/local_secret` -lt-nodes-count=70000 -lt-min-interim-per-node=100000 -lt-concurrency=5
```

During that test I also visited the reports in their default view without any problem.
I also viewed the reports with the data of a year and although they took ~20-30 seconds to render it didn't interrupt the service.

But, I've had interruption of service that occurred randomly without a specific interval where I'd always have 5 packets that would timeout (concurrency is set to 5, so a single transaction per worker was problematic) and then it would come back and work correctly right after. For this I didn't see any errors in the logs of pfacct.log, pfmon.log or packetfence.log so I'll need help from @jrouzierinverse to be able to determine more precisely where this is coming from.

I will also start running a continuous test to fill the DB through pfacct only. 
Will have results tomorrow.

julsemaan:
The slow queries seem to be in the chi_cache table:
```
# Time: 200323 16:06:11
# User@Host: pf[pf] @ localhost []
# Thread_id: 179193  Schema: pf  QC_hit: No
# Query_time: 1.419698  Lock_time: 0.000038  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 2
SET timestamp=1584993971;
INSERT INTO `chi_cache` ( `key`, `value`, `expires_at` ) VALUES ( 'accounting::a0:00:00:00:c1:22', '<B1>^Vy^<FF><FF><FF><FF><FF><FF><FF><FF>^A^A=<F3>rl^C\0(*^LrAcct-Output-Octets <F1><91>^CqAcct-Input-Octets <89><8A>^CoAcct-Session-Idxacct-tester-a0000000c122qAcct-Session-Time^ArCalling-Station-Idqa0:00:00:00:c1:22qCalled-Station-Idq02:00:00:00:00:01iUser-NamesUserOF-a0000000c122oEvent-Timestamp}2020-03-23 16:06:09 -0400 EDTpAcct-Status-Type^AqFramed-IP-Addressk10.0.193.34hNAS-Port ^TnNAS-IP-Addressi127.0.0.1', '2147483641' ) ON DUPLICATE KEY UPDATE `value`=VALUES(`value`), `expires_at`=VALUES(`expires_at`);
# Time: 200323 16:06:45
# User@Host: pf[pf] @ localhost []
# Thread_id: 179184  Schema: pf  QC_hit: No
# Query_time: 1.115606  Lock_time: 0.000044  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 2
SET timestamp=1584994005;
INSERT INTO `chi_cache` ( `key`, `value`, `expires_at` ) VALUES ( 'accounting::a0:00:00:00:d7:6e', '<D4>^Vy^<FF><FF><FF><FF><FF><FF><FF><FF>^A^A=<F3>rl^C\0(*^LrAcct-Output-Octets <BD><BE>^CqAcct-Input-Octets ն^CoAcct-Session-Idxacct-tester-a0000000d76eqAcct-Session-Time^ArCalling-Station-Idqa0:00:00:00:d7:6eqCalled-Station-Idq02:00:00:00:00:01iUser-NamesUserOF-a0000000d76eoEvent-Timestamp}2020-03-23 16:06:44 -0400 EDTpAcct-Status-Type^AqFramed-IP-Addressl10.0.215.110hNAS-Port ^TnNAS-IP-Addressi127.0.0.1', '2147483641' ) ON DUPLICATE KEY UPDATE `value`=VALUES(`value`), `expires_at`=VALUES(`expires_at`);
# User@Host: pf[pf] @ localhost []
# Thread_id: 179165  Schema: pf  QC_hit: No
# Query_time: 1.114968  Lock_time: 0.000040  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 2
SET timestamp=1584994005;
INSERT INTO `chi_cache` ( `key`, `value`, `expires_at` ) VALUES ( 'accounting::a0:00:00:00:d9:42', '<D4>^Vy^<FF><FF><FF><FF><FF><FF><FF><FF>^A^A=<F3>rl^C\0(*^LrAcct-Output-Octets <91><C2>^CqAcct-Input-Octets <A9><BA>^CoAcct-Session-Idxacct-tester-a0000000d942qAcct-Session-Time^ArCalling-Station-Idqa0:00:00:00:d9:42qCalled-Station-Idq02:00:00:00:00:01iUser-NamesUserOF-a0000000d942oEvent-Timestamp}2020-03-23 16:06:44 -0400 EDTpAcct-Status-Type^AqFramed-IP-Addressk10.0.217.66hNAS-Port ^TnNAS-IP-Addressi127.0.0.1', '2147483641' ) ON DUPLICATE KEY UPDATE `value`=VALUES(`value`), `expires_at`=VALUES(`expires_at`);
# User@Host: pf[pf] @ localhost []
# Thread_id: 179162  Schema: pf  QC_hit: No
# Query_time: 1.233945  Lock_time: 0.000037  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 2
SET timestamp=1584994005;
INSERT INTO `chi_cache` ( `key`, `value`, `expires_at` ) VALUES ( 'accounting::a0:00:00:00:d9:59', '<D4>^Vy^<FF><FF><FF><FF><FF><FF><FF><FF>^A^A=<F3>rl^C\0(*^LrAcct-Output-Octets <A8><C2>^CqAcct-Input-Octets <C0><BA>^CoAcct-Session-Idxacct-tester-a0000000d959qAcct-Session-Time^ArCalling-Station-Idqa0:00:00:00:d9:59qCalled-Station-Idq02:00:00:00:00:01iUser-NamesUserOF-a0000000d959oEvent-Timestamp}2020-03-23 16:06:44 -0400 EDTpAcct-Status-Type^AqFramed-IP-Addressk10.0.217.89hNAS-Port ^TnNAS-IP-Addressi127.0.0.1', '2147483641' ) ON DUPLICATE KEY UPDATE `value`=VALUES(`value`), `expires_at`=VALUES(`expires_at`);
# User@Host: pf[pf] @ localhost []
# Thread_id: 179166  Schema: pf  QC_hit: No
# Query_time: 1.236384  Lock_time: 0.000036  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 2
SET timestamp=1584994005;
INSERT INTO `chi_cache` ( `key`, `value`, `expires_at` ) VALUES ( 'accounting::a0:00:00:00:d9:2a', '<D4>^Vy^<FF><FF><FF><FF><FF><FF><FF><FF>^A^A=<F3>rl^C\0(*^LrAcct-Output-Octets <F9><C1>^CqAcct-Input-Octets <91><BA>^CoAcct-Session-Idxacct-tester-a0000000d92aqAcct-Session-Time^ArCalling-Station-Idqa0:00:00:00:d9:2aqCalled-Station-Idq02:00:00:00:00:01iUser-NamesUserOF-a0000000d92aoEvent-Timestamp}2020-03-23 16:06:44 -0400 EDTpAcct-Status-Type^AqFramed-IP-Addressk10.0.217.42hNAS-Port ^TnNAS-IP-Addressi127.0.0.1', '2147483641' ) ON DUPLICATE KEY UPDATE `value`=VALUES(`value`), `expires_at`=VALUES(`expires_at`);
# User@Host: pf[pf] @ localhost []
# Thread_id: 179193  Schema: pf  QC_hit: No
# Query_time: 1.311881  Lock_time: 0.000032  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 2
SET timestamp=1584994005;
INSERT INTO `chi_cache` ( `key`, `value`, `expires_at` ) VALUES ( 'accounting::a0:00:00:00:d9:3f', '<D4>^Vy^<FF><FF><FF><FF><FF><FF><FF><FF>^A^A=<F3>rl^C\0(*^LrAcct-Output-Octets <8E><C2>^CqAcct-Input-Octets <A6><BA>^CoAcct-Session-Idxacct-tester-a0000000d93fqAcct-Session-Time^ArCalling-Station-Idqa0:00:00:00:d9:3fqCalled-Station-Idq02:00:00:00:00:01iUser-NamesUserOF-a0000000d93foEvent-Timestamp}2020-03-23 16:06:44 -0400 EDTpAcct-Status-Type^AqFramed-IP-Addressk10.0.217.63hNAS-Port ^TnNAS-IP-Addressi127.0.0.1', '2147483641' ) ON DUPLICATE KEY UPDATE `value`=VALUES(`value`), `expires_at`=VALUES(`expires_at`);
```

We'll have to investigate how we can adjust this

julsemaan:
More details, I used redis instead of the DB to store the accounting cache and some other queries started popping up during the timeouts:

I picked some of the queries, there are much more than that and a lot of them are repetitive
```
# User@Host: pf[pf] @ localhost []
# Thread_id: 179612  Schema: pf  QC_hit: No
# Query_time: 1.956928  Lock_time: 0.000041  Rows_sent: 0  Rows_examined: 32
# Rows_affected: 0
SET timestamp=1584995108;
INSERT INTO bandwidth_accounting (tenant_id, mac, unique_session_id, time_bucket, in_bytes, out_bytes)
            SELECT 1 AS tenant_id, 'a0:00:00:00:b0:13' AS mac, 'd2134d71cb29a6b3940c88a7a6d70198' AS unique_session_id, '2020-03-23 16:25:00' AS time_bucket, in_bytes, out_bytes FROM (
                SELECT GREATEST(46074 - IFNULL(SUM(in_bytes), 0), 0) AS in_bytes, GREATEST(47074 - IFNULL(SUM(out_bytes), 0), 0) AS out_bytes FROM bandwidth_accounting WHERE tenant_id = 1 AND mac = 'a0:00:00:00:b0:13' AND unique_session_id = 'd2134d71cb29a6b3940c88a7a6d70198' AND time_bucket != '2020-03-23 16:25:00'
            ) AS y WHERE in_bytes > 0 || out_bytes > 0
        ON DUPLICATE KEY UPDATE in_bytes = VALUES(in_bytes), out_bytes = VALUES(out_bytes);

# User@Host: pf[pf] @ localhost []
# Thread_id: 179539  Schema: pf  QC_hit: No
# Query_time: 1.929218  Lock_time: 0.000027  Rows_sent: 1  Rows_examined: 4
# Rows_affected: 0
SET timestamp=1584995153;
SELECT nasname, secret, tenant_id, unique_session_attributes FROM radius_nas WHERE nasname = '02:00:00:00:00:01'
        UNION
          SELECT nasname, secret, tenant_id, unique_session_attributes FROM radius_nas WHERE nasname = '172.20.155.47'
        UNION
        (
            SELECT nasname, secret, tenant_id, unique_session_attributes from radius_nas
            WHERE INET_ATON('172.20.155.47') BETWEEN start_ip AND end_ip
            ORDER BY range_length LIMIT 1
        ) LIMIT 1;
# User@Host: pf[pf] @ localhost []
# Thread_id: 179669  Schema: pf  QC_hit: No
# Query_time: 1.726422  Lock_time: 0.000044  Rows_sent: 0  Rows_examined: 0
# Rows_affected: 0
SET timestamp=1584995153;
DELETE FROM `chi_cache` WHERE ( `expires_at` <= '2020-03-23 16:25:51' ) LIMIT 1000;
# User@Host: pf[pf] @ localhost []
# Thread_id: 179579  Schema: pf  QC_hit: No
# Query_time: 1.067585  Lock_time: 0.000091  Rows_sent: 1  Rows_examined: 0
# Rows_affected: 0
SET timestamp=1584995153;
select count(distinct n.mac) from node n join radacct r on r.callingstationid=n.mac where n.status='reg' and (n.unregdate >= NOW() or n.unregdate = '0000-00-00 00:00:00') and r.acctstoptime is null;

```

It just seems like the DB is temporarily loaded at some intervals and causes extreme slowdown for a few seconds. We'll have to investigate if improvements from @satkunas can help here

PR #5192, opened 2020-03-06
Converting switch modules to default switch template

lzammit:
# Description
Convert existing perl switch module to switch template configuration for more flexibility.

# Impacts
Switches

### DIFF
--- a/conf/template_switches.conf.defaults
+++ b/conf/template_switches.conf.defaults
@@ -110,20 +110,6 @@ EOT
 acceptRole=Filter-ID = $role
 disconnect=Calling-Station-Id = $mac
 
-[Mist]
-description=Mist AP Template chite
-acceptVlan= <<EOT
-Tunnel-Medium-Type  = 6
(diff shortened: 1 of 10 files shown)


conf/template_switches.conf.defaults
```diff
@@ -55,3 +55,41 @@ Tunnel-Medium-Type  = 6
...
+Acct-Session-Id = ${last_accounting.acctsessionid}
+User-Name = ${last_accounting.username}
+EOT
+description=Ruckus ZoneDirector
+voip=
+radiusDisconnect=disconnect
+reject=
+coa=
+
+[Ruckus::SmartZone]
```
julsemaan: You shouldn't use an ID we're already using in our Perl modules. Otherwise that will conflict with it

nqb:
@lzammit, is it normal that this PR doesn't contain .def files ?

lzammit:
Like normal I forgot them ? I will fix it ;-)

PR #5198, opened 2020-03-12
[pkg] don't start packetfence-httpd.admin at end of install in CI env

nqb:
# Description
Don't start automatically `packetfence-httpd.admin` service at end of package installation in a CI environment.

The idea is to start `packetfence-httpd.admin` later (during Ansible provisioning) with limited ressources to avoid consuming a lot of RAM when several virtual machines are running during integration tests.

# Impacts
RPM and DEB package installations.

# Delete branch after merge
YES

### DIFF
--- a/debian/packetfence.postinst
+++ b/debian/packetfence.postinst
@@ -153,8 +153,13 @@ case "$1" in
         set +e
         /usr/local/pf/bin/pfcmd configreload
         /usr/local/pf/bin/pfcmd fixpermissions
-        set -e
-        /bin/systemctl start packetfence-httpd.admin
+        if [ -n "$CI" ]; then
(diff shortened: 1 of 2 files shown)


debian/packetfence.postinst
```diff
@@ -153,8 +153,13 @@ case "$1" in
         set +e
         /usr/local/pf/bin/pfcmd configreload
         /usr/local/pf/bin/pfcmd fixpermissions
-        set -e
-        /bin/systemctl start packetfence-httpd.admin
+        if [ -n "$CI" ]; then
+            echo "CI environment, not starting PacketFence Administration GUI with default config to save ressources"
+        else
+            echo "Starting PacketFence Administration GUI..."
+            set -e
```
fdurand: Maybe we can remove that.

nqb:
Bump

PR #5201, opened 2020-03-12
Do not assume pid=email for sponsor authentication

### LINKED ISSUE #5200: User created by Sponsor respect user PID but devices associated to email
**Describe the bug**

Hello, I configured a sponsor auth source and a portal sponsor module. Inside the sponsor module, I selected to use "custom_field_1" as user PID. It did used the PID to create the user. However, it also created another user using the informed guest email and associated the device with that user (which has no other info but the email) instead of the other new user which uses "custom_field_1" as its PID.

I'm using packetfence-9.3.0-20200113144930.108928498.0007.el7.x86_64 in a ZEN installation. 

**To Reproduce**
1. configure a sponsor source
2. configure a sponsor portal module (Authentication::Sponsor) using "custom_field_1" as PID
3. fill the captive form and submit
4. even before the sponsor authorizes the acces, user tab already lists both a new user using  "custom_field_1" as PID (with all informed fields but no devices) and another new user with email as PID (with the device being registered but no informed fields).

**Expected behavior**
It should respect the PID selected by the admin.

luizluca:
Fixes #5200

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm
@@ -156,6 +156,7 @@ sub do_sponsor_registration {
     my $pid = $self->request_fields->{$self->pid_field};
     my $email = $self->request_fields->{email};
     $info{'pid'} = $pid;
+    $info{'email'} = $email;
     my ( $status, $status_msg ) = $source->authenticate($pid);
     unless ( $status ) {
         $self->app->flash->{error} = $status_msg;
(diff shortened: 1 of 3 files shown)


luizluca:
Forced-push a new version which also fixes the email to the sponsor adding a new PID field when it is different from email.

julsemaan:
Will test in devel and cherry-pick in maintenance once I confirm it fully works

julsemaan:
Cherry-picked up to maintenance/8.3

luizluca:
Shouldn't it be 9.3?

julsemaan:
up to maintenance/8.3 which includes all releases after that one

PR #5208, opened 2020-03-18
Do not ignore 'PID field' from Authentication::Password

luizluca:
Authentication::Password does askes for non-required PID field.
Even when provided, it was ignored by Authentication::Login.

If 'PID field' is missing, it will use username as usual.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>


luizluca:
Thanks!

PR #5211, opened 2020-03-19
[RFC] Add new POTD portal auth module (WIP)

### LINKED ISSUE #3762: POTD: Ability to specify the username to use
Right now for the POTD, you're forced to use the user potd and don't have any control over it.

This also means you can only have a single POTD source since many would use the same user and be useless

luizluca:
It's a copy of Authentication::Password but without a fixed
username (it uses the one provided by the user just like
Authentication::Login). The difference is that PID and username
are treated independently. Username is only considered for
authentication while PID is used to actually create the user
in DB.

It depends on #5208 
Does it fix #3762?

The web module configuration is not working. It is still not listing the new module. I can use it manually changing conf/portal_modules.conf. However, it gives me a blank form when I edit it.
I need some help here. I also still need to remove username string field.

Maybe it should not be Authentication::POTD but something like Authentication::AltUser. However, it does not make sense to have a POTD != username when the user has its own password. I'll wait for comments.

### DIFF
--- a/html/captive-portal/lib/captiveportal/DynamicRouting/Module/Authentication/POTD.pm
+++ b/html/captive-portal/lib/captiveportal/DynamicRouting/Module/Authentication/POTD.pm
@@ -1,0 +1,45 @@
+package captiveportal::DynamicRouting::Module::Authentication::POTD;
+use Moose;
+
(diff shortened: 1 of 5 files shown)


luizluca:
Maybe this new auth module could allow the admin to set "test all possible POTD" and try to authenticate using all configured POTD sources. In this case, it could hide the username field from guest form.

PR #5212, opened 2020-03-19
Admin GUI via haproxy

fdurand:
# Description
Use haproxy to server the admin gui

# Impacts
Admin GUI

# NEW Package(s) required
perl(Catalyst::Plugin::SmartURI)
perl(URI::SmartURI)

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Haproxy server the admin gui, httpd.admin disabled by default


fdurand:
Ready for review.

html/pfappserver/root/static.alt/src/views/Configuration/_components/CertificatesView.vue
```diff
@@ -104,6 +105,7 @@
               <p>
                 {{ $t('Some services must be restarted to load the new certificate.') }}
                 <span v-if="id === 'http'" v-html="$t('The {service} service needs to be restarted from the command-line.', { service: $strong('httpd.admin') })"></span>
```
julsemaan: This isn't true anymore I believe

html/pfappserver/root/static.alt/src/views/Configuration/_components/CertificatesView.vue
```diff
@@ -26,6 +26,7 @@
...
```
julsemaan: This isn't true anymore I believe

julsemaan:
Also one thing I don't see is adding the new admin services in the protected services in the Services section of the admin. And we could also remove the protection of httpd.admin

html/pfappserver/root/static.alt/src/views/Configuration/_components/CertificatesView.vue
fdurand: yes you are right

PR #5278, opened 2020-04-01
use shorter VM name for PacketFence ZEN

### LINKED ISSUE #5146: Add domain on PacketFence-Zen
Default PF zen hostname: PacketFence-ZEN

If we keep the %h for the domain configuration it goes over 14 chars.

julsemaan:
# Description
Make the PacketFence ZEN VM name shorter (now `PF-ZEN`)

# Impacts
new ZEN VMs

# Issue
fixes #5146 

# Delete branch after merge
YES

### DIFF
--- a/addons/ZEN-vagrant-builder/installer/install-pf.sh
+++ b/addons/ZEN-vagrant-builder/installer/install-pf.sh
@@ -19,9 +19,9 @@ yum install perl -y
 yum install --enablerepo=$PFREPO $PFPACKAGE -y
 
 # Setting the hostname
-hostname PacketFence-ZEN
-echo "PacketFence-ZEN" > /etc/hostname
-echo "NETWORKING=yes HOSTNAME=PacketFence-ZEN" > /etc/sysconfig/network
+hostname packetfence
+echo "packetfence" > /etc/hostname
+echo "NETWORKING=yes HOSTNAME=packetfence" > /etc/sysconfig/network
 
 # Setting up rc.local so it modifies /etc/issue to display instructions on setting up the ZEN
 cat /vagrant/installer/rc.local > /etc/rc.local


addons/ZEN-vagrant-builder/installer/install-pf.sh
```diff
@@ -19,9 +19,9 @@ yum install perl -y
 yum install --enablerepo=$PFREPO $PFPACKAGE -y
 
 # Setting the hostname
-hostname PacketFence-ZEN
-echo "PacketFence-ZEN" > /etc/hostname
-echo "NETWORKING=yes HOSTNAME=PacketFence-ZEN" > /etc/sysconfig/network
+hostname PF-ZEN
```
extrafu: Shouldn't we just do that and leave the rest as-is?

julsemaan: To make things cleaner and more consistent I changed everything but if you prefer the full name to be used elsewhere (like the VM ESX name), I can revert those

extrafu: That's my preference, but it doesn't mean we should do it!

I think "PF ZEN" isn't nice-looking for ESX name and others. For the hostname, heck I don't mind - we could even call it bob.

julsemaan: Thinking about it again, `packetfence` is 12 characters which meets the requirement for the fix.

So I'll just change the hostname to that and leave the rest as is

nqb:
Not related to this PR but you should use:
```
yum localinstall http://packetfence.org/downloads/PacketFence/RHEL7/packetfence-release-7.stable.noarch.rpm
```
in `addons/ZEN-vagrant-builder/installer/install-pf.sh`

julsemaan:
I adjusted the VM name to `packetfence` and left all the rest to PacketFence-ZEN to have it cleaner in the ESX and download links.

I didn't do the release RPM install change, its unrelated to this PR

PR #5336, opened 2020-04-08
allow to use existing ip4log entries in web-auth

julsemaan:
# Description
Allows to disable the mechanism that creates an ip4log entry from a web-auth request and instead uses the data that is already in the ip4log. This way ip4log data from DHCP and accounting can have precedence over the data set by the web-auth request parsing

# Impacts
Web-auth

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Added parameter to prevent external portal requests from updating the ip4log

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1270,6 +1270,13 @@ description=<<EOT
 Use the information included in the accounting to update the iplog
 EOT
 
+[advanced.update_iplog_with_external_portal_requests]
+type=toggle
+options=enabled|disabled
+description=<<EOT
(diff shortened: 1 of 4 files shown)


julsemaan:
I've made this a 10.0 since its a fix but we can postpone it to a future release.

AFAIK this is fairly corner-case and only affects Cisco+Meraki web-auth where the web-auth clients are NATed to the PF server

julsemaan:
:tada:

PR #5348, opened 2020-04-10
Admin - Multi-Tenancy

satkunas:
# Description
Add multi-tenancy to web admin.

# Impacts
`aaa-api`: adds optional `X-PacketFence-Tenant-Id` header to API requests.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Add multi-tenancy to web admin

# TODO
 - [ ] Convert API GET /api/v1/token_info -> `tenant_id: int` to `tenant: { domain_name, id, name, portal_domain_name }`


satkunas:
+TODO

We will need to modify `aaa-api` in order to return the entire tenant object in the token_info request. This will allow mortal users to see the "name" of their current tenant when they don't have the neccessary admin roles (`SYSTEM_READ`) to list all the available tenants and are unable to do a name lookup on their tenant_id.

`token_info` now:
```json
{
    "item": {
        "admin_actions": [...],
        "admin_roles": [...],
        "tenant_id": 1,
        "username": "mortaluser",
        "expires_at": "2020-04-10T11:26:58.176130538-04:00"
    }
}
```


`token_info` after:
```json
{
    "item": {
        "admin_actions": [...],
        "admin_roles": [...],
        "tenant_id": {
            "domain_name": "...",
            "id": "1",
            "name": "customtenant",
            "portal_domain_name": "..."
        },
        "username": "mortaluser",
        "expires_at": "2020-04-10T11:26:58.176130538-04:00"
    }
}
```

julsemaan:
Things to handle:
- [x] Tenant ID 1 doesn't have access to the configuration. When there is a single tenant, it should have access to the configuration (or should tenant ID 1 also always have access to the config)
- [x] Tenants will have to be able to list some configuration items as the admin leverages them (like security events config)
- [x] Need to be able to set the `TenantId` attribute of a switch to a valid tenant ID (backend already supports it, just need the attribute in the frontend and perhaps in the OPTIONS meta)
- [x] When creating a switch, the roles tab is empty although the API seems to have returned the list of roles:
image
- [x] The allowed_tenant_id is redundant. We should rely on the tenant_id attribute of the password entry in order to set the tenant associated to the token
>  (moved to new list) - [ ] The tenant ID action shouldn't be allowed when creating a local user. The user will be created with the tenant that is currently scoped

julsemaan:
I added some comments on this. Some adjustments in the backend seem to be necessary for the local accounts + some things that weren't thought of in the backend when it was created before we started working on the new admin

nqb:
An important thing for me:

Currently, frontend doesn't display admin roles assigned by administration rules when you log in on web admin. It's a problem because it could be hard to know which admin roles a user had.

For example, if you use a LDAP source to dynamically assign admin roles, the only way to know admin roles of a user is to check logs because LDAP users are not created in `person` table.

I have seen that admin roles are returned by API in `token_info` response. It would be nice to display them somewhere in web admin next to name of logged user and tenants.

julsemaan:
> An important thing for me:
> 
> Currently, frontend doesn't display admin roles assigned by administration rules when you log in on web admin. It's a problem because it could be hard to know which admin roles a user had.
> 
> For example, if you use a LDAP source to dynamically assign admin roles, the only way to know admin roles of a user is to check logs because LDAP users are not created in `person` table.
> 
> I have seen that admin roles are returned by API in `token_info` response. It would be nice to display them somewhere in web admin next to name of logged user and tenants.

This is more a general issue/feature with the admin, not directly related to multi-tenancy.
I suggest you open an issue for this so that its resolved outside of this PR

satkunas:
> * [x]  When creating a switch, the roles tab is empty although the API seems to have returned the list of roles:

Fixed with https://github.com/inverse-inc/packetfence/commit/198865107b801dab9647b5c56a5a8e3705c8ad9a

julsemaan:
More items:
- [x] Admin needs to ignore errors (or simply not attempt) to get the switch groups in the node page for the quick search. We can't expose this API endpoint to all the tenants as it would leak info from other tenants
- [x] Tenants shouldn't have access to dashboard, it won't contain data that is scoped per tenant
- [x] The configuration section should be hidden when the tenant doesn't have access to it
- [x] The services section (Status tab) should be hidden when the tenant doesn't have access to it
- [x] The queue sections of the status tab should be hidden when the tenant doesn't have access to it
- [x] The tenant ID action shouldn't be allowed when creating a local user. The user will be created with the tenant that is currently scoped

julsemaan:
@satkunas, I went through the items for the backend and posted a new list (including the items of the first list that haven't been done). Can you take a look at this list.

satkunas:
@julsemaan I've pushed some changes for all your comments above.

julsemaan:
>  The configuration section should be hidden when the tenant doesn't have access to it

This isn't working on 7b17b1c. Can you take a look @satkunas 

Note: it works when logged in as tenant master and switching to a tenant. But doesn't work when logged in as a non-master tenant

julsemaan:
I added a unique key on the PID as all local usernames will have to be unique.

(NOTE: the test below needs to be made with the latest schema)

In one tenant I created a username called `mrbobby` and it worked fine

Then in another tenant, I created another username called `mrbobby`. 

The person entry was created since multiple tenants can have multiple person entries called the same. But the password entry failed to be created with a 409. The admin reported it succeeded although the password entry failed. The frontend needs to handle the failure of the creation of a password entry even though the person entry succeeded.

julsemaan:
A feature we need to implement:
Since all password entries are unique, the frontend should suffix the domain name (domain_name property of the tenant) to all users created via the admin. This way it will help scope the local users of each tenant. That could be a setting that is enabled/disabled in pf.conf

julsemaan:
One thing that is missing in the GUI, we need the realm page to be tenant aware. Since the tenant ID is not a property of the sections but part of the section name, it seems it will require some adjustments on the API so that it can be exposed on the frontend. 

Not exactly sure how to do this, @jrouzierinverse can you provide help for this specific issue?

satkunas:
@jrouzierinverse is it possible to override the error message in the case that a user already has a password entry? The following error isn't descriptive enough and instead of handling it on the front-end it would be easier to simply overload the message and display it to the user:

""An attempt to add a duplicate entry was stopped. Entry already exists and should be modified instead of created.""

extrafu:
@jrouzierinverse Please answer @satkunas and @julsemaan 's questions. We need to wrap that up.

html/pfappserver/root/static.alt/src/views/Users/_router/index.js
```diff
@@ -14,10 +15,17 @@ const route = {
...
+      return acl.$can('read', 'users') || acl.$can('create', 'users') // has ACL for 1+ children
+    },
+    fail: { path: '/configuration', replace: true }, // no ACL in this view, redirect to next sibling
+    transitionDelay: 300 * 2 // See _transitions.scss => $slide-bottom-duration
+  },
   props: { storeName: '$_users' },
-  meta: { transitionDelay: 300 * 2 }, // See _transitions.scss => $slide-bottom-duration
   beforeEnter: (to, from, next) => {
     if (!store.state.$_users) {
+console.log('register module', '$_users')
```
cgx: Remove debugging output.

html/pfappserver/root/static.alt/src/App.vue
```diff
@@ -39,6 +39,19 @@
             <b-dropdown-divider></b-dropdown-divider>
             <b-dropdown-item to="/logout">{{ $t('Log out') }}</b-dropdown-item>
           </b-nav-item-dropdown>
+          <b-nav-item-dropdown right v-if="tenant && tenant.id === 0">
+            <template v-slot:button-content>
+              <icon name="layer-group"></icon> {{ tenant_mask_name }}
+            </template>
+            <b-dropdown-item-button v-for="tenant in tenants" :key="tenant.id"
```
cgx, line 47: I propose to add a `<b-dropdown-item-header>` element at the top of the list that would says *Tenants*.

html/pfappserver/root/static.alt/src/views/Users/_router/index.js
```diff
@@ -58,7 +70,8 @@ const route = {
         next()
       },
       meta: {
-        can: 'create users'
+        can: 'create users',
+        fail: { name: 'usersPreview', replace: true } // redirect to next sibling
```
cgx: We should not fallback to `usersPreview` but simply to `users`.

satkunas:
@cgx changes pushed

cgx (approved):
Frontend looks good!

julsemaan:
@satkunas, did you take care of this?

Since all password entries are unique, the frontend should suffix the domain name (domain_name property of the tenant) to all users created via the admin. This way it will help scope the local users of each tenant. That could be a setting that is enabled/disabled in pf.conf

satkunas:
> @satkunas, did you take care of this?
> 
> Since all password entries are unique, the frontend should suffix the domain name (domain_name property of the tenant) to all users created via the admin. This way it will help scope the local users of each tenant. That could be a setting that is enabled/disabled in pf.conf

@julsemaan  I will come back this afternoon.

satkunas:
@julsemaan I implemented some of the basic code needed to append the domain_name when creating a user. However, it still needs to be implemented for the CSV import. And the validations need to conditionally take this in to account.  The search may also need to be modified.

Is it not possible to implement this in the backend so it's transparent to the web admin? I'm seeing plenty of workarounds that may be needed to accommodate this w/ the front-end.

julsemaan:
> @julsemaan I implemented some of the basic code needed to append the domain_name when creating a user. However, it still needs to be implemented for the CSV import. And the validations need to conditionally take this in to account. The search may also need to be modified.
> 
> Is it not possible to implement this in the backend so it's transparent to the web admin? I'm seeing plenty of workarounds that may be needed to accommodate this w/ the front-end.

We could do it in the backend too, do you want to leave the visual indicator in the admin that it will be suffix but strip it off the request? Then I'll check for doing this in the backend

jrouzierinverse (changes requested):
The database upgrade is failing
```t/serialized_unittests/db/apply_upgrade.t```

**Comment by julsemaan** [comment]
> The database upgrade is failing
> `t/serialized_unittests/db/apply_upgrade.t`

Should be all good now

**Comment by extrafu** [comment]
@jrouzierinverse Review the last change and merge if all is fine.

**Comment by jrouzierinverse** [comment]
Additional questions
If the PID is being suffixed with the domain name.
* How do we handle changing the domain name?
* How do we handle the default/admin user for each tenant?
* Why just the PID in password is unique why not also the PID in person also?

**Comment by julsemaan** [comment]
> Additional questions
> If the PID is being suffixed with the domain name.
> 
> * How do we handle changing the domain name?
  
We don't allow this to change right now, so we should handle that case if we decide this can happen

> * How do we handle the default/admin user for each tenant?
  
default doesn't have a password entry, admin doesn't either I believe, they're just there because the code expects admin and default to exist in the person table

> * Why just the PID in password is unique why not also the PID in person also?
  
The PID in person doesn't make sense, otherwise guests registering using gmail would be bob@gmail.com@acme.com. The uniqueness only needs to exist in the password table for authentication purposes but the person table doesn't need that same restriction since it isn't used for authentication purposes, just for reporting and user tracking purposes and is not used in a non-tenant scope. The password table is used for non-tenant scope when a user is logging into the admin or API

**Comment by jrouzierinverse** [comment]
We already can scope the users of a tenant with the tenant_id on the password table.
Why are we adding an additional complexity of having to append the domain of the tenant?

**Comment by julsemaan** [comment]
> We already can scope the users of a tenant with the tenant_id on the password table.
> Why are we adding an additional complexity of having to append the domain of the tenant?

When we logging into the admin, we need a way to uniquely find the user that is logging in.

For example, if I have 2 `bob` users for 2 different tenants, we want the distinction to be clear between them when performing a login in the admin where we would have bob@example.com and bob@acme.com

**Comment by jrouzierinverse** [comment]
What about the other places that create a user should they also suffix the person's PID?

**Comment by julsemaan** [comment]
> What about the other places that create a user should they also suffix the person's PID?

The only other place I can think of is the 'Create local account' and I think it will be up to the admin to manage this so that accounts don't collide but we can't force the suffix since otherwise you end up with the same double domain problem for email/oauth/etc registration

If you have another place in mind, let me know

File: html/pfappserver/root/static.alt/src/views/Users/_components/UsersCreate.vue [hunk 4]
```diff
@@ -361,7 +381,10 @@ export default {
             ...{ quiet: true }
           }
           for (let i = 0; i < this.form.multiple.quantity; i++) {
-            const pid = this.form.multiple.prefix + (i + 1)
+            let pid = this.form.multiple.prefix + (i + 1)
+            if (this.domainName) { // append domainName to pid when available (tenant)
+              pid = `${pid}${this.domainName}`
```
jrouzierinverse, line 386: Shouldn't this be pid@domain instead of piddomain

jrouzierinverse (changes requested):

html/pfappserver/root/static.alt/src/views/Users/_components/UsersCreate.vue
extrafu: @satkunas Needs to make the change?

julsemaan: Yep

satkunas: Suffix delim changed in https://github.com/inverse-inc/packetfence/pull/5348/commits/664d2a88299aacd113353c9002a22dfe38a1b639

PR #5369, opened 2020-04-15
CI: deploy RPM an Deb packages to dedicated repository (branches)

nqb:
# Description
- Deploy RPM built on other branches than devel to packetfence-branches repository
- Deploy Deb built on other branches than devel to debian-branches repository
- Add packetfence-branches and packetfence-gitlab repository to packetfence-release RPM package

This change is necessary to avoid installing packages built on other branches than devel when we only configured -devel repository on machines.

# Impacts
- Packages upload
- Test stage of pipeline

# Code / PR Dependencies
https://github.com/inverse-inc/ansible-packetfence/pull/28 (already merged)

# Delete branch after merge
YES

### DIFF
--- a/addons/vagrant/extra_vars_branches.yml
+++ b/addons/vagrant/extra_vars_branches.yml
@@ -1,0 +1,24 @@
+---
+# Add branches repository
+packetfence_install__centos_release_rpm: 'http://packetfence.org/downloads/PacketFence/RHEL7/packetfence-release-7.branches.noarch.rpm'
+packetfence_install__centos:
+  repos:
+    - packetfence-devel
+    - packetfence-branches
+
+packetfence_install__deb:
+  repos:
+    - debian
+    - debian-devel
+    # - debian-branches
+
+# Replace by *exact* name of your package
+packetfence_install__centos_packages:
+   - packetfence-9.3.9-20200124160156.113848061.0007.fix~5075.el7
+
+packetfence_install__deb_packages:
+  - packetfence=9.3.9+20200415061223+136088176+0009+fix~pipeline~issues+stretch1
+
+# workaround to install a package with version number using "="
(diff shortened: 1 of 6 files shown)


nqb:
BUMP

addons/vagrant/extra_vars_branches.yml
```diff
@@ -0,0 +1,24 @@
...
+
+packetfence_install__deb:
+  repos:
+    - debian
+    - debian-devel
+    - debian-branches
+
+# Replace by *exact* name of your package
+packetfence_install__centos_packages:
+   - packetfence-9.3.9-20200124160156.113848061.0007.fix~5075.el7
```
julsemaan: Should this be hardcoded in the config ?

addons/vagrant/extra_vars_branches.yml
```diff
@@ -0,0 +1,24 @@
...
+
+packetfence_install__deb_packages:
+  - packetfence=9.3.9+20200415061223+136088176+0009+fix~pipeline~issues+stretch1
```
julsemaan: Same as above

addons/vagrant/extra_vars_branches.yml
nqb: `extra_vars_branches.yml` is like an example file that allow you to override, on CLI, some inventory variables.

I created this file in order to edit it by hand when I want to install branch specific package.

Let's me know if that makes sense for you

nqb: In fact, it makes more sense to replace these values by a lookup on environment variables.
```suggestion
packetfence_install__centos_packages: {{ lookup("env", "CENTOS_PKG_NAME") }}
```
I just need to convert this to a list

nqb: Done

nqb:
@julsemaan, ready to merge ?

PR #5392, opened 2020-04-27
Feature/live log viewer

julsemaan:
# Description
Adds a live log viewer to the admin

# Impacts
api-frontend

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries
## New Features
* Added live log viewer in the admin GUI (located in the Auditing tab)


julsemaan:
Basic example: https://support.inverse.ca/~jsemaan/example-log-tailer.html

julsemaan:
@satkunas, its all yours now

go/caddy/log-tailer/handlers.go
```diff
@@ -0,0 +1,227 @@
...
+							"value": pflog("httpd.aaa.access"),
+						},
+						gin.H{
+							"text":  "httpd.portal Apache error log",
+							"value": pflog("httpd.portal.error"),
+						},
+						gin.H{
+							"text":  "httpd.portal Apache error log",
+							"value": pflog("httpd.portal.error"),
+						},
```
satkunas: Duplicate

satkunas:
Session creation and deletion should be logged in the admin audit log. However, I don't think we need to log every request during the session.

satkunas:
Invalid meta parsing seen with:

```json
{
    "timestamp": 1588252692881,
    "category": "039e1328-0241-468a-a2c9-428baa7ebd39",
    "data": {
        "meta": {
            "timestamp": "2020-04-30T09:18:12-04:00",
            "hostname": "packetfence",
            "log_level": "",
            "process": "WARN -e",
            "syslog_name": "packetfence"
        },
        "raw": "Apr 30 09:18:12 packetfence packetfence: WARN -e(24878): Use of uninitialized value in concatenation (.) or string at /usr/local/pf/lib/pf/services/manager/radiusd_child.pm line 622."
    }
}
```

julsemaan:
> Invalid meta parsing seen with:
> 
> ```json
> {
>     "timestamp": 1588252692881,
>     "category": "039e1328-0241-468a-a2c9-428baa7ebd39",
>     "data": {
>         "meta": {
>             "timestamp": "2020-04-30T09:18:12-04:00",
>             "hostname": "packetfence",
>             "log_level": "",
>             "process": "WARN -e",
>             "syslog_name": "packetfence"
>         },
>         "raw": "Apr 30 09:18:12 packetfence packetfence: WARN -e(24878): Use of uninitialized value in concatenation (.) or string at /usr/local/pf/lib/pf/services/manager/radiusd_child.pm line 622."
>     }
> }
> ```

Its fixed, it was the logging configuration that wasn't standard in conf/log.conf

You'll need to copy over conf/log.conf.example to conf/log.conf

satkunas:
@julsemaan please add `log_file` to the meta. Include the full path so it is the same values seen in OPTIONS. This will allow us to post-filter the output on a specified source(s).

satkunas:
Invalid parsing seen:

```json
{
    "events": [{
        "timestamp": 1588365487450,
        "category": "4b10d601-a4c2-43f2-9af1-8b87697032e4",
        "data": {
            "meta": {
                "timestamp": "0001-01-01T00:00:00Z",
                "hostname": "",
                "log_level": "",
                "process": "",
                "syslog_name": "",
                "log_without_prefix": "",
                "filename": "/usr/local/pf/logs/packetfence.log"
            },
            "raw": " (pf::condition::fingerbank::device_is_a::match)"
        }
    }]
}
```

from:
```log
May  1 16:37:08 packetfence packetfence_httpd.webservices: httpd.webservices(22776) WARN: [mac:00:0a:f5:5f:d0:10] Use of uninitialized value $device in concatenation (.) or string at /usr/local/pf/lib/pf/condition/fingerbank/device_is_a.pm line 50.
 (pf::condition::fingerbank::device_is_a::match)
```

satkunas:
A `GET` request blocks `DELETE`. It succeeds, but waits until the first request is resolved. Can we interrupt `GET` with a syncronous `DELETE`?

satkunas:
@julsemaan the front-end is now fully functional.

I still need to develop and test nagivation guards in an attempt to destroy the sessions when the SPA is refreshed (beforeunload).

I'll also work on adding some useful additional options next, some possible ideas include...
* add to clipboard
* save as
* dark/light color schemes (white-text on black-background)
* font-size adjuster
* pretty print (markup styling of meta)
* unread counter on tabs
* client-side text/regexp filter
* alternate client-side buffer that stores client filtering indefinitely (to avoid log line buffer overrun)

julsemaan:
> A `GET` request blocks `DELETE`. It succeeds, but waits until the first request is resolved. Can we interrupt `GET` with a syncronous `DELETE`?

The solution would be bigger than the problem. I suggest interrupting your GET when you are deleting and that should do it since it should cancel out the request

julsemaan:
> Invalid parsing seen:
> 
> ```json
> {
>     "events": [{
>         "timestamp": 1588365487450,
>         "category": "4b10d601-a4c2-43f2-9af1-8b87697032e4",
>         "data": {
>             "meta": {
>                 "timestamp": "0001-01-01T00:00:00Z",
>                 "hostname": "",
>                 "log_level": "",
>                 "process": "",
>                 "syslog_name": "",
>                 "log_without_prefix": "",
>                 "filename": "/usr/local/pf/logs/packetfence.log"
>             },
>             "raw": " (pf::condition::fingerbank::device_is_a::match)"
>         }
>     }]
> }
> ```
> 
> from:
> 
> ```
> May  1 16:37:08 packetfence packetfence_httpd.webservices: httpd.webservices(22776) WARN: [mac:00:0a:f5:5f:d0:10] Use of uninitialized value $device in concatenation (.) or string at /usr/local/pf/lib/pf/condition/fingerbank/device_is_a.pm line 50.
>  (pf::condition::fingerbank::device_is_a::match)
> ```

It seems this is caused when its on 2 different lines.

I'd have to see if log4perl allows to ensure all is on the same line instead of being on 2 lines

satkunas:
Payload used to init a session regarding the issue where DELETEing a 2nd session also kills the first.

POST /api/v1/logs/tail 

```json
{"name":"packetfence.log","files":["/usr/local/pf/logs/packetfence.log"],"filter":null,"filter_is_regexp":false}
```
This only occurs if the 2nd session is the same file as the 1st. If the 2nd file differs the 1st is not affected after DELETE.

jrouzierinverse:
log4perl space indents new lines.
You can also just skip lines that begin with spaces or append any line that begins with a space to the previous line.

julsemaan:
> log4perl space indents new lines.
> You can also just skip lines that begin with spaces or append any line that begins with a space to the previous line.

The problem with this is that I'm fed one line at a time so I'd need to buffer them to know if we're dealing with multi-line. If we can't get it on one line, then lets just have it where this line simply doesn't have meta

julsemaan:
> Payload used to init a session regarding the issue where DELETEing a 2nd session also kills the first.
> 
> POST /api/v1/logs/tail
> 
> ```json
> {"name":"packetfence.log","files":["/usr/local/pf/logs/packetfence.log"],"filter":null,"filter_is_regexp":false}
> ```
> 
> This only occurs if the 2nd session is the same file as the 1st. If the 2nd file differs the 1st is not affected after DELETE.

I pushed a fix, seems the issue is in hpcloud/tail directly. Could you try again and let me know how it looks now

satkunas:
> I pushed a fix, seems the issue is in hpcloud/tail directly. Could you try again and let me know how it looks now

~~Thanks. It looks good now.~~

Run 2 sessions at the same time, creation order doesn't matter. One with a busy log (A: packetfence.log), and one with a quiet log (B: mariadb). Although the timeouts are not equal session B appears to block the response from session A by increasing session A's timeout to that of session B. When session B is DELETED, session A returns to its normal frequency of timeouts.

julsemaan:
> > I pushed a fix, seems the issue is in hpcloud/tail directly. Could you try again and let me know how it looks now
> 
> ~Thanks. It looks good now.~
> 
> Run 2 sessions at the same time, creation order doesn't matter. One with a busy log (A: packetfence.log), and one with a quiet log (B: mariadb). Although the timeouts are not equal session B appears to block the response from session A by increasing session A's timeout to that of session B. When session B is DELETED, session A returns to its normal frequency of timeouts.

Should be good now with the latest commit

satkunas:
> Should be good now with the latest commit

Fixed. Thanks!

satkunas:
Invalid parsing seen with:
```json
{
    "events": [{
        "timestamp": 1588706240141,
        "category": "2ac12be2-6fbc-4371-ae84-b6772e5c5bd4",
        "data": {
            "meta": {
                "timestamp": "2020-05-05T15:17:20-04:00",
                "hostname": "packetfence",
                "log_level": "info",
                "process": "pfipset",
                "syslog_name": "pfipset",
                "log_without_prefix": "5249]: t=2020-05-05T15:17:20-0400 lvl=info msg=\"Reloading ipsets\" pid=5249",
                "filename": "/usr/local/pf/logs/packetfence.log"
            },
            "raw": "May  5 15:17:20 packetfence pfipset[5249]: t=2020-05-05T15:17:20-0400 lvl=info msg=\"Reloading ipsets\" pid=5249"
        }
    }]
}
```

julsemaan:
TODO for me: expose API endpoint to "touch" the session in case its paused

julsemaan:
One thing that looks weird to me is that the top loading bar that is within the page constantly goes back and forth when loading the logs:

image

julsemaan:
List of issues I see:
- [ ] I'm unable to scroll up the logs (I can scroll down when in top to bottom mode)
- [ ] Issue with the top loading bar constantly moving (see comment above)
- [ ] Not an issue, just want a confirmation that the cluster use-case was handled since I'm not testing in cluster
- [ ] The log section seems to be just a little too big to fit in the page which creates a scrollbar for the whole page (Firefox on Ubuntu):
image

satkunas:
> * I'm unable to scroll up the logs (I can scroll down when in top to bottom mode)

There's 2 view modes: `bottom-to-top` and `top-to-bottom`. The former uses a regular element where the scrollbar "sticks" to the topmost position, preventing it from auto-scrolling downward when content is added. This doesn't work well when content is added to the bottom, so the latter uses a CSS trick that inverts (reverses) the behavior of the element so the scrollbar "sticks" to the bottom preventing it from auto-scrolling upward when content is added. The behavior you're experiencing is intentional to force the scroll-bar to naturally "stick" in both directions. The alternative is not feasible as the scrollbar would need to be updated constantly by the user (w/ top-to-bottom) with each mutation of the content.

Since it's a CSS scroll-bar hack the overhead is minimal. Tweaking the scrollbar further will require some Javascript events which may cause other usability issues.

> * [ ]  Issue with the top loading bar constantly moving (see comment above)

I originally added some workarounds so the log viewer wasn't triggering the loading bar when polling the API. However, I found it useful to see when navigating elsewhere on the site as an indicator that the log viewer was still running in the background, and opted to remove this workaround and to display it as a reminder. Let me know if you don't agree and I will re-hide it.

> * [ ]  Not an issue, just want a confirmation that the cluster use-case was handled since I'm not testing in cluster

I haven't tested this either. We'll need to discuss this together.


> * [ ]  The log section seems to be just a little too big to fit in the page which creates a scrollbar for the whole page (Firefox on Ubuntu):

It's hard to find a magic height that works universally for various screen resolutions and aspect ratio's. I'll tweak it a little bit as this is exactly how it currently renders on my setup (Chrome on Fedora).

julsemaan:
>  Issue with the top loading bar constantly moving (see comment above)

I think we should put the workaround to hide this when the logs are updating

julsemaan:
>  The log section seems to be just a little too big to fit in the page which creates a scrollbar for the whole page (Firefox on Ubuntu):

We could make it just a bit smaller than it is and I think we'll hit the sweetspot for most browsers

satkunas:
> > The log section seems to be just a little too big to fit in the page which creates a scrollbar for the whole page (Firefox on Ubuntu):
> 
> We could make it just a bit smaller than it is and I think we'll hit the sweetspot for most browsers

The height was reduced in https://github.com/inverse-inc/packetfence/pull/5392/commits/7ceaefc93c6106286b9dc8931fca627aeef20298

satkunas:
> > Issue with the top loading bar constantly moving (see comment above)
> 
> I think we should put the workaround to hide this when the logs are updating

done

julsemaan:
@satkunas 

I added the API endpoint to touch a session. Have fun!

satkunas:
@julsemaan touch is implemented.I think we're ready to review.

go/caddy/log-tailer/handlers.go
```diff
@@ -0,0 +1,245 @@
...
+				"type":        "array",
+				"required":    true,
+				"placeholder": nil,
+				"default":     nil,
+				"item": gin.H{
+					"type":        "string",
+					"required":    true,
+					"placeholder": nil,
+					"default":     nil,
+					"allowed": []gin.H{
```
fdurand: Most of the files are defined here https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/constants/syslog.pm maybe we can create a pfconfig that expose them and use it.

go/caddy/log-tailer/handlers.go
```diff
@@ -0,0 +1,245 @@
...
+						gin.H{
+							"text":  "Global syslog (/var/log/messages)",
+							"value": "/var/log/messages",
```
fdurand: /var/log/syslog too (debian)

go/caddy/log-tailer/logmeta.go
```diff
@@ -0,0 +1,181 @@
...
+func (lme *LogMetaEngine) ExtractMeta(log string) (lm LogMeta) {
+	if m := lme.GlobalExtractionRe.FindAllStringSubmatch(log, -1); m != nil {
+		lm.Timestamp, _ = time.ParseInLocation("Jan _2 15:04:05 2006", fmt.Sprintf("%s %d", m[0][lme.TimestampPos], time.Now().Year()), time.Local)
+		lm.Hostname = m[0][lme.HostnamePos]
+		lm.SyslogName = m[0][lme.SyslogNamePos]
+		lm.LogWithoutPrefix = strings.Trim(m[0][lme.LogWithoutPrefixPos], " ")
+		if extractor := lme.SyslogMap[lm.SyslogName]; extractor != nil {
+			extractor.ExtractMeta(lm.SyslogName, log, &lm)
+		}
+	} else {
```
fdurand: else {} is useless

go/caddy/log-tailer/logmeta_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"testing"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
+)
+
+func TestRsyslogMetaEngine(t *testing.T) {
+	line := `Apr 07 17:12:00 pf pfipset[2126]: lvl=info msg="Request isn't authorized, performing login against the Unified API" pid=2126 request-uuid=90505d47-8880-11ea-95cd-0050569c7a2d`
+	res := NewRsyslogMetaEngine().ExtractMeta(line)
+	spew.Dump(res)
```
fdurand: Remove

go/caddy/log-tailer/logmeta.go
```diff
@@ -0,0 +1,181 @@
...
+	LogLevelExtractor: logLevelImplicitError,
+	ProcessNameExtractor: func(syslogName string, log string) string {
+		syslogName = strings.Replace(syslogName, "_", ".", -1)
+		return strings.Replace(syslogName, ".err", "", -1)
+	},
+}
+
+func NewRsyslogMetaEngine() *LogMetaEngine {
+	return &LogMetaEngine{
+		SyslogMap: map[string]*LogMetaExtractor{
```
fdurand, line 95: Should be coll to fetch that from pfconfig

fdurand (changes requested):
It miss the systemd stuff and the service manager and the packaging

julsemaan:
> It miss the systemd stuff and the service manager and the packaging

I don't see what we would need for systemd (its part of the existing api-frontend service)

Same for packaging

go/caddy/log-tailer/logmeta.go
julsemaan: This one doesn't list files, it binds the different syslog names to their extractor so I doubt using pfconfig will be useful like for the files as this doesn't exist in our Perl codebase AFAIK and won't have any usage outside of this module

julsemaan:
Went through @fdurand's comments

go/caddy/log-tailer/logmeta_test.go
```diff
@@ -3,14 +3,11 @@ package logtailer
...
 	"testing"
 	"time"
-
-	"github.com/davecgh/go-spew/spew"
 )
 
 func TestRsyslogMetaEngine(t *testing.T) {
 	line := `Apr 07 17:12:00 pf pfipset[2126]: lvl=info msg="Request isn't authorized, performing login against the Unified API" pid=2126 request-uuid=90505d47-8880-11ea-95cd-0050569c7a2d`
 	res := NewRsyslogMetaEngine().ExtractMeta(line)
-	spew.Dump(res)
```
extrafu: Excessive dump: https://www.youtube.com/watch?v=nnun8y7r8_U

go/caddy/log-tailer/logmeta.go
fdurand: The fact is we will need to maintain this list if we add new log files.
Why not adding a type and doing something like:

```
 logs.Element[i].Name: &LogMetaExtractor[ logs.Element[i].Type]
```

julsemaan: That's because the name here is not equal to the name in pf::constants::syslog

The name in pf::constants::syslog is the name of the file. Here is the name of the syslog name which isn't part of the constants.

So I'd need to add something in the Perl constants that has no usage except for our Golang codebase here which doesn't make much sense. Plus one log file in the logs constant can have multiple syslog names (like packetfence.log) so there would be added logic to handle all of this when it has no use in Perl

fdurand (approved):

julsemaan:
@cgx, can you perform a review of the frontend part of this PR?

html/pfappserver/root/static.alt/src/main.js
```diff
@@ -94,13 +96,17 @@ import 'vue-awesome/icons/sign-in-alt'
...
+import 'vue-awesome/icons/sort-numeric-down'
 import 'vue-awesome/icons/spinner'
 import 'vue-awesome/icons/step-backward'
 import 'vue-awesome/icons/regular/square'
 import 'vue-awesome/icons/square'
 import 'vue-awesome/icons/stop'
 import 'vue-awesome/icons/stopwatch'
 import 'vue-awesome/icons/sync'
+import 'vue-awesome/icons/tag'
+import 'vue-awesome/icons/tags'
```
cgx: `sort-numeric-up-alt`, `sort-numeric-down`, `tag` and `tags` don't seem to be used.

html/pfappserver/root/static.alt/src/styles/_utilities.scss
```diff
@@ -5,3 +5,8 @@
     font-size: $code-font-size;
     white-space: pre; /* allow \n to break */
   }
+
+.text-raw {
```
cgx, line 9: Instead of creating a new class name selector, you should simply replace your `<div>` element by a `<pre>`.

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,517 @@
...
+        // noop
+      })
+    },
+    unpauseSession () {
+      this.$store.dispatch(`$_live_logs/${this.id}/unpauseSession`).then(() => {
+        // noop
+      })
+    },
+    clearEvents () {
+      this.$store.dispatch(`$_live_logs/${this.id}/clearEvents`).then(response => {
```
cgx: `response` is not used.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
```diff
@@ -0,0 +1,328 @@
...
+* "$_live_log" store module
+*/
+import Vue from 'vue'
+import api from '../_api'
+import { createDebouncer } from 'promised-debounce'
+import i18n from '@/utils/locale'
+
+const scopes = {
+
+}
```
cgx: `scopes` is empty.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
```diff
@@ -0,0 +1,328 @@
...
+  isPaused: state => state.paused,
+  session: state => state.session,
+  events: state => state.events,
+  scopes: state => state.scopes,
+  filters: state => state.filters,
+  isFiltered: state => (scope, key) => {
+    const { scopes: { [scope]: { values: { [key]: { filter = false } = {} } = {} } = {} } = {} } = state
+    return filter
+  },
+  eventsFiltered: (state, getters) => {
```
cgx: `getters` are not used.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
```diff
@@ -0,0 +1,328 @@
...
+      return api.touchLogTailSession(state.session.session_id).then(response => {
+        commit('LOG_SESSION_SUCCESS')
+        return response
+      }).catch(err => {
+        commit('LOG_SESSION_ERROR', err.response)
+        return err
+      })
+    }
+  },
+  toggleFilter: ({ state, getters, commit }, { scope, key }) => {
```
cgx: `state` is not used.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
```diff
@@ -0,0 +1,328 @@
...
+    if (getters.isFiltered(scope, key)) { // disable
+      commit('LOG_FILTER_DISABLE', { scope, key })
+      commit('UPDATE_FILTERS')
+    }
+    else { //enable
+      commit('LOG_FILTER_ENABLE', { scope, key })
+      commit('UPDATE_FILTERS')
+    }
+  },
+  setSize: ({ state, commit }, size) => {
```
cgx: `state` is not used.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
```diff
@@ -0,0 +1,328 @@
...
+      state.events = state.events.slice(-state.size) // truncate events
+      state.lines = state.size
+    }
+  },
+  CLEAR_EVENTS: (state) => {
+    state.events = []
+  },
+  CLEAR_COUNTS: (state) => {
+    for(let [scope, { values = {} }] of Object.entries(state.scopes)) {
+      for(let [key, { count = 0 } ] of Object.entries(values)) {
```
cgx: `count` is not used.

lib/pf/constants/syslog.pm
```diff
@@ -107,100 +130,133 @@ our @SyslogInfo = (
...
+    (map { $_->{name} = "$log_dir/".$_->{name} ; $_ } @SyslogInfo),
+    {
+        'description' => 'Global syslog (CentOS)',
+        'name'       => '/var/log/messages',
+    },
+    {
+        'description' => 'Global syslog (Debian)',
+        'name'       => '/var/log/syslog',
+    },
+);
```
cgx, line 255: It would be nice to have a unique entry with a dynamic `name` attribute that depends on the running OS.

cgx (changes requested):
As I discussed with @satkunas, it would be more appropriate to move the log viewer toolbar bellow or above the log entries and to use the small version of each form widget. We could even make it configurable to fit all screens and user's preferences.

cgx:
The timestamps are not highlighted when I'm tailing `/var/log/messagse.log`:

Screen Shot 2020-05-25 at 10 22 58 AM

julsemaan:
> The timestamps are not highlighted when I'm tailing `/var/log/messagse.log`:
> 
> Screen Shot 2020-05-25 at 10 22 58 AM

I think by default the highlighting is off and you need to toggle it, @satkunas can confirm that

satkunas:
> The timestamps are not highlighted when I'm tailing `/var/log/messagse.log`:

Highlighting is disabled by default. Change `raw` to `color`:

Screenshot_20200525_144523

extrafu:
@julsemaan Are we waiting for anything else on that PR or it's good to merge?

julsemaan:
As discussed in the meeting, @cgx will do the final review and merge

cgx:
The bottom toolbar is affecting the left toolbar.

Screen Shot 2020-05-27 at 12 00 29 PM

cgx:
Tags of pfFormChosen are not centered.

Screen Shot 2020-05-27 at 11 52 13 AM

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,533 @@
...
+                    </b-list-group-item>
+                  </template>
+                </b-list-group>
+              </template>
+            </template>
+          </div>
+        </b-col>
+        <b-col sm="9" class="pl-0" :class="`direction-${options.order}`">
+
+          <b-row class="mb-3" align-v="center" :class="(options.order === 'forward') ? 'mt-3' : 'mb-3'">
```
cgx: I propose this change:

```
          <b-row align-v="center" :class="(options.order === 'forward') ? 'pt-3 border-top' : 'pb-3 border-bottom'">
```

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,533 @@
...
+
+              <b-button-group class="mx-1 ml-3" size="sm" :disabled="!events || !events.length">
+                <b-button @click="copyEvents()" variant="outline-primary">{{ $t('Copy') }}</b-button>
+                <b-button @click="saveEvents()" variant="outline-primary">{{ $t('Save') }}</b-button>
+                <b-button @click="clearEvents()" variant="outline-danger">{{ $t('Clear') }}</b-button>
+              </b-button-group>
+
+              <b-button-group class="mx-1 ml-3" size="sm" :title="$i18n.t('Choose background')" v-b-tooltip.hover.top.d300>
+                <b-button @click="options.background = 'white'" :active="options.background === 'white'" variant="outline-dark">
+                  <icon name="font" class="text-dark" />
```
cgx: Instead of using `font` which we already use for size options, I propose to use `sun` and `moon`.

html/pfappserver/root/static.alt/src/main.js
cgx: Still relevent.

html/pfappserver/root/static.alt/src/styles/_utilities.scss
cgx: @satkunas What's your opinion on this?

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
cgx: Still relevant.

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,517 @@
...
+        if (allowed) {
+          this.files = allowed.map(item => {
+            return { name: item.text, value: item.value }
+          })
+        }
+      })
+    },
+    toggleFilter (scope, key) {
+      this.$store.dispatch(`$_live_logs/${this.id}/toggleFilter`, { scope, key }).then(() => {
+        // noop
```
cgx: Remove useless noop.

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,517 @@
...
+      })
+    },
+    stopSession () {
+      this.$store.dispatch(`$_live_logs/${this.id}/stopSession`).then(() => {
+        // noop
+      })
+    },
+    startSession () {
```
cgx, line 361: Remove useless noop.

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,517 @@
...
+      })
+    },
+    toggleFilter (scope, key) {
+      this.$store.dispatch(`$_live_logs/${this.id}/toggleFilter`, { scope, key }).then(() => {
+        // noop
+      })
+    },
+    stopSession () {
+      this.$store.dispatch(`$_live_logs/${this.id}/stopSession`).then(() => {
+        // noop
```
cgx: Remove useless noop.

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,517 @@
...
+          this.$router.push({ name: 'live_log', params: { id: session_id } })
+        }
+        this.isStarting = false
+      }).catch(() => {
+        this.isStarting = false
+      })
+    },
+    pauseSession () {
+      this.$store.dispatch(`$_live_logs/${this.id}/pauseSession`).then(() => {
+        // noop
```
cgx: Remove useless noop.

html/pfappserver/root/static.alt/src/views/Auditing/_components/LiveLogView.vue
```diff
@@ -0,0 +1,517 @@
...
+      })
+    },
+    unpauseSession () {
+      this.$store.dispatch(`$_live_logs/${this.id}/unpauseSession`).then(() => {
+        // noop
```
cgx: Remove useless noop.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
cgx: Still relevant.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
cgx: Still relevant.

html/pfappserver/root/static.alt/src/views/Auditing/_store/liveLog.js
cgx: Still relevant.

html/pfappserver/root/static.alt/src/main.js
satkunas: `sort-numeric-up-alt`, `sort-numeric-down` are used here

html/pfappserver/root/static.alt/src/styles/_utilities.scss
satkunas: `pre` does not work well with the colored content, so a CSS definition is needed in at least one case.

Screenshot_20200527_145721

cgx (approved):

PR #5404, opened 2020-05-01
feature/LDAP-not_found_set_role_action

### LINKED ISSUE #5397: Improve ldap authentication rules
**We have sometimes some complex conditions for an authentication rule, like:**

> ($(|(cn=$username)(samaccountname=$radius.username))(|(memberof=student)(memberof=staff)))

and it's not possible to do it right now.

We should have something like a filter engine that generate the ldap filter (in a perfect world) and have condition in the authentication rule (like advanced ldap filter) where we can select the filter we created in the filter engine.

Or have a text area where we can fill the request like the rule above.

### LINKED ISSUE #5398: In the LDAP authentication create a specific action when query results are empty
If the LDAP request returns not found then have a "not_found_set_role_action" to set per example REJECT.

jrouzierinverse:
Description
===========
New action for AD/LDAP sources to set role when user is not found

NEWS file entries
=======================
Enhancements
-------------
* New action for AD/LDAP sources to set role when user is not found

Issue
=====
fixes #5398

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Field/SourceRule.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Field/SourceRule.pm
@@ -126,8 +126,8 @@ sub validate {
     }
 
     if ($class eq 'authentication') {
-        unless ($typesCount{$Actions::SET_ROLE}) {
-            $actions->add_error("You must set a role.");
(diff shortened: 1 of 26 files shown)


fdurand:
Works as expected

PR #5410, opened 2020-05-04
feature/templated_radius_attributes

### LINKED ISSUE #5406: Radius Attributes manipulation
I am facing a problem where i need to do some ldap search based on partial information in some radius attributes.

Let's say we have the radius attribute User-Name = zaym@reboot.acme.com and we need to do a ldap search like (&(cn=rebootVLAN)(memberOf=dn=cn=zaym,dc=reboot,dc=acme,dc=com)) and if there is a result then we set a role and if not found we reject.

The advanced ldap filter and set role on not found are part of another issue numbers: #5397 and #5398.

What i need is to be able to do the following:

> User-Name ~= (\w+)\@([^.]+)\.acme.com/
> 
> PacketFence-Internal-Attribute-1 = $1
> PacketFence-Internal-Attribute-2 = $2.VLAN
> PacketFence-Internal-Attribute-3 = $2
> 

Then is the ldap authentication source:

(&(cn=${PacketFence-Internal-Attribute-2})(memberOf=dn=cn=${PacketFence-Internal-Attribute-1},dc=${PacketFence-Internal-Attribute-3} = $1,dc=acme,dc=com))

jrouzierinverse:
Description
===========
Allow RADIUS filters to also have templated values like switch templates

Impacts
=======
Radius Filters

Enhancements
-------------
* Templated RADIUS filter values
* New RADIUS filter scope preProcess

Issue
=====
fixes #5406

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


fdurand:
It works but i need to have more flexibility.

Per example i have this radius attribute:
'TLS-Client-Cert-Common-Name' => 'inverse-test@staff.it.acme.edu'

and i want to extract staff and suffix it with VLAN in the internal VlanMatch attribute:

answer.0 = VlanMatch = ${BuildFromMatch($radius_request.TLS-Client-Cert-Common-Name, /regex/, “\@(\w+)”,$1."VLAN")}

jrouzierinverse:
> It works but i need to have more flexibility.
> 
> Per example i have this radius attribute:
> 'TLS-Client-Cert-Common-Name' => '[inverse-test@staff.it.acme.edu](mailto:inverse-test@staff.it.acme.edu)'
> 
> and i want to extract staff and suffix it with VLAN in the internal VlanMatch attribute:
> 
> answer.0 = VlanMatch = ${BuildFromMatch($radius_request.TLS-Client-Cert-Common-Name, /regex/, “@(\w+)”,$1."VLAN")}

I pushed a change now try
```answer.0 = VlanMatch = ${BuildFromMatch($radius_request.TLS-Client-Cert-Common-Name,"@(\w+)","$1.VLAN")}```

**Comment by fdurand** [comment]
The feature works, we just need to have the GUI part. @satkunas can you have a look ?

**Comment by fdurand** [comment]
Did all the tests on this PR and it works as expected.

**Comment by fdurand** [comment]
Tested it and work as expected.

File: lib/pf/access_filter/radius.pm [hunk 1]
```diff
@@ -17,16 +17,32 @@ use warnings;
...
 
 use base qw(pf::access_filter);
 tie our %ConfigRadiusFilters, 'pfconfig::cached_hash', 'config::RadiusFilters';
 tie our %RadiusFilterEngineScopes, 'pfconfig::cached_hash', 'FilterEngine::RadiusScopes';
 
+our %LOOKUP = (
+    session_id => \&setSession,
+);
+
+our %FUNCS = (
```
julsemaan: Since this looks exactly like the switch templates functions, we should regroup the common ones together and reuse them in both places

lib/pf/radius.pm
```diff
@@ -225,6 +226,12 @@ sub authorize {
     $args->{'ssid'} = $ssid;
     $args->{'node_info'} = $node_obj;
     $args->{'fingerbank_info'} = pf::node::fingerbank_info($mac, $node_obj);
+    my $filter = pf::access_filter::radius->new;
+    my $rule = $filter->test('fixup', $args);
```
julsemaan: What is fixup ? It isn't a very explicit name about what this does

Perhaps pre_process or something like this

fdurand:
Can you add the support for an array in PacketFence-Raw attributes ?

```
         'ldap_attribute' => {                                                                                                                                                                                                                                                  
                                'uuRadVlan' => [                                                                                                                                                                                                                                 
                                                 'Airespace-Interface-Name=wifi-pumps-hosp-uconnect',
                                                 'Aruba-User-Vlan=273'                                                                                                                                                                                                           
                                               ]                                                                                                                                                                                                                                 
                              },
```
with a radius rule like that:
```
[IF_SET_ROLE_FROM_SOURCE]
status=enabled
answer.0=reply:Packetfence-Raw = $ldap_attribute.uuRadVlan
top_op=and
description=If the role has been computed from the action set_role_from_source then return the value of the role as a radius attribute
scopes=returnRadiusAccessAccept
radius_status=RLM_MODULE_OK
merge_answer=no
condition=action == "set_role_from_source"
```

fdurand:
All have been fixed, ready for a second review.

lib/pf/config/builder/filter_engine.pm
```diff
@@ -194,6 +216,10 @@ sub buildCondition {
         die "op '$op' not handled";
     }
 
+    if ($ast eq 'true') {
```
julsemaan: Use a constant for the string value 'true' everywhere in this PR

lib/pf/radius.pm
```diff
@@ -225,6 +226,12 @@ sub authorize {
     $args->{'ssid'} = $ssid;
     $args->{'node_info'} = $node_obj;
     $args->{'fingerbank_info'} = pf::node::fingerbank_info($mac, $node_obj);
+    my $filter = pf::access_filter::radius->new;
+    my $rule = $filter->test('preProcess', $args);
+    if ($rule) {
+        my ($reply, $status) = $filter->handleAnswerInRule({%$rule, merge_answer => 'enabled' }, $args, $radius_request);
```
julsemaan, line 232: If this overrides the value in the admin, we need to state it while configuring a rule in that scope

t/unittest/access_filter/radius.t
```diff
@@ -0,0 +1,115 @@
...
+
+BEGIN {
+    #include test libs
+    use lib qw(/usr/local/pf/t);
+
+    #Module for overriding configuration paths
+    use setup_test_config;
+    @tests = (
+        {
+            name           => 'reply:Packetfence-Raw',
```
julsemaan, line 28: I believe this attribute supports array, can you test with an array too

jrouzierinverse: These tests now invalid

jrouzierinverse: Done

PR #5417, opened 2020-05-06
Support for regex Realm

fdurand:
# Description
Added support for regex realm

# Impacts
- RADIUS auth/eduroam authorize workflow
- Connection profiles
- Authentication sources

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Support for regex realm

### DIFF
--- a/html/pfappserver/root/static.alt/src/views/Configuration/_config/realm.js
+++ b/html/pfappserver/root/static.alt/src/views/Configuration/_config/realm.js
@@ -146,6 +146,7 @@ export const view = (form = {}, meta = {}) => {
         },
         {
           label: i18n.t('Regex Realm'),
+          text: i18n.t('You can define a regex realm if needed.'),
           cols: [
             {
               namespace: 'regex',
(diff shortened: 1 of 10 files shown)


html/pfappserver/root/static.alt/src/views/Configuration/_config/realm.js
```diff
@@ -139,6 +144,17 @@ export const view = (form = {}, meta = {}) => {
             }
           ]
         },
+        {
+          label: i18n.t('Regex Realm'),
+          text: i18n.t('You can define a regex realm if needed.'),
```
julsemaan: This needs much more detailing on how it should be used along with the ID

lib/pf/config/util.pm
```diff
@@ -659,12 +666,19 @@ Valid context are "portal" and "admin", basically any prefix to "_strip_username
...
     return $username unless(defined($username));
-    
+
     my $logger = get_logger;
 
     my ($stripped, $realm) = strip_username($username);
     $realm = $realm ? lc($realm) : undef;
-    
+
+    foreach my $realm_key ( keys %pf::config::ConfigRealm ) {
```
julsemaan: Shouldn't this use the ordered realms ?

PR #5374, opened 2020-04-17
Feature/radius ldap better integration

fdurand:
# Description

* New service pfcertmanager that generate certificate files based on the configuration.
* EAP configuration template, add a way to define multiples eap profiles in freeradius.
* Choose which eap profile to use based on a radius filter.
* Basedn can be define in the authentication sources rules
* New hooks for the radius filter engine in eduroam virtual server.
* Redefined "restart" in the service manager to allow "PartOf" in systemcd scripts
* Set role from source authentication rule option (needs #5459)
* Flatten the radius request for the authentication sources (attributes like radius_request.User-Name)
* Radius request attributes / Username are part of the common attributes.
* Support of multiples ldap servers in frreeradius ldap_packetfence configuration file
* copy outter user-Name attribute in PacketFence-Outer-User attribute to be able to use it in the authentication rules.
* copy the LDAP-UserDN attribute in PacketFence-UserDN attribute to be able to use it in the authentication rules.
* Added a way to extend the ldap filter for searchattributes configuration

# Impacts
* RADIUS authorize workflow
* LDAP Authentication sources.

# Code / PR Dependencies

PR #5459

# NEW Package(s) required
freeradius >= 3.0.21

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests
- [ ] Add acceptance tests (TestLink)

# NEWS file entries
## New Features
* New service pfcertmanager that generate certificate files based on the configuration.
* EAP configuration template, add a way to define multiples eap profiles in freeradius.

## Enhancements
* Choose which eap profile to use based on a radius filter.
* Basedn can be define in the authentication sources rules
* New hooks for the radius filter engine in eduroam virtual server.
* Redefined "restart" in the service manager to allow "PartOf" in systemcd scripts
* Set role from source authentication rule option (needs #5459)
* Flatten the radius request for the authentication sources (attributes like radius_request.User-Name)
* Radius request attributes / Username are part of the common attributes.
* Support of multiples ldap servers in frreeradius ldap_packetfence configuration file
* copy outter user-Name attribute in PacketFence-Outer-User attribute to be able to use it in the authentication rules.
* copy the LDAP-UserDN attribute in PacketFence-UserDN attribute to be able to use it in the authentication rules.
* Added a way to extend the ldap filter for searchattributes configuration

PR #5459, opened 2020-05-20
feature/role_dynamic_role

### LINKED ISSUE #5401: Use the value of an LDAP attribute as a role
In some cases we will need to be able to use the raw value of an LDAP attribute as a role or as a radius attribute directlt..

jrouzierinverse:
Description
===========
Use the value of an LDAP attribute as a role
Added the radius/ldap attributes in the args (can be used them in radius filter)

Impacts
=======
LDAP Authentication flow

Enhancements
-------------
* Use the value of an LDAP attribute as a role

Issue
=====
fixes #5401

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


fdurand:
It looks to work but we will probably have to add logic.

Let's say we use SET_ROLE_FROM_SOURCE = vlanid , the default behaviour is to match the value of the vlanid attribute to a role we already defined in packetfence.
But we would be able to override that with a radius filter like if action is set_role_from_source -> radius reply => raw:${node_info.category}

So the radius reply will have the value of the role. something like Airespace-Interface-Name=zeroclients

fdurand:
Ready for a review

fdurand:
The admin gui part for SET_ROLE_FROM_SOURCE is in https://github.com/inverse-inc/packetfence/pull/5374

PR #5464, opened 2020-05-21
Allow to run a scan even if we don't have OS

### LINKED ISSUE #5463: Scan engine: unable to trigger if node has not been profiled
**Describe the bug**
If you configure a scan engine on a connection profile and a node matchs this connection profile, PacketFence will trigger a scan only if node's OS is defined due to this [test](https://github.com/inverse-inc/packetfence/blob/d0d44624fa88c6ce28733c344b01372a541e5a2b/lib/pf/Connection/Profile.pm#L571).

It means that, currently, scan engine features will work only if Fingerbank is enabled.

**To Reproduce**
1. Disable Fingerbank
1. Create a scan engine without OS restrictions
2. Assign scan engine to your default connection profile
3. Trigger a RADIUS request from a node 
4. Check `packetfence.log`:

```
Can't find scan engine for xx:xx:xx:xx:xx:xx since we don't have it's OS
```

**Expected behavior**
If there is not OS restriction on scan engine definition, we should be able to scan a node even if we don't have its OS.

nqb:
# Description
Run a scan even if we don't have OS for a node.

# Impacts
Scan engine workflow.

# Issue
fixes #5463 

# Delete branch after merge
YES

### DIFF
--- a/lib/pf/Connection/Profile.pm
+++ b/lib/pf/Connection/Profile.pm
@@ -568,10 +568,6 @@ sub findScan {
 
     $node_attributes ||= node_attributes($mac);
     my $os = $node_attributes->{'device_type'};
-    unless(defined $os){
-        $logger->warn("Can't find scan engine for $mac since we don't have it's OS");
-        return;
-    }
 
     return first { $_->match($os,$node_attributes) } @scanners;
 }
@@ -594,11 +590,6 @@ sub findScans {
 
     $node_attributes ||= node_attributes($mac);
     my $os = $node_attributes->{'device_type'};
-    
-    unless(defined $os){
-        $logger->warn("Can't find scan engine for $mac since we don't have it's OS");
-        return;
-    }
 
     return grep { $_->match($os,$node_attributes) } @scanners;
 }
--- a/lib/pf/scan.pm
+++ b/lib/pf/scan.pm
@@ -344,8 +344,8 @@ sub matchCategory {
     my $category = [split(/\s*,\s*/, $self->{_categories})];
     my $node_cat = $node_attributes->{'category'};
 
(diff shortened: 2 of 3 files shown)


lib/pf/Connection/Profile.pm
```diff
@@ -568,10 +568,6 @@ sub findScan {
 
     $node_attributes ||= node_attributes($mac);
     my $os = $node_attributes->{'device_type'};
-    unless(defined $os){
```
julsemaan, line 571: We need to move this logic in pf::scan::matchOS after the 'return true if @oses == 0' so that it doesn't attempt to match with an undefined OS

nqb:
@julsemaan, I adjusted code as we spoke together. I will need to adjust unit tests before this one get merged.

nqb:
Ready for another review.

nqb:
@julsemaan, should I need to make some adjustements ?

julsemaan:
Looks good to me.

I'm fine with merging in devel but we'd just have to make sure this is tested before cherry-picking in maintenance since a bad change in this code path can have serious consequences

nqb:
> I'm fine with merging in devel but we'd just have to make sure this is tested before cherry-picking in maintenance since a bad change in this code path can have serious consequences

I adjusted unit tests to be sure we don't miss anything and they all passed on my side. What you can do is cherry-pick and run unit tests to be sure all is fine before updating repo.

nqb:
@julsemaan, what do you want to do ?

nqb:
@julsemaan, I suggest we merge this in devel. Testing each scenario to backport on maintenance branches will take a while and I don't have a full environment to test it.

julsemaan:
I will open an issue to patch this in the maintenance after 10.2 is released. That way we'll have tested the fix for the release

PR #5468, opened 2020-05-22
fix pf-maint.pl while running behind a proxy

### LINKED ISSUE #3425: pf-maint.pl doesn't take *_proxy environment variables into account

julsemaan:
# Description
Fix pf-maint.pl while its running behind a proxy

# Impacts
pf-maint.pl

# Issue
fixes #3425 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed pf-maint.pl when its running behind a proxy (#3425 )

### DIFF
--- a/addons/pf-maint.pl
+++ b/addons/pf-maint.pl
@@ -40,6 +40,7 @@ use Pod::Usage;
 use IO::Handle;
 use Term::ReadKey;
 use IO::Interactive qw(is_interactive);
+use LWP::Protocol::connect;
 our $GITHUB_USER = 'inverse-inc';
 our $GITHUB_REPO = 'packetfence';
 our $PF_DIR      = $ENV{PF_DIR} || '/usr/local/pf';
@@ -52,6 +53,9 @@ our $GIT_BIN = '/usr/bin/git';
 our $COMMIT_ID_FILE = catfile($PF_DIR,'conf','git_commit_id');
 our $test;
 our $TERMINAL_WIDTH;
+
+$ENV{PERL_NET_HTTPS_SSL_SOCKET_CLASS} = "Net::SSL";
+$ENV{PERL_LWP_SSL_VERIFY_HOSTNAME} = 0;
 
 # Files that should be excluded from patching
 # Will only work when using git to patch a server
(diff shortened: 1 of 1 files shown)


addons/pf-maint.pl
```diff
@@ -277,7 +281,11 @@ sub get_url {
     my $request  = HTTP::Request->new( GET => $url ), my $response_body;
     my $ua       = LWP::UserAgent->new;
     $ua->show_progress(1);
-    $ua->env_proxy;
+    for my $proxy_var qw(http_proxy HTTP_PROXY) {
+        if($ENV{$proxy_var} && $ENV{$proxy_var} =~ /^http:\/\/(.+)/) {
```
fdurand: ^https ?!

julsemaan: ^Done ?!

addons/pf-maint.pl
```diff
@@ -40,6 +40,7 @@ =head1 SYNOPSIS
 use IO::Handle;
 use Term::ReadKey;
 use IO::Interactive qw(is_interactive);
+use LWP::Protocol::connect;
```
fdurand, line 43: I don't see this package in the spec and in the debian dependencies, is it an extra package or does it come with another one ?

julsemaan: I think its the Fingerbank Perl lib that uses it so its installed on PF installs. Let me add it explicitly in the PF packaging in case we change the Fingerbank Perl packaging at some point

julsemaan: I added that package in our spec, its already in our repo so that should be golden

PR #5578, opened 2020-06-12
Feature/tenant id unifiedapiclient

### LINKED ISSUE #5554: unifiedapiclient Needs to be able to set which tenant for the call

julsemaan:
# Description
Adds the ability to set the X-PacketFence-Tenant-Id in the Perl and Golang UnifiedAPI clients

# Impacts
Unified API clients

# Issue
fixes #5554

# Delete branch after merge
YES

### DIFF
--- a/lib/pf/api/unifiedapiclient.pm
+++ b/lib/pf/api/unifiedapiclient.pm
@@ -112,6 +112,14 @@ The current token that was obtained by the login
 
 has token => (is => 'rw');
 
+=head2 tenant_id
+
+The tenant ID to send in the X-PacketFence-Tenant-Id header if any. When sent to undef (default value), it doesn't send it
+
+=cut
+
+has tenant_id => (is => 'rw', default => sub{undef});
+
 use constant REQUEST => 0;
 use constant RESPONSE => 2;
 use constant NOTIFICATION => 2;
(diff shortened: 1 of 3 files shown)


lib/pf/api/unifiedapiclient.pm
```diff
@@ -183,7 +191,11 @@ sub call {
     $curl->setopt(CURLOPT_CUSTOMREQUEST, $method);
 
     if($self->token) {
-        $curl->setopt(CURLOPT_HTTPHEADER, ["Authorization: Bearer ".$self->token]);
+        $curl->pushopt(CURLOPT_HTTPHEADER, ["Authorization: Bearer ".$self->token]);
+    }
+
+    if(defined($self->tenant_id)) {
```
jrouzierinverse: We should use the global tenant_id if this is undefined to avoid having to change all places where we use pf::api::unifiedapiclient.

jrouzierinverse (changes requested):

lib/pf/api/unifiedapiclient.pm
julsemaan: If you don't specify the header, it makes it default to the tenant ID associated to the token.

This piece of code handles the case where its not specified:
https://github.com/inverse-inc/packetfence/blob/devel/go/api-frontend/aaa/authorization.go#L241-L255

Its better to only specify the header to the api-frontend when its needed than to include it at all time when performing backend work

jrouzierinverse: This would mean every place we use the ```pf::api::unifiedapiclient``` we must explicitly define which tenant_id we are using when 99.9% of the case we would be using the current globally defined one.

It is pretty reasonable to assume if we are making an API call then we should the current globally defined one.

By not using the globally defined one it just makes it easier to make to forget to use.

julsemaan: It uses the tenant associated to the credentials if its not specified.

Since the credentials that are set in the client by default are the ones of the system user (associated to the global tenant), so by using it as we always do, it already does what you describe (i.e. using the global tenant) without specifying the tenant header

jrouzierinverse: I understand the miscommunication the 'global tenant' is not the one in the database.
I meant the globally defined variable ```$pf::config::tenant::CURRENT_TENANT```

julsemaan: I see, we were indeed talking of 2 completely different things

Your idea is good to prevent duplication of efforts

PR #5590, opened 2020-06-15
Added a role selection field for self reg portal

### LINKED ISSUE #5114: Allow users to select a different role in Self Service Portal
We have certain scenarios where it would be nice for users to be able to 'self select' a new role from the /status page.  Currently, all devices get a generic _Gaming_ role through the self service portal.  This works very well 90% of the time, but, because all devices on our network are behind a firewall and NAT'd, users end up getting strict NAT errors when trying to game.

We have created a _NoNAT_ role (and corresponding VLAN) to deal with this where NAT is not used and gamers are happy.  For now, users have to open a ticket with us, and we have to apply the new role to their specific device.

If PacketFence admins could maintain a list of which roles are 'self selectable' and users could go to /status to modify their roles as needed (and only choose a role from the 'special' list), that would helpful to us.

For now, having the _NoNat_ role be the default role for our self service portal is not an option as we don't have enough IP space.

fdurand:
# Description
Choose the role to use to register a device on the self service page

# Impacts
Self-service page

# Issue
fixes #5114

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Select the role of the device when register via self-service portal.

# UPGRADE file entries

=== Self registration portal

device_registration_role has been renamed device_registration_roles in order to apply the change run the following script:
./addons/upgrade/to-10.2-selfservice-conf.pl

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
@@ -49,10 +49,11 @@ has_field 'roles_allowed_to_unregister' =>
 has_field 'device_registration_role' =>
   (
    type => 'Select',
+   multiple => 1,
    label => 'Role to assign',
    options_method => \&options_roles,
    tags => { after_element => \&help,
-             help => 'The role to assign to devices registered from the self-service portal. If none is specified, the role of the registrant is used.' },
+             help => 'The role to assign to devices registered from the self-service portal. If none is specified, the role of the registrant is used. If multiples are defined then the user will have to choose' },
   );
 
 has_field 'device_registration_access_duration' =>
@@ -80,18 +81,6 @@ has_field 'device_registration_allowed_devices' =>
    fingerbank_model => "fingerbank::Model::Device",
   );
 
-has_field 'device_registration_list_roles' =>
-  (
-   type => 'Select',
-   label => 'Allowed roles list',
-   multiple => 1,
-   element_class => ['chzn-deselect'],
-   element_attr => {'data-placeholder' => 'Click to add a role'},
-   options_method => \&options_roles,
-   tags => { after_element => \&help,
-             help => 'The list of roles that are allowed to be used by the user to select in which role the device should be register.' },
-  );
-
 has_block definition =>
   (
    render_list => [ qw(id description) ],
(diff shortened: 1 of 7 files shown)


html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
```diff
@@ -80,6 +80,18 @@ has_field 'device_registration_allowed_devices' =>
    fingerbank_model => "fingerbank::Model::Device",
   );
 
+has_field 'device_registration_list_roles' =>
```
julsemaan: What happens when you set this and a device registration role:
https://github.com/inverse-inc/packetfence/blob/611a4ea93c3b2909273fdbe742f412a69c7c03a8/html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm#L49

I think that instead of creating a new field, lets just expand the existing one and when you put more than 1 role it lets the user pick it. If you put just 1 role, it forces it on the node and 0 role, it uses the one of the user

fdurand: Yes good catch

html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
```diff
@@ -49,10 +49,11 @@ has_field 'roles_allowed_to_unregister' =>
 has_field 'device_registration_role' =>
   (
    type => 'Select',
+   multiple => 1,
    label => 'Role to assign',
    options_method => \&options_roles,
    tags => { after_element => \&help,
-             help => 'The role to assign to devices registered from the self-service portal. If none is specified, the role of the registrant is used.' },
+             help => 'The role to assign to devices registered from the self-service portal. If none is specified, the role of the registrant is used. If multiples are defined then the user will have to choose' },
```
julsemaan, line 56: The help needs to be adjusted in the frontend code, this here isn't honored anymore

html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
```diff
@@ -49,10 +49,11 @@ has_field 'roles_allowed_to_unregister' =>
 has_field 'device_registration_role' =>
```
julsemaan: also lets get a sed going to rename this to `device_registration_roles`

PR #5613, opened 2020-06-18
feature/go-pfmon

jrouzierinverse:
Description
===========
A go based pfmon

NEWS file entries
=======================
New Features
-------------
* go based pfmon

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

Upgrade Instructions
=========
Rename PFMON -> PFCRON
-------------
/usr/local/pf/bin/addons/upgrade/to-10.2-adminroles-conf.pl


julsemaan:
One general comment I have about this is that the new binary shouldn't be called pfmaint since it can easily be confused with pf-maint.pl

We should keep the name pfmon

fdurand:
Does the new pfmon binary is aware of the tasks it can run on all cluster members versus some of them who needs to run on only one server ?

jrouzierinverse:
> Does the new pfmon binary is aware of the tasks it can run on all cluster members versus some of them who needs to run on only one server ?

No, I will add support for running on only the master server since that is the current behavior of pfmon.

jrouzierinverse:
It now only runs on the master in the cluster.

nqb:
After this one is merged, will `pfmon` continues to exist or should we need to replace `pfmon` by `pfcron` in all places (for instance in documentation) ?

addons/dev-helpers/pfmon_to_maintenance.pl
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/perl
```
julsemaan: Do we need to keep this given we have an upgrade script

go/cron/window_batch_task.go
```diff
@@ -0,0 +1,67 @@
+package maint
```
julsemaan, line 1: I think it would be important to have a test for this using sample data in the DB to ensure this logic works

go/cron/pfcronjob.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+type PfcronJob struct {
+	Task
+}
+
+func (j *PfcronJob) Run() {
+	cmd := exec.Command("/usr/local/pf/bin/pfcmd", "pfcron", j.Type)
+	err := cmd.Run()
+	_ = err
```
julsemaan: We should output this error when it happens

go/cron/config.go
```diff
@@ -0,0 +1,94 @@
...
+		`UPDATE radacct SET acctstoptime = NOW() WHERE acctstarttime < DATE_SUB(?, INTERVAL ? SECOND) AND acctstoptime IS NULL LIMIT ?`,
+		`DELETE FROM radacct WHERE acctstarttime < DATE_SUB(?, INTERVAL ? SECOND) AND acctstoptime IS NOT NULL LIMIT ?`,
+		`DELETE FROM radacct_log WHERE timestamp < DATE_SUB(?, INTERVAL ? SECOND) LIMIT ?`,
+	),
+	"bandwidth_maintenance_session": MakeWindowSqlJobSetupConfig(
+		`UPDATE bandwidth_accounting, (SELECT node_id, unique_session_id, MAX(last_updated) FROM bandwidth_accounting GROUP BY node_id, unique_session_id HAVING MAX(last_updated) < DATE_SUB(?, INTERVAL ? SECOND) AND MAX(last_updated) > '0000-00-00 00:00:00' LIMIT ?) as old_sessions SET last_updated = '0000-00-00 00:00:00' WHERE (bandwidth_accounting.node_id, bandwidth_accounting.unique_session_id) = (old_sessions.node_id, old_sessions.unique_session_id)`,
+	),
+}
+
+func GetMaintenanceConfig() map[string]interface{} {
```
julsemaan: It seems that modifying a window in the configuration would require to restart the pfmon service.

We should either:
- Get the GUI adjusted to ensure users are aware that they need to restart when modifying any of the parameters
- Refresh the configuration on every loop start

go/cron/cluster.go
```diff
@@ -0,0 +1,69 @@
...
+				logError(ctx, "Error calling "+clientApi.Host+": "+err.Error())
+			}
+		}
+	}
+}
+
+func NotifyCluster(ctx context.Context, method string, args interface{}) {
+	servers, cluster_mode := enabledServers(ctx)
+	if cluster_mode {
+		clientApi := jsonrpc2.NewClientFromConfig(context.Background())
```
julsemaan: you should use the context that is provided in the function

go/cron/cluster.go
```diff
@@ -0,0 +1,69 @@
...
+		return false
+	}
+
+	return !info.IsDir()
+}
+
+func CallCluster(ctx context.Context, method string, args interface{}) {
+	servers, cluster_mode := enabledServers(ctx)
+	if cluster_mode {
+		clientApi := jsonrpc2.NewClientFromConfig(context.Background())
```
julsemaan: you should use the context that is provided in the function

go/cron/cluster.go
```diff
@@ -0,0 +1,69 @@
+package maint
```
julsemaan: The content of this file should be in a cluster package so that we can reuse this in our codebase

go/cron/batch.go
```diff
@@ -0,0 +1,93 @@
+package maint
```
julsemaan, line 1: This should have a test file

julsemaan (changes requested):

lib/pf/cmd/pf/pfmon.pm
```diff
@@ -67,71 +67,8 @@ pf::cmd::pf::pfmon
 
 use strict;
 use warnings;
-use pf::config::pfmon qw(%ConfigPfmon);
```
fdurand, line 70: Why not just remove the file ?

lib/pf/file_paths.pm
```diff
@@ -330,6 +332,8 @@ $portal_modules_config_file = catfile($conf_dir,"portal_modules.conf");
 $portal_modules_default_config_file = catfile($conf_dir,"portal_modules.conf.defaults");
 $pfmon_config_file = catfile($conf_dir,"pfmon.conf");
```
fdurand: Understand that this line is still here for the upgrade script but can we remove it and use the full path in the upgrade script ?

lib/pf/cmd/pf/pfmon.pm
jrouzierinverse: I left it for backward compatibility of cron jobs.

jrouzierinverse:
Please rereview

go/cron/certificates_check_test.go
```diff
@@ -0,0 +1,191 @@
+package maint
+
+import (
+	"testing"
+)
+
```
fdurand, line 6: Does it make sense to test an expired certificate ?

fdurand:
There is still references to pfmon in the code.

```
addons/backup-and-maintenance.sh
addons/monit/monit_checks_configurations/00_packetfence.tt
conf/log.conf.d/pfmon.conf.example
conf/monitoring/statsd.d/packetfence.conf.example
docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
docs/PacketFence_Clustering_Guide.asciidoc
docs/api/spec/openapi.yaml
lib/pf/constants/admin_roles.pm
....
```

What do we do with that ?

extrafu:
@jrouzierinverse Advised on Fabrice's last comment and let's wrap this up once for all.

jrouzierinverse:
Please re-review

addons/upgrade/to-10.2-adminroles-conf.pl
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+to-9.1-selfservice-conf.pl
```
fdurand: wrong description

addons/upgrade/to-10.2-adminroles-conf.pl
```diff
@@ -0,0 +1,81 @@
...
+
+=cut
+
+=head1 DESCRIPTION
+
+Remove references to DEVICE_REGISTRATION_READ in adminroles.conf if there are any
```
fdurand: wrong description

fdurand (changes requested):

fdurand:
On my side it looks ok, i let keep @julsemaan doing a last shot.

julsemaan:
My comments were marked as resolved so I'm fine with this being merged

PR #5623, opened 2020-06-22
Feature/cisco coa improvement

julsemaan:
# Description
Ensure CoA will be used in Meraki web-auth

# Impacts
CoA for all vendors, Meraki deauth attributes

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed CoA for Meraki web-authentication so that it doesn't disconnect the user from the SSID


julsemaan:
I've put this as a +1 unless someone objects from this being in the maintenance

PR #5646, opened 2020-06-29
Fix Brocade VoIP

lyubomirtraykov:
On Brocade Switch the VoIP vlan need to be tagged.


julsemaan:
You did the right thing by editing the template_switches.conf.defaults but the original author (@lzammit) should have put this in a `.def` file.

I'll go ahead and create the `.def` including your fix

Thanks for the contribution

PR #5655, opened 2020-07-02
Fix RADIUS Audit Log CoA parsing

lyubomirtraykov:
Fix RADIUS Audit Log CoA Request and Reply parsing.

### DIFF
--- a/lib/pf/util/radius.pm
+++ b/lib/pf/util/radius.pm
@@ -195,9 +195,9 @@ record CoA in the radius audit log
 
 sub record_coa {
     my ($connection_info, $radius_code, $attributes, $vsa, %return) = @_;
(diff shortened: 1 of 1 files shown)


julsemaan:
I think its intended that it doesn't use the `\n` to separate lines.

Are you getting display issues in the admin when viewing CoA audit records ?

lyubomirtraykov:
Yes, at the end of every attribute i get ` ", ` and the last attribute and  VSA attributes are one line.

jrouzierinverse:
That is intended, freeradius escapes special characters before the request to the DB.
Can you send a screenshot?

lyubomirtraykov:
This is how it looks:
Screen Shot 2020-07-09 at 10 09 07
And this is how it looks with my fix:
Screen Shot 2020-07-09 at 9 22 43

PR #5675, opened 2020-07-08
fix/5659

### LINKED ISSUE #5659: Network Interface Configuration Issue
**Describe the bug**

Happen if you use the HWADDR parameter to map the interface with a specific name.

When you have a network interface configured like that:

```
DEVICE=em1
ONBOOT=yes
HWADDR=00:50:56:20:0F:85
TYPE=Ethernet
BOOTPROTO=none
IPADDR=10.255.204.90
NETMASK=255.255.252.0
DNS1=10.255.232.204
DNS2=10.255.232.216
```

and when you go in the configurator and set the interface as management (or reg/isol) then the configuration become like that:

```
DEVICE=em1
ONBOOT=yes
TYPE=Ethernet
BOOTPROTO=none
IPADDR=10.255.204.90
NETMASK=255.255.252.0
DNS1=10.255.232.204
DNS2=10.255.232.216
```
and you lost the HWADDR configuration parameter.
Then if you rebbot the server than you will lost the network access.

We should at least fetch the existing parameters (like only the one we doesn't changre from the admin gui) and merge them with the new ones.

jrouzierinverse:
Description
===========
Do not override user defined values in the interface file for centos.

NEWS file entries
=======================
Bug Fixes
-------------
* Don't override user defined values in the interface file for centos.

Issue
=====
fixes #5659

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/html/pfappserver/lib/pfappserver/Model/Config/System.pm
+++ b/html/pfappserver/lib/pfappserver/Model/Config/System.pm
@@ -345,29 +345,33 @@ sub writeNetworkConfigs {
 
     while (my ($interface, $interface_values) = each %$interfaces_ref) {
         next if ( !$interface_values->{is_running} );
+        my %vars = (
+            logical_name => $interface,
+            vlan_device  => $interface_values->{'vlan'},
+            hwaddr       => $interface_values->{'hwaddress'},
+            ipaddr       => $interface_values->{'ipaddress'},
+            netmask      => $interface_values->{'netmask'},
+            ipv6_address => $interface_values->{'ipv6_address'},
+            ipv6_prefix  => $interface_values->{'ipv6_prefix'},
+        );
 
-        my $vars = {
-            logical_name    => $interface,
-            vlan_device     => $interface_values->{'vlan'},
-            hwaddr          => $interface_values->{'hwaddress'},
-            ipaddr          => $interface_values->{'ipaddress'},
-            netmask         => $interface_values->{'netmask'},
-            ipv6_address    => $interface_values->{'ipv6_address'},
-            ipv6_prefix     => $interface_values->{'ipv6_prefix'},
-        };
 
         my $template = Template->new({
             INCLUDE_PATH    => "/usr/local/pf/html/pfappserver/root/interface",
             OUTPUT_PATH     => $var_dir,
         });
-        $template->process( "interface_rhel.tt", $vars, $_interface_conf_file.$interface );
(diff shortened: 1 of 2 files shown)


julsemaan:
This seems like it doesn't enforce NM_CONTROLLED=no like we currently do (this ensures that although the NetworkManager is running, it doesn't interfere with our interfaces)

julsemaan:
I think we should adopt a strategy of tracking which configuration items we don't want to touch

Otherwise we may leave incompatible configuration in place

jrouzierinverse:
It finds all the variables that were not in the new config from the old config and appends them to the new config.
So the case of 'NM_CONTROLLED=no' since it is in the new config.

julsemaan:
> It finds all the variables that were not in the new config from the old config and appends them to the new config.
> So the case of 'NM_CONTROLLED=no' since it is in the new config.

doesn't seem to work

I modified my interface to use:
```
DEVICE=eth0
HWADDR=
ONBOOT=yes
BOOTPROTO=static
NM_CONTROLLED=yes
IPADDR=172.20.20.109
NETMASK=255.255.0.0
TEST=hello


IPV6INIT=yes
IPV6ADDR=fd9b:46e6:e651:0000:020c:29ff:feda:b089/64
```

And then in the admin, I saved it and it stayed with NM_CONTROLLED=yes

jrouzierinverse:
Did you restart your pfperl-api? It works on my side.

julsemaan:
I just retried ensuring that pfperl-api was restarted.

Using the configuration I put above, the setting `NM_CONTROLLED=yes` stays in place although it should have been replaced by `NM_CONTROLLED=no`

julsemaan:
Also, when I removed my IPv6 address in the admin, it left the 2 IPV6 directives in the file

jrouzierinverse:
I did see the issue with the IPV6 directives in the file.
However, I am not seeing the NM_CONTROLLED=yes issue.
Can you paste the before and after of the network config here, please?

jrouzierinverse:
Also, I just pushed the fix for the IPV6 directives

julsemaan:
> I did see the issue with the IPV6 directives in the file.
> However, I am not seeing the NM_CONTROLLED=yes issue.
> Can you paste the before and after of the network config here, please?

Config wasn't changing due to a permission issue for a file in var/ for the NM_CONTROLLED test I was doing

I tested again against your latest commit and all looks good on my end now

PR #5682, opened 2020-07-09
Feature/configurable portal fake ip

### LINKED ISSUE #5679: 192.0.2.1 IP address may be used by another system
**Describe the bug**
I just got a case where that IP was active on a network equipment that was between PF and the endpoint. It was intercepting traffic and preventing users from being on the portal.

**To Reproduce**
1. Put 192.0.2.1 on another equipment than PF that is directly connected to the endpoint (like on the WLC directly)
2. Try to use the reg VLAN

**Expected behavior**

I suggest one of the following:
 1. We buy a public IP address for the exclusive usage of this mechanism (will cost a few bucks a month)
 1. We allow the IP to be changed via configuration so that we can pick something that doesn't cause a problem if we hit similar cases

For now I've worked around by replacing the IP with another one in the code but that won't survive the maintenance (change in pfdns)

julsemaan:
# Description
Use an Inverse owned IP for the captive portal IP (instead of 192.0.2.1)
Configurable captive portal IP

# Impacts
Captive portal detection and access

# Issue
fixes #5679 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Removed hardcoded 192.0.2.1 IP address for the L2 captive portal and switch to an Inverse owned IP
* Allow configuration of the IP address for the L2 captive portal

### DIFF
--- a/html/pfappserver/root/static.alt/src/views/Configuration/_config/captivePortal.js
+++ b/html/pfappserver/root/static.alt/src/views/Configuration/_config/captivePortal.js
@@ -18,7 +18,7 @@ export const view = (form = {}, meta = {}) => {
       rows: [
         {
           label: i18n.t('IP address'),
-          text: i18n.t('The IP address the portal uses in the registration and isolation networks. Do not change unless you know what you are doing.'),
+          text: i18n.t('The IP address the portal uses in the registration and isolation networks. Do not change unless you know what you are doing. Changing this requires to restart all of the PacketFence services.'),
           cols: [
             {
               namespace: 'ip_address',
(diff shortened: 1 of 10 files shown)


fdurand:
Still a ref here:
https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/services/manager/haproxy_admin.pm#L287

julsemaan:
Adjusted the missing ref

html/pfappserver/root/static.alt/src/views/Configuration/_config/captivePortal.js
```diff
@@ -16,6 +16,17 @@ export const view = (form = {}, meta = {}) => {
     {
       tab: null,
       rows: [
+        {
+          label: i18n.t('IP address'),
+          text: i18n.t('The IP address the portal uses in the registration and isolation networks. Do not change unless you know what you are doing.'),
```
fdurand: Maybe just a note that you need to restart haproxy-portal and keepalived.

julsemaan: I added a note to restart all the services since that IP is in a lot of places and may appear in other places at some point

PR #5701, opened 2020-07-16
feature/generalize-makefile

jrouzierinverse:
Description
===========
Use the golang cmd directory convention and generalize the Makefile.

Impacts
=======
Golang complilation

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


config.mk
```diff
@@ -14,7 +14,7 @@ SBINDIR = $(PF_PREFIX)/sbin
 # Golang
 #
 GOVERSION = go1.13.1
-GOBINARIES = pfhttpd pfdhcp pfdns pfstats pfdetect galera-autofix pfacct pfcertmanager
+GOBINARIES = pfhttpd pfdhcp pfdns pfstats pfdetect galera-autofix pfacct pfcertmanager pfmaint
```
julsemaan: I guess pfmaint is the new pfmon but that's clearly not a good name, it will be easily confused with pf-maint.pl

I think the name pfmon should still be kept

extrafu: I agree on this one, I was surprised to see it renamed to pfmaint - I strongly prefer pfmon

jrouzierinverse: pfmon name did not make sense to me since it is for maintenance tasks not monitoring.
However, I do agree that it should it could be confused for pf-maint.pl.
Maybe pfcron would be a better name since make more sense of it's purpose.

julsemaan: pfcron sounds like a good name for me

nqb: Me too

jrouzierinverse:
bump

julsemaan:
I'll launch a build on that branch to ensure it builds fine and then I'll merge

nqb:
Before merge, it will be nice to test a build on maintenance/99.9 to be sure all Golang binaries will be built correctly for maintenance.

go/Makefile
```diff
@@ -100,6 +15,20 @@ test:
...
+	cd cmd/$@ && \
+	go build && \
+	mv $@ ../..
+
+BINARYS_RACE = $(patsubst cmd/%, %-race, $(CMDS))
+$(BINARYS_RACE):
+	cd cmd/$(@:-race=) && \
+	go build -o $@ -race && \
+	mv $@ ../..
+
```
nqb, line 31: `GOBINARIES` will need to be replace by `BINARYS` in this Makefile and also in `config.mk` in order to build `pfcron`. If we do this, we can remove `BINARYS` from `config.mk` because it doesn't seems to be used anywhere.

go/Makefile
```diff
@@ -100,6 +15,20 @@ test:
 	/usr/local/pf/t/pfconfig-test ;\
 	PFCONFIG_TESTING=y go test -count=1 ./...
 
+CMDS = $(wildcard cmd/*)
+BINARYS = $(patsubst cmd/%, %, $(CMDS))
```
julsemaan: its BINARIES not BINARYS

go/Makefile
jrouzierinverse: pfcron is not merged into devel yet.

jrouzierinverse: GOBINARIES is for go binaries that will be placed in sbin.
BINARIES is for all the binaries including db-tester and radius-tester.
Should we rename it to GOBINARIES GOSBINBINARIES?

nqb: Got your point. In that case, I suggest to:
- replace BINARIES by ALL_BINARIES
- replace GOBINARIES by PF_BINARIES

to be more explicit.

Using `_` in variable names is more easy to read.

jrouzierinverse:
Done

extrafu:
@jrouzierinverse Rebase. @julsemaan if all fine after, please merge.

jrouzierinverse:
rebased

PR #5704, opened 2020-07-17
Feature/exos module

julsemaan:
# Description
Add a new EXOS module to support newer features for Extreme switches

# Impacts
Just a new module so not much

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Improved support for Extreme switches running EXOS

### DIFF
--- a/docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
+++ b/docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
@@ -2882,7 +2882,7 @@ First define your policy in the switch:
   configure policy maptable response both
   configure policy vlanauthorization enable
 
-Next, in PacketFence, enable 'Role by Switch Role' your the switch configuration and assign the policies to the roles there. They will be returned inside the Filter-Id attribute.
+Next, in PacketFence, enable 'Role by Switch Role' in your switch configuration and assign the policies to the roles there. They will be returned inside the Filter-Id attribute.
 
 Make sure you use the 'Extreme EXOS' type for your switch to use this feature.
 
@@ -2898,7 +2898,7 @@ SNMP configuration
 AAA Configuration
 
   configure radius netlogin primary server 192.168.1.5 1812 client-ip 10.0.0.8 vr VR-Default
-  configure radius netlogin primary shared-secret 12345
+  configure radius netlogin primary shared-secret useStrongerSecret
   enable radius netlogin
 
 Web-auth profile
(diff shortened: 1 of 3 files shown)


docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -2855,7 +2868,55 @@ Netlogin (802.1X)
...
+You can assign policies defined on the switch via PacketFence.
+
+First define your policy in the switch:
+
+  configure policy profile 1 name "gaming" pvid-status "enable" pvid 3521 untagged-vlans 3521
+  configure policy profile 2 name "guest" pvid-status "enable" pvid 3522 untagged-vlans 3522
+  configure policy maptable response both
+  configure policy vlanauthorization enable
+
+Next, in PacketFence, enable 'Role by Switch Role' your the switch configuration and assign the policies to the roles there. They will be returned inside the Filter-Id attribute.
```
fdurand: Next, in PacketFence, enable 'Role by Switch Role' in your the switch configuration and assign the policies to the roles there. They will be returned inside the Filter-Id attribute.

docs/PacketFence_Network_Devices_Configuration_Guide.asciidoc
```diff
@@ -2855,7 +2868,55 @@ Netlogin (802.1X)
...
+SNMP configuration
+
+  enable snmp access snmp-v1v2c
+  configure snmp add community readonly public
+  configure snmp add community readwrite private
+
+AAA Configuration
+
+  configure radius netlogin primary server 192.168.1.5 1812 client-ip 10.0.0.8 vr VR-Default
+  configure radius netlogin primary shared-secret 12345
```
fdurand: useStrongerSecret

lib/pf/Switch/Extreme/EXOS.pm
```diff
@@ -0,0 +1,189 @@
...
+
+Returns the reference to the deauth technique or the default deauth technique.
+
+=cut
+
+sub wiredeauthTechniques {
+    my ($self, $method, $connection_type) = @_;
+    my $logger = $self->logger;
+    if ($connection_type == $WIRED_802_1X) {
+        my $default = $SNMP::SNMP;
```
fdurand, line 120: Why not $SNMP::RADIUS ?

julsemaan: SNMP::RADIUS doesn't work as well during my tests when mixing 802.1x and MAB

It seems it does MAB for a bit after the deauth, then 802.1x kicks back in after a minute or so

With up/down, 802.1x has priority like it should

julsemaan:
All should be good now

PR #5710, opened 2020-07-20
Added the way to evaluate the conditions based on the args

fdurand:
# Description
Added a way to dynamically compare a value of an attribut with a dynamic value of another attribute.
Like username == ${radius_request.User-Name}

# Impacts
Filter engine

# Delete branch after merge
YES | NO

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Dynamic comparison in filter engine

### DIFF
--- a/lib/pf/condition/includes.pm
+++ b/lib/pf/condition/includes.pm
@@ -43,7 +43,8 @@ sub match {
     my ($self, $arg, $args) = @_;
     return $FALSE if !defined $arg;
     my $reftype = reftype($arg) // '';
-    return any { $self->evalParam($self->value, $args) eq $_ } ($reftype eq 'ARRAY' ? @$arg : $arg);
+    my $value = $self->evalParam($self->value, $args);
+    return any { $value eq $_ } ($reftype eq 'ARRAY' ? @$arg : $arg);
 }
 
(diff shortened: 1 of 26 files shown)


lib/pf/condition/includes.pm
```diff
@@ -40,10 +40,10 @@ Check if the value is part of the array that is passed as an argument
 =cut
 
 sub match {
-    my ($self, $arg) = @_;
+    my ($self, $arg, $args) = @_;
     return $FALSE if !defined $arg;
     my $reftype = reftype($arg) // '';
-    return any { $self->value eq $_ } ($reftype eq 'ARRAY' ? @$arg : $arg);
+    return any { $self->evalParam($self->value, $args) eq $_ } ($reftype eq 'ARRAY' ? @$arg : $arg);
```
jrouzierinverse: Only calculate the value once
```perl
my $value = $self->evalParam($self->value, $args);
return any { $value eq $_ } ($reftype eq 'ARRAY' ? @$arg : $arg);
```

jrouzierinverse (changes requested):

fdurand:
Bump

PR #5711, opened 2020-07-20
Feature/acl generator

julsemaan:
# Description
Code that allows a switch to generate an ACL (for future usage for on the fly ACLs)

# Impacts
Switch modules but it just adds code

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

### DIFF
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -3550,17 +3550,28 @@ sub getCiscoAvPairAttribute {
     return ;
 }
 
-sub generateACL {
-    my ($self, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port) = @_;
-    my $t = '${if($allow, "permit", "deny")} $proto ${if($src_host, join(" ", "host", $src_host), "any")} ${if($src_port, join(" ", "eq", $src_port), "")} ${if($dst_host, join(" ", "host", $dst_host), "any")} ${if($dst_port, join(" ", "eq", $dst_port), "")}';
-    return pf::mini_template->new($t)->process({
+sub generateACLFromTemplate {
+    my ($self, $t, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port) = @_;
+    my $acl = pf::mini_template->new($t)->process({
         allow => $allow,
         proto => $proto,
-        src_host => $src_host,
-        src_port => $src_port,
-        dst_host => $dst_host,
-        dst_port => $dst_port,
+        src_host => $src_host // "",
+        src_port => $src_port // "",
+        dst_host => $dst_host // "",
+        dst_port => $dst_port // "",
     });
+
+    # remove double spaces to cleanup ACL
+    $acl =~ s/  / /g;
+    return $acl;
+}
+
+sub generateACL {
+    my ($self, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port) = @_;
+    return $self->generateACLFromTemplate(
+        '${if($allow, "permit", "deny")} $proto ${if($src_host, join(" ", "host", $src_host), "any")} ${if($src_port, join(" ", "eq", $src_port), "")} ${if($dst_host, join(" ", "host", $dst_host), "any")} ${if($dst_port, join(" ", "eq", $dst_port), "")}',
+        $allow, $proto, $src_host, $src_port, $dst_host, $dst_port, 
+    );
 }
 
 =back
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -62,6 +62,7 @@ use pf::access_filter::radius;
 use File::Spec::Functions;
 use File::FcntlLock;
 use JSON::MaybeXS;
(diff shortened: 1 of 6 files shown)


lib/pf/constants/switch.pm
```diff
@@ -0,0 +1,55 @@
+package pf::constants::switch;
```
jrouzierinverse, line 1: We already have a module for constants for switch.
pf::Switch::constants

lib/pf/Switch.pm
```diff
@@ -3550,6 +3551,39 @@ sub getCiscoAvPairAttribute {
...
+sub generateACLFromTemplate {
+    my ($self, $t, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port) = @_;
+    my $acl = pf::mini_template->new($t)->process({
+        allow => $allow,
+        proto => $proto,
+        src_host => $src_host // "",
+        src_port => $src_port // "",
+        dst_host => $dst_host // "",
+        dst_port => $dst_port // "",
+    });
```
jrouzierinverse: In order to make this future proof.
I would suggest passing in a hash with all the args already.
```perl
    my ($self, $t, $args) = @_
    my $acl = pf::mini_template->new($t)->process($args);
```
So if we ever need to add more args in the future it will be easy.

lib/pf/Switch.pm
```diff
@@ -3550,6 +3551,39 @@ sub getCiscoAvPairAttribute {
...
+
+=head2 generateACL
+
+Generate an ACL using the default template
+
+=cut
+
+sub generateACL {
+    my ($self, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port) = @_;
+    return $self->generateACLFromTemplate($DEFAULT_ACL_TEMPLATE, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port);
```
jrouzierinverse: To further generalize to so non-templated switch modules can use a different template.
```perl
$self->generateACLFromTemplate($self->aclTemplate(), $args);

sub aclTemplate { $DEFAULT_ACL_TEMPLATE }
```

lib/pf/Switch/Template.pm
```diff
@@ -453,6 +453,22 @@ sub deauthenticateMacRadius {
...
+sub generateACL {
+    my ($self, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port) = @_;
+    if(my $t = $self->{_template}->{acl_template}) {
+        return $self->generateACLFromTemplate($t, $allow, $proto, $src_host, $src_port, $dst_host, $dst_port);
+    }
+    else {
+        return $self->SUPER::generateACL($allow, $proto, $src_host, $src_port, $dst_host, $dst_port);
+    }
+}
+
```
jrouzierinverse, line 467: To use the generalized version suggested above.

```perl
sub aclTemplate {
    my ($self) = @_;
    return $self->{_template}{acl_template} || $DEFAULT_ACL_TEMPLATE
}
```

jrouzierinverse (changes requested):
It looks like you can only configure the ACL but it is not used anywhere.

lib/pf/constants/switch.pm
julsemaan: I know but the package name is 'SNMP' and doesn't contain another package name that this could fit in

You prefer that I define another package in pf::Switch::constants and put this in there ?

jrouzierinverse:
Why not just have this as a general attribute of the switch instead of some special case of the switch template?

julsemaan:
> Why not just have this as a general attribute of the switch instead of some special case of the switch template?

It would be like defining what RADIUS attributes go back when accepting a user in a VLAN. Its the input/output logic of the switch module so it doesn't make sense to have it as a configuration of the switch object

lib/pf/constants/switch.pm
jrouzierinverse: That package pf::Switch::constants is a bit messy.
Leave it where it is for we can start to clean it up later.
We can start to move the other constants from pf::Switch::constants later to the new package to keep things consistent.

jrouzierinverse:
> > Why not just have this as a general attribute of the switch instead of some special case of the switch template?
> 
> It would be like defining what RADIUS attributes go back when accepting a user in a VLAN. Its the input/output logic of the switch module so it doesn't make sense to have it as a configuration of the switch object

That makes sense, the only advantage I see in making the aclTemplate a configuration of the switch object is if we want to change the behavior of the dynamic ACL of a non-templated switch via configuration but keep the other logic.

At the end of the day, it is not a hard thing to add if we really need it.

julsemaan:
Adjusted code based on discussion above

jrouzierinverse:
The code looks good.
Do we want to merge it now or later?

julsemaan:
Now would be perfect, that will be the base for starting the on the fly ACL through Fingerbank

PR #5720, opened 2020-07-24
Add universal getPhonesLLDPAtIfIndex

lyubomirtraykov:
# Description
This is the most general getPhonesLLDPAtIfIndex. I did a test on Juniper, Cisco and Brocade

# Impacts
VoIP detection

### DIFF
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -1404,6 +1404,89 @@ sub isPhoneAtIfIndex {
     }
 }
 
+=head2 getPhonesLLDPAtIfIndex
+
+Return list of MACs found through LLDP on a given ifIndex.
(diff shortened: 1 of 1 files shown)


lyubomirtraykov:
@julsemaan please check if this is working with your test equipment.

julsemaan:
This code won't do much I think because the Template module doesn't have the Lldp capability

Also, adding the Lldp capability to the Template module will mean that all switch templates modules will attempt to communicate using SNMP to do the LLDP since the default switch configuration enables LLDP checks if the switch supports it. 

We need to think of a good way to prevent the LLDP check on a switch template that doesn't support it or that the user won't bother to configure SNMP on. Perhaps have a flag in the switch template that states whether or not the LLDP check should be made which would be disabled by default

PR #5734, opened 2020-08-05
feature/roles-delete

### LINKED ISSUE #3940: Unable to delete roles
There is no way to delete roles from the web GUI in 8.3.0

jrouzierinverse:
Description
===========
Roles Delete

Issue
=====
fixes #3940

NEWS file entries
=======================
Enhancements
-------------
* Allow roles to be delete

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -53,7 +53,8 @@ CREATE TABLE class (
   target_category varchar(255),
   delay_by int(11) NOT NULL default 0,
   external_command varchar(255) DEFAULT NULL,
-  PRIMARY KEY (security_event_id)
+  PRIMARY KEY (security_event_id),
+  KEY password_target_category (target_category)
 ) ENGINE=InnoDB;
 
 --
@@ -186,6 +187,7 @@ CREATE TABLE node (
   KEY `node_status` (`status`, `unregdate`),
   KEY `node_dhcpfingerprint` (`dhcp_fingerprint`),
   KEY `node_last_seen` (`last_seen`),
+  KEY `node_bypass_role_id` (`bypass_role_id`),
   CONSTRAINT `0_57` FOREIGN KEY (`tenant_id`, `pid`) REFERENCES `person` (`tenant_id`, `pid`) ON DELETE CASCADE ON UPDATE CASCADE,
   CONSTRAINT `node_category_key` FOREIGN KEY (`category_id`) REFERENCES `node_category` (`category_id`),
   CONSTRAINT `node_tenant_id` FOREIGN KEY(`tenant_id`) REFERENCES `tenant` (`id`)
@@ -477,6 +479,7 @@ CREATE TABLE `password` (
   `unregdate` datetime NOT NULL default "0000-00-00 00:00:00",
   `login_remaining` int DEFAULT NULL,
   PRIMARY KEY (tenant_id, pid),
+  KEY password_category (category),
   UNIQUE KEY pid_password_unique (pid)
 ) ENGINE=InnoDB;
 
--- a/lib/pf/UnifiedApi/Controller/Config/Roles.pm
+++ b/lib/pf/UnifiedApi/Controller/Config/Roles.pm
@@ -17,7 +17,7 @@ pf::UnifiedApi::Controller::Config::Roles
 use strict;
 use warnings;
 use pf::dal::node_category;
-use pf::error (is_error);
+use pf::error qw(is_error);
 
 use Mojo::Base qw(pf::UnifiedApi::Controller::Config);
 
@@ -64,7 +64,7 @@ SQL
 sub can_delete_from_db {
     my ($self) = @_;
     my $id = $self->id;
-    my ($status, $sth) = pf::dal::node_category->db_execute($sql, $id);
+    my ($status, $sth) = pf::dal::node_category->db_execute($CAN_DELETE_FROM_DB_SQL, $id);
     if (is_error($status)) {
         return ($status, "Unable to check role in the database");
     }
@@ -72,7 +72,7 @@ sub can_delete_from_db {
     my ($count) = $sth->fetchrow_array();
     $sth->finish;
     if ($count) {
-        (422, 'Role still in use');
+        return (422, 'Role still in use');
     }
 
     return (200, '');
--- a/lib/pf/UnifiedApi/Controller/Config/Roles.pm
+++ b/lib/pf/UnifiedApi/Controller/Config/Roles.pm
@@ -16,7 +16,8 @@ pf::UnifiedApi::Controller::Config::Roles
 
 use strict;
 use warnings;
-
+use pf::dal::node_category;
+use pf::error (is_error);
 
 use Mojo::Base qw(pf::UnifiedApi::Controller::Config);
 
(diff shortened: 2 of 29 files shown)


lib/pf/UnifiedApi/Controller/Config/Roles.pm
```diff
@@ -26,7 +25,62 @@ has 'primary_key' => 'role_id';
...
+        return ($status, $msg);
+    }
+
+    return (200, '');
+}
+
+sub can_delete_from_config {
+    my ($self) = @_;
+    if (exists $RolesReverseLookup{$self->id}) {
+        return (422, 'Role still in use');
```
julsemaan: Since roles can be used in a lot of places, we need to tell the user where its still being used so he can fix it

lib/pf/UnifiedApi/Controller/Config/Roles.pm
```diff
@@ -26,7 +25,62 @@ has 'primary_key' => 'role_id';
...
+    my ($self) = @_;
+    my $id = $self->id;
+    my ($status, $sth) = pf::dal::node_category->db_execute($CAN_DELETE_FROM_DB_SQL, $id);
+    if (is_error($status)) {
+        return ($status, "Unable to check role in the database");
+    }
+
+    my ($count) = $sth->fetchrow_array();
+    $sth->finish;
+    if ($count) {
```
julsemaan: From what I understand the code does (correct me if I'm wrong), this prevents a user from deleting a role if its still referenced in the DB.

What we need according to the discussions we had for this feature is to tell the user that its still used and allow him to:
- Assign a new role to the nodes that use this role prior to deletion
- Assign a new bypass role to the nodes that use this bypass role prior to deletion
- Assign a new role to the local users that use this role prior to deletion

These 3 should be done automatically for the user during the deletion without him needing to go track down all of these DB entries and modify them.

The security events check can stay as a check as the role would have to be changed in the security events config

lib/pf/nodecategory.pm
```diff
@@ -81,9 +81,23 @@ sub nodecategory_populate_from_config {
...
     while(my ($id, $role) = each(%$config)) {
+        push @keep, $id;
         nodecategory_upsert($id, %$role);
     }
+    _nodecategory_bulk_delete(\@keep);
+}
+
+sub _nodecategory_bulk_delete {
+    my ($keep) = @_;
+    pf::dal::nodecategory->remove_items(
```
julsemaan: This has the potential to fail due to constaints I believe.

I can imagine someone deleting the role from the configuration file manually and reloading the config.

In that case it wouldn't output the fact that the role can't be deleted from the DB due to constaints and that will have the logs show an SQL error on every configreload hard

To prevent that, the configreload hard output should say if this succeeded or not

julsemaan (changes requested):

jrouzierinverse:
Please review.

db/pf-schema-X.Y.Z.sql
```diff
@@ -477,6 +479,7 @@ CREATE TABLE `password` (
   `unregdate` datetime NOT NULL default "0000-00-00 00:00:00",
   `login_remaining` int DEFAULT NULL,
   PRIMARY KEY (tenant_id, pid),
+  KEY password_category (category),
```
julsemaan, line 482: This isn't right, it should be on bypass_role_id

lib/pf/UnifiedApi/Controller/Config/Roles.pm
```diff
@@ -26,7 +25,239 @@ has 'primary_key' => 'role_id';
...
+    $self->reassign_role_with_sql(\@errors, "pf::dal::node", $REASSIGN_NODE_BYPASS_ROLE_ID, $old, $new, "node bypass role");
+    $self->reassign_role_with_sql(\@errors, "pf::dal::person", $REASSIGN_PASSWORD_CATEGORY, $old, $new, "password category");
+    $self->reassign_role_config_store(\@errors, "pf::ConfigStore::AdminRoles", $old, $new, qw(allowed_roles allowed_node_roles));
+    $self->reassign_role_config_store(\@errors, "pf::ConfigStore::Scan", $old, $new, qw(categories));
+    $self->reassign_role_config_store(\@errors, "pf::ConfigStore::Provisioning", $old, $new, qw(category role_to_apply));
+    $self->reassign_role_config_store(\@errors, "pf::ConfigStore::SelfService", $old, $new, qw(roles_allowed_to_unregister device_registration_roles));
+    $self->reassign_role_config_store(\@errors, "pf::ConfigStore::BillingTiers", $old, $new, qw(roles_allowed_to_unregister device_registration_roles));
+    $self->reassign_role_config_store(\@errors, "pf::ConfigStore::Firewall_SSO", $old, $new, qw(categories));
+    $self->reassign_role_config_store_switch(\@errors, $old, $new);
```
julsemaan, line 184: Authentication sources seem to be missing from the list

lib/pf/UnifiedApi/Controller/Config/Roles.pm
```diff
@@ -26,7 +25,239 @@ has 'primary_key' => 'role_id';
...
+    my ($self, $errors, $class, $old, $new, @fields) = @_;
+    my $cs = $class->new;
+    my $i = 0;
+    my $cachedConfig = $cs->cachedConfig;
+    for my $sect ($cs->_Sections()) {
+        for my $f (@fields) {
+            next if !$cachedConfig->exists($sect, $f);
+            my $values = $cachedConfig->val($sect, $f);
+            my @roles = split(/\s*,\s*/, $values);
+            my @new_roles = map { $_ eq $old ? $new : $_ } @roles;
```
julsemaan, line 204: We should make this list unique to prevent the role from being there twice if it was already part of the list

lib/pfconfig/namespaces/resource/RolesReverseLookup.pm
```diff
@@ -0,0 +1,92 @@
...
+use warnings;
+use pfconfig::namespaces::config::Scan;
+use pfconfig::namespaces::config::AdminRoles;
+use pfconfig::namespaces::config::Provisioning;
+use pfconfig::namespaces::config::SelfService;
+use pfconfig::namespaces::config::BillingTiers;
+use pfconfig::namespaces::config::Firewall_SSO;
+use pfconfig::namespaces::config::VlanFilters;
+use pfconfig::namespaces::config::SecurityEvents;
+use pfconfig::namespaces::config::Switch;
```
julsemaan, line 23: Missing authentication sources here too I believe

julsemaan (changes requested):

db/pf-schema-X.Y.Z.sql
jrouzierinverse: There is no bypass_role_id for the password table.

lib/pf/UnifiedApi/Controller/Config/Roles.pm
jrouzierinverse: You are right I also missing portal modules.

lib/pf/UnifiedApi/Controller/Config/Roles.pm
jrouzierinverse: Ok

lib/pfconfig/namespaces/resource/RolesReverseLookup.pm
jrouzierinverse: Ok

db/pf-schema-X.Y.Z.sql
julsemaan: My bad, the GH diff display confused me, I thought this was on the node table

extrafu:
@julsemaan Please do a final review and merge if all is fine.

PR #5735, opened 2020-08-06
feature/switch-template-bounce

### LINKED ISSUE #5633: Restart Switchport with COA
Restart Switchport configuration can be added to Switch Templates to set the COA attributes. This way we don't need SNMP for the Switches that support port bounce with COA.

jrouzierinverse:
Description
===========
Template based bouncePort

NEWS file entries
=======================
Enhancements
-------------
* Template based bouncePort

Issue
=====
fixes #5633

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/docs/developer/template_switches.asciidoc
+++ b/docs/developer/template_switches.asciidoc
@@ -288,6 +288,61 @@ So here is mini templating language to format values.
 ** switch._wsUser
 ** vlan
 
+* bounce
+** disconnectIp
+** mac
+** ifIndex
+** switch._ExternalPortalEnforcement
+** switch._RoleMap
+** switch._SNMPAuthPasswordRead
+** switch._SNMPAuthPasswordTrap
+** switch._SNMPAuthPasswordWrite
+** switch._SNMPAuthProtocolRead
+** switch._SNMPAuthProtocolTrap
+** switch._SNMPAuthProtocolWrite
+** switch._SNMPCommunityRead
+** switch._SNMPCommunityTrap
+** switch._SNMPCommunityWrite
+** switch._SNMPEngineID
+** switch._SNMPPrivPasswordRead
+** switch._SNMPPrivPasswordTrap
+** switch._SNMPPrivPasswordWrite
+** switch._SNMPPrivProtocolRead
+** switch._SNMPPrivProtocolTrap
+** switch._SNMPPrivProtocolWrite
+** switch._SNMPUserNameRead
+** switch._SNMPUserNameTrap
+** switch._SNMPUserNameWrite
+** switch._SNMPVersion
+** switch._SNMPVersionTrap
+** switch._TenantId
+** switch._UrlMap
+** switch._VlanMap
+** switch._VoIPEnabled
+** switch._cliEnablePwd
+** switch._cliPwd
+** switch._cliTransport
+** switch._cliUser
+** switch._coaPort
+** switch._controllerIp
+** switch._deauthMethod
+** switch._disconnectPort
+** switch._id
+** switch._inlineTrigger
+** switch._ip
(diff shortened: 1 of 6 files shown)


lyubomirtraykov:
I think it will be better if we don't use the locationlog to determine the mac of the node, but to pass the mac as argument in:
html/pfappserver/lib/pfappserver/Model/Node.pm - sub restartSwitchport
lib/pf/UnifiedApi/Controller/Nodes.pm - sub do_restart_switchport
lib/pf/api.pm - sub _reassignSNMPConnections

lib/pf/Switch/Template.pm
```diff
@@ -474,17 +474,18 @@ sub bouncePort {
         return $TRUE;
     }
 
-    return $self->_bouncePortCoa($ifindex);
+    my $radiusBounce = $self->{_template}{bounce};
+    if (!defined $radiusBounce) {
```
lyubomirtraykov, line 478: Add `isenabled($self->{_template}->{snmpDisconnect})` in the if

jrouzierinverse: If I add that in the if statement it would be a logical error.
Since I am expecting radiusBounce to be defined here https://github.com/inverse-inc/packetfence/pull/5735/files#diff-e7181847796bf1ef221178bfd4822927R483

jrouzierinverse:
> I think it will be better if we don't use the locationlog to determine the mac of the node, but to pass the mac as argument in:
> html/pfappserver/lib/pfappserver/Model/Node.pm - sub restartSwitchport
> lib/pf/UnifiedApi/Controller/Nodes.pm - sub do_restart_switchport
> lib/pf/api.pm - sub _reassignSNMPConnections

That makes sense I would need to review the case of VOIP.

lib/pf/Switch/Template.pm
lyubomirtraykov: Yes, but if you enable snmpDisconnect and add bounce in the template configuration, the disconnection method will call handleReAssignVlanTrapForWiredMacAuth which uses bouncePort and then it will preform COA not SNMP.

jrouzierinverse: Thank you for the better explanation of the issue.

jrouzierinverse: To avoid the least surprise.
If you define bounce attributes we will use that when performing a bounce.
If you enable snmpDisconnect then we will use SNMP to disconnect.
It is the responsibility of the person configuring the Template Switch to choose the behavior they want.

I pushed the code to do this.

lyubomirtraykov:
@jrouzierinverse i tested passing the mac as argument and everything is working as it should. So i think we should use the argument method instead of locationlog.

nqb:
LGTM, I will push minor adjustements to documentation before I merged.

docs/developer/template_switches.asciidoc
```diff
@@ -288,6 +288,61 @@ So here is mini templating language to format values.
 ** switch._wsUser
 ** vlan
 
+* bounce
```
nqb, line 284: I was able to use `$last_accounting.acctsessionid` in `bounce` scope but it is not referenced here.

Some vendors require to use `Acct-Session-Id` attribute to identify port to bounce.

nqb (changes requested):
Last thing I noticed @jrouzierinverse during my review of Switch Template's documentation is that following syntax: `VendorName:Attribute-Name1 = value1` is not valid anymore. If you use a Vendor attribute in GUI, resulting attribute is written like any other.

extrafu:
@nqb & @jrouzierinverse what are we doing based on Nicolas' last comment?

PR #5708, opened 2020-07-19
feature/switch-template-cli-access

jrouzierinverse:
Description
===========
Add support for cli Access for Switch::Template

NEWS file entries
=======================
Enhancements
-------------
* Add support for CLI Access for Switch::Template

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests
- [ ] Add acceptance tests (TestLink)

PR #5742, opened 2020-08-10
feature/default-switch-packetfence-standard

### LINKED ISSUE #5490: RFC: use "PacketFence::Standard" in place of "Generic" for default Switch Group
**Describe the bug**
Currently, default switch group still use `Generic` as default switch (old switch module). We should move to `PacketFence::Standard` (template module). That's way, all new switches created from default switch group will use `PacketFence::Standard` (template module).

Is there is some limitations ?

I saw two functions in `Generic.pm` related to CLI access to switches (https://github.com/inverse-inc/packetfence/blob/5ba7bf9d8863837d8c81013bf443237fd8691a70/lib/pf/Switch/Generic.pm) and I wonder if we can handle those things in `PacketFence::Standard`.

From my point of view, that's the case.

jrouzierinverse:
Description
===========
Set the default switch type to Packetfence::Standard

NEWS file entries
=======================

Enhancements
-------------
* Set the default switch type to Packetfence::Standard
* Create a PacketFence::SNMP switch to force reevaluate access using SNMP

Issue
=====
fixes #5490

Delete branch after merge
=========================
NO

UPGRADE file entries
=========================

Set the switch type to Generic if it was not previously defined.

/usr/local/pf/addons/upgrade/to-10.2-default-switch-packetfence-standard.pl

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/addons/upgrade/to-10.2-default-switch-packetfence-standard.pl
+++ b/addons/upgrade/to-10.2-default-switch-packetfence-standard.pl
@@ -1,0 +1,99 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+to-10.2-default-switch-packetfence-standard.pl
+
+=cut
+
+=head1 DESCRIPTION
+
+Add Generic type where needed
+
+=cut
+
+use strict;
+use warnings;
+use lib qw(/usr/local/pf/lib);
+use pf::IniFiles;
+use pf::file_paths qw(
+    $switches_config_file
+);
+use List::MoreUtils qw(any);
+use pf::util;
+use File::Copy;
+
+run_as_pf();
+
+my $file = $switches_config_file;
+
+if (@ARGV) {
+    $file = $ARGV[0];
+}
+
+my $cs = pf::IniFiles->new(-file => $file, -allowempty => 1);
+my @groups;
+my @switches;
+
+my $default_type = $cs->val('default', 'type');
+
+if ($default_type && $default_type ne 'Generic') {
+    print "Nothing to do\n";
+    exit;
+}
+
+for my $section ($cs->Sections()) {
+    next if $section eq 'default';
+    if ($section =~ /^group /) {
+        push @groups, $section;
+    } else {
+        push @switches, $section;
+    }
+}
+
+for my $group (@groups) {
+    if (!$cs->exists($group, 'type')) {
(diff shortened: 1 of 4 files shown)


nqb (changes requested):
There is an ambiguity for `addons/upgrade/to-10.2-default-switch-packetfence-standard.pl` script.

Name of script doesn't match its description: "Rename Generic switch type to Packetfence::SNMP" and its actions: script is trying to set Generic value for switches.

jrouzierinverse:
Fixed the verbage

addons/upgrade/to-10.2-default-switch-packetfence-standard.pl
```diff
@@ -0,0 +1,99 @@
...
+
+to-10.2-default-switch-packetfence-standard.pl
+
+=cut
+
+=head1 DESCRIPTION
+
+Add Generic type where needed
+
+=cut
```
nqb, line 13: If purpose of this PR is to "Set the default switch type to Packetfence::Standard", title and description needs to be adjust.

nqb: Also description of PR (Upgrade section) need to be edited.

nqb (commented):
Should we not remove Generic switch module too ?

nqb (changes requested):
If we want to replace `Generic` switch module by `PacketFence::Standard` switch module, we should first merge #5708, rebase and add `cliAuthorizeRead` and `cliAuthorizeWrite` scopes to `PacketFence::Standard` switch module based on what we do in Generic switch module.

I can take care of this change.

jrouzierinverse:
Let's remove the Generic switch in 11.0.

jrouzierinverse:
There is no need to add cliAuthorizeRead and cliAuthorizeWrite scopes to PacketFence::Standard.
As it has the same behavior of the Generic module if none are defined.

addons/upgrade/to-10.2-default-switch-packetfence-standard.pl
nqb: Purpose of this script is to set Generic type to switch which don't have any type.

PR #5781, opened 2020-09-09
Feature/rfc 7710

### LINKED ISSUE #5638: Add support for RFC-7710bis - captive portal detection through DHCP
Trying to find a way to improve captive portal detection, as it seems to be hit-or-miss. As of two days ago, Apple has announced support ([https://developer.apple.com/news/?id=q78sq5rv](https://developer.apple.com/news/?id=q78sq5rv)) for https://tools.ietf.org/html/draft-ietf-capport-rfc7710bis-08.

This allows networks to advertise the URL for the captive portal using DHCP options, which would make detecting a captive portal almost infallible (on supported devices, of course). 

Basically, we would have to add support for DHCP option 114 (for DHCPv4) and, if IPv6 support is eventually added, the corresponding DHCP and RA (Router Advertisment) options.

Looking at the code, it seems as it should be relatively simple to add:

*) We would need to add the option to the dhcp dictionary:
`/go/dhcp/dictionary.go`

```
var Tlv = TlvList{
	Tlvlist: map[int]TlvType{
<snip>
114: TlvType{"OptionCaptivePortal", tlvNstring},
```

and then I believe the following code for the dhcp config itself ( `/go/dhcp/config.go` ) should be (almost) enough:

```
options[dhcp.OptionCaptivePortal] = []byte(ConfNet.PortalFQDN)
```

We might want to add this only to the registration network (not 100% sure) or maybe registration and isolation. For "user" networks, if PF is in in-line mode it probably shouldn't be added.

Also, not 100% sure fi the PortalFQDN would be enough and a connection to the naked domain would be caught by haproxy or do we need more details for the specific portal to redirect to. If we need the exact URL, it might need slightly more work as the Portal URL would need to be exposed to the `/go/pfconfigdriver/structs.go`

**Describe alternatives you've considered**
This would be an enhancement to the current, dns based captive portal detection mechanism

**Additional context**
RFC-7710 (the original one) is from a few years back, but it seems it was never widely adopted by OS vendors. The main difference with the current version is that the DHCP option used to convey this URL has been changed in the new proposal as option 160 (the original) was de-facto used by polycom for provisioning and thus could create problems. The new option, 114, was assigned by IANA to Apple but was never used and is being given back to IANA for assignment for RFC 7710bis

With apple pushing this, hopefully other OS vendors get onboard and thus, finally, simplify the captive portal detection mess that today's DNS logic is.

fdurand:
# Description
rfc-7710 support

# Impacts
DHCP + Portal

# Issue
fixes #5638

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Support for RFC7710-bis


go/dhcp/config.go
```diff
@@ -340,6 +355,15 @@ func (d *Interfaces) readConfig() {
 						options[dhcp.OptionDomainNameServer] = ShuffleDNS(ConfNet)
 						options[dhcp.OptionRouter] = ShuffleGateway(ConfNet)
 						options[dhcp.OptionDomainName] = []byte(ConfNet.DomainName)
+						if portal.SecureRedirect == "enabled" {
+							if ConfNet.PortalFQDN != "" {
+								portalURL := "https://" + ConfNet.PortalFQDN + "/api"
+								options[dhcp.OptionCaptivePortal] = []byte(portalURL)
+							} else {
+								portalURL := "https://" + general.Hostname + "." + general.Domain + "/api"
```
julsemaan: can we use something more meaningful than `/api` ?
Or is this path part of the spec.

Using `/rfc7710` would make it more explicit when analyzing logs

go/dhcp/config.go
```diff
@@ -340,6 +355,15 @@ func (d *Interfaces) readConfig() {
 						options[dhcp.OptionDomainNameServer] = ShuffleDNS(ConfNet)
 						options[dhcp.OptionRouter] = ShuffleGateway(ConfNet)
 						options[dhcp.OptionDomainName] = []byte(ConfNet.DomainName)
+						if portal.SecureRedirect == "enabled" {
+							if ConfNet.PortalFQDN != "" {
```
julsemaan: Could we use the code of detectPortalURL that we move in util/ to build the URL of the portal so that the logic isn't duplicated

extrafu:
@fdurand Please rebase and @julsemaan do a final a review and merge after.

fdurand:
Rebased

PR #5784, opened 2020-09-09
Support EAP-PEAP and EAP-TLS for CLI and VPN authentication

julsemaan:
# Description
Support EAP-PEAP MSCHAPv2 and EAP-TLS for CLI and VPN authentication

# Impacts
CLI and VPN authentication

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Added support for EAP-PEAP MSCHAPv2 and EAP-TLS for CLI and VPN RADIUS authentication

### DIFF
--- a/lib/pf/radius.pm
+++ b/lib/pf/radius.pm
@@ -72,6 +72,7 @@ use pf::dal;
 use pf::security_event;
 use pf::constants::security_event qw($LOST_OR_STOLEN);
 use pf::Redis;
+use pf::constants::eap_type qw($EAP_TLS $MS_EAP_AUTHENTICATION);
 
 our $VERSION = 1.03;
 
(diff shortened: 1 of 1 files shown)


lyubomirtraykov:
Please add support for EAP-MSCHAPv2 (EAP-Type = 29), because it is the only supported method for Android StrongSwan client.

julsemaan:
> Please add support for EAP-MSCHAPv2 (EAP-Type = 29), because it is the only supported method for Android StrongSwan client.

I can only add what I can test but I think this is handled by the MSCHAP-Challenge test

lib/pf/radius.pm
```diff
@@ -169,7 +170,6 @@ sub authorize {
         ifIndex => $port,
         ifDesc => $ifDesc,
         user_name => $user_name,
-        username => $user_name,
```
fdurand: We have to keep it since we use it in the radius filter.

julsemaan: my bad, this shouldn't have been part of the PR

julsemaan:
Adjusted

PR #5796, opened 2020-09-15
Add tenancy to realm configuration

satkunas:
# Description
Adds tenancy to GUI realm configuration

# Impacts
Regress search utility on gui realm configurations

# Delete branch after merge
YES 

# Checklist
- [n/a] Document the feature
- [n/a] Add unit tests

# NEWS file entries
## New Features
* Add tenancy to Realm configuration

### DIFF
--- a/html/pfappserver/root/static.alt/src/components/pfMixinSearchable.js
+++ b/html/pfappserver/root/static.alt/src/components/pfMixinSearchable.js
@@ -16,6 +16,9 @@
  *            type: Object,
  *            default: {
  *              searchApiEndpoint: 'users',
+ *              searchApiHeaders: {
+ *                foo: 'bar'
+ *              },
  *              searchApiEndpointOnly: false,
  *              defaultSortKeys: ['pid'],
  *              defaultSearchCondition: { op: 'and', values: [{ op: 'or', values: [{ field: 'pid', op: null, value: null }] }] },
@@ -47,6 +50,7 @@ export default {
     searchableOptions: {
       type: Object,
       default: {
+        searchApiHeaders: {},
         searchApiEndpointOnly: false,
         defaultSearchCondition: () => {
           return { op: 'and', values: [{ op: 'or', values: [{ field: null, op: null, value: null }] }] }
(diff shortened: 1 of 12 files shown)


julsemaan:
This needs to be rebased off devel so that the RPM build can succeed

julsemaan:
When I'm logged in in single tenant mode as the admin of tenant 1, I get the following error:
image

It seems it wants to display the realms of the global tenant even when we're in single tenant mode

The global tenant realms should not be displayed/fetched when in single tenant mode (i.e. the token accessing this section is for tenant ID 1)

julsemaan:
Also, other issue, when I create a realm and navigate back to the realm list, it doesn't reload the realms

julsemaan:
Also, other issue: I created a couple of realms for the global tenant (with valid global tenant credentials), then tried to reorder them. The GUI said the reorder worked fine but it didn't go through. When looking at the network console, the X-PacketFence-Tenant-Id wasn't sent in the sort_items call

julsemaan:
The sorting is still failing but just for the 3 built-in realms of the `default` tenant (i.e. DEFAULT, LOCAL and NULL)

Perhaps these 3 should be locked since they're part of the default config file, @fdurand and/or @jrouzierinverse can confirm

julsemaan:
There is 1 pending issue above and then its good to merge for me unless @cgx wants to take a look at the JS code

fdurand:
Sorting is important only for realm that contain regex.
Se the DEFAULT, LOCAL and NULL can stay locked.

satkunas:
@jrouzierinverse please send `not_sortable` field with all requests to lock these specific columns.

cgx:
JavaScript looks good. 👍

jrouzierinverse:
Done with mark items as not sortable.

extrafu:
@julsemaan One final review before the merge?

julsemaan:
If we assume that @jrouzierinverse's fix make it work, then its good to merge

Otherwise I need to build a package for this to validate the fix

extrafu:
I suggest we merge it and make it part of the tests to do for the v10.2 release.

PR #5802, opened 2020-09-16
Create a new entry in radacct and in radacct_log if the interim update has been received after a stop.

fdurand:
# Description
Having an interim update after a stop for the same session add wrong values in the radacct_log table.
If we detect this case then we create a new entry in radacct and in radacct_log whit 0 as value (in/out/time).

# Impacts
RADIUS accounting procedure

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Fix acct_update procedure to avoid wrong values.


satkunas (commented):
Looks good. Merging.

PR #5808, opened 2020-09-18
Allow only TLS 1.2 in Freeradius

fdurand:
# Description
Set TLS 1.2 the only TLS version available in Freeradius
(The configuration allows TLS 1.0 and/or TLS 1.1.  We STRONGLY recommned using only TLS 1.2 for security)

# Impacts
RADIUS authorize workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Define TLS 1.2 as the only version available in RADIUS

### DIFF
--- a/conf/radiusd/eap.conf.example
+++ b/conf/radiusd/eap.conf.example
@@ -268,8 +268,29 @@ eap [% key -%] {
         # the EAP keys correctly.  The fix is to upgrade
         # OpenSSL, or disable TLS 1.2 here. 
 #       disable_tlsv1_2 = no
-
+        disable_tlsv1_1 = yes
+        disable_tlsv1 = yes
         #
+
+        #  Set min / max TLS version.  Mainly for Debian
+        #  "trusty", which disables older versions of TLS, and
+        #  requires the application to manually enable them.
+        #
+        #  If you are running Debian trusty, you should set
+        #  these options, otherwise older clients will not be
+        #  able to connect.
+        #
+        #  Allowed values are "1.0", "1.1", "1.2", and "1.3".
+        #
+        #  Note that the server WILL NOT permit negotiation of
+        #  TLS 1.3.  The EAP-TLS standards for TLS 1.3 are NOT
+        #  finished.  It is therefore impossible for the server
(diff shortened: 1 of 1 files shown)


julsemaan:
Is TLS 1.2 supported in all the supplicants ?

Since we set it to our minimum, wouldn't that prevent older devices from connecting ?

fdurand:
No it's probably not but it's highly recommended to do that.
What i can do is to add an option in the Admin gui where you will be able to select the minimum/maximum version and if you want to enable/disable it.

julsemaan:
Yes admin GUI option would be good since it may be common to change it.

We also need to document what are the messages we get so that we know we need to adjust the parameter

And we also need upgrade notes on this to warn users about the change and instructions on how to make it the same as the current default if needed

conf/radiusd/eap.conf.example
```diff
@@ -268,8 +268,29 @@ eap [% key -%] {
...
+        #
+        #  Note that the server WILL NOT permit negotiation of
+        #  TLS 1.3.  The EAP-TLS standards for TLS 1.3 are NOT
+        #  finished.  It is therefore impossible for the server
+        #  to negotiate EAP-TLS correctly with TLS 1.3.
+        #
+        #  The values must be in quotes.
         #
+        tls_min_version = "1.2"
+        tls_max_version = "1.2"
```
jrouzierinverse, line 293: Should this be configurable?

julsemaan: Yes, I commented above regarding that, this should be allowed to be configured in the admin GUI and we should document the behavior of a client that doesn't support 1.2 so that we know that we need to change this value

nqb:
I think it would be nice to have this PR in 11.2

julsemaan:
> I think it would be nice to have this PR in 11.2

Before we do that, we must make this configurable via the admin interface instead of through custom configuration in conf/radiusd

There will be scenarios where this will bite and we must be able to change it easily

nqb:
Closing in favor of #6773

Issue #5804, opened 2020-09-17
PF 10.1 - Does not remove nightly backups (/root/backup)


bbs2web:
We had a disc space utilisation alert originate from our Packet Fence 10.1 host where we had files present in the /root/backup directory for the last 3 weeks, even since upgrading PF to 9 to 10.1.

```[root@packetfence /]# cd /root/backup/
[root@packetfence backup]# dir
total 20472908
-rw-r--r-- 1 root root   14132322 Sep 10 00:34 packetfence-db-dump-2020-09-10_00h34.sql.gz
-rw-r--r-- 1 root root   16769808 Sep 11 00:36 packetfence-db-dump-2020-09-11_00h36.sql.gz
-rw-r--r-- 1 root root   18002633 Sep 12 00:33 packetfence-db-dump-2020-09-12_00h33.sql.gz
-rw-r--r-- 1 root root   15732345 Sep 13 00:32 packetfence-db-dump-2020-09-13_00h32.sql.gz
-rw-r--r-- 1 root root   15782306 Sep 14 00:32 packetfence-db-dump-2020-09-14_00h32.sql.gz
-rw-r--r-- 1 root root   20074808 Sep 15 00:35 packetfence-db-dump-2020-09-15_00h35.sql.gz
-rw-r--r-- 1 root root   20416129 Sep 16 00:36 packetfence-db-dump-2020-09-16_00h35.sql.gz
-rw-r--r-- 1 root root   21951644 Sep 17 00:36 packetfence-db-dump-2020-09-17_00h36.sql.gz
-rw-r--r-- 1 root root 1029242448 Aug 31 00:32 packetfence-files-dump-2020-08-31_00h30.tgz
-rw-r--r-- 1 root root 1035039251 Sep  1 00:32 packetfence-files-dump-2020-09-01_00h30.tgz
-rw-r--r-- 1 root root 1059387922 Sep  2 00:33 packetfence-files-dump-2020-09-02_00h30.tgz
-rw-r--r-- 1 root root 1066001908 Sep  3 00:32 packetfence-files-dump-2020-09-03_00h30.tgz
-rw-r--r-- 1 root root 1055473175 Sep  4 00:33 packetfence-files-dump-2020-09-04_00h30.tgz
-rw-r--r-- 1 root root 1108145289 Sep  5 00:34 packetfence-files-dump-2020-09-05_00h30.tgz
[...]

PR #5811, opened 2020-09-19
Issue 5804 - Tar command has options in wrong order

bbs2web:
The tar backup command has the options in the wrong order, so the exclude options are subsequently ignored. I additionally recommend that old backups still be removed when some files change whilst being read.


nqb:
I will see how we can have this change compatible between Debian 9 and CentOS 7 because they don't use same `tar` version and issue #5804 only affects Debian 9.

bbs2web:
I believe the change may not affect CentOS 7 as this only receives back ported security patches. I did however confirm that the following structure also still works on tar 1.26 as part of CentOS 7.8:
  tar -czf test.gz --exclude a/b/* a;

bbs2web:
I realise my feedback on this was ambiguous. Tar ignoring the 'exclude' options, when they come after the source directory, does not appear to affect CentOS but is referenced in posts from almost 12 years ago. I presume RedHat patch tar, due to it catching so many people out.

ie:  Moving the source folder selection to the end of the command, after the 'exclude' options, worked when I tested this on CentOS 7 and Debian 9.


Create some files:
```
[root@packetfence ~]# for a in a b c d; do mkdir -p $a/$a; touch $a/file $a/$a/file; done;
[root@packetfence ~]# find a b c d -type f;
a/a/file
a/file
b/b/file
b/file
c/c/file
c/file
d/d/file
d/file
```

Then backup with selected sources after 'exclude' options (ie 'exclude' being positional) on CentOS 7 and Debian 9:
```
[root@centos7 ~]# tar -czvf test.tgz --exclude='a/a/*' --exclude='b/b/*' a/* b/* c/* d/*;
a/a/
a/file
b/b/
b/file
c/c/
c/c/file
c/file
d/d/
d/d/file
d/file

=================================================================================

[root@debian9 ~]# tar -czvf test.tgz --exclude='a/a/*' --exclude='b/b/*' a/* b/* c/* d/*;
a/a/
a/file
b/b/
b/file
c/c/
c/c/file
c/file
d/d/
d/d/file
d/file
```

If we however move the source directories before the 'exclude' options, as they currently are in the backup script, it does not skip those files on Debian 9:
```
[root@centos7 ~]# tar -czvf test.tgz a/* b/* c/* d/* --exclude='a/a/*' --exclude='b/b/*';
a/a/
a/file
b/b/
b/file
c/c/
c/c/file
c/file
d/d/
d/d/file
d/file

=================================================================================

[root@debian9 ~]# tar -czvf test.tgz a/* b/* c/* d/* --exclude='a/a/*' --exclude='b/b/*';
a/a/
a/a/file
a/file
b/b/
b/b/file
b/file
c/c/
c/c/file
c/file
d/d/
d/d/file
d/file
```

Sample post from almost 12 years ago:
https://stackoverflow.com/questions/984204/shell-command-to-tar-directory-excluding-certain-files-folders

GNU tar 1.32 documentation specifically lists position sensitive options, which includes 'exclude':
https://www.gnu.org/software/tar/manual/html_node/Position_002dSensitive-Options.html

bbs2web:
Herewith a test on RHEL3 - working as expected:
```
[root@rhel3 root]# for a in a b c d; do mkdir -p $a/$a; touch $a/file $a/$a/file; done;
[root@rhel3 root]# find a b c d -type f;
a/a/file
a/file
b/b/file
b/file
c/c/file
c/file
d/d/file
d/file
[root@rhel3 root]# tar -czvf test.tgz --exclude='a/a/*' --exclude='b/b/*' a/* b/* c/* d/*;
a/a/
a/file
b/b/
b/file
c/c/
c/c/file
c/file
d/d/
d/d/file
d/file
```

julsemaan:
IMO, this should get in, @nqb, is this good for us to merge ?

nqb:
I would like to test on my side.

nqb:
Hello @bbs2web,

Regarding `--exclude` option, it works as it should on CentOS 7 and Debian 9 as you reported.

Regarding change related to cleanup of old archives, I prefered to keep existing backups on disk if `tar` exits with a code different than 0. That's way you're sure to have at least valid backups if tar exits with a code different than 0.
On top of that, if we change it's behavior, it also mean we will not detect same issue you reported. 

Also PR need to be rebased due to a new exclusion added by @lzammit in a98c329487ba0d0e1d64c2b3deb7634c5a698ac0

julsemaan:
@fdurand will take a look and merge today, otherwise it will be for the next release

fdurand:
I am agree with @nqb , we should revert the test on the exist code since the --exclude has been fixed.
@bbs2web can you revert that ? and we will be good to merge.

bbs2web:
Apologies that this took so long...

fdurand:
Thanks for you contribution.

PR #5892, opened 2020-10-07
local account creation expiration should follow the access validity

julsemaan:
# Description
Tie the expiration of the local account to the duration of the access.
Example: you are given access for 2 months, your local account is valid for 2 months
At the moment, all local accounts are only valid 31 days

# Impacts
Create local account on the captive portal

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Validity of the local accounts created on the portal is tied to the access duration of the user.


julsemaan:
:partying_face:

PR #5908, opened 2020-10-09
Venom: check the version and status before installation

JeGoi:
# Description
Test version of venom before install it.
Install the latest version expect if the current version is higher than the latest online

# Impacts
Just update the way venom is installed by checking the version
Not really big impacts

# Delete branch after merge
YES

### DIFF
--- a/t/venom/install-venom.sh
+++ b/t/venom/install-venom.sh
@@ -69,8 +69,7 @@ if [ ! type curl 2> /dev/null ] ; then
     echo "Install curl before running this script"
     exit 0 
 else
-    #VENOM_LATEST_VERSION=$(curl -s ${VENOM_REPO_URL} | grep 'tag_name.*' | cut -d '"' -f 4 | grep -oP '\d.*')
-    VENOM_LATEST_VERSION="0.28.0"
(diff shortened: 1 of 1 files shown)


nqb (changes requested):
I understand what the script did but I don't see a real benefit.

I will make more sense from my point of view to be able to install a specific Venom version like we did with [Venom role](https://github.com/inverse-inc/ansible-utils/blob/ab833a89114f62cce5b44b625f95d1f42982536d/roles/venom/defaults/main.yml). In fact, I wrote this script to mimic behavior of Venom role to enforce a Venom version (only the latest at the moment). So be able to install any Venom version will be a nice improvement.

The `venom update` already allow you to upgrade your venom installation to latest.

nqb:
I also want to mention that v0.28.0 has an [issue](https://github.com/ovh/venom/issues/239) with `venom update`

JeGoi:
> I also want to mention that v0.28.0 has an [issue](https://github.com/ovh/venom/issues/239) with `venom update`

So this script fix it. LOL
Since it is already in ansible, should we keep it?

JeGoi:
Closed in favor of https://github.com/inverse-inc/packetfence/pull/5931

PR #5921, opened 2020-10-16
Bash Script to get latest cd/ci files for public directory

JeGoi:
# Description
Bash script to get latest cd/ci files for public directory
It parse the latest 5 pages from gitlab ci, get the latest public archives file and install it in packetfence/public directory.

# Impacts
Can be used with venom automation

# Dependencies
wget and curl

# Delete branch after merge
YES please

### DIFF
--- a/t/venom/get_public_from_gitlab.sh
+++ b/t/venom/get_public_from_gitlab.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+set -o nounset -o pipefail
+
+REPO_URL=https://gitlab.com/inverse-inc/packetfence/-/jobs?page=
+MAXPAGE=5
+ZIPFILE="./public.zip"
+JOBID=""
+
+if [ ! type curl 2> /dev/null ] ; then
+  echo "Install curl before running this script"
+  exit 0 
+fi
+
+if [ ! type wget 2> /dev/null ] ; then
+  echo "Install wget before running this script"
+  exit 0 
+fi
+
+# Check accross the first ${MAXPAGE} pages
+WEXIT=0
+i=0
+while [ ${i} -le ${MAXPAGE} -a ${WEXIT} -lt 1 ]
+do
+  GITVAL=$(curl -s ${REPO_URL}${i})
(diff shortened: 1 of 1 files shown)


nqb:
Hello @JeGoi,

1. Could you create a new PR on a branch directly in inverse-inc/packetfence repo in place of your fork ?
2. In this new PR, could you put your script in `ci/lib/test/` directory ?
It makes more sense to put it here because it's much related to provision of VMs using Vagrant

nqb:
The script doesn't work on my side. It looks like `$j` is never incremented.

@JeGoi, according to [GitLab's docs](https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html#downloading-the-latest-artifacts), we can get more easily latest artifacts for a branch. On my side, I'm able to get latest artifact available for devel branch using following command:

```
curl -L https://gitlab.com/inverse-inc/packetfence/-/jobs/artifacts/devel/download?job=pages --output /tmp/public.zip
```

I really prefer this approach.

I'm sure that parsing resulting HTML from gitlab.com will not survive to GitLab upgrades.

JeGoi:
Hum, if the j is not incremented, it is because the first job id (published and available) has been found in the page. So no need to search in the rest of the page. It will increment if a job (first or not) is publish and canceled. It will search for the first and latest available.

For the latest artifact, the link will be a problem if the latest artifacts available is not the publish one. You will have the wrong one.
Example, if the pipeline failed at the "sign" step, the latest available artifact will be the "build package" artifacts, which is not the one we are looking for, correct?
I have seen that problem only once when a pipeline has been cancelled.

JeGoi:
Ok no problem, I will move it and publish it in a branch of packfence repository.
Thanks for your time

nqb:
> For the latest artifact, the link will be a problem if the latest artifacts available is not the publish one.

I'm surprised by this behavior, I expected to have the latest artifact available for the job pages. If the latest pipeline has failed, GitLab.com should not provide files to download.

JeGoi:
It was on the third page, at the time, I've script it.
And, it looks that only the first 4 pages have artifacts (Gitlab will not keep it for ever...).

nqb:
Test to cancel "pages" job on: https://gitlab.com/inverse-inc/packetfence/-/pipelines/205188863

JeGoi:
@nqb Ok it is working perfectly with the link you have provided.
I will close the pull request.

PR #5932, opened 2020-10-20
[Venom] Feature/venom public

JeGoi:
# Description
Bash script to get latest cd/ci artifacts files for pages
Change the Makefile
Source: https://github.com/inverse-inc/packetfence/pull/5921

# Impacts
Could be used with venom automation to prepare the public file shared with vm for local tests

# Dependencies
curl

# Delete branch after merge
YES

### DIFF
--- a/ci/lib/test/get_public_from_gitlab.sh
+++ b/ci/lib/test/get_public_from_gitlab.sh
@@ -5,7 +5,7 @@ REPO_URL=https://gitlab.com/inverse-inc/packetfence/-/jobs/artifacts/devel/downl
 ZIP_FILE="/tmp/public.zip"
 DIR=$(pwd | grep -oP '.*?(?=ci/lib/test)')
 
-if [ ! type curl 2> /dev/null ] ; then
+if ! type curl 2> /dev/null ; then
   echo "Install curl before running this script"
   exit 1 
 fi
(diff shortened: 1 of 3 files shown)


ci/lib/test/get_public_from_gitlab.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -o nounset -o pipefail
+
+REPO_URL=https://gitlab.com/inverse-inc/packetfence/-/jobs/artifacts/devel/download?job=pages
+ZIP_FILE="/tmp/public.zip"
+DIR=$(pwd | grep -oP '.*?(?=ci/lib/test)')
+
+if [ ! type curl 2> /dev/null ] ; then
```
nqb:
```suggestion
if ! type curl 2> /dev/null ; then
```

ci/lib/test/get_public_from_gitlab.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -o nounset -o pipefail
+
+REPO_URL=https://gitlab.com/inverse-inc/packetfence/-/jobs/artifacts/devel/download?job=pages
+ZIP_FILE="/tmp/public.zip"
+DIR=$(pwd | grep -oP '.*?(?=ci/lib/test)')
```
nqb:
```suggestion
ZIP_FILE=$(mktemp --suff=".zip")

# get root of git directory
DIR=$(pwd | grep -oP '.*?(?=ci/lib/test)')
```

nqb (changes requested):
1. Could you add a check in `ci/lib/test/get_public_from_gitlab.sh` to verify that `public` directory exists ? If that is the case, there is no need to download latest artifact because Ansible will try to install latest package available.

1. Could you add a `public` directory to  `.gitignore` ?

PR #5936, opened 2020-10-26
[Docs] Reorganize Documentation content

JeGoi:
# Description
This branch reorganize doc content in trees and leaf.
Each major file (cluster, install, network, upgrade) has now a dedicated directory with content.
Each major file is divided in sub-file according to the title level2
If a sub-file is/could be divided in something more readable, a new directory is created with the content in separated files.
Images are in same place as before. Only contents have changed.

Other content modifications:
- Clean all titles and subtitles with the = style
- Make almost all navigation in italic (it could be changed later)
- Try to make all buttons with ` (back quote - probably not exhaustive)
- Try to use  * bold when it is needed (and remove italic or simple/double quotation marks)

This branch has merged modifications and discussions from PR #5873 
(PR #5873 could be closed when this one is merged)

# Impacts
PRO:
- Reduce the size of each files => More readable by human
- Reorganize easily
- Create network-device will be more easily done etc.

CONS:
- a split is some time not that easy to compile if too much split => could be transform in another maze

# Delete branch after merge
YES

# Checklist
- [x] Other PR not yet merged

PR #5968, opened 2020-11-06
Replace xsltproc/fop by asciidoctor-pdf

cgx:
# Description
Use [asciidoctor-pdf](https://github.com/asciidoctor/asciidoctor-pdf) to generate the guides in pdf format.

# Impacts

* No more dependencies on XSL/docbook/fop/Java. The _theme_ is also easier to [maintain](https://github.com/asciidoctor/asciidoctor-pdf/blob/master/docs/theming-guide.adoc).
* I also move the generated HTML documents to the root of the `docs` directory. It simplifies the relative URLs of the images.

# Limitations
The title page had to be simplified because of the current limitations of `asciidoctor-pdf`. It is now a combination of a _cover page_ with the PacketFence logo and a _title page_ with the guide name and PacketFence version. I think the result is totally acceptable.

# New Package(s) Required
We need `asciidoctor-pdf`. It can be packaged, or simply installed using `gem`. 

# Generated PDF Example

[PacketFence_Clustering_Guide.pdf](https://github.com/inverse-inc/packetfence/files/5502352/PacketFence_Clustering_Guide.pdf)


nqb (changes requested):
When #5936 is merged, we will need to remove TOC from PDF generation for NEWS file.

EDIT: not relevant anymore because NEWS.asciidoc is not part of PDF build

rpm/packetfence.spec
```diff
@@ -1131,7 +1116,6 @@ fi
 %exclude                /usr/local/pf/docs/README.asciidoc
 %if %{builddoc} == 1
 %doc                    /usr/local/pf/docs/*.pdf
-%exclude                /usr/local/pf/docs/*.fo
 %endif
```
nqb: We can certainly remove this part.

Makefile
```diff
@@ -64,42 +30,39 @@ HTML = $(patsubst %.asciidoc,docs/html/%.html, $(ASCIIDOCS))
 
 docs/html/%.html: docs/%.asciidoc
 	asciidoctor \
-		-D docs/html \
 		-n \
 		-r ./docs/html/asciidoctor-html.rb \
```
nqb:
```suggestion
		-r ./docs/asciidoctor-html.rb \
```

Makefile
```diff
@@ -64,42 +30,39 @@ HTML = $(patsubst %.asciidoc,docs/html/%.html, $(ASCIIDOCS))
...
 
 html/pfappserver/root/static/doc:
 	make html
-	mkdir html/pfappserver/root/static/doc
-	mkdir html/pfappserver/root/static/images
-	cp -a docs/html/* html/pfappserver/root/static/doc
-	cp -a docs/images/* html/pfappserver/root/static/images
+	mkdir -p html/pfappserver/root/static/doc/images
+	cp -a docs/*.html html/pfappserver/root/static/doc
+	cp -a docs/images/* html/pfappserver/root/static/doc/images
```
nqb: I think this target is not used anymore

docs/README.asciidoc
```diff
@@ -1,14 +1,11 @@
-Documentation README
-====================
+= Documentation README
```
nqb, line 1: Instructions here are not really true because now we need at least asciidoctor > 2 in order to build HTML

rpm/packetfence.spec
```diff
@@ -29,7 +29,7 @@ BuildRequires: gettext, httpd, ipset-devel, pkgconfig, jq
 BuildRequires: libmnl-devel,
 %endif
 %if 0%{?rhel} == 7
-BuildRequires: asciidoc >= 8.6.2, fop, libxslt, docbook-style-xsl, xalan-j2
+BuildRequires: asciidoc >= 8.6.2
```
nqb: This requirement is not true, we need to have asciidoctor > 2 in order to build at least HTML documentation.

docs/includes/global-attributes.asciidoc
```diff
@@ -10,13 +10,24 @@
     License: GFDL 1.2 or later. http://www.gnu.org/licenses/fdl.html
 
 ////
-
-// TODO have the build system take care of this
-
 :release_version: 10.2.0
```
nqb, line 13: Perhaps we can set this value to `devel` and override (in another PR) this value when we released using tag value (will be v11.0.0)

docs/includes/global-attributes.asciidoc
```diff
@@ -10,13 +10,24 @@
...
+:release_month: October
```
nqb, line 14: If possible, I will prefer to not use this attribute in `docinfo.asciidoc` because we will have to compute or update this field each time we release.

Could we rely on another [environment attribute](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#attributes-and-substitutions) like `docdate` ?

docs/includes/global-attributes.asciidoc
```diff
@@ -10,13 +10,24 @@
...
+:encoding: UTF-8
+:lang: en
+:doctype: book
+PacketFence v{release_version}
```
nqb, line 18: I'm not sure it's possible to add text directly here. I know it works but it seems it can cause unexpected behaviors.

On Network guide, it hide all other attributes declared after text.

I suggest to add this text directly after General title (in main AsciiDoc file) with `:release_version:`attribute declared **before** if we want to keep same behavior or include `global-attributes.asciidoc` before text.

nqb (changes requested):

docs/includes/global-attributes.asciidoc
nqb: Second solution doesn't work.

UPGRADE.asciidoc
```diff
@@ -1,21 +1,17 @@
...
-////
+    See docs/includes/global-attributes.asciidoc for authors, copyright and
+    license information.
 
+////
+// Fix includes on GitHub
+:extra_path:
+ifdef::env-github[]
+:extra_path: docs/
+endif::[]
```
nqb: If think we can remove that part and all `github[]` references in files under docs/*/*.asciidoc

nqb (changes requested):

docs/includes/global-attributes.asciidoc
cgx: What are the _unexpected behaviors_?

nqb: On Network Guide, due to a blank line after main title, reading of `docs/includes/global-attributes.asciidoc` was stopped after `PacketFence v{release_version}`. It means that `images` directory was unset and images were not found.

cgx: How about fixing the Network Guide instead?

nqb: Let's go with this solution. It's already done.

rpm/packetfence.spec
cgx: You mean removing the `%{builddoc}` variable from the entire spec file?

Makefile
cgx: This was to make the documentation accessible from the Web admin interface. Can someone confirm we don't need this target anymore?

UPGRADE.asciidoc
cgx: It's still relevant.

rpm/packetfence.spec
nqb: Yes. Unless I'm missing something, we don't build anymore documentation using `builddoc` variable.

Makefile
nqb: At the moment, this target is not used anymore in RPM and Debian packaging. We us following things:
- CentOS: https://github.com/inverse-inc/packetfence/blob/075d5ac98c8f644cb7ff6b0b67ef7edc9bbcf777/rpm/packetfence.spec#L510-L517
- Debian: https://github.com/inverse-inc/packetfence/blob/075d5ac98c8f644cb7ff6b0b67ef7edc9bbcf777/debian/rules#L170-L178

PR #5673, opened 2020-07-08
Update PacketFence_Installation_Guide.asciidoc

rcmcdonald91:
Description
REF: #5670 

Fixes documentation to explicitly recommend rebooting after updating base OS prior to installing kernel development package and then PacketFence

Also warns about disabling secure boot when running UEFI mode.

# Impacts
(REQUIRED)
Documentation

# Issue
Fixes #5670 

# Delete branch after merge
(REQUIRED)
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ n/a ] Document the feature
- [ n/a] Add unit tests
- [ n/a] Add acceptance tests (TestLink)

PR #5996, opened 2020-11-23
Docs/fix git links

JeGoi:
# Description
- Add links for github.
- Fix references to github project
-  Add UEFI secureboot in documentation https://github.com/inverse-inc/packetfence/pull/5673

# Impacts
Make the documentation more readable in github.

# Delete branch after merge
YES


nqb:
Hello,

From my point of view, adding all this code to have a correct display on GitHub is not a good idea:
- it's specific to GitHub: if we use another forge at a time, we don't need it anymore
- it adds same piece of code in each AsciiDoc file, we don't meet [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and I'm sure it will be missed each time someone add a new AsciiDoc file.

As I already mentioned, I'm much in a favor of having a public devel version of your documentation, build using same method as release documentation.

I'm not sure there is a lot of people reading PacketFence devel documentation on GitHub, perhaps I'm wrong. On my side, I build HTML doc directly on my computer.

nqb:
@JeGoi will close this PR and open a new one with only changes related to content.

PR #6011, opened 2020-12-02
Added static routes feature

fdurand:
# Description
Add a way to define additional static routes.

# Impacts
Network config

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Static routes management via admin gui

### DIFF
--- a/html/pfappserver/root/static.alt/src/views/Configuration/_config/network.js
+++ b/html/pfappserver/root/static.alt/src/views/Configuration/_config/network.js
@@ -144,7 +144,7 @@ export const view = (form = {}, meta = {}) => {
         },
         {
           label: i18n.t('Static routes'),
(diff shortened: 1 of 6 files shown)


html/pfappserver/root/static.alt/src/views/Configuration/_config/network.js
```diff
@@ -140,6 +141,22 @@ export const view = (form = {}, meta = {}) => {
               attrs: attributesFromMeta(meta, 'interfaceSNAT')
             }
           ]
+        },
+        {
+          label: i18n.t('Static routes'),
+          text: i18n.t('Add custom static toutes managed by keepalived, one line per static route.'),
```
julsemaan: Although its obvious to me that the format should be the same one as in the route command, saying it and giving an example would be good since this is a text field. That will help users understand what they need to input

Issue #5977, opened 2020-11-12
Audit log: profile has an empty value when doing Ethernet/Wireless-NoEAP


nqb:
**Describe the bug**
When PacketFence received a RADIUS request with Ethernet/Wireless-NoEAP, even if request hits a specific connection profile, this information is not saved in DB. In place we have a `N/A`.

Behavior is the same if your device is registered or unregistered.

**To Reproduce**
1. Create a Mac Auth RADIUS request
2. Check "Profile" value on RADIUS Audit log menu
3. Check `packetfence.log` for this RADIUS request:

=> You can find a `instantiate profile NAME_OF_PROFILE` log

**Expected behavior**
Profile should be save in `radius_audit_log` table for such requests.

**Additional context**
I put FreeRADIUS in debug and I noticed that PacketFence doesn't return a `PacketFence-Profile` RADIUS attribute to FreeRADIUS. Consequently, FreeRADUS save a `N/A` in DB according to [queries.conf](https://github.com/inverse-inc/packetfence/blob/7d97560ee029c8264fba19a5f625ebb4c65ced77/raddb/mods-config/sql/main/mysql/queries.conf#L422)

```
(81) Wed Nov 11 10:11:21 2020: Debug: rest: Processing response header
(81) Wed Nov 11 10:11:21 2020: Debug: rest:   Status : 100 (Continue)
(81) Wed Nov 11 10:11:21 2020: Debug: rest: Continuing...
(81) Wed Nov 11 10:11:21 2020: Debug: rest: Processing response header
(81) Wed Nov 11 10:11:21 2020: Debug: rest:   Status : 200 (OK)
(81) Wed Nov 11 10:11:21 2020: Debug: rest:   Type   : json (application/json)
(81) Wed Nov 11 10:11:21 2020: Debug: rest: Parsing attribute "control:PacketFence-IfIndex"
(81) Wed Nov 11 10:11:21 2020: Debug: rest: EXPAND 8
(81) Wed Nov 11 10:11:21 2020: Debug: rest:    --> 8
(81) Wed Nov 11 10:11:21 2020: Debug: rest: PacketFence-IfIndex := "8"
(81) Wed Nov 11 10:11:21 2020: Debug: rest: Parsing attribute "control:PacketFence-Mac"
(81) Wed Nov 11 10:11:21 2020: Debug: rest: EXPAND 00:03:00:11:11:01
[...]

PR #6030, opened 2020-12-09
Display connection profile matched by a CLI login

nqb:
# Description
When you use CLI login, a connection profile is instantiated using a fake MAC which doesn't exist in PacketFence DB.
Current logic doesn't write a line in logs with connection profile matched in that's case which make troubleshooting difficult.

This PR change this behavior to display a line in a log with connection profile matched when we received a RADIUS request for CLI login.

# Impacts
Connection profile instantiation.

# Delete branch after merge
YES

# Checklist
- [X] Add unit tests

## Enhancements
* Log connection profile matched when using CLI login


nqb:
I also want to mention that we don't use same logic as for `authorize` regarding audit log. So I was not able to save in DB the profile matched, like I did on #5977

nqb:
Pipeline failed, I need to review my code.

jrouzierinverse (changes requested):
I got ```"$FAKE_MAC" is not exported by the pf::constants module``` when I tried to test this.
Add the ```$FAKE_MAC``` in the ```@EXPORT``` array in ```pf::constants```

nqb:
Done.

jrouzierinverse:
Can you retry to see if the pipeline is working now?

nqb:
Yes sure.

nqb:
It works this time.

PR #6031, opened 2020-12-09
feature/syslog-cef

### LINKED ISSUE #5927: Forward logs stored in DB to a central system
When you want to audit some events on a PacketFence server, you will find useful to browse content of `auth_log` table or `admin_api_audit_log` for a long amount of time. Some customers are interested to extract content from these tables to forward then to a central location like a Syslog server.

Be able to add a second backend storage for `_log` tables to facilitate their export.

jrouzierinverse:
Description
===========
Forward logs stored in DB to a central system.

NEWS file entries
=======================
New Features
-------------
* Forward logs stored in DB to a central system.

Issue
=====
fixes #5927

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests


src/mariadb_udf/syslog.c
```diff
@@ -0,0 +1,173 @@
+#include <stdlib.h>
```
extrafu, line 19: Need to add our (c) notice.

src/mariadb_udf/syslog.h
```diff
@@ -0,0 +1,27 @@
+#ifndef __PF_SYSLOG__
```
extrafu, line 19: Need to add our (c) notice.

src/mariadb_udf/test_pf_udf.c
```diff
@@ -0,0 +1,86 @@
+#include <stdio.h>
```
extrafu, line 19: Need to add our (c) notice.

db/pf-schema-X.Y.Z.sql
```diff
@@ -1299,10 +1299,50 @@ CREATE TABLE auth_log (
   KEY  attempted_at (attempted_at)
 ) ENGINE=InnoDB;
 
+DELIMITER //
+CREATE TRIGGER `log_event_auth_log_insert` AFTER INSERT ON `auth_log`
```
extrafu: How about:

- RADIUS audit log?
- DNS audit log?
- DHCP option 82?

julsemaan:
I'm far from being an expert in C so I won't be able to comment much on the code but just to make sure, did you take the necessary steps to ensure this doesn't leak memory (like running it through valgrind or something similar).

Last thing we want is this leaking memory in our DB process

extrafu:
> I'm far from being an expert in C so I won't be able to comment much on the code but just to make sure, did you take the necessary steps to ensure this doesn't leak memory (like running it through valgrind or something similar).
> 
> Last thing we want is this leaking memory in our DB process

I'll review the C code.

src/mariadb_udf/config.c
```diff
@@ -0,0 +1,308 @@
...
+
+int facility_lookup(char* name, ssize_t len)
+{
+    return _lookup(name, len, FACILITY_LOOKUP, sizeof(FACILITY_LOOKUP) / sizeof(FACILITY_LOOKUP[0]));
+}
+
+#define get_val(f, v, l)               \
+    do {                               \
+        char* tmp;                     \
+        char* field = strstr(line, f); \
```
extrafu: strstr is usually unsafe. strnstr should be used

src/mariadb_udf/config.c
```diff
@@ -0,0 +1,308 @@
...
+int loadconfig(char* path, int* count, struct configuration** out)
+{
+    FILE* fp;
+    char host[256];
+    char port[256];
+    char* line = NULL, *end;
+    size_t len;
+    ssize_t read;
+    struct configuration* configurations = NULL;
+    ;
```
extrafu: remove

src/mariadb_udf/config.c
```diff
@@ -0,0 +1,308 @@
...
+
+        while (value < end) {
+            char* tmp = memchr(value, ',', end - value);
+            if (tmp == NULL) {
+                tmp = end;
+            }
+
+            val_len = tmp - value;
+            conf_len++;
+            configurations = realloc(configurations, sizeof(struct configuration) * conf_len);
```
extrafu, line 218: Ultimately, that memory is never released (on the last iteration of the loop). If it's fine, because it's executed only once by mysqld, at least mention it.

src/mariadb_udf/config.c
```diff
@@ -0,0 +1,308 @@
...
+        get_val("port=", value, val_len);
+        if (value == NULL) {
+            cleanup_syslog_header(&conf.syslog_header);
+            cleanup_cef_header(&conf.cef_header);
+            continue;
+        }
+
+        strncpy(port, value, val_len);
+        port[val_len] = 0;
+        rc = get_udp_socket(host, port, &conf.syslog_fd, (struct sockaddr_storage*)&conf.saddr, &conf.saddr_len);
```
extrafu, line 195: close() is never called on conf.syslog_fd

db/pf-schema-X.Y.Z.sql
jrouzierinverse: The radius audit log has lots of fields. Which one should be logged?

src/mariadb_udf/config.c
jrouzierinverse: strnstr is not in the libc in Linux.
I used the equivalent function memmem.

db/pf-schema-X.Y.Z.sql
extrafu: Propose something

src/mariadb_udf/config.c
jrouzierinverse: Since there is no cleanup routine I never close the socket.
Also If I close it that also means I would need to open it back up.
Since it is an unconnected UDP socket I could just leave it open without any have to call socket() again.
Also having multiple threads share the UDP socket is safe since sending is "atomic".

extrafu: Ok then, so document it in the code please.

extrafu:
@nqb Please test this.

extrafu:
@satkunas Do a code review also.

nqb (changes requested):
build failed:
- CentOS: need a rebase
- Debian: `src/mariadb_udf/pf_udf.c:19:19: fatal error: mysql.h: No such file or directory`, see https://gitlab.com/inverse-inc/packetfence/-/jobs/929461568#L345

jrouzierinverse:
On Debian make sure libmariadbd-dev is installed.

jrouzierinverse:
I did a rebase

nqb:
> On Debian make sure libmariadbd-dev is installed.

IMHO, this package needs to be defined as a build depency (not as Pre-Depends) because we need to have it in order to build PacketFence.
That's certainly the same for CentOS 7 but I didn't see new package dependency in `rpm/packetfence.spec`.

Once everything is ok, I will rebuild our Docker images to refresh build dependencies and be able to build that code.

nqb:
Error: ` Error: No Package found for MariaDB-devel >= 10.1`

This package doesn't seems to be present in:
- EPEL
- CentOS
- packetfence-devel

repos.

jrouzierinverse:
You can find the rpm here.

https://archive.mariadb.org//mariadb-10.1.33/yum/centos7-amd64/rpms/MariaDB-10.1.31-centos73-x86_64-devel.rpm

nqb:
Thanks ! I put MariaDB-10.1.21-centos7-x86_64-devel.rpm in our devel repos.

debian/rules
```diff
@@ -128,6 +128,10 @@ install: build
 	install -d $(CURDIR)/debian/packetfence-remote-arp-sensor$(PREFIX)/$(NAME)/sbin
 	install -m0755 addons/pfarp_remote/sbin/pfarp_remote $(CURDIR)/debian/packetfence-remote-arp-sensor$(PREFIX)/$(NAME)/sbin
 	install -m0600 addons/pfarp_remote/conf/pfarp_remote.conf $(CURDIR)/debian/packetfence-remote-arp-sensor$(PREFIX)/$(NAME)/conf
+	#PacketFence Mariadb plugin
+	make src/mariadb_udf/pf_udf.so
+	install -d -m0755 $(CURDIR)/debian/packetfence/$$(pkg-config mariadb --variable=plugindir)/
+	mv src/mariadb_udf/pf_udf.so $(CURDIR)/debian/packetfence/$$(pkg-config mariadb --variable=plugindir)/
```
nqb: use `install` instead of `mv` to set rights at same time.

lib/pf/cef.pm
```diff
@@ -0,0 +1,101 @@
...
+=cut
+
+use strict;
+use warnings;
+use Moo;
+use pf::version;
+
+#CEF:Version|Device Vendor|Device Product|Device Version|Device Event Class ID|Name|Severity|[Extension]
+
+has verison => ( default => 0, is => 'ro');
```
nqb:
```suggestion
has version => ( default => 0, is => 'ro');
```

lib/pf/file_paths.pm
```diff
@@ -288,6 +292,7 @@ $admin_roles_config_file = catfile($conf_dir, "adminroles.conf");
 
 $security_events_config_file       = catfile($conf_dir, "security_events.conf");
 $authentication_config_file   = catfile($conf_dir, "authentication.conf");
+$event_loggers_config_file = catfile($conf_dir, "event_loggers.conf"); # TODO: Adjust to /floating_devices.conf when $floating_devices_file will be deprecated
```
nqb:
```suggestion
$event_loggers_config_file = catfile($conf_dir, "event_loggers.conf");
```

nqb:
Build passed on CentOS 7 and Debian 9, packages updated in -branches repositories.

Starting to test on Debian.

nqb (changes requested):

nqb:
I ran a test step in pipeline: https://gitlab.com/inverse-inc/packetfence/-/pipelines/234104048

debian/rules
nqb: @jrouzierinverse: permissions are not identical between Debian and CentOS for this file.

nqb (changes requested):
Unit tests are failing: https://gitlab.com/inverse-inc/packetfence/-/jobs/941879378#L481

nqb (commented):
IMHO, we should add C files into your maintenance workflow (with a compilation step).

jrouzierinverse:
Upgrading the UDF module require mysql root access or restarting mysql.

nqb (changes requested):
### SOLVED: Creation of UDF

I didn't see any entry in `mysql.func` table which seems to be the behavior when using an UDF. In current schema, we don't use `CREATE FUNCTION` statement to create `pf_logger` UDF.

Seems to work:
```
MariaDB [mysql]> CREATE FUNCTION IF NOT EXISTS pf_logger RETURNS integer SONAME "pf_udf.so"
    -> ;
Query OK, 0 rows affected (0.01 sec)

MariaDB [mysql]> select * from func;
+-----------+-----+-----------+----------+
| name      | ret | dl        | type     |
+-----------+-----+-----------+----------+
| pf_logger |   2 | pf_udf.so | function |
+-----------+-----+-----------+----------+
1 row in set (0.00 sec)
```
According to MariaDB's doc, `pf_logger` is declared to returned an integer.

We need also to add this to your db/upgrade-XX file.

Important: adding a function requires INSERT privilege on `mysql` database. I don't think it's a problem here because we manage SQL schema with MariaDB `root` account.

Sources: https://mariadb.com/kb/en/create-function-udf/

### Upgrading UDF

According to [MariaDB's doc](https://mariadb.com/kb/en/create-function-udf/#upgrading-a-udf), upgrading an UDF needs to follow a specific process.

At the moment, our upgrade process is not compatible with this process because we installed new files *before* doing any action in DB.

It will require to (after release which contain UDF function):
1. run a manual action on DB ( `DROP FUNCTION IF EXISTS pf_logger`)
1. install new files (using package manager)
1. apply upgrades to schema (which should contain `CREATE FUNCTION IF NOT EXISTS`)

### Replication

I read that "Statements making use of user-defined functions are not safe for replication.".

I wonder if it's a problem in primary and replica situation, I'm sure @fdurand can answer.

nqb (changes requested):
### SOLVED: Regression test

If I tried to create a connection profile (in order to generate an entry in `admin_api_audit_log`, my connection profile is created but nothing is recorded in `admin_api_audit_log`.

In logs:

```
Jan  7 08:52:51 pfdeb9dev packetfence: pfperl-api(1109) ERROR: Database query failed with non retryable error: FUNCTION pf.pf_logger does not exist (errno: 1305) [INSERT INTO `admin_api_audit_log` ( `action`, `method`, `object_id`, `request`, `status`, `tenant_id`, `url`, `user_name`) VALUES ( ?, ?, ?, ?, ?, ?, ?, ? )]{api.v1.Config.ConnectionProfiles.create, POST, test, {"access_registration_when_registered":null,"advanced_filter":{"field":null,"op":"and","value":null,"values":[]},"always_use_redirecturl":null,"autoregister":null,"billing_tiers":[],"block_interval":{"interval":"10","unit":"m"},"default_psk_key":null,"description":"test","dot1x_recompute_role_from_portal":"enabled","dot1x_unset_on_unmatch":"disabled","dpsk":"disabled","filter":[],"filter_match_style":"any","id":"test","locale":[],"login_attempt_limit":0,"logo":null,"mac_auth_recompute_role_from_portal":"disabled","network_logoff":null,"network_logoff_popup":null,"preregistration":null,"provisioners":[],"redirecturl":null,"reuse_dot1x_credentials":null,"root_module":"default_policy","scans":[],"self_service":null,"sms_pin_retry_limit":0,"sms_request_limit":0,"sources":[],"status":"enabled","unreg_on_acct_stop":"disabled","vlan_pool_technique":"username_hash"}, 201, 1, /api/v1/config/connection_profiles, admin} (pf::dal::db_execute)
```
=> PF expects `pf_logger` to be a function in `pf` database.

Same results for `auth_log` table.

nqb:
I set up a remote rsyslog server in #6079, I will be able to continue tests next week.

nqb (changes requested):
Debian build failed since we used `install`: https://gitlab.com/inverse-inc/packetfence/-/jobs/957932711

jrouzierinverse:
Fixed

nqb:
Hello,

I want to summarize here my remarks/questions/notes on this PR. I would like to get feedback from other Inverse team members.

### Operation mode of this feature

You configure an event logger in conf/event_logger.conf to forward messages from specific namespaces (admin_api_audit_log, auth_log, etc.) using syslog as transport and CEF as format.

At the end, you're able to get data from specific DB tables on a remote syslog server for your PacketFence server.

### Configuration management

Each time you configure/edit an event logger, you need to:
- regenerate MariaDB config
- restart MariaDB

IMHO, we should try to avoid restarting DB to put new configuration in place. I'm not able to understand why we need restart MariaDB to put change in place.

### Maintenance management

We need to adapt our upgrade process for N+1 release after we merge this PR because UDF needs to be removed from DB **before** installing new files.

Also, as @jrouzierinverse mentioned, it's not possible to add UDF C files to our maintenance workflow because it's requires mysql root access or restarting mysql.

### Cluster behavior

Not tested yet, several remarks:
- I read that "Statements making use of user-defined functions are not safe for replication".
- Because data in DB is replicate across all cluster members, I wonder if each cluster member will try to send syslog messages to a remote syslog server.

### Future of UDF with next release of MariaDB

Our current MariaDB version as reach end of support, we need to be sure UDF function will work in next version.

### Remaining tasks

- [ ] Document feature
- [ ] GUI integration
- [ ] Cluster tests
- [ ] Integration tests

jrouzierinverse:
> Hello,
> 
> I want to summarize here my remarks/questions/notes on this PR. I would like to get feedback from other Inverse team members.
> 
> ### Operation mode of this feature
> You configure an event logger in conf/event_logger.conf to forward messages from specific namespaces (admin_api_audit_log, auth_log, etc.) using syslog as transport and CEF as format.
> 
> At the end, you're able to get data from specific DB tables on a remote syslog server for your PacketFence server.
> 
> ### Configuration management
> Each time you configure/edit an event logger, you need to:
> 
> * regenerate MariaDB config
> * restart MariaDB

> 
> IMHO, we should try to avoid restarting DB to put new configuration in place. I'm not able to understand why we need restart MariaDB to put change in place.

This is done for efficiency reasons.
If not I would have to test for changes for every insert/update statement.
Right now it is only done once.

> 
> ### Maintenance management
> We need to adapt our upgrade process for N+1 release after we merge this PR because UDF needs to be removed from DB **before** installing new files.
> 
> Also, as @jrouzierinverse mentioned, it's not possible to add UDF C files to our maintenance workflow because it's requires mysql root access or restarting mysql.
> 
> ### Cluster behavior
> Not tested yet, several remarks:
> 
> * I read that "Statements making use of user-defined functions are not safe for replication".
> * Because data in DB is replicate across all cluster members, I wonder if each cluster member will try to send syslog messages to a remote syslog server.
This would matter if we were using statement-based replication.
> 
> ### Future of UDF with next release of MariaDB
> Our current MariaDB version as reach end of support, we need to be sure UDF function will work in next version.
The API for UDF has not changed across versions it should not be an issue.
> 
> ### Remaining tasks
> * [ ]  Document feature
> * [ ]  GUI integration
> * [ ]  Cluster tests
> * [ ]  Integration tests

`

lib/pf/cmd/pf/generatemariadbconfig.pm
```diff
@@ -37,6 +38,13 @@ use pf::config qw(
     $management_network
     $DISTRIB
 );
+use pf::constants::eventLogger;
```
julsemaan, line 41: Are the table we log events for configurable ? Or the moment you enable the feature, it installs the triggers for all the tables ?

jrouzierinverse: It is configurable the following tables are supported admin_api_audit_log,  auth_log, dhcp_option82, dns_audit_log, radius_audit_log.

db/pf-schema-10.2.0.sql
```diff
@@ -1397,6 +1397,23 @@ CREATE TABLE `admin_api_audit_log` (
    KEY `created_at` (`created_at`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=COMPRESSED;
 
+DELIMITER $$
+CREATE TRIGGER `log_event_admin_api_audit_log` AFTER INSERT ON `admin_api_audit_log`
```
julsemaan: This should leave now that the triggers are dynamically installed

db/pf-schema-10.2.0.sql
```diff
@@ -1602,6 +1619,16 @@ CREATE TABLE bandwidth_accounting_history (
     KEY bandwidth_accounting_tenant_id_mac (tenant_id, mac)
 );
 
+--
+-- Table structure for table `event_log`
+--
+
+CREATE TABLE event_log (
```
julsemaan: Not sure this is needed anymore

db/upgrade-10.0.0-10.1.0.sql
```diff
@@ -58,6 +58,17 @@ SET STATEMENT sql_mode='NO_AUTO_VALUE_ON_ZERO' FOR
 \! echo "PIDs should be unique in the password table"
 CREATE UNIQUE INDEX IF NOT EXISTS `pid_password_unique` ON password(pid);
 
+--
+-- New table `event_log`
+--
+\!echo "Creating new table event_log"
+CREATE TABLE IF NOT EXISTS event_log (
```
julsemaan: Not sure this is needed anymore

lib/pf/cmd/pf/generatemariadbconfig.pm
julsemaan: I guess I'm missing something in the code because right now what I see is that the trigger will be installed for all the tables and not just the tables that are part of the configured namespaces

jrouzierinverse: Check the logic here https://github.com/inverse-inc/packetfence/blob/c92364112efb1f5948604dd6855207eaca3b88d5/conf/mariadb/mariadb.sql.tt.example to understand better.
The basic logic is the following.
Re add the pf_logger function.
Drop all triggers.
If trigger is configured then add it.

jrouzierinverse: You look at this test to understand a little more.
 https://github.com/inverse-inc/packetfence/blob/bc841ff6a58f2268f5c383c63f950c3426693f2b/t/unittest/cmd/pf/generatemariadbconfig.t

extrafu:
@jrouzierinverse and @julsemaan What's the status on this PR?

julsemaan:
> @jrouzierinverse and @julsemaan What's the status on this PR?

I believe there is still dead code related to an initial full-perl implementation that needs to be removed and the schema cleanup but the latest seems to be fine now

jrouzierinverse:
The schema cleanup is done.
I will remove the old perl code.

debian/rules
```diff
@@ -120,6 +120,10 @@ install: build
 	#Sudoer
 	install -oroot -groot -d -m0750 $(CURDIR)/debian/packetfence/etc/sudoers.d
 	install -oroot -groot -m0440 $(CURDIR)/debian/packetfence.sudoers $(CURDIR)/debian/packetfence/etc/sudoers.d/packetfence
+	rm -fr $(CURDIR)/debian/packetfence$(PREFIX)/$(NAME)/html/pfappserver/root/static/alt
```
julsemaan, line 123: This doesn't seem to belong in this PR

lib/pf/ConfigStore/EventLogger.pm
```diff
@@ -0,0 +1,50 @@
...
+
+=cut
+
+use HTTP::Status qw(:constants is_error is_success);
+use Moo;
+use namespace::autoclean;
+use pf::file_paths qw($event_loggers_config_file);
+extends 'pf::ConfigStore';
+
+sub configFile { $event_loggers_config_file }
```
julsemaan, line 21: It doesn't look like there is anything done in the admin interface to manage the event loggers

satkunas: @jrouzierinverse see me if we need expedite a UI section

julsemaan:
Will this work with MariaDB 10.5 (used on RHEL8) ?

julsemaan:
This cannot be built on Gitlab:
https://gitlab.com/inverse-inc/packetfence/-/jobs/1430671085

Seems we're missing MariaDB-devel

jrouzierinverse:
> Will this work with MariaDB 10.5 (used on RHEL8) ?

In theory yes there was not changes in the API.
Let me test that theory.

julsemaan:
Also, one other thing, I see that this isn't sending the syslog messages externally but instead logging them locally so my guess is that we'll have to setup a syslog forwarder on top of this ? If so, then we need to make sure that its all tied together with the syslog forwarder so we can pick the namespaces we want to forward there too

Also, where do the syslog messages end up now ? I don't see anything in rsyslog.conf.tt.example so my guess is that they're going in /var/log/messages which isn't very convenient.

jrouzierinverse:
> Also, one other thing, I see that this isn't sending the syslog messages externally but instead logging them locally so my guess is that we'll have to setup a syslog forwarder on top of this ? If so, then we need to make sure that its all tied together with the syslog forwarder so we can pick the namespaces we want to forward there too
> 
> Also, where do the syslog messages end up now ? I don't see anything in rsyslog.conf.tt.example so my guess is that they're going in /var/log/messages which isn't very convenient.

It sends it what ever syslog server you confined in the event logger.
It does not send it to the local syslog server.

julsemaan:
packetfence-mariadb is unable to restart when using this branch because the permissions on var/conf/mariadb.sql only allow root to read this file. This file needs to be readable by the mysql user (or simply all users)

julsemaan:
Couple of more comments:

- The Event loggers section should go in Integration
- We need to instruct that MariaDB has to be restarted after creating/modifying an event logger

julsemaan:
Other issue, when creating an event logger and restarting packetfence-mariadb, it doesn't create the triggers via mariadb.sql so it just does nothing

PR #6043, opened 2020-12-14
Improve failover of AD/LDAP authentication sources + domains

### LINKED ISSUE #6040: Improve AD failover for 802.1x requests
Currently, AD failover is not consistent and will fail unless configured exactly right.

The LDAP source should be able to detect which LDAP server is alive when multiple IPs are defined (or detect which one is alive in a DNS record that contains multiple of them) and then consistently use this IP to connect to the LDAP server. Currently it always tries the first entry and fails over to the 2nd which adds 1 second to every RADIUS request if the first IP is unreachable.

We could flag the alive/dead LDAP IP addresses in Redis so that our processes can share that info and that will allow the LDAP connections to establish faster after an LDAP server is declared dead.

When its a DNS record, we could perform the LDAP lookup ourself instead of relying on the LDAP library and then track the IP addresses like we would if the user would have entered it manually as a list in the configuration (this list would need to be refreshed according to the record's TTL value).

### LINKED ISSUE #6041: winbindd fails to failover when DNS server is unreachable
**Describe the bug**
In your domain configuration, if your first DNS server is down, the timeouts in resolv.conf are so high that winbindd always fails to failover to another available DC with an error like this:
```
tdb_chainlock_with_timeout_internal: alarm (40) timed out for key WIN-G73982ONG8J.inverse.inc in tdb /var/cache/samba/mutex.tdb
```

The timeouts need to be reviewed to allow faster failover to another DNS server

julsemaan:
# Description
Improves the failover of winbindd when losing the primary nameserver and improves the failover of the LDAP sources when one of the IP in the list is unreachable. Also resolves the domain in the auth source to the list of IPs instead to use it the same way we would use a list of IPs in the configuration instead of relying on the LDAP library to perform the lookup.

# Impacts
LDAP authentication sources and winbindd domain authentication

# Issue
fixes #6041 
fixes #6040 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Improved the failover mechanisms when an Active Directory or LDAP server is detected as dead


julsemaan:
I'd like to see this full fix going to the maintenance if the team agrees as it only offers improvements although it will behave differently when an LDAP server is dead

julsemaan:
I have some doc to review by the doc reviewmaster (@extrafu) and the code by @fdurand or @jrouzierinverse, whoever has the most time to check it out.

We can also discuss the maintenance backport during the review

html/pfappserver/root/static.alt/src/views/Configuration/_config/authenticationSource.js
```diff
@@ -685,6 +685,19 @@ export const viewFields = {
       ]
     }
   },
+  dead_duration: (form, meta = {}) => {
+    return {
+      label: i18n.t('Dead duration'),
```
jrouzierinverse: I guess this should be translated.

html/pfappserver/root/static.alt/src/views/Configuration/_config/authenticationSource.js
```diff
@@ -685,6 +685,19 @@ export const viewFields = {
...
+      text: i18n.t('How much time in seconds should a server be marked dead before it is retried. When specifying multiple LDAP servers or a DNS name pointing to multiple IPs, then this option can be used to offer more consistent failover. A value of 0 disables this feature.'),
```
jrouzierinverse: Same

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -224,6 +225,10 @@ sub _connect {
   my $logger = Log::Log4perl::get_logger(__PACKAGE__);
   my ($LDAPServer, $LDAPServerPort);
   my @LDAPServers = split(/\s*,\s*/, $self->{'host'});
+  
+  # Lookup the server hostnames to IPs so they can be shuffled better and to improve the failure detection
+  @LDAPServers = map { valid_ip($_) ? $_ : @{resolve($_)} } @LDAPServers;
```
jrouzierinverse, line 230: This does not handle the case where a port is defined in the host list.

jrouzierinverse (changes requested):

lib/pf/Authentication/Source/LDAPSource.pm
julsemaan: I just figured out we can ignore the override port specification for hostnames (it still works for IPs)

That feature is undocumented and not used by any of our customers (makes no sense to use different ports per LDAP hosts). This just looks like a leftover of PF version 3 when authentication sources were Perl modules

html/pfappserver/root/static.alt/src/views/Configuration/_config/authenticationSource.js
julsemaan: The translation will happen pre-release like it always does

jrouzierinverse: If it will be backported. It should be translated.

lib/pf/Authentication/Source/LDAPSource.pm
jrouzierinverse: That feature was added in this commit 4d1c2db2315ca8d084bba3172c3bd3dc592ff6b6 (This is a 4.0.0).
This was a feature added since we added support for multiple LDAP servers.

My thoughts are this if it is going backported do minimal things that can break a feature that might be used by someone. If this is going to be backported support the override port.

If not then to keep the code clean remove the splitting of the host/port below and add a message in the upgrade notes about removing the feature.

Also add upgrade script should be added to clean up and remove the port from the hostnames.

html/pfappserver/root/static.alt/src/views/Configuration/_config/authenticationSource.js
julsemaan: Lets get someone comfortable with the translation on this then if its important to the merge of this PR since I'm not familiar with that

lib/pf/Authentication/Source/LDAPSource.pm
julsemaan: Lets just skip the backporting and manually patch whoever needs this (shouldn't be too many people)

I'll remove the code of multiple ports and add a note in the guide

jrouzierinverse:
rebase

julsemaan:
Just rebased, lets merge this ASAP before another change in the new admin gets this outdated

jrouzierinverse:
rebase

julsemaan:
Let me know if its mergeable to you first, I'll then rebase and merge right away since it keeps getting conflicts from @satkunas' work

lib/pf/Authentication/Source/LDAPSource.pm
```diff
@@ -232,40 +237,70 @@ sub _connect {
...
-    );
+  my $try_connect = sub {
+      my $honor_dead = shift;
+      TRYSERVER:
+      foreach my $s (@LDAPServers) {
+        $LDAPServer = $s;
+        $LDAPServerPort = undef;
+        # check to see if the hostname includes a port (e.g. server:port)
+        if ($LDAPServer =~ /:/) {
+            $LDAPServerPort = (split(/:/, $LDAPServer))[-1];
```
jrouzierinverse: Remove this since we are dropping using the port.

jrouzierinverse (changes requested):
Add in the upgrade guide the removal of Port in the address.

julsemaan:
Far from being a fan of self-merge, I just added this to devel before it gets outdated again. Since the admin component isn't compatible with v10.x, I'll omit this from the maintenance for now

PR #6094, opened 2021-01-14
add rouge in Docker images

nqb:
# Description
add rouge in Docker images

Related to #5968 

# Delete branch after merge
YES


nqb:
I'm removing `asciidoctor-pdf` from Docker images used to build Packetfence because:
* it requires Ruby 2.1 (not packaged on CentOS 7)
* we don't use it during build process. 

In order to build PDF in CI, I will rely on asciidoctor Docker images.

PR #6120, opened 2021-01-29
Added Aruba CX switch module

fdurand:
# Description
Aruba CX switch module

# Impacts
Switch module

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Aruba CX support

### DIFF
--- a/lib/pf/Switch/Aruba/cx.pm
+++ b/lib/pf/Switch/Aruba/cx.pm
@@ -1,0 +1,219 @@
+package pf::Switch::Aruba::cx;
+
+=head1 NAME
+
+pf::Switch::Aruba::cx
+
+=head1 SYNOPSIS
+
+Module to manage rebanded Aruba HP CX Switche
+
+=head1 STATUS
+
+=over
+
+=item Supports
+
+=over
+
+=item MAC-Authentication
+
+=item 802.1X
+
+=item Radius downloadable ACL support
+
+=item Voice over IP
+
+=item Radius CLI Login
+
+=back
+
+=back
+
+Has been reported to work on Aruba CX
+
+=cut
+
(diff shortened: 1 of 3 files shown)


lib/pf/Switch/Aruba/cx.pm
```diff
@@ -0,0 +1,219 @@
+package pf::Switch::Aruba::cx;
+
+=head1 NAME
+
+pf::Switch::Aruba::cx
+
+=head1 SYNOPSIS
+
+Module to manage rebanded Aruba HP CX Switche
```
julsemaan: I think its spelled chwitche

lib/pf/Switch/Aruba/cx.pm
```diff
@@ -0,0 +1,219 @@
+package pf::Switch::Aruba::cx;
+
+=head1 NAME
+
+pf::Switch::Aruba::cx
```
julsemaan: To follow our naming standards, we should use pf::Switch::Aruba::CX

lib/pf/Switch/Aruba/cx.pm
```diff
@@ -0,0 +1,219 @@
...
+
+=head1 SYNOPSIS
+
+Module to manage rebanded Aruba HP CX Switche
+
+=head1 STATUS
+
+=over
+
+=item Supports
```
julsemaan: Can you confirm that this switch indeed supports all the features that are outlined there (like dACL, RADIUS CLI login, etc)

fdurand: It's based on the Aruba::5400 who support that.

PR #6121, opened 2021-01-29
Update to Stripe v3

julsemaan:
# Description
Stripe now required to use the latest version of Stripe checkout (elements) for auto compliance, this updates the integration to use Stripe elements (also known as Stripe checkout v3)

# Impacts
Stripe integration

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Updated Stripe integration to use Stripe Elements

### DIFF
--- a/lib/pf/Authentication/Source/StripeSource.pm
+++ b/lib/pf/Authentication/Source/StripeSource.pm
@@ -109,6 +109,12 @@ sub _send_json {
     return $self->_do_request($curl);
 }
 
+sub _get_json {
+    my ($self, $curl, $path) = @_;
+    $self->_set_url($curl, $path);
+    return $self->_do_request($curl);
+}
+
 =head2 _send_form
 
 send form data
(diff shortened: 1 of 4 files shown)


julsemaan:
There is 1 last detail to figure out, I had to go back to creating one customer entry per subscription (instead of reusing the customer entry) due to how declines are now handled.

Reached out to Stripe regarding that, we'll see if I have a way to put it back to using the same customer object, otherwise we'll merge as-is.

julsemaan:
@jrouzierinverse, please review this for inclusion in 10.3

lib/pf/Authentication/Source/StripeSource.pm
```diff
@@ -240,12 +252,15 @@ sub subscribe_customer {
...
 
     my ($code, $data);
-    if(my $cus = $self->get_customer_by_email($session->{email})) {
-        ($code, $data) = $self->_send_form($self->curl, "v1/customers/".$cus->{id}, $object);
-    }
-    else {
-        ($code, $data) = $self->_send_form($self->curl, "v1/customers", $object);
+    ($code, $data) = $self->_send_form($self->curl, "v1/customers", $object);
+
+    my $subscriptions = $self->_get_json($self->curl, "v1/subscriptions?customer=".$data->{id});
```
jrouzierinverse: ```$data->{id}``` needs to url encoded

jrouzierinverse (changes requested):

lib/pf/Authentication/Source/StripeSource.pm
julsemaan: Done

Issue #5550, opened 2020-06-08
API - Database Create timeout


satkunas:
**Describe the bug**
Create database fails in configurator due to 504 gateway timeout. The database creation succeeds in the background, but the web admin in left in an unsynchronized state that requires a manual drop of the database and a refresh/reset of the configurator GUI

**To Reproduce**
Install Packetfence on a potato and try to pass the configurator.

**Expected behavior**
Server timeout should allow slower hardware to pass the database creation.

**Additional context**
Significantly increasing the server timeout (from `timeout server 50000`) resolves the problem. Database creation on a 5400rpm drive took 120s to complete.

https://github.com/inverse-inc/packetfence/blob/devel/conf/haproxy-admin.conf.example#L42

nqb:
I've seen same behavior when trying to display ~250 switches in a cluster environment that match [our minimum hardware requirements](https://github.com/inverse-inc/packetfence/blob/13d330e1496ede8f0eb968500fec7426b737e489/docs/PacketFence_Installation_Guide.asciidoc#minimum-hardware-requirements).

During [integration tests](https://github.com/inverse-inc/packetfence/blob/13d330e1496ede8f0eb968500fec7426b737e489/t/venom/pfservers/configurator/40_run_custom_step.yml#L34-L37), I've to increase this timeout too to let PacketFence services start (due to ressources).

I suggest to increase this timeout to 120s like @satkunas proposed.

julsemaan:
I think haproxy-admin should match the values from https://github.com/inverse-inc/packetfence/blob/devel/conf/caddy-services/api.conf.example#L61-L64

And let the api-frontend handle timeouts that are shorter than that if applicable

nqb:
It make sense to me.

Should we do a manual sync ?

julsemaan:
> It make sense to me.
> 
> Should we do a manual sync ?

I think we should for now to get this fixed, then potentially move to a value in pf.conf for the admin panel related timeouts (which we can track in another issue)

That will allow for a faster delivery of the fix

satkunas:
Fixed in https://github.com/inverse-inc/packetfence/pull/6167 by adding pfqueue async call

Closing.

Issue #3980, opened 2019-02-04
configurator: provide a way to configure Fingerbank proxy


nqb:
As reported by @kkkimiko, if you use an HTTP proxy and want to configure your Fingerbank API key in PF's configurator, you can't.

If you specify, `http*_proxy` vars, they will not taken into account.

If you skip this step and want to configure the Fingerbank proxy settings in the Web admin, you can't.

The only way to do that is to edit by hand `/usr/local/fingerbank/conf/fingerbank.conf` like this:

```
[upstream]
api_key=YOUR_API_KEY

[proxy]
host=myproxy.domain.lan
port=3128
``` 
Then do a `/usr/local/pf/bin/pfcmd configreload hard`.

We should provide a way to specify proxy settings directly in configurator at the Fingerbank step.

nqb:
It's now possible through ocnfigurator.

Issue #3707, opened 2018-10-24
Installation - Step 3 MySQL Create Database


satkunas:
In Step 3 of the configurator, in the section "Create the database", after entering the database name and clicking the "Create database and tables" button, there is no feedback to indicate that the command succeeded.

Furthermore, a nice to have would be to reproduce the last row in the `pf_version` table in this section to show the user which schema (if any) is currently loaded. This would help provide the user feedback in the issue above.

julsemaan:
@satkunas 

Want to take care of the fix for that issue ?

satkunas:
@jrouzierinverse do you think it's possible to replace the API calls to use pfqueue polling to work around these timeouts?

jrouzierinverse:
Yes

fdurand:
Closing as stale after a review of the open issues: this was reported against the 2018 configurator, the component involved has been rewritten since, and there has been no follow-up. The configurator was rewritten in the current admin.

If you can reproduce it on a current release (15.x), please reopen with the details — or open a new issue — and we will look at it.

PR #6127, opened 2021-02-02
Composition Configurator

satkunas:
# Description
Refactor the UI Congfigurator to utilize the Vue 3 composition API.

# Impacts
Improves (does not alter or re-order) form workflow through configurator. 
Increases visibility of server-side errors.
Utilizes pfqueue polling to address timeouts on long requests.

# Delete branch after merge
YES 

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests

## Enhancements
* Improves (does not alter or re-order) form workflow through configurator. 
* Increases visibility of server-side errors.
* Utilizes pfqueue polling to address timeouts on long requests.
* Add Fingerbank proxy to configurator.

## Bug Fixes
* API - Database Create timeout (#5550)
* Installation - Step 3 MySQL Create Database (#3707)
* configurator: provide a way to configure Fingerbank proxy (#3980)


satkunas:
Waiting on (#6128) before finalizing.

nqb:
I will need to adjust integration tests related to configurator in order to use new pfqueue API calls. This should be simple because I already did it Venom when [joining domain](https://github.com/inverse-inc/packetfence/blob/1d7d7fbc768f952a7e04aec96c23ff64f8017636/t/venom/pfservers/dot1x_eap_peap/05_join_domain.yml#L60-L74).

@jrouzierinverse or @satkunas, which API calls are concerned ?

satkunas:
@nqb the endpoint isn't new, and is now used with 4 endpoints:

You can grep for them using `pfqueue/pollTaskStatus`:
```
#ag pfqueue/pollTaskStatus -B 1

src/views/Configuration/bases/_store.js
891-    }).then(({ task_id }) => {
892:      return store.dispatch('pfqueue/pollTaskStatus', task_id).then(response => {

src/views/Configuration/domains/_store.js
118-    return api.joinDomain(data).then(response => {
119:      return store.dispatch('pfqueue/pollTaskStatus', response.task_id).then(response => {
--
130-    return api.rejoinDomain(data).then(response => {
131:      return store.dispatch('pfqueue/pollTaskStatus', response.task_id).then(response => {
--
142-    return api.unjoinDomain(data).then(response => {
143:      return store.dispatch('pfqueue/pollTaskStatus', response.task_id).then(response => {
```

* [baseurl]/config/base/database/create
* [baseurl]/config/domain/:id/join
* [baseurl]/config/domain/:id/rejoin
* [baseurl]/config/domain/:id/unjoin

The base url for these 4 endpoints can be either /api/v1/configuration/... or /api/v1/configurator/...

Once the above issue is addressed this list will include the service endpoint.

julsemaan (changes requested):
As discussed with @satkunas, since the API calls now require pfqueue, both redis_queue and pfqueue need to be started on install for both CentOS and Debian. 

Otherwise the DB fails to get setup.

Also, we need to ensure that pfqueue and redis_queue get started on reboot during a new install, perhaps by adding them to the packetfence-base target since they are now necessary even when PF is fresh installed

satkunas:
Moved to #6167

PR #6134, opened 2021-02-08
send the deauth to all APs of switch group for Unifi

julsemaan:
# Description
Send the deauth to all APs of switch group for Unifi which allows to authorize/deauthorize a MAC address on all the APs of the switch group the Unifi AP is in

# Impacts
Unifi deauth

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Authorize a MAC address on all APs of the switch group when using the Unifi module

### DIFF
--- a/lib/pf/Switch/Ubiquiti/Unifi.pm
+++ b/lib/pf/Switch/Ubiquiti/Unifi.pm
@@ -40,6 +40,7 @@ use pf::config qw(
 use pf::locationlog;
 use Try::Tiny;
 use JSON::MaybeXS;
+use pf::config::cluster;
 
 # The port to reach the Unifi controller API
 our $UNIFI_API_PORT = "8443";
@@ -257,7 +258,6 @@ sub _deauthenticateMacWithHTTP {
 
     my $json_data = decode_json($response->decoded_content());
 
-    $args->{ap_mac} = $self->{_id};
     foreach my $entry (@{$json_data->{'data'}}) {
         $response = $ua->get("$base_url/api/s/$entry->{'name'}/stat/sta/$mac");
         if ($response->is_success) {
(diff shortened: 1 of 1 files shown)


julsemaan:
Things I'd like feedback on:

- Should we do this by default and make it a built-in behavior or should this be configurable in the switch entry?
- Should we apply this principle to APs part of the 'default' group?

lib/pf/Switch/Ubiquiti/Unifi.pm
```diff
@@ -269,16 +269,25 @@ sub _deauthenticateMacWithHTTP {
...
-    } else {
-        foreach my $entry (@{$json_data->{'data'}}) {
-            $response = $ua->post("$base_url/api/s/$entry->{'name'}/cmd/stamgr", Content => encode_json($args));
-            if ($response->is_success) {
-                $logger->info("Deauth on site: $entry->{'desc'}");
+    tie my %SwitchConfig, 'pfconfig::cached_hash', "config::Switch($host_id)";
+    foreach my $switch_id (keys(%SwitchConfig)) {
+        my $switch = $SwitchConfig{$switch_id};
+        # If the other switch is a MAC based entry and an Ubiquiti AP and is part of the same AP group, then we'll send the deauth
+        if(valid_mac($switch_id) && $switch->{type} eq $self->{_type} && $switch->{group} eq $self->{_group}) {
```
fdurand, line 277: Maybe test the controller ip too.

lib/pf/Switch/Ubiquiti/Unifi.pm
```diff
@@ -269,16 +269,25 @@ sub _deauthenticateMacWithHTTP {
...
+            if ($found) {
+                $response = $ua->post("$base_url/api/s/$site_opts{'name'}/cmd/stamgr", Content => encode_json($args));
+                if ($response->is_success) {
+                    $logger->info("Deauth on site: $site_opts{'desc'}");
+                }
+            } else {
+                foreach my $entry (@{$json_data->{'data'}}) {
+                    $response = $ua->post("$base_url/api/s/$entry->{'name'}/cmd/stamgr", Content => encode_json($args));
+                    if ($response->is_success) {
+                        $logger->info("Deauth on site: $entry->{'desc'}");
```
fdurand: Can we break the loop here ?

julsemaan: The goal is to deauth on all the APs of the switch group, wouldn't breaking here stop it from happening ?

fdurand:
The controller is supposed to find where the device is connected, so it surprise me that we need to do that.
Btw if it's the only way to do make sure that we reevaluate the access then we can go for it.

My concern to make it by default is if you have a lot of AP then it take time.
And maybe we can have an switch config attribute on the "defined controller" where we set if we want to enable this option.

lib/pf/Switch/Ubiquiti/Unifi.pm
fdurand: Yes i know but if it succeed , does it mean i found the mac on the ap ?
Or it succeed every time ?

julsemaan: From what I was told, it will succeed even if the MAC is not on the AP but I still need to test this in the field with my customer to confirm this 100%

fdurand: Ok , so don't break the loop.
Btw we will have a lot of logs.

julsemaan: There should be 1 line for each AP we do the deauth for. We could always get this in debug but I wanted to try to make it explicit it was doing it on multiple APs

julsemaan:
> The controller is supposed to find where the device is connected, so it surprise me that we need to do that.
> Btw if it's the only way to do make sure that we reevaluate the access then we can go for it.

This is not to get it to authenticate on the AP the user is connected on but instead to make it so when the user roams from AP to AP, it doesn't pop open the portal to authorize the user since the AP will already know about the device.

fdurand:
My bad, i misunderstood the goal of this PR.
So we can set it as the default deauth method when we do webauth.
Also for the log, maybe we can just join the ap mac in a single line.

julsemaan:
> My bad, i misunderstood the goal of this PR.
> So we can set it as the default deauth method when we do webauth.
> Also for the log, maybe we can just join the ap mac in a single line.

You're right, I'll collect all the MAC addresses and print them in a single line (or say did it on X amount of APs since it could be too long for syslog anyway)

fdurand:
Just fix the log and i will merge it

julsemaan:
> Just fix the log and i will merge it

I adjusted the logging. Wasn't able to fully test it but the change is relatively straightforward

PR #6135, opened 2021-02-08
scroll devices in Intune to support counts over 1000

julsemaan:
# Description
Scroll devices in Intune to support counts over 1000 since Intune now requires that

# Impacts
Intune integration

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Adjust Intune integration to support pagination of the managed devices

### DIFF
--- a/lib/pf/provisioner/intune.pm
+++ b/lib/pf/provisioner/intune.pm
@@ -207,8 +207,8 @@ sub update_config {
     return $cs->commit();
 }
 
-sub get_device_info {
-    my ($self, $mac) = @_;
+sub perform_get_device_info {
+    my ($self, $mac, $url) = @_;
     my $logger = get_logger();
 
     unless ($self->get_access_token()) {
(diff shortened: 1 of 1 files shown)


julsemaan:
marking as not-ready since I'm just waiting on final confirmation of the fix

lib/pf/provisioner/intune.pm
```diff
@@ -233,6 +232,29 @@ sub get_device_info {
...
+    my $info = $self->perform_get_device_info($mac, $self->protocol.'://' . $self->host . ':' .  $self->port . '/v1.0/deviceManagement/managedDevices?$select=wiFiMacAddress,complianceState,id');
+    if($info == $pf::provisioner::COMMUNICATION_FAILED) {
+        return $pf::provisioner::COMMUNICATION_FAILED;
+    }
+    push @infos, $info;
+
+    while($info && $info != $pf::provisioner::COMMUNICATION_FAILED && $info->{'@odata.nextLink'}) {
+        $info = $self->perform_get_device_info($mac, $info->{'@odata.nextLink'});
+        if($info == $pf::provisioner::COMMUNICATION_FAILED) {
+            return $pf::provisioner::COMMUNICATION_FAILED;
```
jrouzierinverse, line 250: If the previous messages succeeded shouldn't you just stop and return what was received prior instead of throw away all the previous messages?

julsemaan: Yes we do throw it away so that it can be retried from scratch. Otherwise you could have an incomplete list if we would return what we already have

PR #6140, opened 2021-02-13
fix/6136

### LINKED ISSUE #6136: pfacct consumes all available memory
We're running PF 10.2 with the latest patches applied (/usr/local/pf/addons/pf-maint.pl). Since applying updates last week memory consumption on pfacct processes grows to eventually consume all memory and swap, leading the unavailability of the system when oom (out of memory) protection kicks in and starts ending processes.

We'll gladly provide relevant information, herewith an overview from htop:
image

jrouzierinverse:
Description
===========
Don't return until the accounting request was handled.
To avoid massive amount of go routines from running at the same time.

Impacts
=======
pfacct - radius

NEWS file entries
=======================
Bug Fixes
-------------
* Don't return until the accounting request was handled.

Issue
=====
fixes #6136

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/go/cmd/pfacct/pfacct.go
+++ b/go/cmd/pfacct/pfacct.go
@@ -15,8 +15,8 @@ import (
 	"github.com/inverse-inc/packetfence/go/log"
 	"github.com/inverse-inc/packetfence/go/mac"
 	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
-	"github.com/inverse-inc/packetfence/go/tryableonce"
 	"github.com/inverse-inc/packetfence/go/sharedutils"
+	"github.com/inverse-inc/packetfence/go/tryableonce"
 	statsd "gopkg.in/alexcesaro/statsd.v2"
 )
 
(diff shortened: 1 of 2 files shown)


julsemaan:
Maybe I'm missing something but what's the point of queuing the processing if its to wait for it to be done before proceeding.

Using a channel with a max size would likely accomplish the same goal while leaving this async

jrouzierinverse:
We were not queuing the processing to avoid waiting.
It was done to process avoid processing the same mac in different goroutines.

However, not waiting can cause more packets to be processed than necessary.
The code in question is here https://github.com/inverse-inc/go-radius/blob/82756e2d8d473ef090ad16527e1ff00fff846a15/server-packet.go#L157-L163

julsemaan:
> We were not queuing the processing to avoid waiting.
> It was done to process avoid processing the same mac in different goroutines.
> 
> However, not waiting can cause more packets to be processed than necessary.
> The code in question is here https://github.com/inverse-inc/go-radius/blob/82756e2d8d473ef090ad16527e1ff00fff846a15/server-packet.go#L157-L163

I see, disregard that comment then

PR #6146, opened 2021-02-24
Improve CLI login on SG300

nqb:
# Description
- Read-only privilege for SG300 should be set to 1 (in place of 3 on default Cisco switches).
See: https://www.cisco.com/c/en/us/support/docs/smb/switches/cisco-small-business-300-series-managed-switches/smb3927-user-account-configuration-on-300-series-managed-switches.html

- Force NAS-Port-Type to "virtual" when not defined in order to match on a CLI connection profile.
NAS-Port-Type is not set by SG300 when sending a RADIUS request for a CLI login with firmware 1.4.9.4. Otherwise, NAS-Port-Type is set to Ethernet.

# Impacts
SG300 switch module

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Enhance CLI login on SG300 switches

### DIFF
--- a/lib/pf/Switch/Cisco/SG300.pm
+++ b/lib/pf/Switch/Cisco/SG300.pm
@@ -91,6 +91,26 @@ sub NasPortToIfIndex {
     return $NAS_port;
 }
 
+=head2 returnAuthorizeRead
+
+Return radius attributes to allow read access
+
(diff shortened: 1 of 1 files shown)


nqb:
bump @fdurand

PR #6148, opened 2021-02-25
Use user-name from the radius accounting when exists for the firewall sso

fdurand:
# Description
Use the user-name from the radius accounting request to update the firewall.
Fix the case when an accounting stop is sent just after an authentication request (user-name can be different)

# Impacts
Firewall SSO

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Firewall SSO - Fix the case when an accounting stop is sent just after an authentication request (user-name can be different)


nqb:
I will not backport to maintenance to avoid unexpected behaviors.

PR #6060, opened 2020-12-28
feature/tenant-aware-pfdhcplistener

jrouzierinverse:
Description
===========
Make the pfdhcplistener tenant aware

Impacts
=======
pfdhcplistener

NEWS file entries
=======================
Enhancements
-------------
* The pfdhcplistener is tenant aware.

Issue
=====
fixes #5991

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests
- [ ] Add acceptance tests (TestLink)

UPGRADE file entries
=========
To add the tenant_id to all network configurations run 
perl /usr/local/pf/addons/upgrade/to-11.0-network-conf.pl

PR #6204, opened 2021-03-22
feature/tenant-aware-pfdhcplistener (replaces #6060)

### LINKED ISSUE #5991: pfdhcplistener is not tenant aware
When you are in a multi-tenant setup and you use a DHCP sensor, `pfdhcplistener` will process DHCP packets and create/update all nodes in **default** tenant even if nodes already exist in other tenants.

1. Update nodes already created in DB in correct tenant
1. Create node in correct tenant when node is unknown: we could create a mapping between subnets declared in DHCP and tenants

nqb:
Description
===========
Make the pfdhcplistener tenant aware

Impacts
=======
pfdhcplistener

NEWS file entries
=======================
Enhancements
-------------
* The pfdhcplistener is tenant aware.

Issue
=====
fixes #5991

Delete branch after merge
=========================
NO

Checklist
=========
- [ ] Document the feature
- [x] Add unit tests
- [x] Add acceptance tests (TestLink)

UPGRADE file entries
=========
To add the tenant_id to all network configurations run 
perl /usr/local/pf/addons/upgrade/to-11.0-network-conf.pl


nqb (commented):
### backend

I see a lot of messages like this after I configured a network from API:

```
Mar 22 14:55:57 pfdeb9dev packetfence: -e(18461) WARN: Unknown network type for network 172.20.20.0 (pf::config::get_network_type)                                                                                │
```

What I did is:
- create a tenant
- restart packetfence-config and packetfence services
- configure a network type "other" (172.20.20.0/24) like this:
```ini
[172.20.20.0]
type=other
netmask=255.255.255.0
network=172.20.20.0
tenant_id=2
```
- restart pfdns, pfdhcp, iptables and pfdhcplistener
- forge a DHCP request for a node in 172.20.20.0 network
- check if node is created in DB:
   * new tenant => failed
   * defaul tenant => OK

### frontend

- "Other" network type is not available when creating a new routed network. API works as expected.
- When creating an "other" network, we should restrict creation to:
  - type
  - subnet
  - netmask
  - tenant

ATM, frontend expects a DHCP configuration like for a "real" routed network.

nqb:
@jrouzierinverse and @satkunas, let me know if you need more details.

extrafu:
@jrouzierinverse bump

nqb:
@jrouzierinverse, let me know when I can test it.

jrouzierinverse:
You can test

nqb:
### Tenant detection is wrong

I just understand why the tenant part doesn't work.

In fact, code of this PR rely on `src_ip` address of DHCP packets to assign a tenant to a node.
IMHO, it's not a good idea because `src_ip` will be equals to:

PacketFence see traffic at layer 2:
- 0.0.0.0 for DHCP DISCOVER and DHCP REQUEST messages
- IP address of DHCP server for DHCP OFFER and ACK messages

PacketFence received traffic (from DHCP sensor or DHCP relay):
- 0.0.0.0 for DHCP DISCOVER and DHCP REQUEST messages
- IP address of DHCP server or DHCP relay for DHCP OFFER and ACK messages

We need to rely directly on DHCP data (certainly client IP address) to assign a tenant.

### Unknown network type for network 172.20.20.0

I still see this error message certainly because there is no conditional in `lib/pf/config.pm` for "other" networks.

extrafu:
@jrouzierinverse What's the status on this?

jrouzierinverse:
After speaking to @fdurand the best approach for detecting which tenant to use is the following logic.
If the relay address is defined then use that to determine which tenant network is to be used.
Otherwise use the interface's IP address to determine tenant network.

nqb:
> If the relay address is defined then use that to determine which tenant network is to be used.
> Otherwise use the interface's IP address to determine tenant network.

I'm not in favor of this logic for following reasons:

### PacketFence see DHCP traffic at layer 2

You need to have:
- a "other" network defined with that tenant
- a PF interface with an IP in that "other" network

You have a 1:1 mapping between interface and tenant assigned to DHCP traffic receive on that interface. It means that if you want to manage 3 tenants, you have to create 3 interfaces on PacketFence.

### PacketFence receive DHCP traffic from DHCP relay

Logic will work in this scenario.

### PacketFence receive DHCP traffic from DHCP sensor (common use case, really accurate)

Logic could not work in this scenario if DHCP server receive DHCP traffic without any relay. On top that, DHCP sensor send DHCP traffic to management interface so we can't rely on logic related to interface's IP address.

nqb:
Considering time left before 10.3 release, I will merge PR with this logic. It will cover most of cases and can be seen as a first implementation.

PR #6224, opened 2021-03-24
Feature/switch acl and inherit options

fdurand:
# Description
Added back acl in switches (If the ACL doesn't exist then we use the one defined in the role config)
Added inheritance on Role by switch role and for Web Auth.

# Impacts
Roles

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Inheritance on parent role on Role and Web Auth

### DIFF
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -114,6 +114,8 @@ CREATE TABLE `node_category` (
   `fingerbank_dynamic_access_list` varchar(255) default NULL,
   `acls` TEXT NOT NULL,
   `inherit_vlan` varchar(50) default NULL,
+  `inherit_role` varchar(50) default NULL,
+  `inherit_web_auth` varchar(50) default NULL,
   PRIMARY KEY (`category_id`),
   UNIQUE KEY node_category_name (`name`)
 ) ENGINE=InnoDB;
--- a/db/pf-schema-X.Y.Z.sql
+++ b/db/pf-schema-X.Y.Z.sql
@@ -115,7 +115,7 @@ CREATE TABLE `node_category` (
   `acls` TEXT NOT NULL,
   `inherit_vlan` varchar(50) default NULL,
   `inherit_role` varchar(50) default NULL,
-  `inherit_web_auth` varchar(50) default NULL,
+  `inherit_web_auth_url` varchar(50) default NULL,
   PRIMARY KEY (`category_id`),
   UNIQUE KEY node_category_name (`name`)
 ) ENGINE=InnoDB;
--- a/html/pfappserver/root/src/views/Configuration/switches/_components/TheForm.vue
+++ b/html/pfappserver/root/src/views/Configuration/switches/_components/TheForm.vue
@@ -157,6 +157,24 @@
             </div>
           </b-card>
 
+          <b-card v-show="supports(['AccessListBasedEnforcement'])"
+            class="mb-3 pb-0" no-body
+          >
+            <b-card-header>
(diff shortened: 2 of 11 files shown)


lib/pf/Switch.pm
```diff
@@ -586,8 +586,17 @@ sub getRoleByName {
...
-    return if (!defined($self->{'_roles'}) || !%{$self->{'_roles'}});
+
+    if (!defined($self->{'_roles'}) || !defined($self->{'_roles'}{$roleName})) {
+        my $parent = _parentRoleForRole($roleName);
+        if (defined $parent && length($parent)) {
+            return $self->getRoleByName($parent);
+        }
+        # VLAN name doesn't exist
+        $pf::StatsD::statsd->increment(called() . ".error" );
+        $logger->warn("No parameter ${vlanName}Role found in conf/switches.conf for the switch " . $self->{_id});
```
jrouzierinverse: ```Global symbol "$vlanName" requires explicit package name at lib/pf/Switch.pm line 597.```

lib/pf/Switch.pm
```diff
@@ -597,6 +606,20 @@ sub getRoleByName {
     return;
 }
 
+sub _parentRoleForRole {
```
jrouzierinverse, line 609: Update t/hardware-snmp-objects.t so test can pass.

jrouzierinverse (changes requested):
Somethings need to be updated

html/pfappserver/root/src/views/Configuration/switches/_components/TheForm.vue
```diff
@@ -157,6 +157,24 @@
...
+          >
+            <b-card-header>
+              <h4 class="mb-0" v-t="'Role mapping by Access List'"></h4>
+            </b-card-header>
+            <div class="card-body pb-0">
+              <form-group-toggle-access-list-map namespace="AccessListMap"
+                :column-label="$i18n.t('Role by Access List')"
+              />
+
+              <form-group-role-map-access-list v-for="role in roles" :key="`${role}AccessList`" :namespace="`${role}AccessList`"
```
julsemaan, line 172: Make sure that it clearly states in the admin that putting an ACL here will supersede the ACL defined in the role

lib/pf/Switch.pm
```diff
@@ -656,6 +679,9 @@ sub _parentRoleForVlan {
 sub getAccessListByName {
     my ($self, $access_list_name, $mac) = @_;
     my $logger = $self->logger;
+
+    return $self->{'_access_lists'}->{$access_list_name} if (defined($self->{'_access_lists'}->{$access_list_name}));
```
julsemaan: Add a logger debug saying something like "Using ACL from the switch entry instead of the one defined in the role"

db/pf-schema-X.Y.Z.sql
```diff
@@ -114,6 +114,8 @@ CREATE TABLE `node_category` (
   `fingerbank_dynamic_access_list` varchar(255) default NULL,
   `acls` TEXT NOT NULL,
   `inherit_vlan` varchar(50) default NULL,
+  `inherit_role` varchar(50) default NULL,
+  `inherit_web_auth` varchar(50) default NULL,
```
julsemaan: That's me being picky but since in switches its called Url, we should call this inherit_web_auth_url, otherwise its name doesn't refer to the URL field and one could think that it refers to something else like the external portal enforcement setting on the switch

extrafu:
@julsemaan bump

PR #6293, opened 2021-04-15
Feature/move pre process filter

fdurand:
# Description
Added the ability to rewrite the username of the radius request based on the radius prePRocess filter.

# Impacts
- RADIUS authorize workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Rewrite the username of the request from radius preProcess filter

### DIFF
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -2994,7 +2994,7 @@ sub parseRequestUsername {
     foreach my $attribute (@{$Config{radius_configuration}{username_attributes}}) {
         if(exists($radius_request->{$attribute})) {
             my $user_name = $radius_request->{$attribute};
-            get_logger->debug("Extracting username '$user_name' from RADIUS attribute $attribute");
+            get_logger->info("Extracting username '$user_name' from RADIUS attribute $attribute");
             return $user_name;
         }
     }
--- a/lib/pf/Switch.pm
+++ b/lib/pf/Switch.pm
@@ -2994,7 +2994,7 @@ sub parseRequestUsername {
     foreach my $attribute (@{$Config{radius_configuration}{username_attributes}}) {
         if(exists($radius_request->{$attribute})) {
             my $user_name = $radius_request->{$attribute};
-            get_logger->info("Extracting username '$user_name' from RADIUS attribute $attribute");
+            get_logger->debug("Extracting username '$user_name' from RADIUS attribute $attribute");
             return $user_name;
         }
     }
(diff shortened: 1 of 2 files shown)


nqb (changes requested):
I understand the logic.

However, in current state, if you build a RADIUS attribute on the fly to use it as a username, you will have following behavior:
- logic related to machine_account still use username detected before
- logger will display wrong information (`handling autz ..`)

lib/pf/radius.pm
```diff
@@ -233,14 +229,17 @@ sub authorize {
     if ($rule) {
         my ($reply, $status) = $filter->handleAnswerInRule({%$rule, merge_answer => 'enabled' }, $args, $radius_request);
         %$radius_request = %$reply;
+        $args->{'user_name'} = $switch->parseRequestUsername($radius_request);
+        $args->{'username'} = $args->{'user_name'};
```
nqb, line 236: Not sure we need this instruction.

lib/pf/radius.pm
```diff
@@ -233,14 +229,17 @@ sub authorize {
...
+        $self->_machine_auth_detection($user_name,\$node_obj,\$options);
```
nqb:
```suggestion
        $self->_machine_auth_detection($args->{'$user_name'},\$node_obj,\$options);
```

lib/pf/Switch.pm
```diff
@@ -2994,7 +2994,7 @@ sub parseRequestUsername {
     foreach my $attribute (@{$Config{radius_configuration}{username_attributes}}) {
         if(exists($radius_request->{$attribute})) {
             my $user_name = $radius_request->{$attribute};
-            get_logger->debug("Extracting username '$user_name' from RADIUS attribute $attribute");
+            get_logger->info("Extracting username '$user_name' from RADIUS attribute $attribute");
```
nqb: I suggest we keep "debug" level for this instruction but we add same instruction directly in preProcess block in radius.pm

nqb (changes requested):

nqb:
These changes allow us to rewrite username on the fly based on a RADIUS filter in `preProcess` scope. 

However, it seems that it's not possible to modify RADIUS attributes in a RADIUS request and have them be taken into account by PacketFence later. I don't see a simple way to create a connection profile using an Advanced Filter to filter on RADIUS attributes.

nqb:
@jrouzierinverse, is it possible to use an advanced filter like `radius_request.Calling-Station-Id == 192.168.0.1` on a connection profile ?

lib/pf/radius.pm
nqb: Used later in the code.

PR #6303, opened 2021-04-19
Filter netdata incoming connection

fdurand:
# Description
Manage iptables rules for netdata.
DROP packet that not coming from the admin to netdata

# Impacts
iptables rules

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Protect netdata from incoming connection

### DIFF
--- a/lib/pf/iptables.pm
+++ b/lib/pf/iptables.pm
@@ -140,6 +140,17 @@ sub iptables_generate {
     # per interface-type pointers to pre-defined chains
     ($tags{'filter_if_src_to_chain'},$tags{'filter_forward'}) = $self->generate_filter_if_src_to_chain();
 
+    $tags{'netdata'} = "-A input-management-if --protocol tcp --match tcp -s 127.0.0.1 --dport 19999 --jump ACCEPT\n";
+
+    if ($cluster_enabled) {
+        push my @mgmt_backend, map { $_->{management_ip} } pf::cluster::config_enabled_servers();
+
+        foreach my $mgmt_back (uniq(@mgmt_backend)) {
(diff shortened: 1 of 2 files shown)


lib/pf/iptables.pm
```diff
@@ -140,6 +141,17 @@ sub iptables_generate {
...
+    $tags{'netdata'} = "-A input-management-if --protocol tcp --match tcp -s 127.0.0.1 --dport 19999 --jump ACCEPT\n";
+
+    if ($cluster_enabled) {
+        push my @mgmt_backend, map { $_->{management_ip} } pf::cluster::config_enabled_servers();
+
+        foreach my $mgmt_back (uniq(@mgmt_backend)) {
+            $tags{'netdata'} .= "-A input-management-if --protocol tcp --match tcp -s $mgmt_back --dport 19999 --jump ACCEPT\n";
+        }
+    }
+    $tags{'netdata'} .= "-A input-management-if --protocol tcp --match tcp --dport 19999 --jump DROP\n";
```
JeGoi, line 153: Here the idea is to drop everything that is not related to a management IP or localhost, correct?

JeGoi (approved):
Just one question to clarified the drop, but not an issue.
Then it is fine for me.

lib/pf/iptables.pm
fdurand: Yes, since we reverse proxy the netdata interface, only ip addresses from the management and localhost needs to have access.

fdurand:
BUMP

PR #6315, opened 2021-04-22
fix(docs): remove deduped html/images files

cgx:
# Description

Instead of copying the HTML files of the documentation to a different directory, expose the source directory of the documentation from Caddy.

# Delete branch after merge

NO

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,14 +48,14 @@ docs/index.js: $(HTML)
 images:
 	@echo "install images dir and all subdirectories"
 	for subdir in `find docs/images/* -type d -printf "%f\n"` ; do \
-		install -d -m0755 $(DESTDIR)/usr/local/pf/html/pfappserver/root/doc/images/$$subdir ; \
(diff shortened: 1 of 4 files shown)


nqb (changes requested):
- I suggest we keep `images` target in `Makefile` even if it's only used by Debian and we continue to call Makefile from `debian/rules`. Paths in `images` target need to be updated to reflect that change.
- At the moment, on a Debian 9 installation, `/usr/local/pf/docs` and `/usr/local/pf/docs/images` are already installed by `packetfence` package, not by `packetfence-doc` package. Certainly because we take almost all content of git repository to create Debian `packetfence `package, see [debian/rules](https://github.com/inverse-inc/packetfence/blob/dd502b8d38551f87884fa9485b86953c553ed8a6/debian/rules#L53-L58).

cgx:
@nqb I pushed the changes.

nqb (approved):

PR #6340, opened 2021-04-30
Fix/ntlm auth cache

### LINKED ISSUE #6341: NTLM Auth cache(Debian): wrong permissions for /usr/local/pf/addons/AD/secretsdump.py
**Describe the bug**
File is not executable.

See #6340

### LINKED ISSUE #6342: NTLM Auth cache: ldapsearch used in get_sync_samaccountname function is wrong
**Describe the bug**
[LDAP search uses to find sAMAccountName of user to pass to `secretsdump.py`](https://github.com/inverse-inc/packetfence/blob/02b824046aaf21e770c79613b5a3e7bad87562bc/lib/pf/domain/ntlm_cache.pm#L210-L212) script is wrong at two places:
- binddn should be basedn, fixed by #6340 
- search is not limited to username defined in AD source used for NTLM Auth cache. So LDAP query returns all results that match the filter.

I'm not able to understand how this feature can work at the moment.

nqb:
# Description
Refresh NTLM auth cache integration on Debian 11 and EL systems.

# Impacts
NTLM Auth cache

# Issue
fixes #6341
fixes #6342 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests


nqb:
Instead of trying to fix Python errors for `addons/AD/secretsdump.py`, I decided to use `secretsdump.py` script provided by `python3-impacket` package on Debian 11 and EL8. They both provided `impacket` at version 0.9.22

I updated RPM package on inverse.ca, package provided is now: `python3-impacket-0.9.22-4.el8.noarch.rpm`

julsemaan:
> Instead of trying to fix Python errors for `addons/AD/secretsdump.py`, I decided to use `secretsdump.py` script provided by `python3-impacket` package on Debian 11 and EL8. They both provided `impacket` at version 0.9.22
> 
> I updated RPM package on inverse.ca, package provided is now: `python3-impacket-0.9.22-4.el8.noarch.rpm`

I'm happy to go with this but from memory some adjustments were made to the initial implementation of secretsdump.py

These were (still going from memory) related to the background job. We should make sure that the background job still works or deprecate it (it doesn't scale anyway)

Most importantly, as discussed during our group meeting, we should make sure that cache on connection works

Should we decide to deprecate the background job, we'll need an upgrade note (not script, just note) and strip it off the admin for v11. We should also open an issue to strip off the backend code post v11 since it wouldn't be wise to touch this at this point of the QA phase

nqb:
100% agree with your proposal @julsemaan. I built a package, I will test cache on connection to be sure it works as expected.

lib/pf/domain/ntlm_cache.pm
```diff
@@ -188,36 +194,6 @@ sub fetch_all_valid_hashes {
 
 }
 
-=head2 get_sync_samaccountname
```
fdurand, line 191: We will have to keep that, in the AD authentication source it´s suppose to be the bindn (even if it works with the samaccountname).

nqb: Most of the time, the Bind DN is defined as "user@domain.tld" or "user".

But I agree that it's possible to see something like "CN=Username,OU=Users,DC=domain,DC=tld".

Can we just not tell (with a note) to NTLM auth users to use a sAMAccountName as Bind DN in the source ? If they would like to use NTLM Auth cache, it means that we will query an AD and sAMAccountName should work.

I try to remove the function because it didn't work

nqb: Solved after a discussion with @fdurand

nqb:
Tests:

### With a DN as bind DN defined on AD source or with a DN as sAMAccountName defined on AD source
- ✔️  EL8 standalone, NTLM cache on connection
- ❎  EL8 standalone, NTLM cache background job (fetch all): `-usersfile` is certainly the custom option we added to `secretsdump.py` and this is why it failed
- ✔️  EL8 standalone, NTLM cache background job + NTLM cache background job individual fetch

nqb:
I added the upgrade note @julsemaan. Feel free to modify it. Because we will remove option from GUI, I told users to edit file by hand.

@satkunas, could you hide or remove "NTLM cache background job" (only this option) from NTLM Auth cache section ?

julsemaan:
> I added the upgrade note @julsemaan. Feel free to modify it. Because we will remove option from GUI, I told users to edit file by hand.

Perhaps I'm blind, I'm not seeing this in this PR or in devel and its not in the PR description
My 2 cents on the notes: We should simply state the deprecation, we can leave the parameter in domain.conf but just make sure that its presence will be ignored by our backend. Adding an upgrade script at this stage is kind of risky since we're currently testing the upgrade process so its not the time to introduce a regression

satkunas:
@nqb done

nqb:
> Perhaps I'm blind, I'm not seeing this in this PR or in devel and its not in the PR description

With a `git push`, it's always better

nqb:
> My 2 cents on the notes: We should simply state the deprecation, we can leave the parameter in domain.conf but just make sure that its presence will be ignored by our backend.

Agree on that, my approach was different.

Issue #5858, opened 2020-09-29
network should be restart after step 1 of configurator


nqb:
**Is your feature request related to a problem? Please describe.**
When you configure network interfaces during Step 1 of configurator, PacketFence will start to manage network interfaces using network-scripts on CentOS and ifup/down and Debian.

It means that if NetworkManager (or another network manager daemon) managed your network settings before Step 1 of configurator, network changes will only occurs after:
* network has been restarted
or 
* interfaces modified have been restarted
or
* a reboot

**Describe the solution you'd like**
Restart network when we click on next step during Step 1 of configurator.

**Describe alternatives you've considered**
- Restart only interfaces modified during Step 1 of configurator
- Warn users they should restart after they finish configurator (network will be restarted)

julsemaan:
I believe that our code declares the interfaces as unmanaged by NM so even if NM is enabled and started, it should all work.

Is there a specific problem that this attempts to solve ?

nqb:
> I believe that our code declares the interfaces as unmanaged by NM so even if NM is enabled and started, it should all work.

On my side, interfaces are configured in PF, declared as unmanaged by NM but there is no IP addresses configured on interfaces until I restarted network service or reboot. A restart of NM should also do the trick.

julsemaan:
Can you provide an example to replicate. 

When I install a fresh CentOS without touching NM I'm unable to replicate this

nqb:
I will provide a way to replicate.

nqb:
Hello @julsemaan,

In order to replicate, you should:
- install a fresh CentOS with NM and three interfaces in DHCP
- install PacketFence on this server without rebooting
- configure first interface as management and second one as registration with static IP addresses during Step 1 of configurator
- leave third interface as is

On my side, until NM is restarted, I got following behavior:
- `nmcli` show all interfaces as managed by NM even if `NM_CONTROLLED=no` is set in each` /etc/sysconfig/network-scripts/ifcfg- file`
- IP address has been applied on management and registration interface
- Third interface is still in DHCP and NM try to get an IP address for this card each 5 minute.

Before I restarted NM:
```
# nmcli
[..]
eth2: connected to eth2
        "Red Hat Virtio"
        ethernet (virtio_net), 52:54:00:3D:A9:9B, hw, mtu 1500
        inet4 172.17.2.2/24
        route4 172.17.2.0/24

eth3: connected to eth3
        "Red Hat Virtio"
[...]

PR #6390, opened 2021-06-08
[tests] ensure libvirt interface is correctly set up after configurator

nqb:
# Description
On Debian, switch from systemd-networkd to ifdown doesn't seems to work. So interface was still in DHCP even after configurator.

Related to #5858 

# Impacts
Tests

# Delete branch after merge
YES


nqb:
`systemd-networkd` is not used anymore in Vagrant images provided by Debian project. Because we will move these images to support Debian 11, I'm closing this issue.

PR #6421, opened 2021-07-06
fix/dup-logs-username-portal

jrouzierinverse:
Description
===========
Avoid duplicate log entries.

Impacts
=======
PacketFence::DynamicRouting::Module

NEWS file entries
=======================
Bug Fixes
-------------
* Avoid duplicate log entries "User <username> has authenticated on the portal".

Delete branch after merge
=========================
NO


nqb:
I'm backporting this fix to maintenance (10.3 to 10.0)

PR #6407, opened 2021-06-24
Build PacketFence for RHEL8 and Debian 11

nqb:
# Description

### RHEL8
Build PacketFence packages on CentOS 8
Adapt CI to build and upload PacketFence packages for RHEL8

### Debian 11
Build PacketFence packages on Debian 11
Adapt CI to build and upload PacketFence packages for Debian 11

### Common
Prepare build environment to build package with all CPAN modules in CI

# Code / PR Dependencies
https://github.com/fingerbank/perl-client/pull/50

# Delete branch after merge
NO

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests
- [ ] Add acceptance tests (TestLink)

# NEWS file entries
## New Features
* RHEL8 support

Issue #6092, opened 2021-01-12
MariaDB triggers are removed when doing a restore from backup


nqb:
**Describe the bug**
If you do a complete restore of your DB using a backup taken with `./addons/backup-and-maintenance.sh`, triggers are removed by restore.

**To Reproduce**
1. PacketFence server up and running with a recent DB backup
2. Apply following process:
```
mysql -e "drop database pf;"
mysql -e "create database pf;"
mysql pf < /usr/local/pf/db/pf-schema.sql 
mysql pf -e "show triggers;"
```

=> Triggers are present.

3. Restore your DB backup using:
```
mysql pf < /root/backup/packetfence-db-dump-2021-01-12_15h09.sql 
mysql pf -e "show triggers;"
```

=> Triggers have been removed.

**Expected behavior**
Triggers should be present after a restore.

**Additional context**
Only tested on a standalone environment

After some tests, triggers are not present in SQL file generated by backup script certainly because `pf` user doesn't have `TRIGGER` [privilege](https://mariadb.com/kb/en/show-triggers/). Using `root`, TRIGGERS are included into SQL file.

nqb:
I can confirm that missing TRIGGER privilege is the root cause of the issue.

nqb:
I suppose it's caused by the fact that when restoring, we destroy tables if they exists and triggers are certainly destroy too.

nqb:
If you take a dump as `root` triggers are present but when you restore there is a conflict because triggers are part of `pf-schema.sql `

I suggest to remove triggers from backup.

julsemaan:
The thing with that is that if you import the schema and then import the dump, it drops the tables before recreating them which drops the triggers too (at least on MariaDB 10.5)

The import/export scripts have workaround to cope with this, the solution to this issue may just be that we'll have proper import/export scripts in the future

nqb:
I updated documentation to avoid confusion between backups taken as MariaDB user `root` or `pf`.

All backups should now be done using backup script and should not contain triggers and procedures (because they are made by  MariaDB user `pf`). Import script has been adjusted by @julsemaan to handle both cases.

PR #6438, opened 2021-07-16
Feature/full import

julsemaan:
# Description
Create a tool to export and import the PacketFence configuration and database

# Impacts
Its a standalone tool, so not much unless you use it, then it impacts everything!

# Code / PR Dependencies
depends on #6407 

# Delete branch after merge
NO

# Checklist
(REQUIRED) - [yes, no or n/a]
- [WIP] Document the feature
- [N/A] Add unit tests

# NEWS file entries
## New Features
* Added a tool to import and export the configuration and database for 10.3.0 and above


julsemaan:
This is not ready in the sense that it needs more testing and perhaps be merged in devel instead of feature/rhel8 but the code is mostly done unless I missed some stuff to restore

addons/full-import/README.md
```diff
@@ -0,0 +1,85 @@
+
+This is a quick readme to describe the export/import process. We should formalize this in our official documentation.
+
+== Assumptions and limitations
+
+ * You can export on any PacketFence version above 10.3.0
+ * You can import on any PacketFence version above 11.0.0
+ * The import process needs to be runned on a standalone server. Restoring directly to clusters is currently unsupported
```
extrafu: runned -> run

julsemaan: I thought it was runnaide

addons/full-import/README.md
```diff
@@ -0,0 +1,85 @@
...
+  * NOTE: Once you restored to your standalone server, you can make it a cluster by joining other machines to it and creating your cluster.conf but this is relatively advanced and out of scope of this document
+ * Restoring on a fresh install of PacketFence is recommended although restoring on an existing instance can work but your milleage may vary
+ * This process will only restore the files that can be edited via the admin interface which include:
+  * Standard configuration files in /usr/local/pf/conf/*.conf
+  * Connection profiles html templates in /usr/local/pf/html/captive-portal/profile-templates/
```
extrafu: How about if someone uploaded a jpeg/png file (or something else) to be used on his portal?

addons/full-import/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+# Replace 10.3-11.0 with the current stable and current devel version
+cp db/upgrade-X.X-X.Y.sql db/upgrade-10.3-11.0.sql
+```
+
+== Start the export process
+
+NOTE: When you are in a cluster, you need to perform this process on the first member of the incoming addresses of your database cluster. To find the member, run `show status like 'wsrep_incoming_addresses';` inside your MariaDB instance and the first IP will be the one where you need to perform the export process 
+
+If the nightly export done at 0h30 everyday is fine and you don't need the latest data, then you can script this step. Otherwise to have the latest data and configuration in your export, run:
```
extrafu: script -> skip

addons/full-import/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Next, you need to obtain a copy of the full-import tool (can currently be found here: https://github.com/inverse-inc/packetfence/tree/feature/full-import/addons/full-import). This needs to be installed in /usr/local/pf/addons/full-import/
+
+Next, run the export script:
+
+```
+/usr/local/pf/addons/export.sh /tmp/export.tgz
+```
+
+The command above will create your export archive in /tmp/export.tgz. You will now need to copy this file to your new servers via scp or your prefered mechanism
```
extrafu: via scp -> using scp

addons/full-import/README.md
julsemaan: This reminds me I need to grab the file for the `logo` variable of profiles.conf 

But for files referenced in the HTML templates, its pretty hard to track them accurately

I believe that users can use a static/ directory inside their profile-templates/$ID directory which will be migrated through this process.

It may be a bit complex to track assets that are referenced in the HTML template since they won't contain the full path and may be a bit tricky to extract

extrafu: I think we can limit ourself to the logo in profiles.conf for now

addons/full-import/export.sh
```diff
@@ -0,0 +1,81 @@
...
+fi
+
+files_dump=`find /root/backup/ -name 'packetfence-files-dump-*' -mtime -1 | tail -1`
+
+if [ -z "$files_dump" ]; then
+  echo "Unable to find a files dump that was done in the last 24 hours. Add --force to ignore this."
+  exit 1
+fi
+
+build_dir=`mktemp -d`
```
nqb, line 28: I suggest to use `export_dir` in place of `build_dir`

nqb (changes requested):
- The file `addons/full-import/find-upgrade-path.pl` doesn't seem to be used anymore.
- We need to be sure packaging is correctly handled for v11 (I can do it and push commits directly to that branch)
- I always used: `set -o nounset -o pipefail -o errexit` at beginning of Bash scripts as a best practice. I think we should at least add `set -o errexit` to be sure any shell scripts will stop in case of an error which is not handle by script itself (like a full filesystem or a permission issue).

julsemaan:
This PR now points on devel

We should plan on testing this tool and then integrating the readme inside our official docs

addons/full-import/export.sh
```diff
@@ -0,0 +1,83 @@
...
+set -o nounset -o pipefail -o errexit
+
+source /usr/local/pf/addons/full-import/helpers.functions
+source /usr/local/pf/addons/full-import/database.functions
+
+output="$1"
+
+if [ -z "$output" ]; then
+  echo "Missing output parameter"
+  echo "Usage export.sh /path/to/export.tgz [--force]"
```
nqb: I don't think the `--force` argument is handle in `export.sh`

julsemaan: Fixed in 78684d0a6a4e8b0577945bc67529734ff64a8aee

julsemaan:
Rebased against devel, should be all good for testing @nqb

addons/full-import/import.sh
```diff
@@ -0,0 +1,119 @@
...
+echo "Stopping PacketFence services"
+systemctl cat monit >/dev/null 2>&1 && (systemctl stop monit ; systemctl disable monit)
+systemctl isolate packetfence-base
+check_code $?
+systemctl stop packetfence-galera-autofix
+check_code $?
+
+if echo "$db_dump" | grep '\.sql$' >/dev/null; then
+  echo "The database dump uses mysqldump"
+  #TODO /tmp/grants.sql should be included in the export
```
nqb, line 67: I think this one can be removed

nqb (changes requested):

addons/full-import/import.sh
```diff
@@ -0,0 +1,119 @@
...
+check_code $?
+db_dump=`ls packetfence-db-dump-*`
+echo "Uncompressed database dump '$db_dump'"
+
+main_splitter
+echo "Stopping PacketFence services"
+systemctl cat monit >/dev/null 2>&1 && (systemctl stop monit ; systemctl disable monit)
+systemctl isolate packetfence-base
+check_code $?
+systemctl stop packetfence-galera-autofix
```
nqb, line 62: Is it necessary to stop it here because we say in README that we don't support restoring on a cluster ?

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+  sub_splitter
+  echo "Replacing create statements from the dump and removing drop statements"
+  # This is done so that tables aren't dropped 
+  sed -i "s/DROP TABLE IF EXISTS /DELETE FROM /g" $dump_file
+  sed -i "s/CREATE TABLE /CREATE TABLE IF NOT EXISTS /g" $dump_file
+
+  sub_splitter
+  echo "Importing $dump_file into $db_name"
+  mysql ${mariadb_args} $db_name < $dump_file
+	check_code $?
```
nqb: Indentation looks wrong

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+  
+  sub_splitter
+  echo "Importing grants"
+  mysql ${mariadb_args} < $grants_file
+	check_code $?
```
nqb: Indentation looks wrong

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function import_mariabackup() {
+  if ! which qpress > /dev/null 2>&1; then
+    echo "Missing qpress command. Please read the documentation of this script in the Installation guide to prepare your environment before running this script"
+    exit 1
+  fi
+
+  if ! which mariabackup > /dev/null 2>&1; then
+    echo "Missing qpress command. Please read the documentation of this script in the Installation guide to prepare your environment before running this script"
```
nqb:
```suggestion
    echo "Missing mariabackup command. Please read the documentation of this script in the Installation guide to prepare your environment before running this script"
```

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+
+  sub_splitter
+  echo "Starting MariaDB"
+  systemctl start packetfence-mariadb
+  check_code $?
+
+  sub_splitter
+  echo "Performing upgrade of MariaDB. Enter the MariaDB root password if prompted to"
+  echo "NOTE: The root password you must enter is the one of the server of which you're reimporting the data, not this server's MariaDB root password"
+  mysql_upgrade -p
```
nqb, line 126: It looks like `mysql_upgrade` is only done when you did an export on a system which has `mariabackup`. However, standalone installations running 10.3 (without `mariabackup`) need to have their DB upgraded too using this tool.

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+function upgrade_database() {
+  if ! test_db_connection_no_creds; then
+    echo -n "Please enter the root password for MariaDB:"
+    read -s mariadb_root_pass
+    echo
+    mariadb_args="$mariadb_args -p$mariadb_root_pass"
+  fi
+
+  db_name="$1"
+  db_version=`mysql -p $mariadb_args $db_name -e "select version from pf_version order by id desc limit 1;" | egrep -o '^[0-9]+\.[0-9]+'`
```
nqb:
```suggestion
  db_version=`mysql $mariadb_args $db_name -e "select version from pf_version order by id desc limit 1;" | egrep -o '^[0-9]+\.[0-9]+'`  
```

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+
+  sub_splitter
+  echo "Performing upgrade of MariaDB. Enter the MariaDB root password if prompted to"
+  echo "NOTE: The root password you must enter is the one of the server of which you're reimporting the data, not this server's MariaDB root password"
+  mysql_upgrade -p
+
+  popd
+}
+
+function db_upgrade_scripts() {
```
nqb: `get_db_upgrade_scripts` ?

addons/full-import/configuration.functions
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
+
+function config_files() {
```
nqb: `get_config_files` ?

nqb (changes requested):
- import: I suggest to run a `fixpermissions` at end of process:
  * to be compliant with what we do at the moment when restarting PacketFence services
  * to be sure permissions are not different (because we restore files and we change owner)

addons/full-import/export.sh
nqb: I create a new package called `packetfence-export` for v10.3 (Deb and RPM).

This package is a standalone package not dependent to `packetfence` package. It should only be installed on PacketFence installations before 11.0.

This package is built on each commit on devel (can be improve later) and uploaded automatically to Debian and RPM repositories on inverse.ca. A copy to a friendly name like for `packetfence-release` package is done to:
- https://www.packetfence.org/downloads/PacketFence/RHEL8/
- https://www.packetfence.org/downloads/PacketFence/debian/

For 11.0 and next releases, `full-import` directory is part of `packetfence` package.

nqb:
I create a new package called `packetfence-export` for v10.3 (Deb and RPM).

This package is a standalone package not dependent to `packetfence` package. It should only be installed on PacketFence installations before 11.0.

This package is built on each commit on devel (can be improve later) and uploaded automatically to Debian and RPM repositories on inverse.ca. A copy to a friendly name like for `packetfence-release` package is done to:
- https://www.packetfence.org/downloads/PacketFence/RHEL8/
- https://www.packetfence.org/downloads/PacketFence/debian/

For 11.0 and next releases, `full-import` directory is part of `packetfence` package.

addons/full-import/export.sh
nqb: If you run `export.sh filename` with `filename` a name of a file (without path), it works but I think `filename` is removed at end of process. If somebody used `export.sh --force /path/to/export.tgz`, you got this issue.

nqb: Also when using correctly, the `--force` flag doesn't seem to operate as expected because export script continue to copy files from `/root/backup` in place of doing new backups.

julsemaan: The `--force` flag actually accepts backups older than 24 hours for cases where your current PF install is broken

In the instructions, we'll state that you must take a backup before running this tool

addons/full-import/import.sh
julsemaan: I'm scared that someone would do it on a cluster they started building and then galera autofix starts doing stuff with the other nodes. Unlikely but its a safety net

addons/full-import/database.functions
julsemaan: When you restore a mysqldump on the server, it already runs the right version of MariaDB and all the tables are recreated so I don't think there is a need to upgrade the MariaDB files since only raw SQL is reimported, not the actual table files

That is unless I'm missing something else mysql_upgrade is doing

addons/full-import/export.sh
nqb: Sorry I misunderstood usage of `--force`

addons/full-import/import.sh
nqb: I understand, let's keep it.

addons/full-import/database.functions
nqb: Ok, I will check that.

addons/full-import/database.functions
```diff
@@ -0,0 +1,168 @@
...
+  current_version="$1"
+  current_version=`echo $current_version | sed 's/\./\\./g'`
+  db_scripts=`ls /usr/local/pf/db/upgrade-* | egrep '^/usr/local/pf/db/upgrade-[0-9]+\.[0-9]+(\.[0-9]+)?-[0-9]+\.[0-9]+(\.[0-9]+)?\.sql' | sort --version-sort | sed -ne '/'$current_version'/{:a' -e 'n;p;ba' -e '}'`
```
nqb, line 135: Could you add a comment to describe what is doing this one-liner ? It's pretty hard to understand and to maintain.

addons/full-import/configuration.functions
```diff
@@ -0,0 +1,159 @@
...
+  perl -MConfig::IniFiles -I/usr/local/pf/lib_perl/lib/perl5/ -e "print join(\"\n\", Config::IniFiles->new( -file => '/usr/local/pf/conf/pf.conf')->GroupMembers('interface')) . \"\n\"" | sed 's/^interface //g'
+}
+
+function get_os_ip_address() {
+  ip a show dev $1 | egrep -o 'inet [0-9]+\.[0-9]+\.[0-9]+\.[0-9]+' | head -1 | sed 's/^inet //g'
+}
+
+function get_os_netmask() {
+  os_ip=`ip a show dev $1 | grep inet | sed -r 's#.+inet ([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/[0-9]+).+#\1#' | head -1`
+  ipcalc -m $os_ip | sed 's/NETMASK=//'
```
nqb: `ipcalc` should be added has a dependency for EL and Debian.

`ipcalc` is packaged in bullseye, seems also ok on EL.

EDIT: it looks `ipcalc` is already present on a ZEN image (CentOS7) but seems to be missing from SPEC.

addons/full-import/configuration.functions
```diff
@@ -0,0 +1,159 @@
...
+    fi
+  done
+}
+
+function list_pf_interfaces() {
+  perl -MConfig::IniFiles -I/usr/local/pf/lib_perl/lib/perl5/ -e "print join(\"\n\", Config::IniFiles->new( -file => '/usr/local/pf/conf/pf.conf')->GroupMembers('interface')) . \"\n\"" | sed 's/^interface //g'
+}
+
+function get_os_ip_address() {
+  ip a show dev $1 | egrep -o 'inet [0-9]+\.[0-9]+\.[0-9]+\.[0-9]+' | head -1 | sed 's/^inet //g'
```
nqb:
```
# ip -br -o a show dev eth0
eth0             UP             10.0.2.15/24 fe80::a00:27ff:fecb:8dc8/64
```
It could simplify to avoid usage of regex.

addons/full-import/configuration.functions
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function get_os_netmask() {
+  os_ip=`ip a show dev $1 | grep inet | sed -r 's#.+inet ([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/[0-9]+).+#\1#' | head -1`
```
nqb: Idem as above

addons/full-import/helpers.functions
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+function check_code() {
+  if [ $1 -ne 0 ];then
+    if ! [ -z "$2" ]; then
```
nqb:
```suggestion
    if [ -n "$2" ]; then
```

addons/full-import/configuration.functions
```diff
@@ -0,0 +1,159 @@
...
+    else
+      stop_it=""
+      while [ -z "$stop_it" ]; do
+        echo "Cannot find interface $interface on the system. This interface is defined as '`get_pf_interface_type $interface`' in the configuration"
+        sub_splitter
+        ip a show | sed -r 's/^([0-9])/Option \1/'
+        sub_splitter
+        echo -n "Please select a new interface to replace $interface from the list above or press enter to delete the interface from pf.conf: "
+        read new_int_choice
+        if ! [ -z "$new_int_choice" ]; then
```
nqb:
```suggestion
        if [ -n "$new_int_choice" ]; then
```

addons/full-import/import.sh
```diff
@@ -0,0 +1,119 @@
...
+main_splitter
+handle_network_change
+check_code $?
+
+main_splitter
+upgrade_configuration
+check_code $?
+
+main_splitter
+restore_profile_templates
```
nqb: Is it on purpose to restore profile templates after upgrade_configuration ?

In the past, a script like `./addons/upgrade/to-8.3-conf-latin1-to-utf8.sh` modify these files.

addons/full-import/configuration.functions
```diff
@@ -0,0 +1,159 @@
...
+            delete_interface $interface
+            stop_it="yes"
+          fi
+        fi
+      done
+    fi
+  done
+}
+
+function configuration_upgrade_scripts() {
```
nqb:
```suggestion
function get_configuration_upgrade_scripts() {
```

addons/full-import/import.sh
```diff
@@ -0,0 +1,119 @@
...
+main_splitter
+echo "Restoring certificates"
+restore_certificates
+check_code $?
+
+main_splitter
+echo "Completed import of the database and the configuration! Complete any necessary adjustments and restart PacketFence"
+
+# Done with everything, time to cleanup!
+systemctl cat monit > /dev/null 2>&1 && systemctl enable monit
```
nqb, line 132: Is it necessary to revert `systemctl isolate` ?

nqb (changes requested):

addons/full-import/database.functions
nqb: Same goes for `versions_since` in `helper.functions`

addons/full-import/configuration.functions
nqb: I missed to update calls to this function.

addons/full-import/import.sh
julsemaan: If we do so we start PacketFence and the user may want to do last minute tweaks to his system before starting PF so that's why I'm leaving everything off

addons/full-import/import.sh
julsemaan: Good point, lets move it

addons/full-import/configuration.functions
julsemaan: ipcalc is part of the base OS, should we still add it to the spec ?

addons/full-import/import.sh
nqb: Ok let's keep as-is.

addons/full-import/configuration.functions
nqb: Necessary on Debian.

julsemaan:
> * import: I suggest to run a `fixpermissions` at end of process:
>   
>   * to be compliant with what we do at the moment when restarting PacketFence services
>   * to be sure permissions are not different (because we restore files and we change owner)

Done

addons/full-import/configuration.functions
julsemaan: Added to both SPEC

nqb (changes requested):
- If I try to restore PF 10.3 export (done on a cluster of EL7) on a 11.0 (Debian 11), I need to install `mariadb-backup` package bundled with PF 10.3 version on Debian 11. However, it conflicts with mariadb-package currently installed with PF 11.0.

EDIT: I propose to use usual [upgrade instructions](https://www.packetfence.org/doc/PacketFence_Upgrade_Guide.html#_database_backup) to take DB backup on 10.3 even if `mariabackup` is better to avoid locks. We can add a note in README to suggest doing export out of business hours. However, export will certainly be done during a maintenance windows so users have been warned about possible disruptions.

I look at installing `mariabackup-10.5` on a 10.3 setup but [MariaDB documentation](https://mariadb.com/kb/en/mariabackup-overview/#compatibility-of-mariabackup-releases-with-mariadb-server-releases) explains that backward compatibility is not always possible.

EDIT2: https://github.com/inverse-inc/packetfence/issues/6092 is another reason to run a manual db_dump using usual upgrade instructions (in place of using backups created by `backup-and-maintenance.sh` script)

nqb:
I prepare a commit to remove requirement on `qpress` because it doesn't seem to be mandatory.

On both backups taken on Deb 9 and EL7 with mariabackup, I don't see any qp files after I used `mbstream`.

On top of that, there is no mention of such requirement in [official Mariabackup documentation.](https://mariadb.com/kb/en/full-backup-and-restore-with-mariabackup/)

addons/full-import/import.sh
```diff
@@ -0,0 +1,129 @@
...
+check_code $?
+
+main_splitter
+
+echo "Extracting archive..."
+tar -xf export.tgz
+check_code $?
+
+echo "Found the following content in the archive:"
+ls -l
```
nqb: We should exclude "export.tgz", otherwise, `ls -l` displays `export.tgz` and you think  `export.tgz` is part of the archive.

nqb (changes requested):

addons/full-import/configuration.functions
nqb: There is no option `-m` for `ipcalc` on Debian 11 (ipcalc 0.5, pkg: ipcalc 0.42-2)

julsemaan: I'll spin up my Debian to figure out a solution

nqb:
In my export, I have three interfaces with corresponding IP addresses
- eth0
- eth1
- eth2

I do my import on a system which has same identifier for interfaces but eth2 is down

=> Import stop without any error.

Caused by return code of: 
```
# ip -br -o a show dev eth2
# echo $?
130
```
Trace:
```
+ echo 'Found interface eth2 in pf.conf'
Found interface eth2 in pf.conf
+ ip a show dev eth2
+ handle_interface_exists eth2
+ interface=eth2
++ get_os_ip_address eth2
++ ip -br -o a show dev eth2
++ head -1
++ egrep -o '[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+'
+ os_ip=

```

nqb:
This is a side effect of `set -o errexit`

nqb:
@fdurand, @JeGoi and @lzammit, any feedback from your tests ?

extrafu:
Documentation looks good.

nqb:
All looks good to me, I'm ready to merge unless someone want to add something.

fdurand:
you have my blessing to merge it

PR #6548, opened 2021-08-25
Add win10 node in CI environment

nqb:
# Description
Add a Windows 10 node in CI environment

Current state:
- No provisioning using Ansible

TODO:
- need to create a new switch because request come from swp16 (MAC differents from swp12 interface)

Improvements:
- join machine to domain using Ansible


nqb:
Need a rebase and adjustments to reflech CI environment changes.
Would be useful for tests before 11.2 release.

PR #6566, opened 2021-09-07
Fix issues with integration tests

nqb:
# Description
- Delete VM at end of tests (to avoid space issue)
- Ensure interface are up on Debian before doing configuration
- Remove useless things

# Impacts
Integration tests


nqb:
I will backport to maintenance/11.0

Issue #6570, opened 2021-09-07
Google Workspace LDAP authentication not working on PF 11.0


nathanbloemendaal78:
1. Configure the Google Workspace LDAP source as described, including certificates
2. Configure basics like switches, roles, realms, enough to be able to test auth with a client device
3. Perform authentication using EAP-TTLS /w PAP (I tested a random Chromebook and Android phone)
4. Audit log shows error: 

_Reason Config item "realm[null].oauth2.discovery" does not exist_

radius debug shows:

(7) Tue Sep  7 20:55:01 2021: Debug: server packetfence-tunnel {
(7) Tue Sep  7 20:55:01 2021: Debug:   # Executing section authorize from file /usr/local/pf/raddb/sites-enabled/packetfence-tunnel
(7) Tue Sep  7 20:55:01 2021: Debug:     authorize {
*
*
*
(7) Tue Sep  7 20:55:01 2021: Debug: suffix: Authentication realm is LOCAL
(7) Tue Sep  7 20:55:01 2021: Debug:       [suffix] = ok
(7) Tue Sep  7 20:55:01 2021: Debug: ntdomain: Request already has destination realm set.  Ignoring
(7) Tue Sep  7 20:55:01 2021: Debug:       [ntdomain] = noop
(7) Tue Sep  7 20:55:01 2021: Debug:       update control {
(7) Tue Sep  7 20:55:01 2021: Debug:       } # update control = noop
(7) Tue Sep  7 20:55:01 2021: Debug: eap: No EAP-Message, not doing EAP
(7) Tue Sep  7 20:55:01 2021: Debug:       [eap] = noop
(7) Tue Sep  7 20:55:01 2021: Debug:       policy oauth2.authorize {
(7) Tue Sep  7 20:55:01 2021: Debug:         if (&Realm && &User-Password && "%{config:realm[%{Realm}].oauth2.discovery}") {
(7) Tue Sep  7 20:55:01 2021: Debug:         EXPAND realm[****.nl].oauth2.discovery
(7) Tue Sep  7 20:55:01 2021: Debug:            --> realm[****.nl].oauth2.discovery
(7) Tue Sep  7 20:55:01 2021: ERROR:         Config item "****.nl].oauth2.discovery" does not exist
(7) Tue Sep  7 20:55:01 2021: Debug:         EXPAND %{config:realm[%{Realm}].oauth2.discovery}
(7) Tue Sep  7 20:55:01 2021: Debug:            --> 
[...]

PR #6572, opened 2021-09-09
multiple fixes to LDAP EAP-TTLS PAP

### LINKED ISSUE #6571: LDAP TTLS PAP broken
**Describe the bug**
Looks like the LDAP TTLS PAP implementation is broken since our 10.2 release
The field was renamed to `ldap_source_ttls_pap` in the realms but wasn't adjusted in `lib/pf/services/manager/radiusd_child.pm` so it never does anything and actually prevents radiusd from starting when enabled

**To Reproduce**
1. Configure an LDAP ttls PAP source in the NULL realm
2. Restart radiusd

julsemaan:
# Description
A bunch of fixes for EAP TTLS PAP (see news below)

# Impacts
All of EAP-TTLS PAP

# Issue
fixes #6571
starts addressing #6570 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Match the realm more strictly when its not a regex in EAP-TTLS PAP
* Populate the LDAP config for enabled LDAP EAP-TTLS PAP realms
* Only call oauth2 in authorize for the realms that have an Azure AD EAP-TTLS PAP configuration
* Use source username in LDAP module for EAP-TTLS PAP instead of always using sAMAccoutName
* Support LDAP certificate client auth for LDAP EAP-TTLS PAP authentication
* Allow to use Google Workspace LDAP sources in EAP-TTLS PAP authentication

### DIFF
--- a/lib/pf/services/manager/radiusd_child.pm
+++ b/lib/pf/services/manager/radiusd_child.pm
@@ -731,7 +731,7 @@ sub generate_radiusd_ldap {
     foreach my $ldap (keys %ConfigAuthenticationLdap) {
         my $active = $FALSE;
         foreach my $realm ( @pf::config::ConfigOrderedRealm ) {
-            if (defined($pf::config::ConfigRealm{$realm}->{ldap_source}) && ($pf::config::ConfigRealm{$realm}->{ldap_source} eq $ldap) ) {
+            if (defined($pf::config::ConfigRealm{$realm}->{ldap_source_ttls_pap}) && ($pf::config::ConfigRealm{$realm}->{ldap_source_ttls_pap} eq $ldap) ) {
                 $active = $TRUE;
             }
         }
@@ -789,7 +789,7 @@ $edir_options
     }
     user {
         base_dn = "\${..base_dn}"
-        filter = "(&(|$searchattributes(sAMAccountName=%{%{Stripped-User-Name}:-%{User-Name}}))$append)"
+        filter = "(&(|$searchattributes($ConfigAuthenticationLdap{$ldap}->{usernameattribute}=%{%{Stripped-User-Name}:-%{User-Name}}))$append)"
     }
     options {
         chase_referrals = yes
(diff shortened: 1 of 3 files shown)


lib/pf/services/manager/radiusd_child.pm
```diff
@@ -731,7 +731,7 @@ sub generate_radiusd_ldap {
     foreach my $ldap (keys %ConfigAuthenticationLdap) {
         my $active = $FALSE;
         foreach my $realm ( @pf::config::ConfigOrderedRealm ) {
-            if (defined($pf::config::ConfigRealm{$realm}->{ldap_source}) && ($pf::config::ConfigRealm{$realm}->{ldap_source} eq $ldap) ) {
+            if (defined($pf::config::ConfigRealm{$realm}->{ldap_source_ttls_pap}) && ($pf::config::ConfigRealm{$realm}->{ldap_source_ttls_pap} eq $ldap) ) {
```
fdurand: If i am not wrong you should keep the ldap_source too since we use them for "search attributes"

julsemaan: Yes, you're right, I missed that

julsemaan:
@fdurand, can you take another look and see if its fine

I think this will have to go in the maintenance too since there were a lot of regressions on EAP TTLS PAP in 11.x
Right now, only Azure AD EAP TTLS PAP is working, any LDAP based TTLS PAP is broken

lib/pf/services/manager/radiusd_child.pm
```diff
@@ -799,18 +802,29 @@ $edir_options
         start = 0
     }
 EOT
+
+        my $client_auth = "";
```
fdurand, line 806: tab

lib/pf/services/manager/radiusd_child.pm
```diff
@@ -799,18 +802,29 @@ $edir_options
...
+	if($ConfigAuthenticationLdap{$ldap}{client_cert_file} && $ConfigAuthenticationLdap{$ldap}{client_key_file}) {
+            $client_auth = <<"EOT";
+       certificate_file = $ConfigAuthenticationLdap{$ldap}{client_cert_file}
+       private_key_file = $ConfigAuthenticationLdap{$ldap}{client_key_file}
+EOT
+	}
```
fdurand: tab

lib/pf/services/manager/radiusd_child.pm
julsemaan: replaisaide

lib/pf/services/manager/radiusd_child.pm
julsemaan: replaisaide

fdurand:
The logic looks to be ok on my side.

julsemaan:
I did extensive tests with my LDAP directory and this seems to hold up and I tested Azure AD as well in another realm and it all works fine

Unless we want this tested by another team member, we'd be ready to merge and get this into the maintenance/11.0 branch

fdurand:
i am ok to merge on my side.

PR #6573, opened 2021-09-10
Tests: run tests separately using Ansible as orchestrator

nqb:
# Description
- Drop `venom-wrapper.sh` to run Venom test suites using Ansible
  * Vagrant is used to start VM and do initial provisioning with Ansible
  * Provisioning related to test is done using Ansible
  * Test execution is done using Ansible
- Change structure of `t/venom` to run tests more easily
  * Each scenario has a dedicated directory with a specific Ansible configuration
- First draft of export/import scenario
- Upgrade Venom to `1.0.0`

# Impacts
Unit and integration tests

# Delete branch after merge
YES


nqb:
I need to fix a few things to avoid false positive during tests and this PR will be ready to be merged.
I'm running a new pipeline and I will analyze issues after.

PR #6574, opened 2021-09-13
Tests: other VM are created only once

nqb:
# Description
To speed up tests and to avoid failed tests caused by instability of Vagrant and/or Ansible Galaxy, we create other VM only once and use Ansible provisioner to change configuration between each test.

We install Ansible roles and collections only one time.

# Impacts
Integration tests

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [x] Document the feature


nqb:
It works as expected.

I'm merging that one and I will backport to maintenance/11.0 to improve stability of integration tests.

PR #6586, opened 2021-09-17
Allow to use the sAMAccountName from the searchattributes in MSCHAP machine authentication

julsemaan:
# Description
Allow to use the sAMAccountName from the searchattributes in MSCHAP machine authentication

# Impacts
802.1x EAP PEAP MSCHAPv2 machine auth

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests

# NEWS file entries
## Enhancements
* Allow to use the sAMAccountName from the searchattributes in MSCHAP machine authentication

### DIFF
--- a/conf/radiusd/mschap.conf.example
+++ b/conf/radiusd/mschap.conf.example
@@ -386,7 +386,7 @@ mschap chrooted_mschap_machine {
 
         require_strong = yes
         ntlm_auth = "/usr/bin/sudo /usr/sbin/chroot /chroots/%{PacketFence-Domain} /usr/local/pf/bin/ntlm_auth_wrapper -p %%statsd_port%% -- \
-             --request-nt-key --username=%{mschap:User-Name:-None} --challenge=%{mschap:Challenge:-00} --nt-response=%{mschap:NT-Response:-00} %{PacketFence-NTLMv2-Only}"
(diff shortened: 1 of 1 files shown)


julsemaan:
@fdurand, could you see if its safe to backport this in the maintenance?

From my perspective it is but I don't want to break anything

fdurand:
Is it to fix the windows 10 machine pre-login / port-login ?
It it´s that it make completely sense.

julsemaan:
This fix is there to fix the following scenario

1. We have a computer named: `laptop-zammitgrowshaft`
2. This name is 22 chars long
3. When joining it to the domain, the servicePrincipalName is `host/laptop-zammitgrowshaft.zammitcorp.com`
4. Its sAMAccountName though is: `laptop-zammitgr$` due to the character limit of the field
5. When performing machine auth, the computer uses `host/laptop-zammitgrowshaft.zammitcorp.com`
6. FreeRADIUS translates this to the sAMAccountName `laptop-zammitgrowshaft$` and this gets rejected by the AD

So it looks like FreeRADIUS isn't limiting the amount of chars of the sAMAccountName when creating it based on the servicePrincipalName. With this PR, you'd be able to use the custom search attributes to translate the servicePrincipalName to the sAMAccountName which would bypass FreeRADIUS' broken logic.

We should also try to patch the mschap module in FreeRADIUS itself to handle this but it also makes sense to allow for custom search attributes while doing machine auth IMO to keep things consistent between user and machine auth

PR #6589, opened 2021-09-20
Change behavior pf::condition::not_equal to always succeed when match value is undef

jrouzierinverse:
# Description

Change behavior of ```pf::condition::not_equal``` to always succeed when match value is undef.

# Delete branch after merging
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries

## Enhancements
* Change the behavior of ```pf::condition::not_equal``` to always succeed when match value is undef.

# UPGRADE file entries
If any condition for filters (VLAN, RADIUS, Switch, DNS, DHCP, and Profile) uses a ```not equals`` operator.
Check if the logic is still ok if the value is null/undef.

If the filter needs to ensure that the value is defined you would need to add an additional defined condition to that filter.

### DIFF
--- a/t/unittest/condition/not_equal.t
+++ b/t/unittest/condition/not_equal.t
@@ -1,46 +1,37 @@
-package pf::condition::not_equals;
+#!/usr/bin/perl
+
 =head1 NAME
 
-pf::condition::not_equals
+Tests for pf::condition::not_equals
 
 =cut
 
 =head1 DESCRIPTION
 
-pf::condition::not_equals
+Tests for pf::condition::not_equals
 
 =cut
 
 use strict;
 use warnings;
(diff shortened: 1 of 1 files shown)


julsemaan:
The upgrade note is not clear on what the actual change is about and what to expect from the new behavior

extrafu:
@jrouzierinverse Please rephrase the implications of the change.

extrafu:
Bump

extrafu:
@jrouzierinverse Make sure you adjust this PR before noon so it gets into 11.2

jrouzierinverse:
@julsemaan check the upgrade instructions.

PR #6591, opened 2021-09-20
Fixes for #6588

### LINKED ISSUE #6588: galera-autofix: Race condition kills cluster when autofix kicks in during a restore
**Describe the bug**
If galera-autofix kicks in while the cluster is performing an SST, it kills the current primary mysqld process to elect it in --force-new-cluster but the other nodes may not be killed to trash the state which results in the --force-new-cluster node to attempt to connect to the existing cluster that was created prior to it being restarted

**To Reproduce**
1. Stop cluster in order: 3, 2, 1
2. Start node 1 (minute:0)
4. Start node 2 and 3 at minute 8
5. Node 1, 2 and 3 are creating their cluster without the help of galera-autofix at minute 9
6. Galera autofix kicks in at minute 10
7. Galera autofix determines node 1 has the lastest data and kills mysqld and starts it in force-new-cluster
8. Nodes 2 and 3 haven't been killed yet because they didn't reach minute 10 so they stay alive and broadcast the old state of the cluster so the cluster becomes partitioned
9. Node 1 starts up but stays in NON-PRMARY which means the DB is started but is not healthy so its unable to sync its state to the rest of the cluster
10. On top of the DB being unhealthy the seqno reporting gets broken on node 1 because it doesn't read the live seqno which means the DB is stuck broken forever until a manual intervention is made

julsemaan:
# Description
Fixes for #6588

# Impacts
Galera autofix

# Issue
fixes #6588 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Ensure galera-autofix can report sequence numbers when the DB is online in non-primary
* Have galera-autofix do a last minute check for the DB being online before booting in new cluster mode

### DIFF
--- a/go/cmd/galera-autofix/main.go
+++ b/go/cmd/galera-autofix/main.go
@@ -90,8 +90,8 @@ func seqnoReporting(ctx context.Context) {
 	servers := pfconfigdriver.AllClusterServers{}
 	for {
 		seqno := mariadb.DefaultSeqno
-		if mariadb.IsLocalDBAvailable(ctx) {
-			seqno = getRecordLiveSeqno(ctx)
+		if liveSeqNo := mariadb.GetLiveSeqNo(ctx); liveSeqNo != DefaultSeqno {
+			seqno = liveSeqNo
 		} else {
 			var err error
 			seqno, err = mariadb.GetColdSeqno(ctx)
--- a/go/cmd/galera-autofix/main.go
+++ b/go/cmd/galera-autofix/main.go
@@ -90,7 +90,7 @@ func seqnoReporting(ctx context.Context) {
 	servers := pfconfigdriver.AllClusterServers{}
 	for {
 		seqno := mariadb.DefaultSeqno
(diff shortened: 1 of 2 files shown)


julsemaan:
The fix should be complete, I did QA on my end in a cluster to ensure it works

We should have a good code review on this and potentially additional QA but I tried to keep the change small to reduce the risk of regressions

julsemaan:
I've omitted `When a node starts in --force-new-cluster mode, it needs to instruct the other nodes to kill mysqld` from this fix since it doesn't seem to be necessary and added potential complexity/side effects to galera-autofix in general

We will see if that is necessary once this hits the field, worst that can happen is that the DB takes a few more minutes to recover

jrouzierinverse:
Is there a way to create an automated test for this?

julsemaan:
> Is there a way to create an automated test for this?

Clusters aren't automated yet in the CI so its not possible to create test for them yet

fdurand:
The code and logic looks ok for me.

PR #6592, opened 2021-09-20
Feature/sql mode defaults

jrouzierinverse:
# Description
Fixed pf::dal to work in the default MariaDB sql_mode

# Impacts
Data Access Layer

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries

## Enhancements
* Improve the Data Access Layer to work in MariaDB's default sql_mode.

### DIFF
--- a/addons/dev-helpers/bin/generator-data-access-layer.pl
+++ b/addons/dev-helpers/bin/generator-data-access-layer.pl
@@ -37,6 +37,8 @@ my $schema = $OPTIONS{schema};
 system("mysql -u\"$dbuser\" -p\"$dbpass\" $db_name < $schema");
 $dbh->do("USE $db_name;") or $dbh->errstr;
 my $tables = table_data($dbh, $db_name, @ARGV);
+#print Dumper($tables);
+#exit;
 my $output_path = "$PF_DIR/lib/pf/dal";
 my $tt = Template->new({
     OUTPUT_PATH  => "$PF_DIR/lib/pf/dal/",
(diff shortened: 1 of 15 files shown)


addons/dev-helpers/bin/generator-data-access-layer.pl
```diff
@@ -37,6 +37,8 @@ =head1 DESCRIPTION
 system("mysql -u\"$dbuser\" -p\"$dbpass\" $db_name < $schema");
 $dbh->do("USE $db_name;") or $dbh->errstr;
 my $tables = table_data($dbh, $db_name, @ARGV);
+#print Dumper($tables);
```
julsemaan: remove this

lib/pf/dal/_bandwidth_accounting.pm
```diff
@@ -57,7 +57,6 @@ BEGIN {
         out_bytes => '',
         mac => '',
         tenant_id => '',
-        last_updated => 'current_timestamp()',
```
julsemaan, line 60: Why are all the date fields being removed?

Does that mean we can't use the DAL for them anymore or it doesn't change anything to remove them from there?

jrouzierinverse: This for date fields that has the default values set to the current time.
All this means is you cannot insert a value into the date.
However, you can still update and view them.

PR #6600, opened 2021-09-24
Be more precise in auth match when finding access for CLI login

### LINKED ISSUE #6349: CLI authentication rules matching doesn't filter on the rules action
**Describe the bug**
When you match the rules for a VPN user, it doesn't seem to be filtering out the rules that only have a "Mark as sponsor" action. So if you match the first rule that is a sponsor rule and the second that is for CLI access, then it gives you the "Mark as sponsor" from the first rule even though its not accurate in that context

Issue seems here:
https://github.com/inverse-inc/packetfence/blob/devel/lib/pf/radius.pm#L951

Solution would be to do something similar to the following:
```
/usr/local/pf/lib/captiveportal/PacketFence/DynamicRouting/Module/Authentication/Sponsor.pm:    my $value = pf::authentication::match( \@sources, { email => $sponsor_email, 'rule_class' => $Rules::ADMIN , 'context' => $pf::constants::realm::NO_CONTEXT}, $Actions::MARK_AS_SPONSOR );
```

Temporary workaround is to put all the CLI rules at the top and the other rules at the bottom

julsemaan:
# Description
Be more precise in auth match when finding access for CLI login

# Impacts
CLI login

# Issue
fixes #6349 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Ignore 'Mark as sponsor' administration rules when finding the access level of a VPN/CLI user (#6349)

### DIFF
--- a/lib/pf/authentication.pm
+++ b/lib/pf/authentication.pm
@@ -408,6 +408,8 @@ sub match2 {
         $params->{'rule_class'} = pf::Authentication::Rule->meta->get_attribute('class')->default;
         $logger->warn("Calling match with empty/invalid rule class. Defaulting to '" . $params->{'rule_class'} . "'");
     }
+    
+    my $desired_action = delete $params->{action};
 
     #TODO: evaluate if we want to die or have another behavior
     my $context = $params->{'context'} // die "No authentication context provided";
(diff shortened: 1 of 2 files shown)


julsemaan:
I suggest we do not backport this to the maintenance since this alters the behavior of administration rules matching in CLI login so users that were previously not allowed could be allowed to login via that change.

fdurand:
rebase is needed.

julsemaan:
🍚  💋 zed

PR #6656, opened 2021-10-25
have a monit check to ensure winbindd is still connected

### LINKED ISSUE #6655: 11.1 - AD failover doesn't work
**Describe the bug**
AD failover doesn't seem to work because when the active connected DC fails, winbindd doesn't attempt to find a new DC 

**To Reproduce**
1. Execute test PF-164 in testlink

julsemaan:
# Description
have a monit check to ensure winbindd is still connected

# Impacts
Monit configuration + samba/winbindd

# Issue
fixes #6655 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [X] Document the feature
- [N/A] Add unit tests

# NEWS file entries
## Enhancements
* Added a monit check to ensure winbindd is still connected (#6655)

### DIFF
--- a/addons/monit/monit_checks_configurations/00_packetfence.tt
+++ b/addons/monit/monit_checks_configurations/00_packetfence.tt
@@ -217,7 +217,7 @@ CHECK PROCESS packetfence-winbind-[% domain %] MATCHING "/usr/sbin/winbindd -s /
     if changed ppid then alert
     depends on packetfence-winbind
 
-CHECK PROGRAM packetfence-ntlm-[% domain %] with path "/usr/bin/timeout 5 /usr/sbin/chroot /chroots/[% domain %]/ /usr/bin/ntlm_auth --username=df3e129a722a865cc3539b4e69507bad --password="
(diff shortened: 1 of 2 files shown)


julsemaan:
That's not my preferred way of fixing this but I couldn't find a samba issue related to this and I'm running the latest samba on my test machine where I'm experiencing this.

At least with this we can catch this and any other potential issue out of our control with winbindd

extrafu:
I am fine with that. But should we ship it disabled by default?

nqb:
https://github.com/inverse-inc/packetfence-monitoring-scripts/pull/29

julsemaan:
I'll look at `wbinfo -t`. It looks like it will do the trick.

julsemaan:
I changed it to `wbinfo -P` since `wbinfo -t` would fail for a longer time during winbindd's startup which would trigger a never ending restart loop. `wbinfo -P` just checks that the current selected DC is connectable which is exactly what we need.

I just tested it again and it works as expected on my side, ready to merge if everyone is happy with this

PR #6669, opened 2021-10-28
Completes the cleanup of the Apache filters

### LINKED ISSUE #6652: Empty out apache_filters.conf.defaults to prevent redirection issues
**Describe the bug**
The apache filters used to be in the dispatcher code path but now that its in Golang they aren't called in that context anymore, only when the HTTP call is on the portal. I'm working on an issue where they are causing problems with a Samsung device (surprise, surprise!) so removing them completely would make sense to me

julsemaan:
# Description
Completes the cleanup of the Apache filters

# Impacts
Packaging since I cleared apache filters files from it

# Issue
fixes #6652


jrouzierinverse:
```coding-style.t``` fails because of ```lib/pf/UnifiedApi/Controller/Config/Certificates.pm```

julsemaan:
> `coding-style.t` fails because of `lib/pf/UnifiedApi/Controller/Config/Certificates.pm`

I'm not sure this is related to this PR since it doesn't touch that file

jrouzierinverse:
> > `coding-style.t` fails because of `lib/pf/UnifiedApi/Controller/Config/Certificates.pm`
> 
> I'm not sure this is related to this PR since it doesn't touch that file

This was fix in devel already please ignore.

PR #6670, opened 2021-10-28
fix/6645

### LINKED ISSUE #6645: Attempting to release endpoint after security event even if its unreg
I've always found it odd and thought it was fixed but when you are jailed in the isolation network due to an open security event and you hit 'release', if you're unreg, it attempts to detect the network access on the release page instead of bringing you back to the portal (you're unreg so you won't have internet)

When a node is unreg and is releasing from a security event, it should go to /captive-portal, otherwise nothing changes from the current behavior (goes to /access)

julsemaan:
# Description
If node is not reg after security event, go to portal instead of going to /access which has no change of succeeding and will result in an error

# Impacts
Security events portal flow

# Issue
fixes #6645 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* If a node is not registered after clearing its security event, it goes to the captive portal instead of attempting to enable the access

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/SecurityEvent.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/SecurityEvent.pm
@@ -9,6 +9,7 @@ use pf::log;
 use pf::web;
 use pf::node;
 use pf::util;
+use pf::constants::node qw($STATUS_REGISTERED $STATUS_UNREGISTERED);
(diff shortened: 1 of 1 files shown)


jrouzierinverse:
Looks good to me.
When do you want to merge it?

julsemaan:
We can start merging 11.2 PRs now so feel free to merge if it looks good to you

PR #6684, opened 2021-11-05
RADIUS Role attribute for VPN

fdurand:
# Description
If the VPN support role then return it in the radius reply

# Impacts
RADIUS VPN Flow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Support for roles in VPN access


lib/pf/Switch/Fortinet/FortiGate.pm
```diff
@@ -221,7 +221,7 @@ sub identifyConnectionType {
     my @require = qw(Fortinet-Vdom-Name);
     my @found = grep {exists $radius_request->{$_}} @require;
 
-    if (@require == @found) {
+    if ( (@require == @found) || (exists $radius_request->{'Connect-Info'} && $radius_request->{'Connect-Info'} eq 'vpn-ssl') ) {
```
lyubomirtraykov: It wold be nice to include $radius_request->{'Connect-Info'} eq 'vpn-ikev2' as well

lib/pf/Switch/Fortinet/FortiGate.pm
```diff
@@ -221,7 +221,7 @@ sub identifyConnectionType {
     my @require = qw(Fortinet-Vdom-Name);
     my @found = grep {exists $radius_request->{$_}} @require;
```
lyubomirtraykov, line 234: I checked and  Fortinet-Vdom-Name is present in CLI login and VPN login. The only difference between CLI and VPN is Connect-Info. For CLI Connect-Info = admin-login, for VPN Connect-Info = vpn-ssl | vpn-ikev2

lyubomirtraykov:
@fdurand place consider that Fortinet-Vdom-Name attribute is present in CLI login. The Connect-Info attribute value for CLI login is admin-login.

fdurand:
> @fdurand place consider that Fortinet-Vdom-Name attribute is present in CLI login. The Connect-Info attribute value for CLI login is admin-login.

I just forgot to push my commit ...

lib/pf/Switch/Fortinet/FortiGate.pm
```diff
@@ -247,7 +251,25 @@ sub returnAuthorizeVPN {
     my $kick = $self->handleRadiusDeny($args);
     return $kick if (defined($kick));
 
-    my $node = $args->{'node_info'};
+    my $role;
+    if ( isenabled($self->{_RoleMap}) && $self->supportsRoleBasedEnforcement()) {
```
julsemaan: I have one potential concern.

Let's say you do CLI login authentication with a Catalyst switch and RoleMap is enabled, its going to return a Filter-Id for the CLI login from what I can tell.

Shouldn't we have a supportsVPNRoleBasedEnforcement method to handle this case specifically?

Or perhaps just let each switch module handle this in its returnRadiusVpn function?

fdurand: yes good point

fdurand: I have the VpnMap now so it´s an explicit switch config now.

lib/pf/Switch/Fortinet/FortiGate.pm
```diff
@@ -48,7 +48,7 @@ use pf::SwitchSupports qw(
     WirelessMacAuth
     WiredMacAuth
     WirelessDot1x
-    RoleBasedEnforcement
```
lyubomirtraykov, line 51: RoleBasedEnforcement should be supported as well because of WSSO. https://docs.fortinet.com/document/fortiap/6.4.0/fortiwifi-and-fortiap-cookbook/414919/wifi-with-wsso-using-windows-nps-and-user-groups

cree8:
I know this has been closed and merged for a few months already, but thus far I still see the same behavior playing out during VPN authentication. 
Once a user has been successfully authenticated via FortiClient, the RADIUS-reply is as follows
`
RADIUS Reply
MS-CHAP2-Success = 0x6a533d34353830304645304636434434464133453131393239423935413245343933433533453135364539
REST-HTTP-Status-Code = 200`

I have not been able to get the system to return anything else in the reply message no matter what setting is changed. 

For completion's sake, here is the RADIUS-request:
`RADIUS Request
User-Name = "testuser1@puffer.com"
NAS-IP-Address = 10.2.112.25
NAS-Port = 2
Framed-IP-Address = 2.2.2.2
Called-Station-Id = "10.2.112.25"
Calling-Station-Id = "2.2.2.2"
NAS-Identifier = "Puffer-Cluster"
NAS-Port-Type = Virtual
Acct-Session-Id = "22b98268"
Event-Timestamp = "Jul  8 2022 10:34:07 CDT"
Connect-Info = "vpn-ssl"
Fortinet-Vdom-Name = "root"
MS-CHAP-Challenge = 0x22927b05bf89405362ef46b361480157
MS-CHAP2-Response = 0x6a009ea6c1dd037b53b5dce74e8bb044b6dd00000000000000006782a226f93dc0b582112cc7e150838952fc4065093a86ed
Stripped-User-Name = "testuser1"
Realm = "puffer.com"
FreeRADIUS-Client-IP-Address = 10.2.112.25
PacketFence-Domain = "PufferAD"
PacketFence-KeyBalanced = "a0d4593b1ce77d5055254b23ec9ed02a"
PacketFence-Radius-Ip = "10.2.120.20"
PacketFence-NTLMv2-Only = ""
User-Password = "******"
SQL-User-Name = "testuser1@puffer.com"
`

PR #6687, opened 2021-11-08
pfpki: workaround for Windows devices configured through Intune

nqb:
# Description
Reply to Windows devices configured through Intune even if they requested a non-existing URL.
Not configurable on Intune side.

# Impacts
pfpki

# Delete branch after merge
YES

# NEWS file entries
## Bug fixes
* Reply to Windows devices configured through Intune even if they requested a non-existing URL.


nqb:
Waiting a feedback to merge.

PR #6690, opened 2021-11-09
Add RADIUS audit log entry in correct tenant when switches are defined by MAC address

### LINKED ISSUE #6540: packetfence-set-tenant-id ignores switchs defined by MAC address
**Describe the bug**
In a multi-tenant installation, when PacketFence receives a RADIUS request for a switch defined by MAC address, RADIUS audit log entry is added in default tenant in place of tenant configured on switch.

**To Reproduce**
1. Create a new tenant
2. Define a switch by MAC address attached to new tenant
3. Send a RADIUS request with 
* `Called-Station-Id` equals to MAC Address of switch
* `NAS-IP-Address` not defined

=> Node is created in correct tenant but RADIUS log entry is created in default tenant

**Expected behavior**
RADIUS audit log entry is created in correct tenant.

**Additional context**
Because the RADIUS request doesn't contain `NAS-IP-Address` value, our FreeRADIUS configuration will set `NAS-IP-Address` to `PacketFence-Src-IP-Address` (source IP address of packet). In order to assign tenant ID to RADIUS request, FreeRADIUS will lookup into `radius_nas` table using `PacketFence-Src-IP-Address`, because there is no entry, it fallbacks to tenant ID 1

nqb:
# Description
Add RADIUS audit log entry in correct tenant when switches are defined by MAC address

# Impacts
FreeRADIUS tenant management

# Issue
fixes #6540

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Add RADIUS audit log entry in correct tenant when switches are defined by MAC address

### DIFF
--- a/raddb/policy.d/packetfence
+++ b/raddb/policy.d/packetfence
@@ -50,10 +50,16 @@ packetfence-set-tenant-id {
         }
     }
     if ( "%{%{control:PacketFence-Tenant-Id}:-0}" == "0") {
-        update control {
-            &PacketFence-Tenant-Id = "%{sql: SELECT IFNULL((SELECT tenant_id FROM radius_nas WHERE nasname = '%{NAS-IP-Address}'), 0)}"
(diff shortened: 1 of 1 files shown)


nqb:
@fdurand, I'm not able to test this fix correctly because it looks like `radius_nas` table is not updated after I add a new switch or if I send a RADIUS request coming from that switch.
I'm investigating.

nqb:
EDIT: my previous test was incorrect.

I was able to test it and it works as expected when switch are defined by IP address or by MAC address.

PR #6714, opened 2021-11-23
Feature/admin cert san

### LINKED ISSUE #6689: Display Subject Alternative Name in SSL certificates menu
Subject Alternative Name attribute is almost mandatory right now for SSL certificates but we don't display it on web admin.

Display Subject Alternative Name attribute in SSL certifitcates menu.

julsemaan:
# Description
Add support for SAN in the certificates section

# Impacts
Certificates display and new CSRs that will be issued via the admin interface

# NEW Package(s) required
A new version of packetfence-perl must be built with Crypt-OpenSSL-X509-1.910.tar.gz

# Issue
fixes #6689 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [N/A] Document the feature
- [X] Add unit tests

# NEWS file entries

## Enhancements
* Certificate signing requests created via the admin interface now include a Subject Alternative Name (SAN)
* The Subject Alternative Names of a certificate are now displayed in the admin interface

### DIFF
--- a/addons/packetfence-perl/dependencies.csv
+++ b/addons/packetfence-perl/dependencies.csv
@@ -208,7 +208,7 @@ Digest::HMAC_MD5,1.04,ARODLAND/Digest-HMAC-1.04.tar.gz,True,A/AR/ARODLAND/Digest
 Switch,2.17,CHORNY/Switch-2.17.tar.gz,True,C/CH/CHORNY/Switch-2.17.tar.gz,2
 Rose::DB::Object,0.820,JSIRACUSA/Rose-DB-Object-0.820.tar.gz,True,J/JS/JSIRACUSA/Rose-DB-Object-0.820.tar.gz,2
 Data::Phrasebook::Loader::YAML,0.13,BARBIE/Data-Phrasebook-Loader-YAML-0.13.tar.gz,True,B/BA/BARBIE/Data-Phrasebook-Loader-YAML-0.13.tar.gz,2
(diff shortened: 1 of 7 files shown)


julsemaan:
I just need a bit of frontend magic from @satkunas to display the SAN cleanly instead of displaying as an array

Currently displays as:
image

But there is certainly a cleaner way to display these

nqb (commented):
@julsemaan, do you remember why we need to upgrade Crypt::OpenSSL::X509 Perl module ?

I don't see any specific notes in [Changelog](https://github.com/dsully/perl-crypt-openssl-x509/blob/master/Changes).

nqb (changes requested):
Looks like display is not correct in Certification Authority Certificates field (only):

image


@satkunas, could you take a look ?

nqb (commented):
In the code, I can see that we copied Common Name into SAN (`DNS`).

I think it's a good idea as a default behavior. Later, we could let users define their own extended attributes.

julsemaan:
> @julsemaan, do you remember why we need to upgrade Crypt::OpenSSL::X509 Perl module ?
> 
> I don't see any specific notes in [Changelog](https://github.com/dsully/perl-crypt-openssl-x509/blob/master/Changes).

Because to_string isn't implemented in the current version for extensions

nqb:
@satkunas, I will need a fix on https://github.com/inverse-inc/packetfence/pull/6714#pullrequestreview-832872045 if we want to have this PR merged before feature freeze.

nqb:
Works as expected, I will update `packetfence-perl` to include new dependency.

nqb:
Crypt::OpenSSL::X509 should be part of `packetfence-perl` 1.2.0

PR #6717, opened 2021-11-24
explicitely support APs defined by CIDR/IP for Unifi web-auth

### LINKED ISSUE #6686: Unifi: deauthentication doesn't work anymore if AP are declared by IP address
**Describe the bug**
It looks like since #6134, deauthentication of a user doesn't work if it is connected on a Unifi access point which is declared by IP address (not by MAC address).

An AP declared by IP address will not pass `if` test:
https://github.com/inverse-inc/packetfence/blob/758bce50d9fb12c7d538745452507e16448b9d93/lib/pf/Switch/Ubiquiti/Unifi.pm#L272-L277

**To Reproduce**
1. Declared a Unifi AP by IP address
2. Authenticate on captive portal using Web Auth
3. Check logs in `packetfence.log`:

```
│Nov  5 08:58:39 pf01 pfqueue[132890]: pfqueue(132890) INFO: [mac:4c:32:75:70:0e:19] Found site: Test (pf::Switch::Ubiquiti::Unifi::_deauthenticateMacWithHTTP)                                                         │
│Nov  5 08:58:39 pf01 pfqueue[132890]: pfqueue(132890) INFO: [mac:4c:32:75:70:0e:19] Deauth on 0 access points (pf::Switch::Ubiquiti::Unifi::_deauthenticateMacWithHTTP)     
```

**Expected behavior**
Deauthentication should also work on AP declared by IP address.

**Additional context**
In this setup, AP are declared using a CIDR.

julsemaan:
# Description
explicitely support APs defined by CIDR/IP for Unifi web-auth

# Impacts
Ubiquiti Unifi module

# Issue
fixes #6686 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [N/A] Document the feature
- [N/A] Add unit tests

# NEWS file entries
## Bug Fixes
* Fix regression in the Unifi module for deauthentication of webauth clients when the APs are defined using an IP or CIDR in the configuration (#6686)
*

### DIFF
--- a/lib/pf/Switch/Ubiquiti/Unifi.pm
+++ b/lib/pf/Switch/Ubiquiti/Unifi.pm
@@ -256,9 +256,9 @@ sub _deauthenticateMacWithHTTP {
         return;
     }
 
-    my $json_data = decode_json($response->decoded_content());
+    my $sites = decode_json($response->decoded_content());
 
-    foreach my $entry (@{$json_data->{'data'}}) {
+    foreach my $entry (@{$sites->{'data'}}) {
         $response = $ua->get("$base_url/api/s/$entry->{'name'}/stat/sta/$mac");
         if ($response->is_success) {
             $found = $TRUE;
@@ -268,7 +268,25 @@ sub _deauthenticateMacWithHTTP {
             last;
         }
     }
+    
+    # There are two flows of deauth that will be attempted
 
+    # First...
+    # The first one is doing the deauth on the site itself without adding the ap_mac in it which will grant access on the AP the user is currently connected on
+    # This only happens if we found where the client is connected
+    # This supports access points defined via IP or CIDR in the PF configuration
+    if ($found) {
+        $response = $ua->post("$base_url/api/s/$site_opts{'name'}/cmd/stamgr", Content => encode_json($args));
+        if ($response->is_success) {
+            $logger->info("Deauth on site: $site_opts{'desc'}");
+        }
+    }
+
+    # Second...
+    # We go through all the entries in switches.conf and find the ones that are MAC address based and part of the group of the current entry
+    # For each of these, we send the command including the AP MAC which allows to grant access to the device on all APs of that site (to enable web-auth roaming)
+    # If we don't know which site the endpoint is connected on, we send the messages to all the sites
+    # This only supports entries defined by their MAC address in switches.conf
     tie my %SwitchConfig, 'pfconfig::cached_hash', "config::Switch($host_id)";
     my $count = 0;
     foreach my $switch_id (keys(%SwitchConfig)) {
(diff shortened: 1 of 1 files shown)


julsemaan:
I've marked this as a +1 since it doesn't alter the behavior that was released in 11.1, it just adds another behavior that is opinionated on APs defined using IP or CIDR

lib/pf/Switch/Ubiquiti/Unifi.pm
```diff
@@ -292,11 +310,8 @@ sub _deauthenticateMacWithHTTP {
...
-    $logger->info("Deauth on $count access points");
-
-    unless($response->is_success) {
-        $logger->error("Can't send request on the Unifi controller: ".$response->status_line);
-        return;
+    if($count > 0) {
+    	$logger->info("Deauth on $count access points");
     }
 
     $logger->info("Switched status on the Unifi controller using command $command");
```
nqb: This sentence could be wrong if your device has not been found and your access points are defined by AP in PacketFence config.

nqb (changes requested):
Thanks for your detailled comments.

IMHO, it's a problem to have different behavior depending on how AP are declared into PacketFence configuration. 

If I correctly understand your code, in a case where AP are declared by IP addresses in PacketFence configuration and MAC address of device is not found on a Unifi site, we don't deauthenticate. However, in such scenario for AP declared by MAC addresses, we deauthenticate on all sites.

julsemaan:
> Thanks for your detailled comments.
> 
> IMHO, it's a problem to have different behavior depending on how AP are declared into PacketFence configuration.
> 
> If I correctly understand your code, in a case where AP are declared by IP addresses in PacketFence configuration and MAC address of device is not found on a Unifi site, we don't deauthenticate. However, in such scenario for AP declared by MAC addresses, we deauthenticate on all sites.

That is correct, we could make it so we send it to all sites if we don't find the device but from what I could see this wouldn't be necessary on the recent Unifi versions since we're sure to find it if its authorized or connected. This means when granting access we'll find it (since the user is connected and on the portal), and when we're revoking access, the entry should be found in a site since its currently authorized.

My understanding is that the block to deauth on all sites would be targetted at old Unifi versions but its hard to tell at the moment. We could always add the block to deauth on all sites for IP/CIDR APs, it won't cost more than me taking the time to do a copy/paste.

nqb:
> My understanding is that the block to deauth on all sites would be targetted at old Unifi versions but its hard to tell at the moment. We could always add the block to deauth on all sites for IP/CIDR APs, it won't cost more than me taking the time to do a copy/paste.

I agree.

nqb:
@julsemaan, let me know if you plan to adjust this PR.

julsemaan:
PR adjusted

lib/pf/Switch/Ubiquiti/Unifi.pm
nqb: This is still a problem for me.

nqb: I misunderstand logic. In fact, this sentence will never appear in logs if AP are defined by IP addresses.

PR #6725, opened 2021-11-26
Feature/venom fingerbank

julsemaan:
# Description
Adds a venom scenario for testing the Fingerbank invalid DB tests in testlink

It's currently missing the part of the test against the captive portal since that isn't yet available in venom AFAIK but everything else is tested and working

# Impacts
Venom tests

# Delete branch after merge
YES


t/venom/Makefile
```diff
@@ -128,10 +128,18 @@ example_el8:
 	SCENARIOS_TO_RUN=example \
 	$(MAKE_TARGET)
 
+fingerbank_invalid_db_deb11:
+	make \
+	PF_VM_NAME=pfdeb11dev \
```
nqb:
```suggestion
	PF_VM_NAME=pfdeb11$(DEV_ENV) \
```

t/venom/scenarios/fingerbank_invalid_db/site.yml
```diff
@@ -0,0 +1,6 @@
+---
+# - import_playbook: provision.yml
+
```
nqb, line 3: Call to `rsync.yml` is missing

nqb:
```suggestion
# rsync before tests when doing local development
- import_playbook: playbooks/rsync.yml
  when: lookup("env", "CI") != 'true'
```

nqb (changes requested):
1. Did you test it on EL8 ? Do you think it's worth to add support for EL8 ?
1. Is it necessary to test both wireless and wired ?
1. I saw you call ` clear_cache` after `wired_mac_auth` but not between other tests, is it normal ?
1. There is no assertions in  test suites because you only used these test suites to perform actions (like destroy table or clear cache). Adding assertions would be helpful to document what is the expected state. At the moment, I suppose that if we add a regression in Fingerbank, this scenario will fail at some point but I don't know where and why.

julsemaan:
> 1. Did you test it on EL8 ? Do you think it's worth to add support for EL8 ?

IMO doing it on both OSes will make it time consuming while running the tests so my 2 cents are that we should focus on testing the 'secondary' tests on a single OS while ensuring primary and critical tests are tested on both. Happy to do both if the team disagrees

> 2. Is it necessary to test both wireless and wired ?

Perhaps not but I wanted to get the best coverage possible since I had them for free and they were already written

> 3. I saw you call ` clear_cache` after `wired_mac_auth` but not between other tests, is it normal ?

Nope, I'll need to do that

> 4. There is no assertions in  test suites because you only used these test suites to perform actions (like destroy table or clear cache). Adding assertions would be helpful to document what is the expected state. At the moment, I suppose that if we add a regression in Fingerbank, this scenario will fail at some point but I don't know where and why.

The only assertion this needs to test is that even if the DB of Fingerbank is corrupt, the service must stay available so the assertions are actually running the MAB/802.1x/pfdhcplistener tests and ensuring they still work even with a corrupt DB. 

I can add an assertion to ensure the DB is really corrupt via a DB call in the Fingerbank lib if you deem it necessary.

nqb:
I agree with 1, 2 and 3

> The only assertion this needs to test is that even if the DB of Fingerbank is corrupt, the service must stay available so the assertions are actually running the MAB/802.1x/pfdhcplistener tests and ensuring they still work even with a corrupt DB.
> I can add an assertion to ensure the DB is really corrupt via a DB call in the Fingerbank lib if you deem it necessary.

I don't think an assertion to check if DB is really corrupt is necessary. However, are we sure that if PacketFence start to fail with a corrupted Fingerbank DB, wired and wireless tests will really failed ?

julsemaan:
> I agree with 1, 2 and 3
> 
> > The only assertion this needs to test is that even if the DB of Fingerbank is corrupt, the service must stay available so the assertions are actually running the MAB/802.1x/pfdhcplistener tests and ensuring they still work even with a corrupt DB.
> > I can add an assertion to ensure the DB is really corrupt via a DB call in the Fingerbank lib if you deem it necessary.
> 
> I don't think an assertion to check if DB is really corrupt is necessary. However, are we sure that if PacketFence start to fail with a corrupted Fingerbank DB, wired and wireless tests will really failed ?

They don't actually fail to start, they will fail to offer service because the SQL queries will error out in lib/fingerbank/. RADIUS requests will fail if that safety guard breaks. That caused a week-end outage at a dozen customers before we added that safeguard so I'm pretty sure it will break. 

At the end of the day, the main focus is to ensure it doesn't break, whether its through that safeguard or another one that could be added "by accident" in the PF code

julsemaan:
Just pushed changes, ready for another round of review

PR #6769, opened 2021-12-13
Test suites for inlinel2 and inlinel3

fdurand:
# Description
Cover test for inlinel3 and inlinel2

# Impacts
Tests

# Delete branch after merge
YES


.gitlab-ci.yml
```diff
@@ -639,6 +639,22 @@ configurator_deb11:
     - .test_script_job
     - .release_only_rules
 
+inline_deb11:
+  variables:
+    VAGRANT_PF_DOTFILE_PATH: /var/local/gitlab-runner/vagrant/vagrant-${CI_COMMIT_TAG}
+  extends:
+    - .test_job
+    - .test_script_job
+    - .release_only_rules
```
nqb:
```suggestion
    - .test_devel_branches_and_maintenance_rules
```

.gitlab-ci.yml
```diff
@@ -639,6 +639,22 @@ configurator_deb11:
...
+
+inline_el8:
+  variables:
+    VAGRANT_PF_DOTFILE_PATH: /var/local/gitlab-runner/vagrant/vagrant-${CI_COMMIT_TAG}
+  extends:
+    - .test_job
+    - .test_script_job
+    - .release_only_rules
```
nqb:
```suggestion
    - .test_devel_branches_and_maintenance_rules
```

t/venom/scenarios/inline/playbooks/configure.yml
```diff
@@ -0,0 +1,18 @@
+---
+- hosts: pfservers
+  name: configure Golang environment for unit tests
```
nqb: name is wrong

t/venom/test_suites/inline/l3/10_setup_packetfence.yml
```diff
@@ -0,0 +1,311 @@
...
+       "ifindex": "8",
+       "ipaddress": "100.64.0.2",
+       "ipv6_address": null,
+       "ipv6_prefix": null,
+       "is_running": true,
+       "master": null,
+       "name": "inlinel3-b",
+       "nat_dns": "disabled",
+       "nat_enabled": "enabled",
+       "netflow_accounting_enabled": "disabled",
```
nqb: Could be enabled if we want to check NetFlow features ?

Would require a restart of `packetfence-pfacct` below.

t/venom/test_suites/inline/l3/25_local_user_authentication.yml
```diff
@@ -0,0 +1,26 @@
+name: Local User Authentication
+testcases:
+  - name: Register device on the portal
+    steps:
+    - type: exec
+      script: ssh -o "ConnectTimeout 3" -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" "root@192.168.1.10"  -t "/root/register_node  -portal=http://100.64.0.2"
```
nqb: `/root/register_node` is a script provided directly in ulinux image ?

Can I have a look at this script ?

t/venom/test_suites/inline/l3/30_trigger_violation.yml
```diff
@@ -0,0 +1,61 @@
...
+  steps:
+  - type: http
+    method: PUT
+    url: '{{.pfserver_webadmin_url}}/api/v1/node/02:06:19:98:00:00/apply_security_event'
+    ignore_verify_ssl: true
+    body: >-
+     {
+       "id": "02:06:19:98:00:00",
+       "mac": "02:06:19:98:00:00",
+       "security_event_id": "1200003"
```
nqb: Because we want to isolate I suggest to use `Generic` security event to avoid confusion with Bandwidth expiration.

t/venom/test_suites/inline/l3/teardown/05_setup_packetfence.yml
```diff
@@ -0,0 +1,168 @@
...
+    method: DELETE
+    url: '{{.pfserver_webadmin_url}}/api/v1/config/routed_network/192.168.1.0'
+    ignore_verify_ssl: true
+    headers:
+      "Authorization": "{{.get_login_token.json.result.token}}"
+      "Content-Type": "application/json"
+    assertions:
+      - result.statuscode ShouldEqual 200
+
+- name: Configure inlineil2 as other
```
nqb:
```suggestion
- name: Deconfigure inlinel3 interface
```

t/venom/test_suites/inline/l3/teardown/05_setup_packetfence.yml
```diff
@@ -0,0 +1,168 @@
+name: Delete PacketFence configuration
```
nqb: Title of file doesn't match title of test suite

t/venom/test_suites/inline/l2/05_setup_packetfence.yml
```diff
@@ -0,0 +1,300 @@
...
+       "networks": [],
+       "not_editable": false,
+       "reg_network": null,
+       "split_network": null,
+       "tenant_id": 1,
+       "type": "inlinel2",
+       "vip": null,
+       "vlan": null,
+       "ipaddress": "192.168.2.1",
+       "netmask": "255.255.255.0"
```
nqb, line 43: NetFlow option is missing

t/venom/test_suites/inline/l2/teardown/05_setup_packetfence.yml
```diff
@@ -0,0 +1,156 @@
+name: Delete configuration in PacketFence
```
nqb: Title of file is wrong

nqb (changes requested):
I put some single comment during review, I will add here general comments related to the whole test suites:

1. [X] Could you rebase on `devel `?
1. [X] We need to have variables for each test suite defined in `vars/all.yml` 
1. [ ] When using API calls your JSON payloads contain:
   * default values: could you remove it ?
   * sometimes `id` key which is not necessary if already present in URL
1. [X] Could you create a diagram for inline L2 ?
1. [X] On diagrams, could you mention where is the PacketFence server to clarify ?
1. [/] I run that scenario on a runner and both test suites failed at `Test-if-the-device-is-in-the-ipset-role-ID-2-Guest-role`
   * Doesn't seem to be an issue anymore if we run test suites after unit tests https://gitlab.com/inverse-inc/packetfence/-/jobs/1932845215
1. [/] I and @JeGoi will move some test cases into executor to avoid duplication between both test suites
=> will be done later on a specific PR.

I will add another comment later which more specific to the logic of the scenario

nqb (changes requested):
Also I suggest to rename test cases without any spaces and uppercase letters because it create long variables (when you want to reuse output from a test case)

EDIT: DONE

t/venom/test_suites/inline/l2/05_setup_packetfence.yml
fdurand: In the case of inline l2 interface i am not sure it needs to be there. It´s more a network option.

t/venom/test_suites/inline/l3/25_local_user_authentication.yml
fdurand: it´s a golang binary i made.

```
package main

import (
	"flag"
	"fmt"
	"net/http"
	"net/url"
	"os"
	"sync"
)

type Jar struct {
	lk      sync.Mutex
	cookies map[string][]*http.Cookie
}

func NewJar() *Jar {
	jar := new(Jar)
	jar.cookies = make(map[string][]*http.Cookie)
	return jar
}

// SetCookies handles the receipt of the cookies in a reply for the
// given URL.  It may or may not choose to save the cookies, depending
// on the jar's policy and implementation.
func (jar *Jar) SetCookies(u *url.URL, cookies []*http.Cookie) {
	jar.lk.Lock()
	jar.cookies[u.Host] = cookies
	jar.lk.Unlock()
}

// Cookies returns the cookies to send in a request for the given URL.
// It is up to the implementation to honor the standard cookie use
// restrictions such as in RFC 6265.
func (jar *Jar) Cookies(u *url.URL) []*http.Cookie {
	return jar.cookies[u.Host]
}

func main() {

	var portal string
	flag.StringVar(&portal, "portal", "", "Portal URL like http://100.64.0.2")
	flag.Parse()

	if len(portal) == 0 {
		fmt.Println("Usage: register_node -portal")
		flag.PrintDefaults()
		os.Exit(1)
	}

	jar := NewJar()

	client := http.Client{nil, nil, jar, 0}

	resp, err := client.Get(portal + "/switchto/default_policy+default_registration_policy+default_login_policy")

	if err != nil {
		fmt.Println(err.Error())
		os.Exit(1)
	}

	resp.Body.Close()

	resp, err = client.PostForm(portal+"/signup", url.Values{
		"fields[username]": {"iastigmate"},
		"fields[password]": {"password"},
		"fields[aup]":      {"1"},
		"submit":           {""},
	})

	if err != nil {
		fmt.Println(err.Error())
		os.Exit(1)
	}

	resp.Body.Close()

}

File: t/venom/scenarios/inline/playbooks/configure.yml (hunk 3 above)
Comment (JeGoi) [inline t/venom/scenarios/inline/playbooks/configure.yml]:
```suggestion
  name: configure Packetfence VM
```

nqb (changes requested):
At the moment, `inline` test suites don't require other VM than `pf` VM so they could be part of `unit_tests` scenario to speed tests.

EDIT: DONE

nqb (changes requested):
Also we should create a connection profile with an `inline` filter.

EDIT: dependency to default connection profile in `/root/register_node` so we keep things like this for now.

t/venom/test_suites/inline/l3/25_local_user_authentication.yml
nqb: `password` and `user` are hard-coded in script at the moment we will have to take this in consideration if we want to use other variables.

t/venom/vars/all.yml
```diff
@@ -464,3 +464,18 @@ pfdhcplistener_multi_tenant.nodes.node11.dhcp_server: 10.10.10.1
...
+## Inline test suite specific variables
+#################################################################################
+inline_l3.node.macaddress: 02:06:19:98:00:00
+inline_l3.node.ipaddress: 192.168.1.10
+inline_l3.pf_portal: 192.168.1.1
+inline_l3.routed_networks: 192.168.1.0
+inline_l2.node.macaddress: 02:06:19:98:00:01
+inline_l2.node.ipaddress: 192.168.2.10
+inline_l2.pf_portal: 192.168.2.1
+inline_l2.routed_networks: 192.168.2.0
```
nqb: @JeGoi: This is not a routed network

nqb (changes requested):

t/venom/test_suites/inline/l2/05_setup_packetfence.yml
nqb: The option is available on the form when you edit an interface.
It is defined here, it's automatically enabled on the network.

t/venom/vars/all.yml
JeGoi: Well, if I remember correctly, I used the variable name from the first time I found it.
Here it was just after: 
```
url: '{{.pfserver_webadmin_url}}/api/v1/config/routed_networks'
```
like in: t/venom/test_suites/inline/l3/10_setup_packetfence.yml but you are right probably not the best choice.

t/venom/test_suites/inline/l2/05_setup_packetfence.yml
nqb: Let's keep like this for now.

PR #6773, opened 2021-12-14
Fix/6766

### LINKED ISSUE #6766: TLS 1.0 and 1.1 are disabled in FreeRADIUS configuration starting from PacketFence v11
**Describe the bug**
On a PF 11.1.0 (Debian 11 or EL8), I'm not able to connect (with default configuration) a supplicant which try to use TLS 1.0 to secure EAP communication. It works on PF 10.3.0 (CentOS 7).

According to [FreeRADIUS 3.0.22 changelog](https://github.com/FreeRADIUS/freeradius-server/blob/cd8cd3719193304241cea99d853d936ab70129c5/doc/ChangeLog#L109-L121), only TLS 1.2 is enabled by default (because we don't override default configuration)

On PF 11.1.0, we are using:
- FreeRADIUS 3.0.25
- OpenSSL 1.1.1k

**To Reproduce**
1. Forge a RADIUS request with following `eapol_test `configuration:
```
network={
phase1="tls_disable_tlsv1_0=0 tls_disable_tlsv1_1=1 tls_disable_tlsv1_2=1 tls_disable_tlsv1_3=1"
ssid="CORPORATE SSID"
key_mgmt=WPA-EAP
eap=TLS
identity="inverse"
ca_cert="/usr/local/pf/raddb/certs/ca.pem"
client_cert="/usr/local/pf/raddb/certs/client.pem"
private_key="/usr/local/pf/raddb/certs/client.key"
private_key_passwd="whatever"
eapol_flags=3
}
```
=> You got following message in `/usr/local/pf/logs/radius.log`: 

`Dec  6 09:45:04 pf auth[257735]: (49670) eap_peap: ERROR: (TLS) Alert write:fatal:protocol version
`

**Expected behavior**
I'm not sure what should be the expected behavior.

We don't publish anything regarding that change so I expected this to work without any changes. However, as mentioned in #5808, it's better to keep only TLS 1.2 enabled and we don't receive so much complaint about that change. 

I suggest to keep approach mentioned in #5808 (with GUI changes to adjust configuration when it's necessary) and to add documentation related to that changes (Upgrade guide).

**Additional context**
I got that issue on a macOS running Monterey.

fdurand:
# Description
Added a way to defined the tls min and max version in the tls profil

# Impacts
Eap configuration

# Issue
fixes #6766

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* TLS 1.0 is disabled in FreeRADIUS configuration starting from PacketFence v11 (#6766)


conf/radiusd/tls.conf.defaults
```diff
@@ -4,7 +4,8 @@ dh_file = ${certdir}/dh
 ca_path = ${cadir}
 cipher_list = DEFAULT
 ecdh_curve = prime256v1
-disable_tlsv1_2 = yes
+tls_min_version = 1.0
```
nqb: I suggest to stay with TLS 1.2 as `tls_min_version` like you proposed on #5808. Users who want to keep TLS 1.0 or TLS 1.1 need to changed it here.

conf/radiusd/tls.conf.defaults
```diff
@@ -13,5 +14,6 @@ dh_file = ${certdir}/dh
 ca_path = ${cadir}
 cipher_list = ALL:!EXPORT:!eNULL:!SSLv2
 ecdh_curve = prime256v1
-disable_tlsv1_2 = yes
+tls_min_version = 1.0
```
nqb: I suggest to stay with TLS 1.2 as `tls_min_version` like you proposed on #5808. Users who want to keep TLS 1.0 or TLS 1.1 need to changed it here.

html/pfappserver/lib/pfappserver/Form/Config/Radiusd/TLSProfile.pm
```diff
@@ -67,6 +78,10 @@ sub options_ocsp {
     return  map { { value => $_, label => $_ } } @{pf::ConfigStore::Radiusd::OCSPProfile->new->readAllIds};
 }
 
+sub options_tls_version {
+    return map { { value => $_, label => $_ } } qw(1.0 1.1 1.2 1.3);
```
nqb: - Is it possible to have values defined in another place ? Below, I saw that we take values from ConfigStore.
- I read some things about TLS 1.3, are we sure it makes sense to allow enable it ?

nqb (changes requested):
IMHO, we should also add notes in upgrade guide:
* TLS 1.0 and TLS 1.1 are disabled by default
* How to change default behavior 

I can take care of that.

html/pfappserver/lib/pfappserver/Form/Config/Radiusd/TLSProfile.pm
fdurand: I use constant to define the version now.
It´s a supported option in freeradius so it´s why i added it. (1.3)

nqb:
I need to write upgrade note.

nqb:
I will backport to maintenance in January.

PR #6774, opened 2021-12-14
Feature/ruckus floating devices

julsemaan:
# Description
Adds MAB floating device support to Ruckus/Brocade switches

# Impacts
Brocade switch module

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Added MAB floating device support to Ruckus/Brocade switches


julsemaan:
Still have to test this with the customer that wants the feature and it will then be ready

PR #6775, opened 2021-12-15
Feature/venom 1.0.0

nqb:
# Description
- Upgrade Venom to 1.0.0
- Refactor Venom code to use more executors

# Impacts
Tests

# Delete branch after merge
YES


t/venom/lib/customA.yml
```diff
@@ -0,0 +1,15 @@
+executor: customA
+input:
+  myArg: "world2"
+steps:
+- type: customB
+  myWord: "{{.input.myArg}}"
+  vars:
+    content:
+      from: result.foobarb
```
nqb, line 9: If I understand correctly your examples, the key point is here. 

In order to reuse a variable from executor B into executor A, you have to use `result.name_of_output_variable_in_executor_b` here.

I'm surprised by the syntax because I expected: `result.foobarB`

t/venom/test_suites/test_getlogintoken/15_testcBcA.yml
```diff
@@ -0,0 +1,28 @@
...
+    - result.foobarb ShouldContainSubstring world
+
+- name: customB_custom
+  steps:
+  - type: customB
+    myWord: "universe"
+    assertions:
+    - result.foobarb ShouldContainSubstring universe
+
+- name: customA_default
```
nqb: This is the demonstration that we can use an executor inside a test suite which is calling another executor: customA is calling customB.

For our use cases, it means that we can call `get_login_token` inside another executor (for example: `delete_node`) and  use token provided by `get_login_token` inside `delete_node` executor.

Last thing we need to confirm is the syntax to reuse a variable assigned by an executor, inside a test suite, which is calling that executor (current usage of `get_login_token` executor).

JeGoi: Exactly

t/venom/lib/customA.yml
JeGoi: This is the way used in venom, examples.
Let me try this way.

JeGoi: Ok it looks that it can work without rename/reassign it.

t/venom/lib/check_host_internet_access.yml
```diff
@@ -0,0 +1,11 @@
...
+input:
+  host: "{{.node01_mgmt_ip}}"
+  user: "{{.ssh_user}}"
+steps:
+  - type: ssh
+    host: '{{.input.host}}'
+    user: '{{.input.user}}'
+    command: |
+      cd /usr/local/pf/t/venom ; \
+      sudo /usr/local/pf/t/venom/venom-wrapper.sh test_suite//common/check_internet_access.yml
```
nqb: I think you lost path to `nodes` directory here.

addons/vagrant/inventory/group_vars/pfservers/venom.yml
```diff
@@ -1,4 +1,4 @@
 ---
-venom__version: 'v1.0.0-rc.7'
+venom__version: 'v1.0.0'
```
nqb: We will have to use latest stable version from v1.0.X

t/venom/lib/check_host_internet_access.yml
JeGoi: Not exactly.
The main idea is to remove the `node` directory and just use `executor` and `test_suites/common`. But to be discuss. It really depends how we will use executors.
It could be a good move to remove extra specific layers.

addons/vagrant/inventory/group_vars/pfservers/venom.yml
JeGoi: Updated

nqb:
As mentioned on https://github.com/ovh/venom/issues/470, now we need to sort test suites before calling `venom run` when we pass a directory as an argument.

Some workarounds added in #6573 can also be removed with Venom 1.0.0. Could be a good opportunity.

t/venom/lib/customC.yml
```diff
@@ -0,0 +1,14 @@
+executor: customC
+input:
+  myarg: { 
+    "name": "default",
+    "surname": "default2"
+  }
+steps:
+- script: echo "{{.inputi.myarg.name}}" > t.log
```
nqb:
```suggestion
- script: echo "{{.input.myarg.name}}" > t.log
```

t/venom/lib/check_host_internet_access.yml
nqb: - Full paths

t/venom/test_suites/test_venom/15_testcBcA.yml
```diff
@@ -0,0 +1,79 @@
...
+# This will fails, the variable varcustomb is in the executor and can not be called outside
+# 
+- name: customH_custom
+  steps:
+  - type: customE
+    myArg: "{{.customD_custom.result.foobard}}"
+    assertions:
+    - varcustomb ShouldContainSubstring hello
+    - custome.varcustomb ShouldContainSubstring hello
+    - result.custome.varcustomb ShouldContainSubstring hello
```
nqb: I suggest to:
* use a negative assertion: test will pass but we know what is the expected state
* update comment above

t/venom/test_suites/test_venom/20_testVenom.yml
```diff
@@ -0,0 +1,15 @@
...
+    assertions:
+    - result.foobar ShouldEqual "hello world"
+    - result.contento ShouldEqual "hello"
+
+- name: testfoobar multisteps custom
+  steps:
+  - type: foobarcustommultisteps
+    assertions:
+    - result.foobar ShouldEqual "hello world"
+    - content ShouldEqual "hello"
```
nqb: This assertion failed on my side because `content` is not visible outside executor, we should update assertion.

nqb:
```suggestion
    - content ShouldBeEmpty
```

nqb:
@JeGoi, this PR should be ready to merged.

I'm running a [pipeline ](https://gitlab.com/inverse-inc/packetfence/-/pipelines/434703192) to be sure we don't miss anything.

The only remaining thing I see is the new custom executors you want to use to call Venom on other hosts. I think we can do that later.

nqb:
New pipeline after a fix: https://gitlab.com/inverse-inc/packetfence/-/pipelines/434779376

t/venom/lib/check_host_internet_access.yml
JeGoi: Done add vars

JeGoi (approved):
Looks good to me

PR #6778, opened 2021-12-15
Hide download and email if the certificate has been signed by scep request

fdurand:
# Description
Hide download and email if the certificate has been signed by scep request

# Impacts
PKI

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Hide download and email for a certificate that has been signed by a scep request

### DIFF
--- a/db/upgrade-X.X-X.Y.sql
+++ b/db/upgrade-X.X-X.Y.sql
@@ -52,7 +52,7 @@ ALTER TABLE pki_certs
     ADD COLUMN IF NOT EXISTS `scep` BOOLEAN DEFAULT FALSE AFTER ip_addresses;
 
 \! echo "set pki_certs.scep to true if private key is empty"
-UPDATE TABLE pki_certs
+UPDATE pki_certs
     SET `scep`=1 WHERE `key` = "";
 
 \! echo "Incrementing PacketFence schema version...";
--- a/db/upgrade-X.X-X.Y.sql
+++ b/db/upgrade-X.X-X.Y.sql
@@ -51,6 +51,10 @@ DROP PROCEDURE IF EXISTS ValidateVersion;
 ALTER TABLE pki_certs
     ADD COLUMN IF NOT EXISTS `scep` BOOLEAN DEFAULT FALSE AFTER ip_addresses;
 
+\! echo "set pki_certs.scep to true if private key is empty"
+UPDATE TABLE pki_certs
+    SET `scep`=1 WHERE `key` = "";
+
 \! echo "Incrementing PacketFence schema version...";
 INSERT IGNORE INTO pf_version (id, version, created_at) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION), NOW());
 
--- a/db/upgrade-X.X-X.Y.sql
+++ b/db/upgrade-X.X-X.Y.sql
@@ -47,6 +47,9 @@ DELIMITER ;
 call ValidateVersion;
 DROP PROCEDURE IF EXISTS ValidateVersion;
 
+\! echo "altering pki_certs"
+ALTER TABLE pki_certs
+    ADD COLUMN IF NOT EXISTS `scep` BOOLEAN DEFAULT FALSE AFTER ip_addresses;
 
 \! echo "Incrementing PacketFence schema version...";
 INSERT IGNORE INTO pf_version (id, version, created_at) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION), NOW());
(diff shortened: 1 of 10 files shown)


go/caddy/pfpki/models/models.go
```diff
@@ -517,8 +518,8 @@ func (c CA) Put(cn string, crt *x509.Certificate, options ...string) error {
 	if CaDB := c.DB.First(&ca, profiledb[0].CaID).Find(&ca); CaDB.Error != nil {
 		c.DB.First(&ca)
 	}
-
-	if err := c.DB.Create(&Cert{Cn: cn, Ca: ca, CaName: ca.Cn, ProfileName: profiledb[0].Name, SerialNumber: crt.SerialNumber.String(), Mail: attributeMap["emailAddress"], StreetAddress: attributeMap["streetAddress"], Organisation: attributeMap["O"], OrganisationalUnit: attributeMap["OU"], Country: attributeMap["C"], State: attributeMap["ST"], Locality: attributeMap["L"], PostalCode: attributeMap["emailAddress"], Profile: profiledb[0], Key: "", Cert: publicKey.String(), ValidUntil: crt.NotAfter}).Error; err != nil {
+	vrai := true
+	if err := c.DB.Create(&Cert{Cn: cn, Ca: ca, CaName: ca.Cn, ProfileName: profiledb[0].Name, SerialNumber: crt.SerialNumber.String(), Mail: attributeMap["emailAddress"], StreetAddress: attributeMap["streetAddress"], Organisation: attributeMap["O"], OrganisationalUnit: attributeMap["OU"], Country: attributeMap["C"], State: attributeMap["ST"], Locality: attributeMap["L"], PostalCode: attributeMap["emailAddress"], Profile: profiledb[0], Key: "", Cert: publicKey.String(), ValidUntil: crt.NotAfter, Scep: &vrai}).Error; err != nil {
```
nqb: Can we find another name for this variable :-) ?

nqb (changes requested):

nqb (changes requested):
1. Do we really want to prevent download of SCEP certificates from web admin ?

IMHO, we should allow admin users to download SCEP certificates from web admin because web admin doesn't display all informations of certificates (OCSP URL, key usage, etc.) and if you need these informations, you will have to look into DB.

Certificates generated through SCEP are stored in DB without private keys so it's not a problem. Perhaps we can have a "Copy Certificate" button like we have for CA certificates.

2. Download button is still available when you click on a SCEP certificate

fdurand:
Just added a way to download the public key.
Also you need to recompile the pki in order to have the Download button removed from the admin gui.

nqb (changes requested):
- Existing SCEP certs will have a wrong value of `scep` field in DB. We need to update existing SCEP certs if we are able to distinguish them but I'm not sure it's possible.

nqb:
> * Existing SCEP certs will have a wrong value of `scep` field in DB. We need to update existing SCEP certs if we are able to distinguish them but I'm not sure it's possible.

I don't see an automated way to identify SCEP certificate in DB before.

EDIT: it means that "Download" button will be available for old SCEP certificates which can lead to confusion.

EDIT: fixed by 7153f0551f7d4fd0c25cfaf9269b94ad351b3d2a

nqb:
Just did new tests:
1. Download button is still available for SCEP certificates when you click on a SCEP certificate (not on the listing, only "Copy certificate")
1. SCEP filter doesn't work:
image

=> I expect to be able to select SCEP

@satkunas, could you take a look ?

satkunas:
> Just did new tests:
> 
> 1. Download button is still available for SCEP certificates when you click on a SCEP certificate (not on the listing, only "Copy certificate")
> 2. SCEP filter doesn't work:
>    image
> 
> => I expect to be able to select SCEP
> 
> @satkunas, could you take a look ?

@nqb the `scep` column has been pinned to ensure that the scep column is returned with each item in order to show/hide the download/email buttons. If this column is not showing up you'll need to clear your `pkiCerts::defaultSearch` preference.

db/upgrade-X.X-X.Y.sql
```diff
@@ -47,6 +47,13 @@ DELIMITER ;
 call ValidateVersion;
 DROP PROCEDURE IF EXISTS ValidateVersion;
 
+\! echo "altering pki_certs"
+ALTER TABLE pki_certs
+    ADD COLUMN IF NOT EXISTS `scep` BOOLEAN DEFAULT FALSE AFTER ip_addresses;
+
+\! echo "set pki_certs.scep to true if private key is empty"
+UPDATE TABLE pki_certs
```
nqb:
```suggestion
UPDATE pki_certs
```

nqb (changes requested):

nqb:
Work as expected after I remove my preferences and clean my web browser cache.

PR #6786, opened 2021-12-20
Feature/venom fingerbank proxy

julsemaan:
# Description
Test the Fingerbank proxy feature in venom

# Impacts
Venom tests

# Delete branch after merge
YES


julsemaan:
One thing I'm wondering is should this be part of the `configurator` scenario?

Currently, its in its own scenario

julsemaan:
I ran this successfully on runner2 but a pipeline is currently running to assess it works in the official CI environment: https://gitlab.com/inverse-inc/packetfence/-/pipelines/433611476

addons/vagrant/inventory/group_vars/linux_servers/tinyproxy.yml
```diff
@@ -1,3 +1,4 @@
 ---
 tinyproxy__allow:
+  - "{{ users_vars[dict_name]['networks'][0]['subnet'] }}"
```
nqb, line 3: You can remove `[1]['subnet']`, it's a mistake.

nqb:
> One thing I'm wondering is should this be part of the configurator scenario?

No, this scenario is specific and only used for the release.

IMHO, it's better to have a dedicated scenario because we boot an additional VM (`linux02 `here). I suggest to rename your scenario to something more generic related to external services provided by other VM. We will likely reused that scenario to add more tests which cover integration like syslog forwarding, RADIUS, etc.

addons/vagrant/playbooks/linux_servers/apt_update.yml
```diff
@@ -0,0 +1,9 @@
+- hosts: service_rsyslog, service_dhcpd, service_freeradius, service_tinyproxy
```
nqb:
```suggestion
- hosts: linux_servers
```

t/venom/scenarios/fingerbank_proxy/playbooks/run_tests.yml
```diff
@@ -0,0 +1,36 @@
+---
+- hosts: pfservers
+  name: Run Fingerbank Proxy tests
+  become: True
+
+  vars:
+    # put your test suites to try here
+    test_suites:
+      - configurator
+      - global_config
```
nqb: I think `global_config` is not necessary here

t/venom/test_suites/fingerbank_proxy/setup_proxy/setup_proxy.yml
```diff
@@ -0,0 +1,63 @@
...
+- name: Setup proxy in fingerbank.conf
+  steps:
+  - type: http
+    method: PATCH
+    url: '{{.pfserver_webadmin_url}}/api/v1/config/fingerbank_setting/proxy'
+    ignore_verify_ssl: true
+    body: >-
+      {
+        "host": "{{.linux02_mgmt_ip}}",
+        "port": "8888",
```
nqb: Could you put this as a variable ?

t/venom/test_suites/fingerbank_proxy/fingerbank_connectivity/fingerbank_connectivity.yml
```diff
@@ -0,0 +1,35 @@
+name: FingerbankConnectivity
```
nqb, line 1: It's not possible to use PacketFence API to validate account info and connectivity to the API ?

nqb (changes requested):
- For test cases name, could you use something like `name_of_my_testcase` like `get_login_token` in place of sentences with spaces ?
- Could you add a `TESTSUITE.md` in `fingerbank_proxy` which describe logic of the test (with custom iptables rules, test can be run offline, etc.). You can find an example [here](https://github.com/inverse-inc/packetfence/blob/4d8146825ea4b6c293d93dfc053272dcf6fc846d/t/venom/test_suites/wired_dot1x_eap_peap/TESTSUITE.md).
- An improvement could be to create a new test suite like `configurator_proxy` to configure Fingerbank proxy at this step.

t/venom/test_suites/fingerbank_proxy/fingerbank_connectivity/fingerbank_connectivity.yml
julsemaan: I wanted to stay as close to the Fingerbank library as possible in order to prevent any caching mechanism we have in PF and the layers of the Fingerbank librarires

Not that we can't, it's just that I found it closer to want we really want to test here while ensuring we have the least possible chances of hitting a cache (keeping in mind we may add other caching mechanisms later on and forget to update this suite)

nqb: I understand the idea. If we stay like this, I wonder if we can say we really test Fingerbank like a user will do.

I mean if we break API call in PacketFence, are we be able to detect it ? But I'm 99% sure API call is calling directly piece of code you add here.

julsemaan: I'm not talking about the API code in PF being broken but more that we can hit a layer of caching inside PF that will prevent to detect the failure properly if the proxy integration stops working

Making it further away from the actual collector/perl lib code gives us more chances of hitting a cache

julsemaan:
I've addressed most of the comments except the job name

@nqb, in your opinion, how should we name the scenario that is proxy centric (or linux02 centric)?

And do we rename the scenario, make entry and gitlab entry ? or just the make+gitlab entry and run multiple scenarios in that make entry?

nqb:
> @nqb, in your opinion, how should we name the scenario that is proxy centric (or linux02 centric)?

I would suggest something related to external integrations (nothing prevent us to boot `ad` VM later) . `external_integrations` as scenario name looks good to me. You need to rename gitlab entry, Makefile entry and name of directory into `scenarios` directory. You can keep your test suites as is.

julsemaan:
New pipeline with adjustments: https://gitlab.com/inverse-inc/packetfence/-/pipelines/434354693

Now we just need to state on if the strategy for testing the Fingerbank connectivity is acceptable or not

t/venom/test_suites/fingerbank_proxy/TESTSUITE.md
```diff
@@ -0,0 +1,16 @@
+# captive_portal
```
nqb:
```suggestion
# fingerbank_proxy
```

nqb:
@julsemaan, thanks for your adjustments.

I think we are good to merge based on answers you provided related to Fingerbank Connectivity.

PR #6803, opened 2021-12-30
Feature/debian installer

julsemaan:
# Description
Adds tooling to create an ISO installer for PacketFence based on Debian 11.2

# Impacts
Nothing, just adds tooling

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [N/A] Add unit tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -150,7 +150,6 @@ variables:
 # - all branches, except maintenance/X.Y and devel (web) if variable BUILD_PF_IMG_ZEN sets to yes or if CI_COMMIT_MESSAGE contains "build_pf_img_zen=yes".
 .build_pf_img_zen_devel_rules:
   rules:
-    - if: '$CI_COMMIT_REF_NAME == "feature/debian-installer"'
     - if: '$CI_COMMIT_REF_NAME == "devel" && ( $BUILD_PF_IMG_ZEN == "yes" || $CI_COMMIT_MESSAGE =~ /build_pf_img_zen=yes/ )'
     - if: '$CI_COMMIT_REF_NAME != "devel" && $CI_COMMIT_REF_NAME !~ /^maintenance\/[[:digit:]]+\.[[:digit:]]+$/ && $CI_COMMIT_TAG == null && $CI_PIPELINE_SOURCE == "web" && ( $BUILD_PF_IMG_ZEN == "yes" || $CI_COMMIT_MESSAGE =~ /build_pf_img_zen=yes/ )'
 
@@ -159,6 +158,7 @@ variables:
 # - all branches, except maintenance/X.Y and devel (web) if variable BUILD_PF_IMG_ISO sets to yes or if CI_COMMIT_MESSAGE contains "build_pf_img_iso=yes".
 .build_pf_img_iso_devel_rules:
   rules:
+    - if: '$CI_COMMIT_REF_NAME == "feature/debian-installer"'
     - if: '$CI_COMMIT_REF_NAME == "devel" && ( $BUILD_PF_IMG_ISO == "yes" || $CI_COMMIT_MESSAGE =~ /build_pf_img_iso=yes/ )'
     - if: '$CI_COMMIT_REF_NAME != "devel" && $CI_COMMIT_REF_NAME !~ /^maintenance\/[[:digit:]]+\.[[:digit:]]+$/ && $CI_COMMIT_TAG == null && $CI_PIPELINE_SOURCE == "web" && ( $BUILD_PF_IMG_ISO == "yes" || $CI_COMMIT_MESSAGE =~ /build_pf_img_ISO=yes/ )'
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -37,8 +37,8 @@ variables:
   VAGRANT_IMG_DIR: $CIDIR/packer/vagrant_img
   TESTDIR: t/venom
   UPLOAD_DIR: $CILIBDIR/upload
-  SF_REPO_URL: https://sourceforge.net/projects/packetfence/files/PacketFence%20ZEN
-  SF_REPO_URL: https://sourceforge.net/projects/packetfence/files/PacketFence%20ISO
+  SF_ZEN_REPO_URL: https://sourceforge.net/projects/packetfence/files/PacketFence%20ZEN
+  SF_ISO_REPO_URL: https://sourceforge.net/projects/packetfence/files/PacketFence%20ISO
   # env variables
   ANSIBLE_FORCE_COLOR: 1
   ANSIBLE_STDOUT_CALLBACK: yaml
@@ -355,7 +355,7 @@ variables:
   stage: build_pf_img
   environment:
     name: sourceforge
-    url: ${SF_REPO_URL}/${CI_COMMIT_REF_NAME}
+    url: ${SF_ZEN_REPO_URL}/${CI_COMMIT_REF_NAME}
   after_script:
     - make -e -C ${ZENDIR} clean
     - make -e -C ${ZENDIR} clean_cache
@@ -367,7 +367,6 @@ variables:
   stage: build_pf_img
   environment:
     name: sourceforge
-    url: ${SF_REPO_URL}/${CI_COMMIT_REF_SLUG}
   after_script:
     - make -e -C ${ISODIR} clean
   dependencies: []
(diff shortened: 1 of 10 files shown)


julsemaan:
This still needs to be integrated in our release pipeline, I tested the ISO manually and I get through the configurator

julsemaan:
This should be ready, @nqb, can you review?

You can download the ISO from here if you want to test:
https://sourceforge.net/projects/packetfence/files/PacketFence%20ISO/feature-debian-installer/

.gitlab-ci.yml
```diff
@@ -33,10 +33,12 @@ variables:
   RELEASEDIR: ${CILIBDIR}/release
   PACKERDIR: $CIDIR/packer
   ZENDIR: $CIDIR/packer/zen
+  ISODIR: $CIDIR/debian-installer
   VAGRANT_IMG_DIR: $CIDIR/packer/vagrant_img
   TESTDIR: t/venom
   UPLOAD_DIR: $CILIBDIR/upload
   SF_REPO_URL: https://sourceforge.net/projects/packetfence/files/PacketFence%20ZEN
+  SF_REPO_URL: https://sourceforge.net/projects/packetfence/files/PacketFence%20ISO
```
nqb: It will override previous line (same variable name)

ci/debian-installer/build-and-upload.sh
```diff
@@ -0,0 +1,35 @@
...
+    declare -p src_dir dst_dir
+    echo "rsync: $src_dir -> $dst_dir"
+
+    # quotes to handle space in filename
+    rsync -avz $src_dir "$dst_dir"
+}
+
+mkdir -p ${SF_RESULT_DIR}
+
+# Manifest need to be generate by hand because we modify OVF during last step
```
nqb: Can be removed

ci/debian-installer/preseed.cfg.tmpl
```diff
@@ -0,0 +1,507 @@
...
+                                                fr_FR.UTF-8, \
+                                                fr_CA.UTF-8, \
+                                                de_DE.UTF-8, \
+                                                he_IL.UTF-8, \
+                                                it_IT.UTF-8, \
+                                                nb_NO.UTF-8, \
+                                                nl_NL.UTF-8, \
+                                                pl_PL.UTF-8, \
+                                                pt_BR.UTF-8
+
```
nqb: Can be removed, see https://github.com/inverse-inc/packetfence/commit/3298704a27b444eb3319debe8bec2ea987c96ff1

ci/debian-installer/create-debian-installer.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
nqb, line 1: This script will not fail in case of error.

.gitlab-ci.yml
```diff
@@ -353,6 +363,17 @@ variables:
   tags:
     - shell
 
+.build_pf_img_iso_job:
+  stage: build_pf_img
+  environment:
+    name: sourceforge
+    url: ${SF_REPO_URL}/${CI_COMMIT_REF_SLUG}
```
nqb: This URL will be wrong for a release (tag) because we use ${CI_COMMIT_TAG} later.
This has a minimal effect

nqb (changes requested):
Minor changes but it looks good. I didn't test ZEN ISO.

General question: I compare `preseed.cfg.tmpl` with `preseed.cfg` used for ZEN. I see password is not set but I think it's to let user define it.

.gitlab-ci.yml
julsemaan: I added logic to use the tag for releases and the ref for devel related builds

julsemaan:
> Minor changes but it looks good. I didn't test ZEN ISO.
> 
> General question: I compare `preseed.cfg.tmpl` with `preseed.cfg` used for ZEN. I see password is not set but I think it's to let user define it.

That is correct, since the install is interactive in this case, I'm letting users pick their own root password instead of having to change it post-install.

I addressed your comments and rebuilt the ISO and tested it in my lab and it works.
If you want to test it, its available on sourceforge

nqb:
I tested ISO with success. I have one remaining comment: should we not disable predictable network names ? 

It's the case in ZEN image but it's more to simplify provisioning. However, I'm thinking about cluster installation where you can have different network names and you will need to disable it anyways.

Option to add to preseed should be: https://github.com/inverse-inc/packetfence/commit/78f903e917dd304c4800efd2ff0c00054fc1caca#diff-2e9669c2d68bf0daf463f91cf709f56da70b2ba792af71b0e5728f0f7ea7c670

nqb:
We will need to update packetfence.org to mention new image and add a section in Installation guide.

nqb:
Could you disable AppArmor (like in ZEN image) ? I think you just need to add `apparmor=0` after `d-i debian-installer/add-kernel-opts string`

julsemaan:
I noticed that the 'Graphical install' option ignores the preseed. I'm working on removing that option from the ISO too

julsemaan:
> I tested ISO with success. I have one remaining comment: should we not disable predictable network names ?
> 
> It's the case in ZEN image but it's more to simplify provisioning. However, I'm thinking about cluster installation where you can have different network names and you will need to disable it anyways.
> 
> Option to add to preseed should be: [78f903e#diff-2e9669c2d68bf0daf463f91cf709f56da70b2ba792af71b0e5728f0f7ea7c670](https://github.com/inverse-inc/packetfence/commit/78f903e917dd304c4800efd2ff0c00054fc1caca#diff-2e9669c2d68bf0daf463f91cf709f56da70b2ba792af71b0e5728f0f7ea7c670)

The downside with this is that all the network configuration you do during the installer is based on the 'non-predictable' interface name so when you boot into the OS, eth0 is down and not configured in /etc/network/interfaces.

nqb:
This was the issue we had with ZEN image. In that case, we can:
* keep things as is
* pass options directly to the installer to start install with correct options

julsemaan:
The installer accepts `net.ifnames=0` and combined with it in the preseed kernel options we're all good. So the interfaces can be configured using the installer and will keep predictable names both in the installer and after the install

I retested an ISO and it works but haven't republished since I'd need to realter the .gitlab-ci file. I can provide an ISO to test if needed.

Once we're happy with this, I'll write the steps in our install guide to get this installed

nqb:
Tested again, work as expected.
@julsemaan, I will let you adjust documentation.

julsemaan:
> Tested again, work as expected. @julsemaan, I will let you adjust documentation.

Just did the instructions, went pretty generic so that it survives over time

@extrafu, if you want to also check the instructions for verbiage and grammar

docs/installation/installation.asciidoc
```diff
@@ -54,6 +54,27 @@ Make sure that there is only one virtual network card created, and also make sur
...
+
+The ISO edition of PacketFence allows you to install PacketFence on Debian 11 with minimal effort. Instead of manually installing Debian 11 and installing PacketFence after, this will perform both tasks and select the optimal parameters and best practices for installing the operating system.
+
+You can download the ISO here: https://www.packetfence.org/download.html#/iso
+
+==== Machine specifications
+
+This setup has been tested using VMware ESXi, Proxmox VE and VirtualBox but will also work with any hypervisor PacketFence supports as well as bare-metal servers.
+
+You will need a virtual machine or server with 12 GB of RAM dedicated to machine as well as 2 CPUs. Make sure you also have a hard-drive of at least 100GB.
```
extrafu: Make sure you also have a hard-drive of at least 100GB. => Make sure you allocate at least 100GB of disk space for PacketFence.

docs/installation/installation.asciidoc
```diff
@@ -54,6 +54,27 @@ Make sure that there is only one virtual network card created, and also make sur
...
+
+==== Installing the ISO to a virtual machine
+
+Provision a virtual machine with the specifications above, mount the ISO in the CD/DVD drive of the machine and start it. 
+The installer will open and you will simply have to follow the instructions on screen to complete the installation.
+
+==== Installing the ISO to a bare-metal server
+
+First, make sure your server follows the specifications above and then burn the ISO onto a DVD and boot your server using the DVD.
```
extrafu: Would that work from a USB stick too? If so, it's worth mentioning!

extrafu:
Minor nits mentioned

docs/installation/installation.asciidoc
julsemaan: Done! I also talked about putting it on a tape-cassette and floppy disks

julsemaan:
@nqb, all yours for the final review and merge

PR #6804, opened 2021-12-31
Feature/vagrant cluster setup

### LINKED ISSUE #6292: tests during a release pipeline should be done against pristine VM
**Describe the bug**
Since #6099, test stage of devel pipelines used Vagrant images built with all dependencies installed (only for Debian).

It's also the case for release pipeline which is not a good option.

For release pipelines, we should start pristine VM and install release packages from scratch.

nqb:
# Description
- Use our own Debian Vagrant images to speed up builds
- Add new VM to Vagrant and Ansible inventory in order to create cluster installations
- Create a draft of cluster scenario
- Fixes #6292 

# Impacts
Tests

# Dependencies
- https://github.com/inverse-inc/ansible-packetfence/pull/49

# Delete branch after merge
YES


nqb:
This PR needs a merge of https://github.com/inverse-inc/ansible-packetfence/pull/49 before being merged.

addons/vagrant/inventory/hosts
```diff
@@ -137,37 +248,49 @@ all:
...
               mgmt_netmask: "{{ users_vars[dict_name]['vms']['pfel8localdev']['netmask'] }}"
               ansible_host: "{{ mgmt_ip }}"
               ansible_python_interpreter: '/usr/bin/python3'
               cpus: 2
               memory: 6144
             pfdeb11localdev:
-              box: debian/bullseye64
-              box_version: 11.20211018.1
+              box: inverse-inc/pfdeb11dev
+              box_version: 11.2.20220107224856
```
JeGoi, line 266: Does it mean, we will have to update this version number on a regular basis?

addons/vagrant/inventory/hosts
```diff
@@ -99,20 +99,131 @@ all:
...
+            pf3deb11localdev:
+              box: inverse-inc/pfdeb11dev
+              box_version: 11.2.20220107224856
+              mgmt_ip: "{{ users_vars[dict_name]['vms']['pf3deb11localdev']['ip'] }}"
+              mgmt_netmask: "{{ users_vars[dict_name]['vms']['pf3deb11localdev']['netmask'] }}"
+              ansible_host: "{{ mgmt_ip }}"
+              ansible_python_interpreter: '/usr/bin/python3'
+              cpus: 2
+              memory: 6144
+
```
JeGoi, line 212: Could it be more readable to have it cluster per cluster type? like pf1,2,3deb11dev then pf1,2,3el8dev etc.
Here, the split is done only between dev and localdev.

t/venom/Makefile
```diff
@@ -50,76 +52,95 @@ run_w_clean:
...
+	CLUSTER_NAME="$(CLUSTER_NAME)" \
 	RESULT_DIR=$(RESULT_DIR) \
 	INT_TEST_VM_NAMES="$(INT_TEST_VM_NAMES)" \
 	./test-wrapper.sh halt
 
 clean:
 	./test-wrapper.sh teardown
 
+### Pristine images
+# use pristine images to install PacketFence from stratch
```
JeGoi: Typo

JeGoi (approved):

addons/vagrant/inventory/hosts
nqb: Each time we rebuild the image. It means:
 * each time Debian is updated (11.2 => 11.3)
 * each time we updated PacketFence dependencies

I have documented this into our internal wiki.

addons/vagrant/inventory/hosts
nqb: I don't think it's necessary because you can detect difference between hosts (Debian and RHEL) using facts.

JeGoi: Even if it is true, it is not the same order than in:
```
    dev:
      hosts:
 ```
and in:
```
    vars:
        vms:
```

JeGoi (approved):

PR #6808, opened 2022-01-05
Fix/don't resolve IPs for LDAP source using SSL and requiring verification

talltechdude:
# Description
Prevent LDAP Source host resolving to IPs if using SSL and server verification required. This allows hostname to be sent through to the Net::LDAPS module to be added as SNI header and matched against server certificate hostname at the cost of round robin server IP load balancing. Regression introduced in cc22dd9c9f1be228a5fa8ba8d978be6186d6c922.

# Impacts
LDAP Authentication Source server enumeration and IP based multi server connection for SSL and "Require" verification only

# Delete branch after merge
YES

# Checklist
- [X] Document the feature
- N/A Add unit tests
- N/A Add acceptance tests (TestLink)

# NEWS file entries

## Enhancements
* Allows LDAPS Authentication Source to use SSL verification based on certificate hostname

### DIFF
--- a/lib/pf/Authentication/Source/LDAPSource.pm
+++ b/lib/pf/Authentication/Source/LDAPSource.pm
@@ -236,8 +236,14 @@ sub _connect {
   my $connection;
   my $logger = Log::Log4perl::get_logger(__PACKAGE__);
   my $LDAPServer;
-  # Lookup the server hostnames to IPs so they can be shuffled better and to improve the failure detection
-  my @LDAPServers = map { valid_ip($_) ? $_ : @{resolve($_) // []} } @{$self->{'host'} // []};
+  my @LDAPServers;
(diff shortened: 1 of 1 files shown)


julsemaan:
I understand the need for this fix but I'm not sure we should sacrifice the reliability of this code path, especially not silently.

IMO we should either:
- Keep this but make the frontend display a warning that this configuration (encryption + server validation) will disable round robin and LDAP faillover
- Find a way to use the IP for the connection but pass the SNI as an option to the LDAP library with the DNS name

The best would clearly be the 2nd but we have to see if this can be accomplished via Net::LDAPS

trs80:
If you pass `sslserver => LDAPSERVERHOSTNAME` in `addSSLArgs` then Net::LDAP will use that when verifying the certificage.

https://metacpan.org/pod/Net::LDAP#sslserver-=%3E-SSLHOST

Issue #6673, opened 2021-10-29
haproxy-db timeouts don't match the MariaDB timeouts


jrouzierinverse:
**Describe**
The timeouts for haproxy-db don't match the timeouts for MariaDB this can cause connections to be closed abruptly.

PR #6809, opened 2022-01-06
Async pfacct + ping db for pfcron

fdurand:
# Description
Make pfacct async on db update (avoid pfacct mark as dead)

# Impacts
Accounting

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* pfacct async on db update (on pfcron job it locks a table and the insert stuck)

### DIFF
--- a/go/cmd/pfacct/radius.go
+++ b/go/cmd/pfacct/radius.go
@@ -134,26 +134,30 @@ func (h *PfAcct) handleAccountingRequest(rr radiusRequest) {
 	timestamp = timestamp.Truncate(h.TimeDuration)
 	node_id := mac.NodeId(uint16(switchInfo.TenantId))
 	unique_session_id := h.accountingUniqueSessionId(r)
-	if err := h.InsertBandwidthAccounting(
-		status,
-		node_id,
-		switchInfo.TenantId,
-		mac.String(),
-		unique_session_id,
-		timestamp,
-		in_bytes,
-		out_bytes,
-	); err != nil {
-		logError(ctx, "InsertBandwidthAccounting: "+err.Error())
-	}
+	go func() {
+		if err := h.InsertBandwidthAccounting(
+			status,
+			node_id,
+			switchInfo.TenantId,
+			mac.String(),
+			unique_session_id,
+			timestamp,
+			in_bytes,
+			out_bytes,
+		); err != nil {
+			logError(ctx, "InsertBandwidthAccounting: "+err.Error())
+		}
 
-	if status == rfc2866.AcctStatusType_Value_Stop {
-		h.CloseSession(node_id, unique_session_id)
-	}
+		if status == rfc2866.AcctStatusType_Value_Stop {
+			h.CloseSession(node_id, unique_session_id)
+		}
 
-	h.sendRadiusAccounting(r, switchInfo)
-	h.handleTimeBalance(r, switchInfo, unique_session_id)
-	h.handleBandwidthBalance(r, switchInfo, in_bytes+out_bytes)
+		h.sendRadiusAccounting(r, switchInfo)
+		h.handleTimeBalance(r, switchInfo, unique_session_id)
+		h.handleBandwidthBalance(r, switchInfo, in_bytes+out_bytes)
+	}()
+	return
+
 }
 
 func (h *PfAcct) handleTimeBalance(r *radius.Request, switchInfo *SwitchInfo, unique_session uint64) {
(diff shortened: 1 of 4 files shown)


jrouzierinverse:
The DB ping is just a bandaid and will not work completely.

Since multiple connections are being opened by the SQL backend layer.
The ping will only keep open one connection but not others.
The problem still exists but is more hidden.

In order to fix the DB connection issue properly, we should fix #6673.

julsemaan:
I 100% agree that the async is not the solution to this.

We not only need to fix #6673 but also fix the pfacct code so its able to scale as it was supposed to and how it used to in prior releases since it looks like its performances have regressed in recent releases judging by the amount of issues we're seeing around it lately.

jrouzierinverse:
We previously did things in an async manner and it caused problems with the CPU usage.

That was fixed in 8d7db52e4ab527346763747293c05b2b2498733f

I have some thoughts on what should be done let discuss later.

fdurand:
This code has been deployed on a client setup and there is no cpu issue.
Btw it fixs the "Mark as dead" issue we can see on the radius-load-balancer.
We need to answer right away to the accounting packet even if the insert in the db is not finished.

If we keep like that it mean that the equipment will try again and again to send the same accounting packet and it will end with a bogus accounting information in the db.

julsemaan:
I agree that this can fix the issue by being a band-aid but its not the right fix for the issue.

And as @jrouzierinverse pointed out, pfacct was previously async and this was causing problems with some customers so although it works with one customer doesn't guarantee us it will behave correctly everywhere.

Regarding: "If we keep like that it mean that the equipment will try again and again to send the same accounting packet and it will end with a bogus accounting information in the db.", the right fix would be to do like freeradius and drop duplicate accounting requests if we're already processing it so we don't process them twice.

fdurand:
With the last commit it scale correctly with 85 Request/s and doesn´t consume all the connections to the db.

go/cmd/pfacct/radius.go
```diff
@@ -134,25 +143,34 @@ func (h *PfAcct) handleAccountingRequest(rr radiusRequest) {
...
+	go func() {
+		var ok bool
+		select {
+		case h.JobChan <- node_acct_info:
+			ok = true
+		default:
+			ok = false
+		}
+		if !ok {
+			logError(ctx, "handleAccountingRequest: accouting queue - max capacity reached")
```
julsemaan: I'm guessing you should return here, otherwise you'll call the code below which substracts time balance and all the other stuff. I would also be good to not reply to the packet so that the controller sends a new request

go/cmd/pfacct/radius.go
```diff
@@ -134,25 +143,34 @@ func (h *PfAcct) handleAccountingRequest(rr radiusRequest) {
...
-	}
+	node_acct_info := acct_info{node_id: node_id,
+		tenant_id:      switchInfo.TenantId,
+		mac:            mac.String(),
+		unique_session: unique_session_id,
+		bucket:         timestamp,
+		in_bytes:       in_bytes,
+		out_bytes:      out_bytes,
+	}
+	go func() {
```
julsemaan, line 157: Now that the async is handled by the JobChan, do we really need to wrap this into a 'go func' ?

go/cmd/pfacct/radius.go
julsemaan: Also, on top of my comment above, just a small typo to fix in the log line

```suggestion
			logError(ctx, "handleAccountingRequest: accounting queue - max capacity reached")
```

go/cmd/pfacct/radius.go
fdurand: The JobChan is to queue the sql requests but if the api call stuck i still want to reply to the accounting packet right away.

go/cmd/pfacct/radius.go
```diff
@@ -160,7 +160,8 @@ func (h *PfAcct) handleAccountingRequest(rr radiusRequest) {
 			ok = false
 		}
 		if !ok {
-			logError(ctx, "handleAccountingRequest: accouting queue - max capacity reached")
+			logError(ctx, "handleAccountingRequest: accounting queue - max capacity reached")
+			return
```
julsemaan: Will that return but **not answer** so that the controller retries?

Otherwise we miss out on the information in that packet

fdurand: At least i think i need to requeue it since the radius reply has already been sent.

extrafu:
Do we keep that for v11.2?

fdurand:
> Do we keep that for v11.2?

no

PR #6816, opened 2022-01-11
Venom: Test security event autoreg using DHCP fingerprint

julsemaan:
# Description
Test security event autoreg using DHCP fingerprint

# Impacts
Venom tests

# Delete branch after merge
YES


julsemaan:
Pipeline: https://gitlab.com/inverse-inc/packetfence/-/pipelines/445856632

nqb:
I re-run a [pipeline ](https://gitlab.com/inverse-inc/packetfence/-/jobs/1965419635) and it works.

I will review this PR soon.

t/venom/test_suites/security_event_autoreg/prepare/00_create_autoreg_security_event.yml
```diff
@@ -0,0 +1,36 @@
...
+
+- name: create_security_event
+  steps:
+  - type: http
+    method: POST
+    url: '{{.pfserver_webadmin_url}}/api/v1/config/security_events'
+    ignore_verify_ssl: true
+    body: >-
+      {"id":"88888888","isClone":false,"isNew":true,"access_duration":"5D","actions":["autoreg","role"],"desc":"Test DHCP autoreg venom","enabled":"Y","target_category":"machine_employee","triggers":[{"dhcp_fingerprint":"{{.security_event_autoreg.node.dhcp_fingerprint_id}}"}],"whitelisted_roles":[]}
+    headers:
```
nqb: Could you:
- [X] Lint JSON
- [X] Remove `isClone` and `isNew` attributes
- [X] Use a variable for security event id, access duration and description
- [X] Use `{{.mac_auth.roles.headless_device.id}}` as `target_category` in place of `machine_employee` (makes more sense because we are doing MAC Authentication)

t/venom/test_suites/security_event_autoreg/prepare/00_create_autoreg_security_event.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: clear_fingerbank_cache
+  steps:
+  - type: exec
+    script: '/usr/local/pf/bin/pfcmd cache fingerbank clear'
+
+#- name: restart_pfqueue
+#  steps:
+#  - type: exec
+#    script: '/usr/local/pf/bin/pfcmd service pfqueue restart'
+
```
nqb: Can be removed.

t/venom/vars/all.yml
```diff
@@ -501,3 +501,11 @@ inline_l2.node.ipaddress: 192.168.2.10
 inline_l2.pf_portal: 192.168.2.1
 inline_l2.networks: 192.168.2.0
 
+################################################################################
+#Security event autoreg test suite specific variables
+#################################################################################
+security_event_autoreg.node.mac: 00:03:00:11:11:01
+security_event_autoreg.node.mac_url_encoded: 00%3A03%3A00%3A11%3A11%3A01
```
nqb: You can re-use existing variables: `node01_ens7_mac_address` and `node01_ens7_mac_address_url_encoded` with following syntax:

```yaml
security_event_autoreg.node.mac: '{{.node01_ens7_mac_address}}'
```

t/venom/test_suites/security_event_autoreg/verify/run.yml
```diff
@@ -0,0 +1,46 @@
+name: SecurityEventAutoreg
+testcases:
+- name: get_login_token
+  steps:
+  - type: get_login_token
+
+- name: sleep_some_time
+  steps:
+  - type: exec
+    script: sleep 30
```
nqb:
```suggestion
- name: sleep_some_time
  steps:
  - type: sleep
    time_to_sleep: 30
```

nqb (changes requested):
General remarks:
1. [X] Can we merge this test suite to another scenario ? I'm thinking about captive_portal scenario
1. [X] I think there is a risk that DHCP fingerprint can change when we will upgrade node01 OS or make some changes on infrastructure.
1. [ ] I think it could be interesting to check if security event has been triggered on node using API. Can be done later.
1. [X] Could you add a short `TESTSUITE.md` which explain test suite ?

julsemaan:
> 1. Can we merge this test suite to another scenario ? I'm thinking about captive_portal scenario

The "problem" with this is that since this suite involves switching from a VLAN to another it may break the other scenarios as they expand. Having one dedicated to security events (that would include isolation for suricata events, etc) made sense to me since it will be implicit that it involves multiple VLAN switches. I'm fine to merge it if you disagree, I'm just scared of the potential breakage that can happen due to it (either now or in the future)

> 2. I think there is a risk that DHCP fingerprint can change when we will upgrade node01 OS or make some changes on infrastructure.

It's a variable so we can always change it in the future when that happens. If I use dhcp_test, it won't be as "real" as using node01

> 3. I think it could be interesting to check if security event has been triggered on node using API. Can be done later.

👍 

> 4. Could you add a short `TESTSUITE.md` which explain test suite ?

Will do

nqb:
Agree with your answers @julsemaan.

julsemaan:
PIpeline with fixes: https://gitlab.com/inverse-inc/packetfence/-/pipelines/456839663

nqb:
LGTM.

I will just adjust syntax of variables to avoid spaces **inside** curly brackets like we have for all Venom test suites.
* Current: `security_event_autoreg.node.mac: '{{ .node01_ens7_mac_address }}'`
* Expected: `security_event_autoreg.node.mac: '{{.node01_ens7_mac_address}}'`

PR #6817, opened 2022-01-11
cache the PMK in Ruckus unbound DPSK

julsemaan:
# Description
Cache the PMK of each PSK+SSID that we see in Ruckus unbound DPSK

# Impacts
Ruckus unbound DPSK

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Drastically improved the performance of the Ruckus unbound DPSK implementation

### DIFF
--- a/lib/pf/Switch/Ruckus/SmartZone.pm
+++ b/lib/pf/Switch/Ruckus/SmartZone.pm
@@ -358,9 +358,15 @@ sub check_if_radius_request_psk_matches {
         return $FALSE;
     }
 
+    my $pmk = $self->cache->compute(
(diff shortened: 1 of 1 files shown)


lib/pf/Switch/Ruckus/SmartZone.pm
```diff
@@ -358,9 +358,15 @@ sub check_if_radius_request_psk_matches {
         return $FALSE;
     }
 
+    my $pmk = $self->cache->compute(
```
fdurand, line 361: Do you think it will be interesting to use the distributed cache ? (in case of a cluster with a large number of account)

julsemaan: I deliberately picked redis because it was faster to access so if you have a lot of PMK to compare against, then the query time is much lower when we use Redis vs the DB

PR #6822, opened 2022-01-12
Venom: Firewall SSO

satkunas:
# Description
Implement Firewall SSO Venom test suite

# Impacts
Extends `dot1x_eap_peap` scenario

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Feature 1
* Feature 2


t/venom/scenarios/dot1x_eap_peap/playbooks/configure.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: Install packages on el8
+      package:
+        name: "{{ el8Packages }}"
+        state: present
+      when: ansible_facts['os_family'] == "RedHat"
+
+    - name: Install yq python package
+      pip:
+        name: yq
```
nqb:
```suggestion
        name: yq
        state: present
```

t/venom/test_suites/wired_dot1x_eap_peap_firewall_sso_https/03_create_http_smocker.yml
```diff
@@ -0,0 +1,105 @@
+name: Create smocker HTTP mock
+testcases:
+
+- name: create_smocker_dir
+  steps:
+    - type: exec
+      script: "mkdir -p {{.smocker_dir}}"
+      vars:
+        temp_dir:
+          from: result.systemout
```
nqb: can be removed

nqb (changes requested):

nqb:
@satkunas and @julsemaan identified a limitation to run firewall SSO tests sequentially inside `wired_dot1x_eap_peap` scenario.

All tests use one node ( `node01`) and expect `node01` to have new IP address on production VLAN (192.168.121.0/24) between each tests in order to trigger correctly SSO workflow.

Currently, `node01` continues to use same IP address between each tests even after:
- forcing a DHCP release on `node01` 
- forcing a ifdown/ifup on `switch01`

I think this behavior is caused by `libvirt` DHCP server which assigned IP address for one hour by default and just reassign this IP address each time. This lease time is also a problem when a device go on production VLAN and we want to move it to another VLAN: node doesn't try to get a new IP address because lease is still valid.

A possible solution could be to create a another production network (similar to 192.168.121.0/24) with a DHCP server provided by a Vagrant VM (like `linux02`). Another solution could be to use a different node for each test.

nqb:
@julsemaan and @satkunas, I forgot to ask the basics: why do we need a new IP address on node01 ? Is it because we have an existing ip4log entry ? Would it work as-is if we clean ip4log during teardown ?

julsemaan:
> @julsemaan and @satkunas, I forgot to ask the basics: why do we need a new IP address on node01 ? Is it because we have an existing ip4log entry ? Would it work as-is if we clean ip4log during teardown ?

We need to have the device request a new IP address so that the pfdhcplistener sees it and triggers the firewall SSO in the second test.

We tried using the RADIUS accounting data switch01 sends but unfortunately it doesn't contain the Framed-IP-Address so we can't do SSO using that leaving only DHCP as the trigger

t/venom/test_suites/wired_dot1x_eap_peap_firewall_sso_https/teardown/03_kill_http_smocker.yml
```diff
@@ -0,0 +1,22 @@
+name: Create smocker HTTP mock
+testcases:
+
+- name: kill_ssl_termination_haproxy
+  steps:
+  - type: exec
+    script: |
+      sudo kill -9 `cat {{.smocker_dir}}/haproxy.pid`
```
nqb:
```suggestion
      kill -9 `cat {{.smocker_dir}}/haproxy.pid`
```

t/venom/test_suites/wired_dot1x_eap_peap_firewall_sso_https/teardown/03_kill_http_smocker.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: kill_smocker
+  steps:
+  - type: exec
+    script: |
+      sudo pkill smocker
```
nqb:
```suggestion
      pkill smocker
```

t/venom/test_suites/wired_dot1x_eap_peap_firewall_sso_https/teardown/03_kill_http_smocker.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: remove_smocker_dir
+  steps:
+    - type: exec
+      script: "rm -Rf {{.smocker_dir}}"
+      vars:
+        temp_dir:
+          from: result.systemout
```
nqb: Can be removed. This variable will not be usable after end of test.

t/venom/test_suites/wired_dot1x_eap_peap_firewall_sso_radius/03_create_radius_mock.yml
```diff
@@ -0,0 +1,95 @@
+name: Create RADIUS mock
+testcases:
+
+- name: install_configure_radius_mock
+  steps:
+  - script: |
+      cd {{.firewall_sso.radius.radius_mock_dir}}
+      npm install
+      npm install pm2 -g
```
nqb, line 9: Is it still necessary after Ansible provisioning (see configure.yml) ?

nqb (changes requested):
- [X] During teardown of two test suites, I suggest to restart pfsso **after** disable_sso_on_advanced

t/venom/test_suites/wired_dot1x_eap_peap_firewall_sso_radius/03_create_radius_mock.yml
satkunas: Yes, the ansible provisioning did not properly install pm2 as a global package

nqb: I'm removing `pm2` installation in `configure.yml` to avoid confusion

PR #6834, opened 2022-01-17
mysql_probe service in go to control haproxy-db backend

fdurand:
# Description
mysql_probe service in go to control haproxy-db backend

# Impacts
haproxy-db

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* mysql_probe service in go to control haproxy-db backend


go/galeraautofix/mariadb/mariadb.go
```diff
@@ -188,6 +188,68 @@ func IsDBAvailable(ctx context.Context, host string) bool {
 	return false
 }
 
+func IsLocalDBReady(ctx context.Context) bool {
+	return IsDBReady(ctx, "localhost")
+}
+
+func IsDBReady(ctx context.Context, host string) bool {
```
julsemaan, line 195: Just to make sure, these are added here but the actual galera-autofix code is not being altered?

go/cmd/mysql_probe/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	srv := &http.Server{
+		Addr:        ":3307",
+		IdleTimeout: 5 * time.Second,
+		Handler:     router,
+	}
+	// Systemd
+	daemon.SdNotify(false, "READY=1")
+
+	go func() {
```
julsemaan: Looks like if there is a panic in this block, this service will die. Is that what we want?

Not saying its bad or good, just want to make sure we're happy with how it will behave

go/cmd/mysql_probe/main.go
```diff
@@ -0,0 +1,75 @@
...
+		interval, err := daemon.SdWatchdogEnabled(false)
+		if err != nil || interval == 0 {
+			return
```
julsemaan: I never liked silent returns for errors

Let's make it explicit that there was an error or that the interval was 0 and log the behavior.

addons/packages/build-go.sh
```diff
@@ -54,7 +54,7 @@ cd "$PFSRC/go"
 
 if build_mode; then
   # Create any binaries here and make sure to move them to the BINDST specified
-  for service in pfhttpd pfdhcp pfdns pfstats pfdetect galera-autofix;do
+  for service in pfhttpd pfdhcp pfdns pfstats pfdetect galera-autofix mysql_probe;do
```
julsemaan: This is very 💅 but IMO this should be named mysql-probe

conf/haproxy-db.conf.example
```diff
@@ -42,11 +42,8 @@ frontend  main
 backend mysql
     mode tcp
     option tcplog
-    # disabled for now since it proved to be useless - jsemaan@inverse.ca
-    # TODO: remove it after cluster testing
-    #option external-check
-    #external-check command /usr/local/pf/var/run/db-check
-    #external-check path "/usr/bin:/bin"
+    #option httpchk
+    #default-server port 3307 inter 2s downinter 5s rise 3 fall 2 slowstart 60s maxconn 64 maxqueue 128 weight 100
```
julsemaan: This should be either:
- Enabled by default (so uncomment the lines)
- Be configurable via the webadmin in the database_advanced section

lib/pf/services/manager/mysql_probe.pm
```diff
@@ -0,0 +1,58 @@
...
+use pf::cluster;
+use Moo;
+extends 'pf::services::manager';
+
+has '+name' => (default => sub { 'mysql_probe' } );
+
+
+sub isManaged {
+    my ($self) = @_;
+    return $self->SUPER::isManaged() && $cluster_enabled && !pf::cluster::isSlaveMode();
```
julsemaan, line 26: We should have a way to explicitly disable the service via pf.conf like we have for our other services

go/cmd/mysql_probe/main.go
fdurand: I want to test if the service still reply, if it dies then systemd will restart it.

go/galeraautofix/mariadb/mariadb.go
fdurand: It won't change anything, i just want to rely on the great job that has been done in galera-autofix.

extrafu:
@julsemaan All good to merge?

PR #6836, opened 2022-01-19
Add the ip6tables service

### LINKED ISSUE #6351: zen does not setup ipv6 firewall
**Describe the bug**

While doing a network scan, I noticed that packetfence did answered some services through IPv6.
I checked and ip6tables is clean:

```
[root@vms188l etc]# ip6tables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination   
```      

Shouldn't packetfence be responsible to, at least, block incoming IPv6 traffic?

I'm using zen running with packetfence-10.2.0-20201007132254.199263253.0007.v10.2.0.el7.x86_64.
It is not the latest but I didn't find anything related in changelog.

**To Reproduce**
* Setup IPv6 address using packetfence web interface.
* check ip6tables

julsemaan:
# Description
Adds the ip6tables service

# Impacts
New service

# Issue
fixes #6351 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [N/A] Document the feature
- [N/A] Add unit tests
- [TODO] Add acceptance tests (TestLink)

# NEWS file entries
## Enhancements
* The ip6tables rules are now managed by PacketFence

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -465,6 +465,13 @@ Location of the apache binary. Only necessary to change if you are
 not running the RPMed version.
 EOT
 
+[services.iptables_binary]
+type=hidden
+description=<<EOT
+Location of the iptables binary. Only necessary to change if you are
+not running the RPMed version.
+EOT
+
 [services.ip6tables_binary]
 type=hidden
 description=<<EOT
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -230,6 +230,13 @@ description=<<EOT
 Should iptables be managed by PacketFence? Keep enabled unless you know what you're doing.
 EOT
 
+[services.ip6tables]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Should ip6tables be managed by PacketFence? Keep enabled unless you know what you're doing.
+EOT
+
 [services.httpd_portal]
 type=toggle
 options=enabled|disabled
@@ -458,10 +465,10 @@ Location of the apache binary. Only necessary to change if you are
 not running the RPMed version.
 EOT
 
-[services.iptables_binary]
+[services.ip6tables_binary]
 type=hidden
 description=<<EOT
-Location of the iptables binary. Only necessary to change if you are
+Location of the ip6tables binary. Only necessary to change if you are
 not running the RPMed version.
 EOT
 
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -476,7 +476,7 @@ EOT
 type=hidden
 description=<<EOT
 Location of the ip6tables binary. Only necessary to change if you are
-not running the RPMed version.
+not running the official packaged version.
 EOT
 
 [services.winbindd_binary]
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -460,6 +460,16 @@ iptables=enabled
 # location of the iptables binary. only necessary to change if you are not running the rpmed version.
 iptables_binary=/sbin/iptables
 #
+# services.ip6tables
+#
+# Should ip6tables be managed by PacketFence?
+ip6tables=enabled
+# 
+# services.ip6tables_binary
+# 
+# location of the ip6tables binary. only necessary to change if you are not running the rpmed version.
+ip6tables_binary=/usr/sbin/ip6tables
+#
 # services.httpd_portal
 #
 # Should httpd.portal be started?
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -467,7 +467,7 @@ ip6tables=enabled
 # 
 # services.ip6tables_binary
 # 
-# location of the ip6tables binary. only necessary to change if you are not running the rpmed version.
+# location of the ip6tables binary. only necessary to change if you are not running the official packaged version.
 ip6tables_binary=/usr/sbin/ip6tables
 #
 # services.httpd_portal
--- a/lib/pf/ip6tables.pm
+++ b/lib/pf/ip6tables.pm
@@ -1,0 +1,109 @@
+package pf::ip6tables;
+
+=head1 NAME
+
+pf::ip6tables - module for ip6tables rules management.
+
+=cut
+
+=head1 DESCRIPTION
+
+pf::ip6tables contains the functions necessary to manipulate the ip6tables service
+
+=head1 CONFIGURATION AND ENVIRONMENT
+
+F<pf.conf> configuration file and ip6tables template F<ip6tables.conf>.
+
+=cut
+
+use strict;
+use warnings;
+
+use pf::log;
+use Readonly;
+use pf::constants;
+use File::Slurp qw(read_file);
+use pf::util;
+
+use pf::config qw(
+    $management_network
+    @portal_ints
+);
+use pf::file_paths qw($generated_conf_dir $conf_dir);
+
+# This is the content that needs to match in the iptable rules for the service
+# to be considered as running
+Readonly our $FW_FILTER_INPUT_MGMT      => 'input-v6-management-if';
+Readonly our $FW_FILTER_INPUT_PORTAL    => 'input-v6-portal-if';
+
+sub generate {
+    my ($class) = @_;
+    my $logger = get_logger();
+    my %tags;
+    $tags{'input_management_include'} = "#BEGIN include ip6tables-input-management.conf.inc\n" . read_file("$conf_dir/ip6tables-input-management.conf.inc") . "#END include ip6tables-input-management.conf.inc\n";
+
+    $tags{management_chain} = <<EOF;
+-N $FW_FILTER_INPUT_MGMT
+-A INPUT -i $management_network->{Tint} -j $FW_FILTER_INPUT_MGMT
+EOF
+
+    $tags{portal_chain} = "-N $FW_FILTER_INPUT_PORTAL\n";
+    foreach my $portal_interface ( @portal_ints ) {
+        $tags{portal_chain} .= <<EOF;
+-A INPUT -i $portal_interface->{Tint} -j $FW_FILTER_INPUT_PORTAL
+EOF
+    }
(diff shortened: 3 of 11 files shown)


julsemaan:
Tested the packaging on Debian + RHEL and its fine

We'll have an acceptance test in the integration tests with a little magic from @nqb

nqb (changes requested):
Packaging looks good. 

Regarding binary location for Debian, I think we just need to apply a patch on `pf.conf` like we do for FreeRADIUS: https://github.com/inverse-inc/packetfence/blob/84892e085adc6b342790f805bc61fc9361f7a149/debian/patches/debianize.patch#L42-L49

to have a working ip6tables on Debian without any visible override on `pf.conf`

conf/documentation.conf
```diff
@@ -465,6 +472,13 @@ Location of the iptables binary. Only necessary to change if you are
 not running the RPMed version.
 EOT
 
+[services.ip6tables_binary]
+type=hidden
+description=<<EOT
+Location of the ip6tables binary. Only necessary to change if you are
+not running the RPMed version.
```
extrafu: "RPMed version" - how about the Debian packages? Should we simply say "if you are not running the official packaged version" ?

conf/pf.conf.defaults
```diff
@@ -451,6 +451,16 @@ iptables=enabled
...
 iptables_binary=/sbin/iptables
 #
+# services.ip6tables
+#
+# Should ip6tables be managed by PacketFence?
+ip6tables=enabled
+# 
+# services.ip6tables_binary
+# 
+# location of the ip6tables binary. only necessary to change if you are not running the rpmed version.
```
extrafu: Same as above

julsemaan:
> Packaging looks good.
> 
> Regarding binary location for Debian, I think we just need to apply a patch on `pf.conf` like we do for FreeRADIUS:
> 
> https://github.com/inverse-inc/packetfence/blob/84892e085adc6b342790f805bc61fc9361f7a149/debian/patches/debianize.patch#L42-L49
> 
> to have a working ip6tables on Debian without any visible override on `pf.conf`

Based on my findings, the binary is located at the same place on RHEL 8 and Debian 11

lib/pf/ip6tables.pm
```diff
@@ -0,0 +1,109 @@
...
+);
+use pf::file_paths qw($generated_conf_dir $conf_dir);
+
+# This is the content that needs to match in the iptable rules for the service
+# to be considered as running
+Readonly our $FW_FILTER_INPUT_MGMT      => 'input-v6-management-if';
+Readonly our $FW_FILTER_INPUT_PORTAL    => 'input-v6-portal-if';
+
+sub generate {
+    my ($class) = @_;
```
fdurand, line 40: self instead of class

lib/pf/ip6tables.pm
```diff
@@ -0,0 +1,109 @@
...
+EOF
+    }
+
+    parse_template( \%tags, "$conf_dir/ip6tables.conf", "$generated_conf_dir/ip6tables.conf" );
+    $class->restore("$generated_conf_dir/ip6tables.conf");
+}
+
+
+sub save {
+    my ($class, $save_file) = @_;
```
fdurand, line 63: self instead of class

lib/pf/ip6tables.pm
```diff
@@ -0,0 +1,109 @@
...
+    my $logger = get_logger();
+    $logger->info( "saving existing ip6tables to " . $save_file );
+    pf_run("/usr/sbin/ip6tables-save -t nat > $save_file");
+    pf_run("/usr/sbin/ip6tables-save -t mangle >> $save_file");
+    pf_run("/usr/sbin/ip6tables-save -t filter >> $save_file");
+}
+
+sub restore {
+    my ($class, $restore_file) = @_;
```
fdurand, line 72: self instead of class

lib/pf/services/manager/ip6tables.pm
```diff
@@ -0,0 +1,166 @@
...
+
+=cut
+
+sub startService {
+    my ($self) = @_;
+    unless ($self->isAlive()) {
+        pf::ip6tables->save($install_dir . '/var/ip6tables.bak');
+    }
+    pf::ip6tables->generate();
+    return 1;
```
fdurand: Use constant

lib/pf/services/manager/ip6tables.pm
```diff
@@ -0,0 +1,166 @@
...
+
+=head2
+
+generateConfig
+
+=cut
+
+sub generateConfig {
+    pf::ip6tables->generate();
+    return 1;
```
fdurand: Use constant

lib/pf/services/manager/ip6tables.pm
```diff
@@ -0,0 +1,166 @@
...
+=head2 stop
+
+Wrapper around systemctl. systemctl should in turn call the actual _stop.
+
+=cut
+
+sub stop {
+    my ($self) = @_;
+    system('sudo systemctl stop packetfence-ip6tables');
+    return 1;
```
fdurand: Use constant

lib/pf/services/manager/ip6tables.pm
```diff
@@ -0,0 +1,166 @@
...
+
+stop ip6tables (called from systemd)
+
+=cut
+
+sub _stop {
+    my ($self) = @_;
+    my $logger = get_logger();
+    pf::ip6tables->restore( $install_dir . '/var/ip6tables.bak' );
+    return 1;
```
fdurand: Use constant

lib/pf/services/manager/ip6tables.pm
```diff
@@ -0,0 +1,166 @@
...
+=cut
+
+sub isAlive {
+    my ($self) = @_;
+    my $logger = get_logger();
+    my $result;
+    my $pid = $self->pid;
+    my $_EXIT_CODE_EXISTS = "0";
+    my $rules_applied = defined( pf_run( "sudo " . $Config{'services'}{"ip6tables_binary"} . " -S | grep " . $pf::ip6tables::FW_FILTER_INPUT_MGMT ,accepted_exit_status => [$_EXIT_CODE_EXISTS]) );
+    return ($pid && $rules_applied) ? 1 : 0;
```
fdurand: Use constant

lib/pf/services/manager/ip6tables.pm
```diff
@@ -0,0 +1,166 @@
...
+=head2 start
+
+Wrapper around systemctl. systemctl should in turn call the actuall _start.
+
+=cut
+
+sub start {
+    my ($self,$quick) = @_;
+    system('sudo systemctl start packetfence-ip6tables');
+    return $? == 0;
```
fdurand: Use constant

fdurand (changes requested):
It miss the debianize.patch change.
Do we need to log in ip6tables ?

lib/pf/ip6tables.pm
julsemaan: AFAIK, we use `$class` when we don't have to call `new` on the module (its the case here)

@jrouzierinverse can correct me if I'm wrong

julsemaan:
> It miss the debianize.patch change. 

AFAIK there is nothing necessary there because the paths are the same on Debian and RHEL (I tested a package on both)
That is unless something else than the binary paths has to be debianized

> Do we need to log in ip6tables ?

You mean ip6tables.pm? Or the actual ip6tables chain?

fdurand:
> > It miss the debianize.patch change.
> 
> AFAIK there is nothing necessary there because the paths are the same on Debian and RHEL (I tested a package on both) That is unless something else than the binary paths has to be debianized

It´s for the systemd script, the perl path is not the same.
> 
> > Do we need to log in ip6tables ?
> 
> You mean ip6tables.pm? Or the actual ip6tables chain?

I mean that:
-j LOG --log-prefix "STATE_INVALID_DROP: "

julsemaan:
I've moved the venom tests into [fix/6351](https://github.com/inverse-inc/packetfence/tree/fix/6351-test)

julsemaan:
> > > It miss the debianize.patch change.
> > 
> > 
> > AFAIK there is nothing necessary there because the paths are the same on Debian and RHEL (I tested a package on both) That is unless something else than the binary paths has to be debianized
> 
> It´s for the systemd script, the perl path is not the same.

According to my Debian 11 machine, `/bin/perl` points to the Perl binary:
```
root@test-pf:~# /bin/perl --version

This is perl 5, version 32, subversion 1 (v5.32.1) built for x86_64-linux-gnu-thread-multi
(with 46 registered patches, see perl -V for more detail)

Copyright 1987-2021, Larry Wall

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at http://www.perl.org/, the Perl Home Page.
```

With that in mind, is it really worth changing debianize.patch for patching something that isn't required

> > > Do we need to log in ip6tables ?
> > 
> > 
> > You mean ip6tables.pm? Or the actual ip6tables chain?
> 
> I mean that: -j LOG --log-prefix "STATE_INVALID_DROP: "

I'll remove this

fdurand:
If the perl path is ok on debian 11 then we are good so.

julsemaan:
Everything should be adjusted except the `$self` vs `$class` which is due to calling the functions on the class and not on a blessed instance. 

Happy to change it if we state we must use `$self` even in this case

Some examples of `$class`:
- lib/pf/api.pm
- lib/pf/Connection/Profile.pm
- lib/pf/Portal/Session.pm

PR #6844, opened 2022-01-25
Fix/6827

### LINKED ISSUE #6827: pfpki: notify users or admins on certificate expiration and permit renewal
If you generate client certificates using:
* `pfpki` and a PKI provider 
* `pfpki` and SCEP clients

At some point, you need to renew certificates and deploy them again on devices.

#### pfpki and a PKI provider

A mail is mandatory to generate a certificate. This mail can be used to notify end-users that their certificates will expire.
But we need to allow them to renew certificate using captive portal which is not possible at the moment unless certificate has been revoked (see #6377)

#### pfpki and SCEP clients

Depending on PKI template configuration, certificates could be generated without a mail attribute. That's expected when certificates are generated for machines and managed by MDM. In that situation, it will make sense to define an admin mail address on PKI template to notify admin about certificate expiration. That's way, admins can force devices to make new SCEP request but renewal on PacketFence should be possible.

- Notify users (certificates generated from captive portal) or admins (certificates generated through SCEP) about certificate expiration
- Permit certificate renewal on captive portal and using SCEP

fdurand:
# Description
Allow to renew the certificate x days before the expiration date.
Send email x days before the expiration date to the user email/ profil email / administrator.
Pki cn provide certificate for the same cn but for different profiles. (profile name added in Subject)
Auto revoke certificate if expired.

# Impacts
pfpki

# Issue
fixes #6827

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Allow to renew certificate x days before the expiration date.
* Send email x days before the expiration date to the user email/ profil email / administrator.
* Pki cn provide certificate for the same cn but for different profiles. (profile name added in Subject)
* Auto revoke certificate if expired
* 
# UPGRADE file entries
* Email field is mandatory in the pki profile now. So if you have previously a pki profile with an empty Email field then you need to update it. If you don´t do it then renewal email won´t work for new certificates.


html/pfappserver/root/src/views/Configuration/pki/profiles/_components/TheForm.vue
```diff
@@ -119,6 +119,38 @@
           :text="$i18n.t('Cloud Service to integrate.')"
         />
       </base-form-tab>
+      <base-form-tab :title="$i18n.t('Renewal Configuration')">
+        <form-group-days-before-renewal namespace="days_before_renewal"
+          :column-label="$i18n.t('Days before renewal')"
+          :text="$i18n.t('Number of days before the pki authorize renewal. Set to 0 mean renewal is always allowed')"
```
extrafu: Number of days before the pki authorize renewal. Set to 0 mean renewal is always allowed => Number of days before the PKI authorizes renewal. Setting it to 0 means renewal is always allowed.

html/pfappserver/root/src/views/Configuration/pki/profiles/_components/TheForm.vue
```diff
@@ -119,6 +119,38 @@
...
+        />
+        <form-group-renewal-mail namespace="renewal_mail"
+          :column-label="$i18n.t('Renewal Email')"
+          :text="$i18n.t('Send email to the owner when the certificate is about to expire.')"
+        />
+        <form-group-days-before-renewal-mail namespace="days_before_renewal_mail"
+          :column-label="$i18n.t('Days before sending renewal email')"
+          :text="$i18n.t('Number of days before expiration of the certificate that trigger sending email.')"
```
extrafu: Number of days before expiration of the certificate that trigger sending email. => Number of days before certificate expiration to trigger sending email.

extrafu:
@satkunas Also review the wording in html/pfappserver/root/src/views/Configuration/pki/profiles/_components/TheForm.vue - I am not fully satisfied with it.

db/pf-schema-X.Y.sql
```diff
@@ -1456,7 +1456,6 @@ CREATE TABLE `pki_certs` (
   `dns_names` varchar(255) DEFAULT NULL,
   `ip_addresses` varchar(255) DEFAULT NULL,
   `scep` BOOLEAN DEFAULT FALSE,
-  UNIQUE KEY `cn` (`cn`),
```
julsemaan, line 1459: I think we may need a new UNIQUE index on cn+profile_name or at least an index just to speed up the querying

julsemaan:
I have one general question about this PR

When I renew my certificate on the PKI, does it immediately revoke my old certificate even if its not expired yet?

I'm thinking of a case where I'm connected to an SSID with my existing certificate and I'm renewing it but I lose access to the wifi during the renewal because it immediately revokes my current cert

fdurand:
> 



> I have one general question about this PR
> 
> When I renew my certificate on the PKI, does it immediately revoke my old certificate even if its not expired yet?
Yes
> 
> I'm thinking of a case where I'm connected to an SSID with my existing certificate and I'm renewing it but I lose access to the wifi during the renewal because it immediately revokes my current cert

Most of the time when you ask/renew your certificate your device is on an open ssid.
But if your certificate is revoked it doesn´t mean that you are disconnected and it doesn´t mean that you can´t reconnect with your cert (you need to enable ocsp to check the certificate validity).

julsemaan:
> > 
> 
> > I have one general question about this PR
> > When I renew my certificate on the PKI, does it immediately revoke my old certificate even if its not expired yet?
> > Yes
> > I'm thinking of a case where I'm connected to an SSID with my existing certificate and I'm renewing it but I lose access to the wifi during the renewal because it immediately revokes my current cert
> 
> Most of the time when you ask/renew your certificate your device is on an open ssid. But if your certificate is revoked it doesn´t mean that you are disconnected and it doesn´t mean that you can´t reconnect with your cert (you need to enable ocsp to check the certificate validity).

I'm thinking of a case where you'd receive an email notification about your renewal and click it while being connected to the secure SSID. You proceed to renew the certificate but for whatever reason like going to the bathroom for 30 minutes (I know a guy) and you get reauthenticated, its going to reject you before you install the new cert.

IMO, we should leave the existing certificate valid until it expires even if we renew it. That's usually how renewals work (like when we renew public certs on Let's Encrypt or GoDaddy, it leaves the current cert valid to give you a chance to replace it)

fdurand:
> > > I have one general question about this PR
> > > When I renew my certificate on the PKI, does it immediately revoke my old certificate even if its not expired yet?
> > > Yes
> > > I'm thinking of a case where I'm connected to an SSID with my existing certificate and I'm renewing it but I lose access to the wifi during the renewal because it immediately revokes my current cert
> > 
> > 
> > Most of the time when you ask/renew your certificate your device is on an open ssid. But if your certificate is revoked it doesn´t mean that you are disconnected and it doesn´t mean that you can´t reconnect with your cert (you need to enable ocsp to check the certificate validity).
> 
> I'm thinking of a case where you'd receive an email notification about your renewal and click it while being connected to the secure SSID. You proceed to renew the certificate but for whatever reason like going to the bathroom for 30 minutes (I know a guy) and you get reauthenticated, its going to reject you before you install the new cert.
> 
> IMO, we should leave the existing certificate valid until it expires even if we renew it. That's usually how renewals work (like when we renew public certs on Let's Encrypt or GoDaddy, it leaves the current cert valid to give you a chance to replace it)

There is no link in the email, it will just alert the owner that his certificate is about to expire.
So it´s only when the device will hit the portal and ask for a new certificate that we will revoke the old one and provide a new one.
Btw if you don´t have any certificate revoke list or ocsp then the certificate will continue to work until the valid date.

Now it´s more about how the ocsp work, do we want to give a chance to the end user to still continue to use his old certificate for a certain amount of time (to give the guy a chance to finish his urgent task in the bathroom) ?

julsemaan:
> Now it´s more about how the ocsp work, do we want to give a chance to the end user to still continue to use his old certificate for a certain amount of time (to give the guy a chance to finish his urgent task in the bathroom) ?

I think its important to have the option of not immediately revoking the cert so I'd go with one of these:
- Have a setting (enable/disable) to immediately revoke the old cert when a new one is issued
- Have a setting to revoke the old cert after X time when a new one is issued (with 0 being an immediate revocation)

The 2nd seems better but more complex to put in place and probably not absolutely necessary since the 1st idea would accomplish the goal of giving the control to the user based on his needs

What are your thoughts?

fdurand:
Solution 2 will be simplier, i will revoke the certificate but set a revoke date in the future (based on a profile config) and change the logic in ocsp.

julsemaan:
> Solution 2 will be simplier, i will revoke the certificate but set a revoke date in the future (based on a profile config) and change the logic in ocsp.

Happy with that, it was my favorite solution!

nqb:
EDIT: SOLVED

I run existing integration tests to see if it break something. There is an issue to authenticate using a manually generated certificate but not with SCEP certificate.

Subject of manually generated certificate:
`subject=C = CA, ST = Quebec, L = Montreal, O = Inverse, OU = PacketFence + OU = InverseCA1_user, CN = InverseCA1_user_cert` (plus sign doesn't seem legit)

It's also a problem if you have EAPTLS authentication source with rule that check subject: it will not work without a change.

For SCEP certificate, this information is not added in resulting certificate.

I don't think we should include this information in generated certificates.

nqb:
IMHO, we must rely on serial number of certificates to have unique certificates in DB. 

It's an existing field in certificates (no need to add extra informations coming from PacketFence) and it will never cause issues.

conf/pfcron.conf.defaults
```diff
@@ -596,47 +596,69 @@ delay=30D
 # SSL certificate(s) to be monitored
 certificates=/usr/local/pf/conf/ssl/server.pem,/usr/local/pf/raddb/certs/server.crt
 
+[pki_certificates_check]
```
nqb, line 577: Need to be added to the help of `pfcmd pfcron`

db/upgrade-X.X-X.Y.sql
```diff
@@ -49,7 +49,10 @@ DROP PROCEDURE IF EXISTS ValidateVersion;
 
 \! echo "altering pki_certs"
 ALTER TABLE pki_certs
-    ADD COLUMN IF NOT EXISTS `scep` BOOLEAN DEFAULT FALSE AFTER ip_addresses;
+    DROP INDEX cn,
+    ADD CONSTRAINT cn_profile UNIQUE KEY(`cn`,`profile_name`),
+    ADD COLUMN IF NOT EXISTS `scep` BOOLEAN DEFAULT FALSE AFTER ip_addresses,
+    ADD COLUMN IF NOT EXISTS `alert` BOOLEAN DEFAULT FALSE AFTER scep;
```
nqb: `alert` column is not added on a fresh install.

nqb (changes requested):
- [X] Bad display in Maintenance menu and not editable:
image
- [X] Even after I alter table `pki_certs` to add `alert` column, I'm not able to receive emails about a expired certifciate generated from web admin. What I did:
   * create a PKI template with `Validity` = 1 day and `Days before sending renewal email` = 2 days
    * run manually `pki_certificates_check` maintenance task

nqb (changes requested):
- [/] `Renewal Email` should be set to `On` by default.
- We should make Email mandatory when creating manually a certificate from web admin. Email is already mandatory when creating certificate on captive portal. For SCEP, I'm not sure about approach. I see two solutions:
  * We make mandatory email when creating PKI templates. That's way, if CSR sent through SCEP doesn't contain email, we have at least an email to notify
  *  We reject SCEP requests which doesn't contain email address in CSR

nqb:
I will not write integration tests on this PR.

fdurand:
> * Bad display in Maintenance menu and not editable:
>   image
> * Even after I alter table `pki_certs` to add `alert` column, I'm not able to receive emails about a expired certifciate generated from web admin. What I did:
>   
>   * create a PKI template with `Validity` = 1 day and `Days before sending renewal email` = 2 days
>   * run manually `pki_certificates_check` maintenance task

Can you retry with the latest code ?
On my side it works.

fdurand:
> * `Renewal Email` should be set to `On` by default.
> * We should make Email mandatory when creating manually a certificate from web admin. Email is already mandatory when creating certificate on captive portal. For SCEP, I'm not sure about approach. I see two solutions:
>   
>   * We make mandatory email when creating PKI templates. That's way, if CSR sent through SCEP doesn't contain email, we have at least an email to notify
>   * We reject SCEP requests which doesn't contain email address in CSR

We can just have the email mandatory in the profile, so if it´s not define on the portal / in scep then we use the profile one.

fdurand:
> IMHO, we must rely on serial number of certificates to have unique certificates in DB.
> 
> It's an existing field in certificates (no need to add extra informations coming from PacketFence) and it will never cause issues.

we need to take a decision on that, do we want to be able to provide multiples certificate for the same CN  (from other profiles) ? The Subject of the certificate needs to be unique so maybe we can relly on that and check if there is no other certificate in the db with the same subject.

nqb:
> We can just have the email mandatory in the profile, so if it´s not define on the portal / in scep then we use the profile one.

I will not cover use case where you use PKI to generate manually (from web admin) a certificate. For example, if you generate a server certificate, it's mandatory to have an email address to notify admins that certificate will expire.

nqb:
> The Subject of the certificate needs to be unique so maybe we can relly on that and check if there is no other certificate in the db with the same subject.

If I'm not wrong, at the moment, we only rely on the `cn` which is only a part of the subject ?

nqb (changes requested):
- With `Renewal Email` sets to `On` by default, we need to have a default `Renewal Mail Subject` when creating PKI profiles through API or for existing PKI profiles. Otherwise, you will receive an email with empty subject and empty title.

nqb (changes requested):
- It looks like now, when creating a PKI template using frontend, Email is mandatory to have an email when requesting a SCEP certificate. However, if you previously created a PKI template (in an older version) without Email and you request SCEP certificates without emails, resulting certificate will not have any emails.

fdurand:
> > The Subject of the certificate needs to be unique so maybe we can relly on that and check if there is no other certificate in the db with the same subject.
> 
> If I'm not wrong, at the moment, we only rely on the `cn` which is only a part of the subject ?

I did a change and it rely on the cn and all the attributes that are part of the subject. So it mean that if the profile attributes are different it will generate a unique Subject (like OU = IT versus OU = RH).

satkunas:
@fdurand did you address the wording issues @extrafu mentioned above?

fdurand:
> @fdurand did you address the wording issues @extrafu mentioned above?

i believe so

nqb (changes requested):
- In current state of code, we don't cover renewal of manually generated certificate (for example: RADIUS server in PacketFence). Before your certificate expire, you will want to generate a new one but it will not possible because they will have same CN or Subject.

fdurand:
> * In current state of code, we don't cover renewal of manually generated certificate (for example: RADIUS server in PacketFence). Before your certificate expire, you will want to generate a new one but it will not possible because they will have same CN or Subject.

You just have to revoke it first.

nqb:
I will not cause any issues ? I'm thinking about use case where you generated a RADIUS certificate for PacketFence server itself so your OCSP server is also `pfpki`. If you revoke certificate to generate a new one, you have to be sure RADIUS server will continue to work with a revoked certificate until you configure new certificate.

IMHO, it should be possible to have something similar to:

> Have a setting to revoke the old cert after X time when a new one is issued (with 0 being an immediate revocation)

to keep both certificates in place.

fdurand:
> 

The old RADIUS certificate is not tested against ocsp so it will still work until you replace.
So for me there is no issue.

nqb (changes requested):
- If I renew an existing SCEP certificate using a PKI profile with following settings:
```sql
scep_enabled: 1
scep_challenge_password: secret
scep_days_before_renewal: 0
```

A new certificate is correctly generated but my PKI profile is updated and I lost all renewal parameters I did previously.

- [X] With a mail configured on a PKI template and a SCEP request without email attribute, my resulting certificate doesn't contain any mail. But I think it's related to error reported above.

fdurand:
> > We can just have the email mandatory in the profile, so if it´s not define on the portal / in scep then we use the profile one.
> 
> I will not cover use case where you use PKI to generate manually (from web admin) a certificate. For example, if you generate a server certificate, it's mandatory to have an email address to notify admins that certificate will expire.

It has been addressed in https://github.com/inverse-inc/packetfence/pull/6844/commits/bbcb6869b205baeb19671e683d7655311489ce9d

fdurand:
> * It looks like now, when creating a PKI template using frontend, Email is mandatory to have an email when requesting a SCEP certificate. However, if you previously created a PKI template (in an older version) without Email and you request SCEP certificates without emails, resulting certificate will not have any emails.

It´s probably something that we can notify in the upgrade notes, isn´t it ?

extrafu:
I think it's fine to mention this in the upgrade notes.

fdurand:
> * If I renew an existing SCEP certificate using a PKI profile with following settings:
> 
> ```sql
> scep_enabled: 1
> scep_challenge_password: secret
> scep_days_before_renewal: 0
> ```
> 
> A new certificate is correctly generated but my PKI profile is updated and I lost all renewal parameters I did previously.

I tried multiples times and i am not able to replicate that.

> 
> * With a mail configured on a PKI template and a SCEP request without email attribute, my resulting certificate doesn't contain any mail. But I think it's related to error reported above.

With the latest commit https://github.com/inverse-inc/packetfence/pull/6844/commits/33d0c460cf963f2f8cef2e1baea4801d7490c2bd it´s fixed.

fdurand:
> * With `Renewal Email` sets to `On` by default, we need to have a default `Renewal Mail Subject` when creating PKI profiles through API or for existing PKI profiles. Otherwise, you will receive an email with empty subject and empty title.

I already set it in this commit (https://github.com/inverse-inc/packetfence/pull/6844/commits/cd59f14021ac0552d956048cdc29d679bb28b1a1) and it provide that:

image

extrafu:
@jrouzierinverse Log the renewal and resign API calls in your separate PR.

nqb:
There is an issue with the build: https://gitlab.com/inverse-inc/packetfence/-/jobs/2040079076

nqb:
> I already set it in this commit ([cd59f14](https://github.com/inverse-inc/packetfence/commit/cd59f14021ac0552d956048cdc29d679bb28b1a1)) and it provide that:

As mentioned during our meeting, I think `renewal_mail_subject` default value is only defined in frontend. When creating a PKI template using API without `renewal_mail_subject`, value is empty.

nqb:
I'm still having **first** issue reported here:

https://github.com/inverse-inc/packetfence/pull/6844#pullrequestreview-867964251

### Steps to replicate 

- Create CA, PKI template with SCEP and renewal setting (Renewal mail = On, Renewal Mail subject = "Certificate expiration"
=> check values in DB: they are correct
- Get a SCEP certificate using this template
- Check your PKI profile in web browser (you need to refresh)
- Check values of your PKI profile in DB
=> renewal settings are back to:
```
            renewal_mail: 0
days_before_renewal_mail: 0
    renewal_mail_subject:
       renewal_mail_from:
     renewal_mail_header:
     renewal_mail_footer:
     revoked_valid_until: 0
```

fdurand:
> I'm still having **first** issue reported here:
> 
> [#6844 (review)](https://github.com/inverse-inc/packetfence/pull/6844#pullrequestreview-867964251)
> 
> ### Steps to replicate
> * Create CA, PKI template with SCEP and renewal setting (Renewal mail = On, Renewal Mail subject = "Certificate expiration"
>   => check values in DB: they are correct
> * Get a SCEP certificate using this template
> * Check your PKI profile in web browser (you need to refresh)
> * Check values of your PKI profile in DB
>   => renewal settings are back to:
> 
> ```
>             renewal_mail: 0
> days_before_renewal_mail: 0
>     renewal_mail_subject:
>        renewal_mail_from:
>      renewal_mail_header:
>      renewal_mail_footer:
>      revoked_valid_until: 0
> ```

I was able to replicte the issue on a fresh install.

Fixed in https://github.com/inverse-inc/packetfence/pull/6844/commits/34b899f9b9e6b9c8ae272075594a0d779f591b5b

PR #6853, opened 2022-01-27
Allow to centralize the virtual IPs on the same server

### LINKED ISSUE #6849: Allow for registration/isolation virtual IP to be on 1st node of cluster
When creating a dedicated load-balancer in a cluster, having the VIP for reg/isol on the 2nd server defeats the purpose of that use-case. We should have an option to make all VIP (mgmt+reg+isol+inline) to be on the 1st server of the cluster like it used to be

This new option would be disabled by default and would leave the reg+isol VIP on the 2nd server unless enabled which keeps the current behavior

Add the setting to have all VIP be on the 1st server

julsemaan:
# Description
Allow to centralize the virtual IPs on the same server

# Impacts
keepalived

# Issue
fixes #6849 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Allow to centralize the virtual IPs on the same server

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1558,6 +1558,13 @@ description=<<EOT
 Defines the MariaDB password for Galera replication
 EOT
 
+[active_active.centralize_vips]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Centralize the virtual IP addresses on the same node instead of distributing them on the two first nodes of the cluster
+EOT
+
 [parking]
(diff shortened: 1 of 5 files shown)


julsemaan:
@fdurand, I'd like to see this in 11.2 if you're happy with it

fdurand:
Looks good, let´s merge

PR #6855, opened 2022-01-27
Feature: UI Cluster Services 

### LINKED ISSUE #5576: The service restart widget in the admin only restarts the service on a single node
**Describe the bug**
When you use the service restart widget (like in the SSL certs section) and you restart a service from there, it only restarts it on the server you're on, not for the whole cluster.

**To Reproduce**
1. Go in SSL certificates section on a cluster
2. Restart one of the listed services
3. Only restarted on a single node

**Expected behavior**
We should either block this option when in cluster and instruct the user to use the command line, or better call each cluster member to restart the service.

### LINKED ISSUE #6867: redis_queue service restart
**Describe the bug**
Restarting `redis_queue` produces an error

`POST /api/v1/service/redis_queue/restart ` with payload `{async: true}`

Polling with task_id:

`GET /api/v1/pfqueue/task/ApiTask:36D348D0-85CD-11EC-B8CF-5EB0D8505FB5/status/poll`

Restart completes, the process has a new PID, but the response includes the error:

```json
{"message":"Unable to check if job status exists in redis database","status":500}
```

Furthermore, restarting `redis_queue` synchronously on the same host with another service (eg: `windbindd`) produces a payload where the error is included in the response more than once:

```json
{"message":"Unable to get job status from redis database","status":500}{"message":"Unable to get job status from redis database","status":500}
```

**Expected behavior**
* pfqueue polling should not return an error when the service command succeeds.
* pfqueue response error should be singular, even when run synchronously with other service commands.

satkunas:
# Description
Implement a cluster store module with per-host services handling for any/all members of a cluster. Unify service, system_service and systemd API calls. Deprecates 2 existing store modules for service (local only) and cluster_services.

# Impacts
* Configurator service handling
* Status > Services
* Status > Cluster Services

# Issue
- The service restart widget in the admin only restarts the service on a single node (#5576)
- web admin(Services): Restart all button ignore services not required (#6942)
- v12: hostname of server doesn't appear on Status page (#7096)
- redis_queue service restart (#6867)

# Delete branch after merge
YES 

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Unified service store module allowing control of both local - and cluster members - services.

## Bug Fixes
Fixes #5576 
Fixes #6855
Fixes #6867
Deprecates #7072


julsemaan:
I can't wait to use this on my cluserts

extrafu:
@cgx Make sure to review this quickly

html/pfappserver/root/src/store/modules/cluster.js
```diff
@@ -0,0 +1,562 @@
...
+      const async = (idx = 0) => {
+        dispatch('getService', { server: state.servers[idx], id })
+          .finally(() => {
+            if (idx < state.servers.length-1) {
+              async(++idx)
+            }
+          })
+      }
+      async()
+*/
```
cgx: Can we remove this commented block?

satkunas:
@jrouzierinverse using the `X-PacketFence-Server` header to proxy a request through haproxy to another member of the cluster works, but the login token does not appear to be shared among the members, so this fails with a `401 "Invalid token. Login again using /api/v1/login"`. Is there a way that the same login token can be used with all members?

satkunas:
Fixes also needed with #6867

satkunas:
@jrouzierinverse 

`{async: true}` support is needed in `/api/v1/configurator/system_service/`

satkunas:
@cgx conflicts fixed and merged.

html/pfappserver/root/src/components/new/BaseButtonService.vue
```diff
@@ -145,159 +151,107 @@ const setup = (props, context) => {
...
+    emit('start', { server, id: service.value })
+  }).catch(() => {
+    $store.dispatch('notification/danger', { url: server, message: i18n.t('Failed to start service <code>{service}</code>.  See the server error logs for more information.', { service: service.value }) })
   })
 
-  const doEnable = () => $store.dispatch('services/enableService', service.value).then(() => {
-    $store.dispatch('notification/info', { message: i18n.t('Service <code>{service}</code> enabled.', { service: service.value }) })
-    emit('enable', service.value)
+  const doStop = server => $store.dispatch('cluster/stopService', { server, id: service.value }).then(() => {
+    $store.dispatch('notification/info', { url: server, message: i18n.t('Service <code>{service}</code> killed.', { service: service.value }) })
```
cgx: I recommend replacing *kill* by *stop* everywhere.

html/pfappserver/root/src/views/Status/services/_components/BaseService.vue
```diff
@@ -0,0 +1,214 @@
...
+        </b-col>
+      </b-row>
+      <b-row v-if="service.message"
+        class="mt-2 mx-0">
+        <b-col class="small text-danger">
+          {{ service.message }}
+        </b-col>
+      </b-row>
+    </b-container>
+    <template v-slot:overlay v-if="service.status && !['loading'].includes(service.status)">
```
cgx: Simplify with `service.status && service.status !== 'loading'`

html/pfappserver/root/src/views/Status/services/_components/BaseService.vue
```diff
@@ -0,0 +1,214 @@
+<template>
+  <b-overlay :show="service.status && !['success', 'error'].includes(service.status)" variant="white">
+    <b-container fluid class="px-0">
+      <b-row class="row-nowrap mx-0" align-v="start">
+        <b-col cols="7">
+          <b-row class="row-nowrap">
```
cgx: `row-nowrap` doesn't seem to be valid class. Please remove it everywhere.

cgx: Add `cols="2"` to force both columns to appear on the same row.

html/pfappserver/root/src/views/Status/services/_components/BaseButtonBulkActions.vue
```diff
@@ -0,0 +1,178 @@
...
+
+import { computed, nextTick, ref, toRefs } from '@vue/composition-api'
+import i18n from '@/utils/locale'
+
+const setup = (props, context) => {
+
+  const buttonRef = ref(null)
+
+  const onClick = event => {
+    event.stopPropagation()
```
cgx: Useless since you used the `stop` event modifier.

cgx:
It would be nice to adopt either `$t()` or `$i18n.t()` but not both.

cgx (changes requested):

satkunas:
> It would be nice to adopt either `$t()` or `$i18n.t()` but not both.

I will change all to the latter since it's more explicit.

html/pfappserver/root/src/views/Status/services/_components/BaseService.vue
satkunas: This was inherited from a class defined in Portal Modules. I moved the class def to the root /styles.

Issue #6710, opened 2021-11-18
Move to Crypt::OpenSSL::PKCS12 version 1.9


jrouzierinverse:
Drop our custom version of Crypt::OpenSSL::PKCS12 when https://github.com/dsully/perl-crypt-openssl-pkcs12 releases the version that supports create_as_string.

nqb:
Thanks for your effort @jrouzierinverse !

jrouzierinverse:
The newer version is now out.

nqb:
I think we will need to update `packetfence-perl` to a newer version. Anyway, it's necessary for #6714

nqb:
Should now be part of `packetfence-perl` 1.2.0

PR #6869, opened 2022-02-07
packefence-perl enhancements

### LINKED ISSUE #6469: packetfence-perl: install a metadata files to list version of each Perl modules
When packetfence-perl package is installed, you have to browse `.pm` file to know at which version a module is installed.

Add a metadata file which list all versions of Perl modules.

nqb:
# Description
- install_cpan.sh script
  - Use CPAN 2.29 version on all distributions with identical configuration
  - Stop at first error
  - Display a progress
  - Allow to downgrade CPAN modules (if one module installed a dependency at latest version and we want an old version, we allow this action)
  - Assert that all dependencies (in `dependencies.csv`) have been installed correctly
- dependencies.csv file
  - Upgrade some dependencies to avoid downgrades
  - Disable tests which failed due to outdated dependencies or IPv6

# Impacts
packetfence-perl (Perl modules shipped with PacketFence)

# Issue
- Related to #6710 and #6714
- fixes #6469 

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Update PacketFence perl dependencies


nqb:
I'm almost good with part 1 : pull correctly all Perl dependencies from CPAN using `install_cpan.sh`.

Once done, I will update packaging files for RHEL et Debian on this PR (minor changes) and push new packages to repositories.

addons/packetfence-perl/install_cpan.sh
```diff
@@ -10,35 +11,98 @@ if [[ ! -f $CsvFile || "$CsvFile" == "" ]]; then
...
+    # 1. configure CPAN with defaults (answer yes)
+    # 2. override default conf to UNINST cpan after upgrade (seems mandatory for EL8)
+    (echo o conf make_install_arg 'UNINST=1'; echo o conf commit)|PERL_MM_USE_DEFAULT=1 ${CPAN_BIN_PATH} &> /dev/null
+
+    # upgrade CPAN
+    ${CPAN_BIN_PATH} -i ANDK/CPAN-${CPAN_VERSION}.tar.gz &> /dev/null
+
+    # display CPAN version
+    ${CPAN_BIN_PATH} -D CPAN
+    echo "CPAN upgraded"
```
JeGoi: Perhaps check the output of the previous line before the echo, no?

nqb: Following line: https://github.com/inverse-inc/packetfence/blob/0c928bf32d48a03de8c1aa764d8606cd5a7969cc/addons/packetfence-perl/install_cpan.sh#L2

will stop script if `cpan` exit with a return code different from 0. So this line will never be displayed if there is an error.

nqb:
`packetfence-perl` packages have been uploaded to 11.2 repositories.

I'm running a [pipeline ](https://gitlab.com/inverse-inc/packetfence/-/pipelines/465753209) to be sure new installations will behave as expected.

nqb:
Ready to review @JeGoi. You can find several comments in commit messages.

addons/packetfence-perl/install_cpan.sh
JeGoi: Hum, I wasn't thinking about an error more about if it is an upgrade or not. Like compare versions, or something like that.

addons/packetfence-perl/install_cpan.sh
```diff
@@ -96,10 +177,11 @@ OLDIFS=$IFS
 IFS=','
 while read cpanName cpanVersion cpanInstall cpanTest cpanAll
 do
-  ListCsvModInstall+=( $cpanInstall )
   ListCsvModName+=( $cpanName )
+  ListCsvModVersion+=( $cpanVersion )
```
JeGoi, line 185: To be more consistant this one `ListCsvModVersion` should be declared with the others to be more readable. Like this:
```
ListCsvModInstall=()
ListCsvModName=()
ListCsvModTest=()
ListCsvModVersion=()
```
(172-174)

addons/packetfence-perl/get_modules_installed.pl
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/perl -w
+use ExtUtils::Installed;
+my $inst    = ExtUtils::Installed->new();
+my @modules = $inst->modules();
+ foreach $module (@modules){
```
JeGoi:
```suggestion
foreach $module (@modules){
```

JeGoi (commented):
Just few comments.
Good idea to test if it has been installed like expected. The check on cpan output during the installation process, is not that easy.

addons/packetfence-perl/install_cpan.sh
nqb: I only adjusted text because:
- CPAN installation is idempotent: it will reinstall CPAN at specific version even if CPAN is already installed at that version
- This text should not appear if there is an error because script will exit before

nqb:
@JeGoi, I've adjusted code based on your comments.

I need to adjust one more thing because it looks like my latest tests failed.

JeGoi (approved):
Ok, nothing pops up.
Looks good for me.

nqb:
@JeGoi, no more things to add on my side. Please merge if all is fine for you.

PR #6874, opened 2022-02-10
Disable vulnerable ciphers in HAProxy configuration

nqb:
# Description
Disable vulnerable ciphers in HAProxy configuration (portal and admin)

# Impacts
Captive portal and web admin

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Explicitely disable TLS 1.0 in HAProxy config
* Remove vulnerable CBC ciphers and FIXME


julsemaan:
Some old devices will still use TLS 1.0 and may not display the portal correctly.

Similarly to disabling TLS 1.0 in radiusd, we must offer the choice to the user to enable it in the configuration (captive portal section looks logic)

For haproxy-admin I think we can enforce them without allowing users to change it in the admin interface so my comment only applies to haproxy-portal

nqb:
I agree. In that case, perhaps we can merge changes in this PR, except disable of TLS 1.0 on captive portal to handle it in another PR.

julsemaan:
> I agree. In that case, perhaps we can merge changes in this PR, except disable of TLS 1.0 on captive portal to handle it in another PR.

Are we sure the ciphers we'll push are compatible with TLS 1.0?

nqb:
I follow 2016-08 policy described in https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html to set up ciphers. Because this policy is compatible with TLS 1.0, I would say yes.

julsemaan:
> I follow 2016-08 policy described in https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-policy-table.html to set up ciphers. Because this policy is compatible with TLS 1.0, I would say yes.

let's go with that

PR #6879, opened 2022-02-11
fix/6701

### LINKED ISSUE #6701: Add the online/offline to everywhere the node is seen in the admin
Since online/offline is considered by a lot of customers as an important metric for the node, adding it everywhere we see the node would make sense

We should add it at least in:
- The list of devices of a user (we currently see reg/unreg and should add online/offline)
- The 'Info' tab of a node so we don't have to go back to the list of nodes to view that info when viewing/editing a node

satkunas:
# Description
Adds node statuses to UI

Fixes #6701


satkunas:
@jrouzierinverse `online` columns is needed w/ `GET /api/v1/user/default/nodes` is order to add a status to user devices. I can include a `?fields=online` to the request if you want to defer default joining the ip4log table.

I'll need to make some additional changes after the endpoint is updated.

extrafu:
@satkunas What's the status of this PR?

satkunas:
> @satkunas What's the status of this PR?

Still waiting on the additional column.

PR #6900, opened 2022-02-16
fix/6883

### LINKED ISSUE #6883: Can't move cursor in filter engine 'time period within' value
**Describe the bug**
The 'time period within' value will usually contain large values and its impossible to use the mouse to either select or move into that field. Moving using the keyboard left/right works

**To Reproduce**
1. Create a filter engine with a condition 'time period within' with value 'wd {Mon Tue Wed Thu Fri} hr {9am-11am}, wd {Mon Tue Wed Thu Fri} hr {2pm-4am}'
2. Try to select a part of this value with your mouse

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Firefox
 - Version 96.0.3

satkunas:
# Description
Use textareea within filter engine condition in order to display entire value within draggable parent. Add CSS to collapse textarea by. default and expand on hover or focus. Changes all input to textarea.

# Issue
fixes #6883 

# Delete branch after merge
YES

### DIFF
--- a/html/pfappserver/root/src/views/Configuration/filterEngines/_components/BaseConditionValue.vue
+++ b/html/pfappserver/root/src/views/Configuration/filterEngines/_components/BaseConditionValue.vue
@@ -29,12 +29,22 @@
         :disabled="disabled"
       />
 
-      <base-input-group-textarea v-else
-        class="flex-grow-1 m-1"
-        :namespace="`${namespace}.value`"
-        :disabled="disabled"
-        autoFit
-      />
+      <template v-else>
+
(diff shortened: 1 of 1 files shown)


julsemaan:
Is this limited to when we're using time_period_within?

Because like we discussed this morning, having the field be a text-area for MAC addresses and other short values won't be ideal

satkunas:
> Is this limited to when we're using time_period_within?

@julsemaan it's not limited. I'll adjust.

satkunas:
@julsemaan it's now restricted to time period only

PR #6957, opened 2022-04-01
Added Aruba-MPSK-password attribute

nqb:
# Description
Add support for MPSK on Aruba

# Impacts
General Aruba Perl (Aruba.pm) switch module

# Delete branch after merge
YES

# NEWS file entries
## Enhancements
* Support Aruba MPSK

### DIFF
--- a/lib/pf/Switch/Aruba.pm
+++ b/lib/pf/Switch/Aruba.pm
@@ -657,6 +657,77 @@ sub returnAuthorizeRead {
    return [$status, %$radius_reply_ref];
 }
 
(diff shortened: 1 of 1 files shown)


nqb:
Works as expected in my devel environment. Aruba-MPSK-Passphrase value is returned as bytes to network devices.

I will perform additional tests next week and report.

fdurand:
If it´s still ok then let´s merge

fdurand:
We should add it in PF14, this PR is there for a while.

PR #6969, opened 2022-04-15
Remote unused potentially dangerous caddy middlewares

julsemaan:
# Description
Remote unused potentially dangerous caddy middlewares

# Impacts
All caddy based services

# Delete branch after merge
YES


jrouzierinverse:
Are there more we can remove?

PR #7001, opened 2022-05-18
Feature/pfcron container

JeGoi:
# Description
Add pfcron in a container

# Impacts
pfcron will be in a container

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -526,6 +526,7 @@ pfdeb_based_dev:
           - "pfperl-api"
           - "pfconnector"
           - "pfpki"
+          - "pfcron"
 
 img_dev:
   extends:
@@ -588,6 +589,7 @@ pfdeb_based_br_maint:
           - "pfperl-api"
           - "pfconnector"
           - "pfpki"
+          - "pfcron"
 
 img_br_maint:
   extends:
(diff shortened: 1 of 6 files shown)


julsemaan:
Just to make sure, you did validate the tasks were able to run correctly?
(Including the ones that get done via a call to pfperl-api)

JeGoi:
Tests done against api and DB. It looks fine.
Thanks @jrouzierinverse and @julsemaan

jrouzierinverse:
Is the fingerbank files shared among the containers?

julsemaan:
> Is the fingerbank files shared among the containers?

Not unless its explicitly exposed (its not in this case)

Looks like the Fingerbank DB update in pfcron goes through the httpd.webservices so that is the one that would require the mount point `-v /usr/local/fingerbank/db:/usr/local/fingerbank/db` from a quick code read. @jrouzierinverse, I'll let you confirm my analysis

jrouzierinverse:
You are correct httpd.webservices need access to the fingerbank db

JeGoi:
It looks that is a default:
https://github.com/inverse-inc/packetfence/blob/devel/containers/systemd-service#L15

julsemaan:
@JeGoi, please rebase and it should be ready to merge

JeGoi:
@julsemaan Rebase on devel is done.

PR #7015, opened 2022-06-01
Fingerbank Upstream DB image builder

julsemaan:
# Description
Build a docker image that contains the Fingerbank upstream DB so that it can be mounted in k8s deployments
This only creates the image, we'll need to setup a schedule that publishes it on a regular basis to keep it up to date

# Impacts
Gitlab builds

# Delete branch after merge
YES

### DIFF
--- a/containers/fingerbank-db/Dockerfile
+++ b/containers/fingerbank-db/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine
+
+ARG FINGERBANK_API_KEY
+
+RUN apk update && apk add curl
+
+RUN mkdir -p /usr/local/fingerbank/db
+
+RUN curl -H "Authorization: Bearer ${FINGERBANK_API_KEY}" https://api.fingerbank.org/api/v2/download/db > /usr/local/fingerbank/db/fingerbank_Upstream.db
+
--- a/containers/fingerbank-db/Dockerfile
+++ b/containers/fingerbank-db/Dockerfile
@@ -6,5 +6,5 @@ RUN apk update && apk add curl
 
 RUN mkdir -p /usr/local/fingerbank/db
 
-RUN curl -H "Authorization: Bearer ${FINGERBANK_BUILD_API_KEY}" https://api.fingerbank.org/api/v2/download/db > /usr/local/fingerbank/db/fingerbank_Upstream.db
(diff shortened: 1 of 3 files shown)


containers/fingerbank-db/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM alpine
+
+ARG FINGERBANK_BUILD_API_KEY
+
+RUN apk update && apk add curl
+
+RUN mkdir -p /usr/local/fingerbank/db
+
+RUN curl --fail -H "Authorization: Bearer ${FINGERBANK_BUILD_API_KEY}" https://api.fingerbank.org/api/v2/download/db > /usr/local/fingerbank/db/fingerbank_Upstream.db
```
nqb: I suggest to add a retry option like we have during build of Fingerbank perl client:

https://github.com/fingerbank/perl-client/blob/e1ae4b2d6aacf8c90c18b1f3e82ec8dfc449cfe6/Makefile#L135-L137

nqb (changes requested):

nqb:
Two remarks:
- regarding schedule to build this image, at which frequency should we schedule it ?

I'm asking because if it's a different schedule than other images, it could be interesting to create a dedicated job for that image in place of reusing a job shared by several images (haproxy-portal at the moment)

- Before merging, I think we should adjust build of https://github.com/fingerbank/perl-client/ to remove DB from packaging.

julsemaan:
> * regarding schedule to build this image, at which frequency should we schedule it ?
> 
> I'm asking because if it's a different schedule than other images, it could be interesting to create a dedicated job for that image in place of reusing a job shared by several images (haproxy-portal at the moment)

The DB is updated on a daily basis so 24h would make sense

> * Before merging, I think we should adjust build of https://github.com/fingerbank/perl-client/ to remove DB from packaging.

The intent (atm) is to have this be used only in k8s environment and leave the classic PF untouched where it gets a DB at install and updates it via pfcron. That is going to stay true for 12.0 until the Fingerbank integration is reworked to support contenarization. This just paves the way for that but doesn't actually do anything.

julsemaan:
I added the retries and launched a pipeline to test it out: https://gitlab.com/inverse-inc/packetfence/-/pipelines/555111306

I think that if it passes we're good to merge unless @nqb has other adjustments he can suggest

nqb (approved):

nqb:
I adjusted weekly schedule on `devel` to rebuild all container images each day.

PR #7017, opened 2022-06-03
add radius filter to use ttls-pap username from inner tunnel in reply

julsemaan:
# Description
Makes the User-Name sent in a TTLS-PAP reply use the identity of the inner tunnel

# Impacts
EAP-TTLS PAP

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* The User-Name value in an EAP-TTLS PAP reply will always be the identity of the inner tunnel

### DIFF
--- a/conf/radius_filters.conf.defaults
+++ b/conf/radius_filters.conf.defaults
@@ -1,3 +1,13 @@
+[pap-username-from-tunnel]
+merge_answer=yes
+status=enabled
+scopes=returnRadiusAccessAccept
+condition=connection_type == "Wireless-802.11-EAP" && connection_sub_type == "EAP-TTLS"
+description=papusername
+top_op=and
+radius_status=RLM_MODULE_OK
+answer.0=reply:User-Name = ${radius_request.User-Name}
+
 [EXAMPLE_Ethernet-EAP-Accept]
 status=disabled
 description=Return Access-Accept with VSA when the connection is Ethernet-EAP and when there is no security event
--- a/conf/radius_filters.conf.defaults
+++ b/conf/radius_filters.conf.defaults
@@ -1,13 +1,3 @@
-[pap-username-from-tunnel]
-merge_answer=yes
-status=enabled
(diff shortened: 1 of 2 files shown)


julsemaan:
One downside I see using this approach (RADIUS filter) is that if the user defines another filter that matches after this one in EAP-TTLS PAP, it will never match. I see 2 solutions for this:

- Allow more than 1 filter to match in the RADIUS filters (could have a use outside of the scope of this PR too)
- Implement this in the code directly in radius.pm instead of through a filter

Happy to hear the opinion of the team around this

fdurand:
i have a concern about that, if we do that it mean that it will remove the anonymous identity for the eap-ttls.
Btw with peap we use that in the peap section: "use_tunneled_reply = yes" just because we compute the answer in packetfence-tunnel/post-auth.
So i am not sure if it´s ok or no.

Also we can do it directly in unlang:

```
if ( outer.EAP-Type == TTLS) {
update {
 &reply:User-Name = %{User-Name}
}

julsemaan:
> i have a concern about that, if we do that it mean that it will remove the anonymous identity for the eap-ttls. Btw with peap we use that in the peap section: "use_tunneled_reply = yes" just because we compute the answer in packetfence-tunnel/post-auth. So i am not sure if it´s ok or no.
> 
> Also we can do it directly in unlang:
> 
> ```
> if ( outer.EAP-Type == TTLS) {
> update {
>  &reply:User-Name = %{User-Name}
> }
> ```

I think the unlang way would be better. AFAIK, the attributes are properly tunneled but the outer tunnel handler adds the User-Name even though nowhere in the configuration it says to do so.

I think I'll try to implement your solution instead, would `post_auth` be the best section to put that in `packetfence-tunnel`?

fdurand:
I think you can add it in packetfence-tunnel in the authorized section (there is already a test with TTLS)

fdurand:
The logic as been ported in unlang instead of the radius filter.

PR #6813, opened 2022-01-11
Feature/venom eap tls pfpki

JeGoi:
# Description
Add venom eap tls pfpki with venom 1.x feature

# Impacts
Add new test

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Pass CI test

PR #7018, opened 2022-06-07
Improve stability of tests on maintenance (11.2 and after)

nqb:
# Description
Fixes:
- keep VM available if a test failed, destroy VM if a test passed, extract logs in both cases. Cleanup job has been deployed on runners
- disable apt-daily timers to prevent lock during initial upgrade
- increase SSH timeout when booting Vagrant VM
- start services at end of configurator using "poll" mechanism to avoid 504 errors
- create dedicated AD domain (example1, example2, etc.) for each test suite
- keep more logs on node* VM in place of log of latest test suite
- share common VM (ad, switch01) per branch

TODO:
- ~[ ] upgrade Ansible on runners to 2.12 to benefit from (https://github.com/ansible/ansible/commit/ee725846f070fc6b0dd79b5e8c5199ec652faf87) which should limit timeout when calling ansible-galaxy API~

**Please note** that this PR is opened against `maintenance/11.2` branch and not `devel` branch.

# Impacts
Tests

# Delete branch after merge
*NO*

# TODO post-merge
- [X] Need to backport things to maintenance/11.1 and maintenance/11.0
- [ ] Need to backport possible things to devel. Venom code has changed a lot between `maintenance/11.2` and `devel` due to #6813
- [X] Re-enable tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -401,7 +401,15 @@ variables:
   script:
     - MAKE_TARGET=run make -e -C ${TESTDIR} ${CI_JOB_NAME}
   after_script:
-    - MAKE_TARGET=halt make -e -C ${TESTDIR} ${CI_JOB_NAME}
+    - >
+      if [ $CI_JOB_STATUS == 'success' ]; then
+        echo 'Passed tests, no need to keep VM'
+        MAKE_TARGET=clean make -e -C ${TESTDIR} ${CI_JOB_NAME}
+      else
+        echo 'Failed tests, shutdown VM to analyze issue'
+        MAKE_TARGET=halt make -e -C ${TESTDIR} ${CI_JOB_NAME}
+      fi
+
 
 ################################################################################
 # JOBS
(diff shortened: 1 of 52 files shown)


nqb:
Requirements to upgrade Ansible (ansible-base 2.10 to ansible-core 2.12) are too important:
- https://docs.ansible.com/ansible-core/devel/porting_guides/porting_guide_core_2.11.html
- https://docs.ansible.com/ansible-core/devel/porting_guides/porting_guide_core_2.12.html

Especially: Python interpreter needs to be 3.8 and we are running 3.7.3 on runners because we still use Debian Buster (oldstable). The correct way to handle that is to upgrade our runners from Buster to Bullseye but it involve a lot of efforts.

I'm putting Ansible upgrade aside for now. We will have to deal with some timeout during CI jobs.

nqb:
Nothing more to change, I'm running a new pipeline:

https://gitlab.com/inverse-inc/packetfence/-/pipelines/557667295

JeGoi (approved):
Ok the code looks fine.
Nothing pops up, except: how it will be ported to devel...
And we really should reduce the amount of replicated code! But that is another story, we have to write. 
Lets go!

PR #7019, opened 2022-06-07
Feature/pfpki container

JeGoi:
# Description
Add pfpki in container

# Impacts
pfpki is now in a container

# Delete branch after merge
YES

# Checklist
- [./ ] Tests done


julsemaan:
You seemed to have mixed the pfcron branch with this one here, make sure it's properly split since you do have a PR for pfcron as well

JeGoi:
Will use devel instead of pfcron branch

PR #7020, opened 2022-06-07
[Containers] pfpki: add as a container

JeGoi:
# Description
Add pfpki in container

# Impacts
pfpki is now in a container

# Delete branch after merge
YES

# Checklist
- [./ ] Tests done


containers/pfpki/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY ./lib /usr/local/pf/lib
+COPY ./config.mk /usr/local/pf/config.mk
+COPY ./Makefile /usr/local/pf/Makefile
+RUN cd /usr/local/pf/go/ && \
+    make pfhttpd
+
+FROM ${KNK_REGISTRY_URL}/pfdebian:${IMAGE_TAG}
+WORKDIR /usr/local/pf/
+COPY --from=0 /usr/local/pf/go/pfhttpd /usr/local/pf/sbin/pfhttpd
+ENTRYPOINT /bin/bash -c "export $(cat /usr/local/pf/var/conf/pfpki.env | xargs) && /usr/local/pf/sbin/pfhttpd -conf /usr/local/pf/conf/caddy-services/pfpki.conf -log-name=pfpki"
```
julsemaan: You do not need to load /usr/local/pf/var/conf/pfpki.env in the env, it will be done implicitely (api-frontend is an example of that)

FYI, where it's done:
https://github.com/inverse-inc/packetfence/blob/devel/containers/systemd-service#L17-L19

julsemaan (changes requested):

containers/pfpki/Dockerfile
JeGoi: Fixed

julsemaan:
Launched a pipeline, if it succeeds I'll merge

https://gitlab.com/inverse-inc/packetfence/-/pipelines/558556203

JeGoi:
Unless I missed something, it looks good.

PR #7024, opened 2022-06-14
Improve pfconnector support in switch modules

julsemaan:
# Description
Allow to use the connector to connect to switch via SNMP
Allow configurability of using the connector during the CoA/Disconnect
Move RADIUS related fields in switches to the RADIUS tab (in the admin)

# Impacts
SNMP switch connect, RADIUS deauth

# Delete branch after merge
YES


fdurand:
The code is ok.
Just maybe set the value to disabled by default.

julsemaan:
> The code is ok. Just maybe set the value to disabled by default.

The intent with it being enabled by default is to get the connector to be broadly used by our 'classic' user base so that we can prove the technology without users adopting remote connectors. That said, if it proves to be unreliable, we can push a disable through the maintenance easily.

extrafu:
@fdurand OK to merge?

PR #7025, opened 2022-06-14
Auth sources via pfconnector

julsemaan:
# Description
Allow to use the connector in LDAP(all types) and RADIUS sources

# Impacts
LDAP and RADIUS sources

# Delete branch after merge
YES

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Source/LDAP.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Source/LDAP.pm
@@ -198,6 +198,14 @@ has_field 'shuffle',
    default => $META->get_attribute('shuffle')->default,
 );
 
+has_field 'use_connector',
+  (
+   type => 'Toggle',
+   checkbox_value => '1',
+   unchecked_value => '0',
+   default => $META->get_attribute('use_connector')->default,
+);
+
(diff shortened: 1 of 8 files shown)


fdurand:
For internal PacketFence authentication sources it´s ok.
For eap-ttls we will need to find a way to make it works.

Also we should add a upgrade script to append use_connector value.

Last thing, isn´t it better to set it to disabled by default ?

julsemaan:
> For internal PacketFence authentication sources it´s ok. For eap-ttls we will need to find a way to make it works.

True, I did not think of EAP-TTLS, we could expose static tunneled binds for LDAP sources and point FreeRADIUS to them. They wouldn't need to be dynamic since the amount of LDAP auth sources is always relatively small (can't see a customer with more than 100 sources which means 100 used ports). If we're all happy with that, we can see if we fit this in the scope of 12.0

> Also we should add a upgrade script to append use_connector value.

Thanks to @jrouzierinverse, there is already a script that runs post-upgrade that does that for all values in the auth sources

> Last thing, isn´t it better to set it to disabled by default ?

Let's discuss that here: https://github.com/inverse-inc/packetfence/pull/7024#issuecomment-1155658188

extrafu:
@fdurand OK to merge?

PR #7026, opened 2022-06-14
Feature/container systemd notify

JeGoi:
# Description
Allow systemd notify from a container
It is using the following wrapper https://github.com/bcsaller/sdnotify with https://github.com/coreos/sdnotify-proxy

# Impacts
Service can use notify.
Among other positif impact, it has an impact on venom tests.

# Delete branch after merge
YES

# Checklist
- [x] Need to find a way to integrate https://github.com/coreos/sdnotify-proxy
- [x] Tests done on dev machine

# NEWS file entries
## New Features
* Services in containers with notifications

### DIFF
--- a/containers/systemd-service
+++ b/containers/systemd-service
@@ -60,7 +60,7 @@ function run {
 }
 
 function sdproxy() {
-  SD_NOTIFY_PROXY=/usr/local/pf/bin/sdnotify-proxy
+  SD_NOTIFY_PROXY=/usr/local/pf/sbin/sdnotify-proxy
   SD_SOCK=/usr/local/pf/var/run/$1-systemd-notify.sock
   WATCHDOG_USEC_D=10000000
   $SD_NOTIFY_PROXY $SD_SOCK /usr/bin/docker run --volume=$SD_SOCK:$SD_SOCK --env NOTIFY_SOCKET=$SD_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}
--- a/containers/systemd-service
+++ b/containers/systemd-service
@@ -61,11 +61,9 @@ function run {
 
 function sdproxy() {
   SD_NOTIFY_PROXY=/usr/local/pf/bin/sdnotify-proxy
-  SD_SOCK=/tmp/$1.sock
-  CR_SOCK=/var/run/$1.sock
+  SD_SOCK=/usr/local/pf/var/run/$1-systemd-notify.sock
   WATCHDOG_USEC_D=10000000
-  echo "SDPROXY cli: $SD_NOTIFY_PROXY $SD_SOCK /usr/bin/docker run --volume=$SD_SOCK:$CR_SOCK --env NOTIFY_SOCKET=$CR_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}"
-  $SD_NOTIFY_PROXY $CR_SOCK /usr/bin/docker run --volume=$CR_SOCK:$CR_SOCK --env NOTIFY_SOCKET=$CR_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}
+  $SD_NOTIFY_PROXY $SD_SOCK /usr/bin/docker run --volume=$SD_SOCK:$SD_SOCK --env NOTIFY_SOCKET=$SD_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}
 }
 
 function run_img {
@@ -75,6 +73,6 @@ function run_img {
   cmd="$4"
   cd /usr/local/pf/
   docker container rm -f $name 1&2> /dev/null
-  #echo "Running with args $args"
+  echo "Running with args $args"
   sdproxy $name $args --name=$name $img $cmd
 }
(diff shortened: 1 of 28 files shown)


containers/systemd-service
```diff
@@ -59,13 +57,22 @@ function run {
   fi
 }
 
+function sdproxy() {
+  SD_NOTIFY_PROXY=/usr/local/pf/bin/sdnotify-proxy
+  SD_SOCK=/tmp/$1.sock
```
julsemaan: IMO we should change this to `/tmp/$1-systemd-notify.sock`

containers/systemd-service
```diff
@@ -59,13 +57,22 @@ function run {
...
+  CR_SOCK=/var/run/$1.sock
```
julsemaan: IMO we use the same socket path in the container and outside to help things stay consistent

containers/systemd-service
```diff
@@ -59,13 +57,22 @@ function run {
...
+  WATCHDOG_USEC_D=10000000
+  echo "SDPROXY cli: $SD_NOTIFY_PROXY $SD_SOCK /usr/bin/docker run --volume=$SD_SOCK:$CR_SOCK --env NOTIFY_SOCKET=$CR_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}"
+  $SD_NOTIFY_PROXY $CR_SOCK /usr/bin/docker run --volume=$CR_SOCK:$CR_SOCK --env NOTIFY_SOCKET=$CR_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}
```
julsemaan: With the changes above this line would become:
```
  $SD_NOTIFY_PROXY $SD_SOCK /usr/bin/docker run --volume=$SD_SOCK:$SD_SOCK --env NOTIFY_SOCKET=$SD_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}
```

containers/systemd-service
```diff
@@ -59,13 +57,22 @@ function run {
   fi
 }
 
+function sdproxy() {
+  SD_NOTIFY_PROXY=/usr/local/pf/bin/sdnotify-proxy
+  SD_SOCK=/tmp/$1.sock
+  CR_SOCK=/var/run/$1.sock
+  WATCHDOG_USEC_D=10000000
+  echo "SDPROXY cli: $SD_NOTIFY_PROXY $SD_SOCK /usr/bin/docker run --volume=$SD_SOCK:$CR_SOCK --env NOTIFY_SOCKET=$CR_SOCK --env WATCHDOG_USEC=$WATCHDOG_USEC_D ${@:2}"
```
julsemaan: I'd remove this line here since keeping this line and the one below may lead them to be out of sync

containers/systemd-service
```diff
@@ -59,13 +57,22 @@ function run {
...
   img="$1"
   name="$2"
   args="$3"
   cmd="$4"
   cd /usr/local/pf/
-  docker container rm -f $name
-  echo "Running with args $args"
-  docker run $args --name=$name $img $cmd
+  docker container rm -f $name 1&2> /dev/null
+  #echo "Running with args $args"
```
julsemaan: Keeping this line here can be useful for troubleshooting

julsemaan (changes requested):

containers/systemd-service
JeGoi: I had an issue with that for the haproxy-portal. I had to move it in /var/run/ in the container to make it works.
I have to retest that part to be sure.

containers/systemd-service
JeGoi: It has been replaced by "SDPROXY cli" but as you mentioned it could be out of sync with the command line.
I will bring it back.

containers/systemd-service
JeGoi: I still do not know why but /tmp/xx.sock is failing with haproxy-portal but i was able to make it work with /usr/local/pf/var/run/ for exemple.

julsemaan: In this case we can use the same value for both and use `/usr/local/pf/var/run/$1-systemd-notify.sock`

What do you think?

JeGoi: Agree. I will try with all containers.

containers/systemd-service
JeGoi: Done

containers/systemd-service
JeGoi: Done

containers/systemd-service
JeGoi: Done

JeGoi:
Pipeline is fine: https://gitlab.com/inverse-inc/packetfence/-/pipelines/565856752

nqb:
@JeGoi, my 2 cents but are we sure haproxy is installed at same version as in previous image ?

julsemaan:
Prior to merging we need sdnotify-proxy packaged as a Golang binary in our packaging and installed in /usr/local/pf/sbin/ and then change the path in containers/systemd-service

Because right now if we merge as-is, my understanding is that we have to compile and install this manually on our systems

JeGoi:
> @JeGoi, my 2 cents but are we sure haproxy is installed at same version as in previous image ?

Love your 2 cents. I will fix it and I am 99% sure that is the 2.4.17-1...
And it looks that we need something >=2.2 https://github.com/inverse-inc/packetfence/blob/v11.2.0/debian/control#L148
So, it should be fine.

julsemaan:
Configurator works and all services start/stop correctly.

I'll do the rebase and merge it

PR #7027, opened 2022-06-15
feature/multi-sql-reports

jrouzierinverse:
# Description
Remove the use of BEGIN NOT ATOMIC

# Impacts
reports.conf

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries

## Enhancements
* Make reports compatible with MySQL


satkunas (approved):
SQL looks good.

We should also update the docs in the parts where we mention "BEGIN NOT ATOMIC", so it's MySQL8 friendly. 

https://github.com/inverse-inc/packetfence/blob/devel/docs/installation/advanced_topics.asciidoc

PR #7028, opened 2022-06-15
Docker iptables systemd service

fdurand:
# Description
Create a new systemd service to manage the docker iptables rules.
All the services who run in docker is dependent of this service and launch the packetfence-docker-iptables service once started 

# Impacts
iptables rules

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests


julsemaan:
I think it would make sense to have a lib/pf/services/manager/ for that service so that it can be managed via pfcmd, unless you see a reason not to

fdurand:
pfcmd generatedockeriptables subcommand added.

PR #7035, opened 2022-06-20
Only apply NatDNS if it´s an inline network

fdurand:
# Description
(REQUIRED)
ie: Allow PacketFence to trigger the coffee machine on wireless EAP 802.1X connection to brew a fresh espresso.

# Impacts
pfdhcp

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Fixed ShuffleDNS logic (only apply to inline network)

### DIFF
--- a/go/cmd/pfdhcp/utils.go
+++ b/go/cmd/pfdhcp/utils.go
@@ -193,9 +193,12 @@ func NodeInformation(ctx context.Context, target net.HardwareAddr) (r NodeInfo) 
 
 // ShuffleDNS return the dns list
 func ShuffleDNS(ConfNet pfconfigdriver.RessourseNetworkConf) (r []byte) {
-	if !sharedutils.IsEnabled(ConfNet.NatDNS) {
-		var excluded []string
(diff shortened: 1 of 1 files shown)


nqb:
This PR will fix following issue: in a cluster environment with `dns_on_vip_only` enabled, `pfdhcp` returns IP address of current server in cluster in place of VIP.

nqb:
I wonder if we should backport that fix because it's has the potential to cause issues:
* before fix: DNS server was IP address of each cluster member
* after fix: DNS server will be VIP only

If somebody has put firewall rules which doesn't allow DNS requests on VIP, DNS inside registration VLAN will not work anymore.
I doubt that is the case because if you enabled `dns_on_vip_only`, I doesn't make sense to restrict DNS traffic to VIP.

nqb (approved):

nqb:
@fdurand,

FYI, on a setup with L3 registration VLANs, it looks like `nat_dns` is enabled by default which mean for these setups, we already return DNS VIP in DHCP response.

PR #7044, opened 2022-06-28
Replace haproxy_db with proxysql + modification for cluster

fdurand:
# Description
Replaced haproxy_db with proxysql (mariadb galera cluster)
Fixes to make packetfence cluster on prem working

# Impacts
A lot

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Proxysql support

# UPGRADE file entries
Upgrade script to execute:
/usr/local/pf/addons/upgrade/to-12.0-use-proxysql.pl


.gitlab-ci.yml
```diff
@@ -652,7 +654,7 @@ pfdeb_based_rel:
           - "pfperl-api"
           - "pfconnector"
           - "pfpki"
-          - "haproxy-portal"
```
julsemaan, line 655: We need to keep haproxy-portal there

conf/proxysql.conf.example
```diff
@@ -0,0 +1,121 @@
...
+#
+########################################################################################
+
+datadir="/usr/local/pf/var/proxysql/"
+errorlog="/usr/local/pf/logs/proxysql.log"
+
+admin_variables=
+{
+    admin_credentials="admin:admin"
+    mysql_ifaces="0.0.0.0:6032"
```
julsemaan: This should be 127.0.0.1 or 100.64.0.1 to reduce the exposure

conf/proxysql.conf.example
```diff
@@ -0,0 +1,121 @@
...
+    refresh_interval=2000
+    web_enabled=true
```
julsemaan: IMO, this should be disabled by default

julsemaan (commented):
Some quick preliminary comments, will do a full review tomorrow

lib/pf/services/manager/radiusd_child.pm
```diff
@@ -1596,6 +1625,32 @@ sub generate_container_environments {
...
+    my ($self, $port) = @_;
+    my @eduroam_authentication_source = @{pf::authentication::getAuthenticationSourcesByType('Eduroam')};
+    my $eduport = $eduroam_authentication_source[0]{'auth_listening_port'};
+    $self->{auth_port} = "1812";
+    $self->{acct_port} = "1813";
+    $self->{cli_port} = "1815";
+    $self->{eduroam_port} = $eduport || "11812";
+    $self->{radsec_port} = "2083";
+    if ($cluster_enabled) {
+        $self->{auth_port} = "1822";
```
julsemaan: I'd apply the same logic of `$self->{auth_port} + 10` here and in all the ones below to keep this cleaner

conf/systemd/packetfence-proxysql.service
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) Inverse inc.
+[Unit]
+Description=PacketFence ProxySQL Load Balancer for connecting to clustered databases
+Before=packetfence-httpd.portal.service packetfence-api-frontend.service
+Wants=packetfence-config.service 
+Before=packetfence-docker-iptables.service
+Requires=packetfence-docker-iptables.service
+PartOf=packetfence-docker-iptables.service
+
+[Service]
```
julsemaan, line 10: Does proxysql support sdnotify? If so we should definitely consider it since it's a critical service

conf/systemd/packetfence-proxysql.service
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) Inverse inc.
+[Unit]
```
julsemaan, line 2: We should add a monit check for this service

debian/packetfence.preinst
```diff
@@ -56,8 +56,12 @@ case "$1" in
             else
                 useradd -r -d "/usr/local/pf" -s /bin/sh -c "PacketFence" -M pf -g pf
             fi
+            usermod -u 6969 pf
```
julsemaan: We shouldn't be using the static UID/GID based on our latest conversations unless there is another reason why  this should still apply

debian/packetfence.preinst
```diff
@@ -56,8 +56,12 @@ case "$1" in
...
+            groupmod -g 6969 pf
             echo "create pf user"
         else
+                usermod -u 6969 pf
```
julsemaan: Same as my comment above

debian/packetfence.preinst
```diff
@@ -107,6 +111,14 @@ case "$1" in
         /bin/systemctl mask mariadb
         /bin/systemctl stop mariadb
 
+        #Change the uid/gid
+        set +e
```
julsemaan, line 111: And here too for my UID/GID comment

lib/pf/pfcmd/checkup.pm
```diff
@@ -1215,26 +1215,6 @@ sub cluster {
         }
     }
 
-    _check_database_host_in_cluster();
-}
-
-
-sub _check_database_host_in_cluster {
```
julsemaan: Why are we removing this check? This is still a valid check since we do not want to use `localhost` as the host value in the database configuration

rpm/packetfence.spec
```diff
@@ -589,6 +592,19 @@ if ! /usr/bin/id pf &>/dev/null; then
                 ( echo Unexpected error adding user "pf" && exit )
     fi
 fi
+/usr/sbin/usermod -u 6969 pf || \
```
julsemaan: The UID/GID conversation applies here too

sbin/proxysql-docker-wrapper
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+source /usr/local/pf/containers/systemd-service
+
+name=proxysql
+
+args=`base_args $name`
+args="$args -v/usr/local/pf/var/conf/:/usr/local/pf/var/conf/"
+args="$args -v /usr/local/pf/var/proxysql/:/usr/local/pf/var/proxysql/"
+args+=" -p 6032:6032 -p 6033:6033 -p 6080:6080"
```
julsemaan: For 6032, it should actually be `-p 127.0.0.1:6032:127.0.0.1:6032`

julsemaan:
We need an upgrade script that
- Rewrites database.host to 100.64.0.1 when the value is 127.0.0.1
- Rewrites database.port to 6033 when the value is 3306 and database.host is 127.0.0.1 pre-upgrade
- Does the same thing in pfconfig.conf

That means we will be using proxysql by default which is in my opinion the right thing to do in the upgrade.

If we're happy with this logic, we also need to adjust the cluster install guide with the new port value (6033 instead of 3306)

conf/systemd/packetfence-proxysql.service
fdurand: No support for sdnotify by default.

fdurand:
All done

julsemaan:
Adjusted the clustering guide for port 6033

Merging 💪

PR #7048, opened 2022-06-30
feature/batch-802.1X-sql

jrouzierinverse:
# Description
Reduce the number of inserts into the node table.

# Impacts
RADIUS authorize workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries
## Enhancements
* Reduce the number of inserts in the node table.


extrafu:
@jrouzierinverse @julsemaan While doing these changes and gradually merging them, it would be nice to see the performance impacts. We should use a L3 geodistributed cluster and test for example with 100,000 802.1X connections before and after this change.

We could test the same with a local cluster, a geodistributed with one with absolutely no change and update measurements in a Wiki page after each change.

julsemaan:
This change here has a very small impact, I had a discussion with @jrouzierinverse around what it's going to take to achieve the goal and he is going to work on it when he's back. Still wanted to get this in since it has a positive (although small) impact but it's too soon to start quantifying it.

When we get there though, I agree 100% with what you described to quantify the impact of each change

PR #7078, opened 2022-08-05
Package upgrades and container images

nqb:
# Description
Adjust logic related to package upgrades and container images

# Impacts
Installation and upgrades

# Delete branch after merge
YES

### DIFF
--- a/containers/manage-images.sh
+++ b/containers/manage-images.sh
@@ -15,6 +15,8 @@ source <(grep 'LOCAL_REGISTRY' ${PF_SRC_DIR}/config.mk | tr -d ' ')
 source ${PF_SRC_DIR}/conf/build_id
 
 configure_and_check() {
+    # yes=default
+    CLEANUP_IMAGES=${CLEANUP_IMAGES:-yes}
     # find all directories with Dockerfile
     # excluding non necessary images
     DOCKERFILE_DIRS=$(find ${SCRIPT_DIR} -type f -name "Dockerfile" \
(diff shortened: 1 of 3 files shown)


nqb:
I suggest one improvement on this PR: we tagged Docker images on branches and devel with two tags.

So a commit on `devel` with rebuild of Docker images will generate images with two tags:
- `devel-COMMIT-SHA`
- `devel`

In `conf/build_id`, we use `devel-COMMIT-SHA` but if someone want to pull latest Docker images on `devel`, he can use `devel` without having to know ID of latest commit.

EDIT: I did it, test in progress

nqb:
Another important change related to that PR: Docker images build is mandatory for each pipeline. Otherwise, packaging will install a `conf/build_id` in the form: `devel-COMMIT-SHA`. If Docker images have not been rebuild, this tag is invalid.

nqb:
Ready for review.
Use cases tested on Debian & RHEL:
- first installation
- upgrade
- downgrade

I got only images tagged with `conf/build_id` tag each time. Previous images are removed.
Downgrade is performed by installing a previous package which will trigger pull of previous images.

@julsemaan, one thing I didn't really check is the impact on our package registry. 

This PR will obviously created a tag each time we run a pipeline so it will increase a lot registry size but IIRC, as an open source project we are not limited by GitHub.

julsemaan:
I took a quick look at the changes and I'm wondering if we can adjust this so that we only have the commit SHA in the tag for releases (tags) and for specific branches like the maintenance. 

I don't believe keeping all the images for all the commits will cost us anything but it certainly won't be very clean. 

If we can't make it so we keep the images for specific tags and branches, then I think we could have a cleanup script we run a few days after a release to cleanup the devel related images and start fresh since the older images won't be needed anymore (and could be rebuilt if needed)

nqb:
@julsemaan, code adjusted to reflect changes we discussed.

I will create another PR to add build ID inside Docker images.

containers/manage-images.sh
```diff
@@ -47,15 +50,27 @@ tag_images() {
...
 
     echo "$(date) - Tag of images finished"
 }
 
 cleanup_images() {
+    if [ "$CLEANUP_IMAGES" = "yes" ]; then
+        if delete_images; then
+            echo "$(date) - Previous images cleaned"
+        else
+            echo "$(date) - Cleanup has failed"
```
julsemaan:
```suggestion
            echo "$(date) - Cleanup of outdated docker images has failed"
```

debian/packetfence.postinst
```diff
@@ -147,16 +158,15 @@ case "$1" in
...
+        systemctl start packetfence-config
         /usr/local/pf/bin/pfcmd generatemariadbconfig --force
+        # only packetfence-config is running after this command
         systemctl isolate packetfence-base.target
 
         # Override the MySQL socket location to match our standard
         printf '[client-server]\nsocket = /var/lib/mysql/mysql.sock\n' > /etc/mysql/mariadb.conf.d/999-socket-override.cnf
         sed -i 's#^socket\s*=.*#socket=/var/lib/mysql/mysql.sock#' /etc/mysql/my.cnf
 
-        perl /usr/local/pf/addons/upgrade/add-default-params-to-auth.pl
```
julsemaan, line 160: AFAIK, this is necessary to keep, otherwise when we add new parameters to the auth sources, the config entries created on older version become out of sync and this creates various issues

nqb: @jrouzierinverse, can you confirm what @julsemaan said ?

For me, if we add new parameters to auth sources:
- new sources will be created with new parameters
- old sources will need to be updated using an upgrade script to add this new parameter

nqb:
https://github.com/inverse-inc/packetfence/commit/317d64bd69a297d553dfd97c34bb6cdd8bf5a974

debian/packetfence.postinst
nqb: @jrouzierinverse, any comment ?

nqb:
Rebased.

@fdurand, I rebased on `devel` and I adjusted a bit your logic in commit https://github.com/inverse-inc/packetfence/commit/317d64bd69a297d553dfd97c34bb6cdd8bf5a974.

debian/packetfence.postinst
nqb: Script is necessary. I will adjust code.

nqb:
@julsemaan: code adjusted.

PR #7081, opened 2022-08-05
sd_notify ProxySQL Docker image

fdurand:
# Description
Added support for sd_notify in ProxySQL

# Impacts
ProxySQL

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

### DIFF
--- a/containers/proxysql/Dockerfile
+++ b/containers/proxysql/Dockerfile
@@ -31,6 +31,7 @@ RUN apt-get update && apt-get install -y \
         python3 \
         uuid-dev \
         libsystemd-dev \
+        systemd \
   && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/
--- a/containers/proxysql/Dockerfile
+++ b/containers/proxysql/Dockerfile
@@ -54,7 +54,7 @@ ARG release=feature/sd_notify_v2.4.2
 
 RUN mkdir -p /root/ctl
 ARG COMMIT_ID=67f866745d8537519725b1bec37e1483959d2176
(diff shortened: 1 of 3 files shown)


containers/proxysql/Dockerfile
```diff
@@ -1,3 +1,79 @@
...
+
+#
+#  Shallow clone the proxysql source
+#
+ARG source=https://github.com/inverse-inc/proxysql.git
+ARG release=feature/sd_notify_v2.4.2
+
+RUN mkdir -p /root/ctl
+ARG COMMIT_ID=67f866745d8537519725b1bec37e1483959d2176
+RUN git clone -qq --depth 1 --single-branch --branch ${release} ${source}
```
julsemaan: I think we should get rid of the extra args that make it lean because if the COMMIT_ID is not the latest commit of that branch, it will fail to find it when doing the checkout below, so I'd suggest:

```
RUN git clone -qq ${release} ${source}
```

julsemaan:
Changes look good, launched a final pipeline to validate the build and I'll merge:
https://gitlab.com/inverse-inc/packetfence/-/pipelines/607577925

nqb:
@fdurand, can this PR be merged without the ability to disable Kaniko cache ?

fdurand:
@nqb yes it can be merged without that. As @julsemaan say , by adding the commit id, it disable the kaniko cache.

fdurand:
new pipeline is on the way https://gitlab.com/inverse-inc/packetfence/-/pipelines/607594749

PR #7084, opened 2022-08-08
use 100.64.0.1:0 instead of 0.0.0.0:0 so the containers can communicate with each other

### LINKED ISSUE #7083: Portal Preview is not working.
**Describe the bug**
After installing Packetfence and setting the management interface to serve the portal I am unable to preview the portal.

I get the following response.
```json { "status":503, "message":"No server is available to handle this request" } ```

jrouzierinverse:
# Description
Fixes the portal preview

# Issue
fixes #7083

# Delete branch after merge
YES


fdurand:
Tested and works on my setup. Let´s merge

PR #7085, opened 2022-08-09
Show the PSK is the status page

fdurand:
# Description
Added the way to present the dpsk in the status page

# Impacts
Status page

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Present the dpsk if enabled on the status page

### DIFF
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
@@ -4,6 +4,7 @@ use namespace::autoclean;
 use pf::util;
 use pf::constants;
 use pf::config;
+use pf::config qw(%ConfigSelfService);
 use pf::node;
 use pf::person;
 use pf::web;
@@ -31,8 +32,18 @@ sub auto :Private {
     $c->session->{release_bypass} = $TRUE;
     $c->stash->{isDeviceRegEnable} = $c->forward(DeviceRegistration => "isDeviceRegEnabled");
     $c->stash->{stripeCustomerPortalEnabled} = $c->profile->stripeCustomerPortalEnabled;
+    $c->stash->{dynamicPreSharedKeyEnabled} = is_dpsk_enabled($c->profile);
     $c->forward('setupCurrentNodeInfo');
     return 1;
+}
+
+sub is_dpsk_enabled {
+    my ( $profile ) = @_;
+
+    if (exists $profile->{'_self_service'} && isenabled($ConfigSelfService{$profile->{'_self_service'}}{'dynamic_pre_shared_key'}) ) {
+        return $TRUE;
+    }
+    return $FALSE;
 }
 
 =head2 index
@@ -96,6 +107,18 @@ sub userIsAuthenticated : Private {
     $c->stash(
         nodes    => \@nodes,
     );
+    if (is_dpsk_enabled($c->profile)) {
+        my @provisioners = $c->profile->provisionerObjects();
+        unless(@provisioners){
+            return;
+        }
+
+        foreach my $provisioner (@provisioners) {
+            next unless $provisioner->isa("pf::provisioner::dpsk");
+            my $dpsk = $provisioner->generate_dpsk($pid);
+            $c->stash( psk => $dpsk);
+        }
+    }
 }
 
 sub setupCurrentNodeInfo : Private {
--- a/html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
+++ b/html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
@@ -109,9 +109,6 @@ sub userIsAuthenticated : Private {
     );
     if (is_dpsk_enabled($c->profile)) {
         my @provisioners = $c->profile->provisionerObjects();
-        unless(@provisioners){
-            return;
-        }
 
         foreach my $provisioner (@provisioners) {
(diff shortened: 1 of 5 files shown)


html/pfappserver/root/src/views/Configuration/selfServices/_components/TheForm.vue
```diff
@@ -38,6 +38,12 @@
           :text="$i18n.t('List of OS which will be allowed to be register via the self service portal.')"
         />
       </base-form-tab>
+      <base-form-tab :title="$i18n.t('Dynamic Pre Shared Key')">
+        <form-group-dynamic-pre-shared-key namespace="dynamic_pre_shared_key"
+          :column-label="$i18n.t('Show the user pre shared key')"
+          :text="$i18n.t('Enable the view of the pre shared key of the user in the status page.')"
```
julsemaan:
```suggestion
          :text="$i18n.t('Will display the WPA pre-shared key on the status page when a DPSK provisioner is configured.')"
```

html/captive-portal/templates/status.html
```diff
@@ -156,6 +156,16 @@ <h2>[% i18n("Your registered devices") %]</h2>
 </div>
 [%- END %]
 
+[% IF psk.defined %]
+<div class="u-padding">
+  <div class="o-media o-media--notice u-padding u-margin-bottom">
+    <div class="o-media__img">[% flashIcon(level='notice', size='tiny') %]</div>
+    <p class="o-media__body">[% i18n("Your psk key: ") %] [% psk %]</p>
```
julsemaan:
```suggestion
    <p class="o-media__body">[% i18n("Your WPA Wi-Fi key: ") %] [% psk %]</p>
```

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -96,6 +107,18 @@ sub userIsAuthenticated : Private {
...
     );
+    if (is_dpsk_enabled($c->profile)) {
+        my @provisioners = $c->profile->provisionerObjects();
+        unless(@provisioners){
+            return;
+        }
+
+        foreach my $provisioner (@provisioners) {
+            next unless $provisioner->isa("pf::provisioner::dpsk");
+            my $dpsk = $provisioner->generate_dpsk($pid);
```
julsemaan, line 115: If there is more than one, it's going to generate it each time but only one item is in the stash after.
Will it behave correctly if more than one DPSK provisioner is setup on the connection profile?

html/pfappserver/lib/pfappserver/Form/Config/SelfService.pm
```diff
@@ -81,6 +81,15 @@ has_field 'device_registration_allowed_devices' =>
    fingerbank_model => "fingerbank::Model::Device",
   );
 
+has_field 'dynamic_pre_shared_key',
+  (
+   type => 'Toggle',
+   label => 'Enable PSK',
```
julsemaan: Remove the label from here since it has no effect

html/pfappserver/root/src/views/Configuration/selfServices/_components/TheForm.vue
```diff
@@ -38,6 +38,12 @@
           :text="$i18n.t('List of OS which will be allowed to be register via the self service portal.')"
         />
       </base-form-tab>
+      <base-form-tab :title="$i18n.t('Dynamic Pre Shared Key')">
+        <form-group-dynamic-pre-shared-key namespace="dynamic_pre_shared_key"
+          :column-label="$i18n.t('Show the user pre shared key')"
```
julsemaan:
```suggestion
          :column-label="$i18n.t('Show the user WPA pre-shared key')"
```

html/pfappserver/root/src/views/Configuration/selfServices/_components/TheForm.vue
```diff
@@ -38,6 +38,12 @@
           :text="$i18n.t('List of OS which will be allowed to be register via the self service portal.')"
         />
       </base-form-tab>
+      <base-form-tab :title="$i18n.t('Dynamic Pre Shared Key')">
```
julsemaan:
```suggestion
      <base-form-tab :title="$i18n.t('DPSK')">
```

(To stay consistent with all the other places we display it like the connection profile)

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
```diff
@@ -96,6 +107,18 @@ sub userIsAuthenticated : Private {
     $c->stash(
         nodes    => \@nodes,
     );
+    if (is_dpsk_enabled($c->profile)) {
+        my @provisioners = $c->profile->provisionerObjects();
+        unless(@provisioners){
```
julsemaan: I don't think this is necessary since the foreach will just loop through an empty array without it

html/captive-portal/lib/captiveportal/PacketFence/Controller/Status.pm
fdurand: In fact the better option is probably to put the dpsk in the stack if the dpsk exist in the person field.
If psk doesnt exist for the specific user then put an empty dpsk in the stash.

satkunas (approved):

PR #7087, opened 2022-08-09
Add better error handling for _deauthOnPrevious

### LINKED ISSUE #6985: captive portal redirect page return Caught exception in captiveportal::Controller::Root->dynamic_application "Can't use string ("0") as a HASH ref while "strict refs" in use at /usr/local/pf/lib/pf/enforcement.pm line 206
I'm running packetfence 11.2 on debian 11- standalone server. I've update to the latest release using the script.

**the captivePortal redirection page do not work and show the error*

Caught exception in captiveportal::Controller::Root->dynamic_application "Can't use string ("0") as a HASH ref while "strict refs" in use at /usr/local/pf/lib/pf/enforcement.pm line 206

I think there is a missing or wrong value but I don't  was able to recognize
Cattura

jrouzierinverse:
# Description
Better error handling when performing a deauth on the previous switch.

# Impacts
enforcement

# Issue
fixes #6985

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Bug Fixes
* Better error handling when performing a deauth on the previous switch. (#6985)

### DIFF
--- a/lib/pf/enforcement.pm
+++ b/lib/pf/enforcement.pm
@@ -173,12 +173,15 @@ sub _vlan_reevaluation {
 
     if (isenabled($switch->{_deauthOnPrevious})) {
         $locationlog_entry = locationlog_last_entry_previous_switch($mac,$switch);
-        if ($locationlog_entry) {
-            $switch = pf::SwitchFactory->instantiate( { switch_mac => $locationlog_entry->{'switch_mac'}, switch_ip => $locationlog_entry->{'switch_ip'} } );
-            if ( !$switch ) {
-                $logger->error("Can't instantiate switch (".$locationlog_entry->{'switch_ip'}.")! Check your configuration!");
-                return $FALSE;
-            }
+        if (!$locationlog_entry) {
+            $logger->warn("previous location log entry not found for and $mac $switch->{_id}");
+            return $FALSE;
+        }
+
+        $switch = pf::SwitchFactory->instantiate( { switch_mac => $locationlog_entry->{'switch_mac'}, switch_ip => $locationlog_entry->{'switch_ip'} } );
+        if ( !$switch ) {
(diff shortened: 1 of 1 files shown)


lib/pf/enforcement.pm
```diff
@@ -173,7 +173,13 @@ sub _vlan_reevaluation {
 
     if (isenabled($switch->{_deauthOnPrevious})) {
         $locationlog_entry = locationlog_last_entry_previous_switch($mac,$switch);
-        $switch = pf::SwitchFactory->instantiate( { switch_mac => $locationlog_entry->{'switch_mac'}, switch_ip => $locationlog_entry->{'switch_ip'} } );
+        if ($locationlog_entry) {
```
julsemaan: We should have a warning that no location entry was found and that the deauth will not happen

jrouzierinverse: Do we also skip the deauth on the current switch?

julsemaan: I think @fdurand will answer better than me but I'd say we don't deauth on the current switch if it's configured to be the previous one and there isn't a way to do it. Otherwise it doesn't respect the behavior that is configured.

jrouzierinverse: Is a warning needed? Since that scenario can happen in a normal workflow I can see how it would be useful to know if it happened.

fdurand: Yes the goal of that is to be able to deauth on the previous equipment. Like you have 2 equipments and the only way to reevaluate the acces is to do the deauth on the n-1 then do it if possible , otherwise do nothing and log a warning.

jrouzierinverse: Ok done please review

PR #7109, opened 2022-08-16
adjust proxysql config for when the DB is in read-only mode

### LINKED ISSUE #7105: v12: Read-only mode will not work with ProxySQL
**Describe the bug**
Read-only mode will not work with ProxySQL because it declares the backend as unhealthy and stops proxying to it

**To Reproduce**
1. Execute test PF-41 in testlink

**Additional context**

Example log radiusd produces:
```
Aug 15 20:00:37 cluster-nj-2 freeradius[3473133]: Mon Aug 15 20:00:37 2022 : ERROR: (53) sql_reject: ERROR: rlm_sql_mysql: ERROR 9001 (Max connect timeout reached while reaching hostgroup 10 after 10000ms): HY000```

ProxySQL logs of the backend being down:
```
Aug 15 20:00:07 cluster-nj-2 proxysql-docker-wrapper[3464486]: HID: 10 , address: 170.187.159.126 , port: 3306 , gtid_port: 0 , weight: 100 , status: OFFLINE_HARD , max_connections: 1000 , max_replication_lag: 0 , use_ssl: 0 , max_latency_ms: 0 , comment:
```

### PR DESCRIPTION
**Pull request by julsemaan**
# Description
Adjust proxysql config for when the DB is in read-only mode

# Impacts
ProxySQL configuration

# Issue
fixes #7105

# Delete branch after merge
YES

### DIFF
--- a/conf/proxysql.conf.example
+++ b/conf/proxysql.conf.example
@@ -56,7 +56,8 @@ mysql_variables=
     default_schema="information_schema"
     stacksize=1048576
     server_version="5.1.30"
-    connect_timeout_server=10000
+    connect_timeout_server=50
+    connect_timeout_server_max=100
     monitor_history=60000
     monitor_connect_interval=200000
     monitor_ping_interval=200000
@@ -118,8 +119,15 @@ mysql_query_rules =
     {
         rule_id=200
         active=1
-        match_pattern=".*read_only.*"
-        destination_hostgroup=10
+        match_pattern="SELECT @@global.read_only"
+        destination_hostgroup=30
+        apply=1
+    },
+    {
+        rule_id=200
+        active=1
+        match_pattern="show status like.*"
+        destination_hostgroup=30
         apply=1
     },
     {
--- a/conf/proxysql.conf.example
+++ b/conf/proxysql.conf.example
@@ -124,7 +124,7 @@ mysql_query_rules =
         apply=1
     },
     {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by julsemaan** [comment]
I still need to test when wsrep_ready == OFF by getting a single DB without quorum to be the only one available.

Let's hold-off the merge until I test that since it may require more tweaks in the config depending on if proxysql treats this differently

**Comment by julsemaan** [comment]
> I still need to test when wsrep_ready == OFF by getting a single DB without quorum to be the only one available.
> 
> Let's hold-off the merge until I test that since it may require more tweaks in the config depending on if proxysql treats this differently

The behavior of proxysql is the same in this case as with read_only = 1, this can be reviewed + merged with @fdurand's blessing

File: conf/proxysql.conf.example [hunk 1]
```diff
@@ -118,8 +119,15 @@ mysql_query_rules =
...
         rule_id=200
         active=1
-        match_pattern=".*read_only.*"
-        destination_hostgroup=10
+        match_pattern="SELECT @@global.read_only"
+        destination_hostgroup=30
+        apply=1
+    },
+    {
+        rule_id=200
```
fdurand: you should raise it, the rule id 200 already exist.

julsemaan: The roule has been raised

fdurand:
roule ma poule, let´s merge

PR #7118, opened 2022-08-18
fix/7108

### LINKED ISSUE #7108: v12: The api-frontend cannot store tokens when DB is in R/O
**Describe the bug**
The api-frontend cannot store tokens when DB is in R/O and doesn't store them in memory either on error which means the admin interface cannot be accessed when the DB is in R/O mode

**To Reproduce**
1. Set DB server in R/O mode
2. Try to login in the admin

**Expected behavior**
We should be able to login in the admin even if the DB is in R/O mode. 
We can't expect the token to be stored in the DB and be shared by the servers but it should at least be valid in memory like it used to before we started storing the tokens in the DB

jrouzierinverse:
# Description
Fall back to the memory token backend if the database is in readonly mode.

# Impacts
api frontend authentication.

# Issue
fixes #7108

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries
## Enhancements
* Fall back to the memory token backend if the database is in readonly mode (#7108).

### DIFF
--- a/go/api-frontend/aaa/multi_token_backend.go
+++ b/go/api-frontend/aaa/multi_token_backend.go
@@ -1,0 +1,63 @@
+package aaa
+
+import (
+	"time"
+)
+
+type MultiTokenBackend struct {
+	backends []TokenBackend
+}
+
+func NewMultiTokenBackend(backends ...TokenBackend) TokenBackend {
+	return &MultiTokenBackend{
+		backends: append([]TokenBackend{}, backends...),
+	}
+}
+
+func (tb *MultiTokenBackend) TokenInfoForToken(token string) (*TokenInfo, time.Time) {
+	for _, b := range tb.backends {
+		token, t := b.TokenInfoForToken(token)
+		if token != nil {
+			return token, t
+		}
+	}
+	return nil, time.Unix(0, 0)
+}
+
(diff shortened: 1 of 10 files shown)


go/api-frontend/aaa/multi_token_backend.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (tb *MultiTokenBackend) StoreTokenInfo(token string, ti *TokenInfo) error {
+	var errs []error
+	for _, b := range tb.backends {
+		err := b.StoreTokenInfo(token, ti)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return nil
```
julsemaan, line 49: It should be returning an error when errs is not empty

julsemaan (changes requested):

go/api-frontend/aaa/multi_token_backend.go
julsemaan: Or if we can't return an error because that would prevent read-only from working, we could:
- Return an error when `len(errs) == len(backends)`
- Log warnings regardless so that we know backend xyz produced an error

jrouzierinverse: Ok done

PR #7135, opened 2022-08-23
Temporary fix for password encoding in rest

fdurand:
# Description
Fix the the cli/vpn access the password encoding is this one contain special char like ç .

# Impacts
RADIUS Cli/VPN

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Fix the wrong encoding of special char in the rest call to packetfence (use base64)

### DIFF
--- a/lib/pf/radius.pm
+++ b/lib/pf/radius.pm
@@ -75,6 +75,7 @@ use pf::Redis;
 use pf::constants::eap_type qw($EAP_TLS $MS_EAP_AUTHENTICATION $EAP_PSK);
 use pf::person;
 use pf::factory::mfa;
+use MIME::Base64;
 
 our $VERSION = 1.03;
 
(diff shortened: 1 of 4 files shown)


lib/pf/radius.pm
```diff
@@ -790,6 +791,9 @@ sub switch_access {
     my $connection_type = $connection->attributesToBackwardCompatible;
     my $connection_sub_type = $connection->subType;
     my $password = $radius_request->{'User-Password'};
+    if (exists($radius_request->{'PacketFence-UserPassword'})) {
```
nqb, line 794: Should we not rename `password` variable to distinguish:
- unencoded variable coming from `User-Password`
- encoded variable coming from `PacketFence-UserPassword` 

It's just a suggestion because the line 793 is confusing for me.

julsemaan:
Applied on maintenance/11.1 and maintenance/11.2

It didn't patch cleanly on 11.0 due to the MFA integration added in 11.1

We can certainly cherry-pick in 11.0 and below with some effort if a customer really needs it

nqb:
Thanks @julsemaan, I merged it quickly and missed backports.

PR #7137, opened 2022-08-23
fix/7120

### LINKED ISSUE #7120: http portal cannot connect to the SMTP server
If the STMP server is set to localhost then the httpd.portal service cannot send an email

jrouzierinverse:
# Description
Send all emails through the queue.

# Impacts
Sending emails

# Issue
fixes #7120

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [x] Add unit tests

### DIFF
--- a/conf/log.conf.d/pfqueue.conf.example
+++ b/conf/log.conf.d/pfqueue.conf.example
@@ -18,19 +18,22 @@ log4perl.category.pf.dhcp = INFO, QUEUE_DHCP
 log4perl.additivity.pf.dhcp = 0
 
 ### Logging for pfqueue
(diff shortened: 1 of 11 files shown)


julsemaan:
The code looks good, can you confirm which flows you did test and which you didn't test so we can adapt the QA accordingly.

jrouzierinverse:
I tested the password of the day workflow which technically covers all flows that use templated emails except for email/sponsor activation.

So we should test 
* Email/Sponsor activation.
* Security event action 'email_admin'

PR #7141, opened 2022-08-24
adjust the NAS-IP-Address in the RADIUS source

julsemaan:
# Description
Adjust the NAS-IP-Address value when sending an Access-Request in the RADIUS source

# Impacts
RADIUS authentication source (not RADIUS auth server)

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [X] Document the feature
- [N/A] Add unit tests

# NEWS file entries
## Enhancements
* The NAS-IP-Address can now be configured for RADIUS authentication sources

# UPGRADE file entries
Included in the PR

### DIFF
--- a/lib/pf/Authentication/Source/RADIUSSource.pm
+++ b/lib/pf/Authentication/Source/RADIUSSource.pm
@@ -12,7 +12,8 @@ use pf::constants qw($TRUE $FALSE);
 use pf::Authentication::constants qw($LOGIN_CHALLENGE);
 use pf::constants::authentication::messages;
 use pf::log;
-use pf::config qw(%Config);
+use pf::config qw(%Config $management_network);
+use pf::cluster qw($cluster_enabled);
 
 our $RADIUS_STATE = 'State';
 our $RADIUS_REPLY_MESSAGE = 'Reply-Message';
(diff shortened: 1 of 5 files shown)


lib/pf/Authentication/Source/RADIUSSource.pm
```diff
@@ -172,7 +174,14 @@ sub _fetch_attributes {
 sub check_radius_password {
     my ($self, $radius, $name, $pwd, $nas, @extra) = @_;
 
-    $nas = eval {$radius->{'sock'}->sockhost()} unless defined($nas);
+    if(!defined($nas)) {
+        if($self->nas_ip_address) {
+            $nas = $self->nas_ip_address;
+        }
+        else {
+            $nas = $cluster_enabled ? pf::cluster::management_cluster_ip : $management_network->{Tip};
```
jrouzierinverse, line 182: Should it be 
```$nas = $cluster_enabled ? pf::cluster::management_cluster_ip :  ($management_network->{Tvip} // $management_network->{Tip});```
Or is the cluster IP is the VIP?

julsemaan: There is no more use of 'Tvip' anymore AFAIK, it would be for when you put a `vip` in pf.conf and that isn't being done since many years. We should likely just get that completely removed at some point.

And yes the management_cluster_ip is where the actual VIP for management is now

fdurand (approved):

PR #7153, opened 2022-08-29
Fix/proxysql read only again

julsemaan:
# Description
Improve proxysql's behavior when DB is in read-only mode

# Impacts
ProxySQL

# Delete branch after merge
YES


lib/pf/services/manager/proxysql.pm
```diff
@@ -104,6 +104,9 @@ EOT
         $i--;
         }
     }
+
+    $tags{'mysql_pf_user'} = $DB_Config->{user};
+    $tags{'mysql_pf_pass'} = $DB_Config->{pass};
```
jrouzierinverse, line 110: Escape the password and maybe the username.
```perl
$tags{'mysql_pf_pass'} =~ s/"/\\"/g;
```

conf/proxysql.conf.example
```diff
@@ -150,3 +150,17 @@ mysql_users =
...
+     id=1
+     active=1
+     interval_ms=5000
+     filename="/proxysql-read-only-handler.sh"
+     arg1="{\"username\":\"admin\", \"password\":\"admin\", \"host\":\"127.0.0.1\", \"port\":6032}"
+     arg2="{\"username\":\"[% mysql_pf_user %]\", \"password\":\"[% mysql_pf_pass %]\"}"
+     arg3="{\"read_write_hostgroup\":10, \"read_hostgroup\":30, \"rule_ids\":\"100,400\"}"
+     arg4="/proxysql-read-only-handler.log"
+  }
+)
```
jrouzierinverse, line 166: change the file permissions to make it executable.

go/cmd/pfacct/radius.go
```diff
@@ -582,7 +582,9 @@ type RadiusStatements struct {
 func setupStmt(db *sql.DB, stmt **sql.Stmt, sql string) {
 	var err error
 	if *stmt, err = db.Prepare(sql); err != nil {
-		panic(err)
+		log.Logger().Error(fmt.Sprintf("Failed to prepare statement '%s': %s", sql, err))
+		time.Sleep(5 * time.Second)
+		go setupStmt(db, stmt, sql)
```
jrouzierinverse: Maybe a better approach would be to use a WaitGroup in RadiusStatements::Setup.
So we can wait until all of the statements succeed before moving forward.

jrouzierinverse (changes requested):

go/cmd/pfacct/radius.go
julsemaan: Doing this would prevent pfacct from starting and accepting requests until this completes which is not desirable. We should be able to startup in R/O mode and the `INSERT/UPDATE` statements can't be prepared in this mode when going through ProxySQL

conf/proxysql.conf.example
julsemaan: It is already via the Dockerfile:
https://github.com/inverse-inc/packetfence/pull/7153/files#diff-66d4b4df5930c5daf1764ead4c20e20ab7406ea37b521051cf1b860cdaa5de93R79

lib/pf/services/manager/proxysql.pm
julsemaan: Fixed

bin/proxysql-read-only-handler.sh
```diff
@@ -0,0 +1,56 @@
...
+  if [ "${READ_ONLY}" = "0" ] ; then
+    echo `date` SERVER $server:$port is READ_ONLY=0 >> ${ERR_FILE}
+    HAS_ONE_READ_WRITE=yes
+  fi
+done <<< "$($PROXYSQL_CMDLINE "SELECT hostname,port,status FROM mysql_servers WHERE hostgroup_id='$READ_WRITE_HOSTGROUP'")"
+
+#echo "$HAS_ONE_READY $HAS_ONE_READ_WRITE"
+
+if [ "${HAS_ONE_READY}" = "no" ] || [ "${HAS_ONE_READ_WRITE}" = "no" ]; then
+  echo `date` All the servers are either wsrep_ready=off or read_only=1, stopping writes >> ${ERR_FILE}
```
extrafu: Small nitpick but we should s/are/have/

bin/proxysql-read-only-handler.sh
```diff
@@ -0,0 +1,56 @@
...
+  $PROXYSQL_CMDLINE "update mysql_query_rules set replace_pattern='ERROR DB IS IN R/O', destination_hostgroup=$READ_HOSTGROUP where rule_id IN ($QUERY_RULE_IDS); LOAD MYSQL QUERY RULES TO RUNTIME;" 2>> ${ERR_FILE}
+else
+  echo `date` At least one server is wsrep_ready=on and read_only=0 >> ${ERR_FILE}
```
extrafu: Small nitpick but we should s/is/has/

go/cmd/pfacct/radius.go
jrouzierinverse: We would need to detect If the statement is nil.
When starting pfacct in readonly mode, what behavior did you see?

julsemaan: If we leave it with the previous way it was, it would make pfacct crash on startup when in read-only mode. With this fix, it starts but doesn't respond well (which is expected) and then when the read-only goes away, it starts processing packets. Although I agree this may not be the perfect solution, we need something for the release. If you have a better proposition that can work and that you can thoroughly test before the release, let's implement it in this branch ASAP. Otherwise, let's keep this as-is so we can get the fix in the release

jrouzierinverse: I updated the setupStmt function

julsemaan: Reviewed that change, I'm happy with it.

If all is good for you too now, let's merge this baby!

jrouzierinverse (approved):

PR #7160, opened 2022-08-31
fix/3768

### LINKED ISSUE #3768: authentication sources: catchall with match=any
In a AD source, if you create a catchall authentication rule with a match directive sets as "any" and no conditions, this rule will never match. With match directive sets as "all" and no conditions, it works.

IMO, we should have same behavior between "any" and "all" in this specific case.

Config to reproduce (easy with `pftest`):
```
[ADUsers rule catchall-any]
match=any
action1=set_unreg_date=2020-01-01
action0=set_role=gaming
class=authentication
description=catchall

[ADUsers rule catchall-all]
action1=set_unreg_date=2020-01-01
match=all
description=catchall
class=authentication
action0=set_role=guest

```
With this configuration, users will always get `guest` role.

Tested on 8.2

jrouzierinverse:
# Description
When a rule match is 'any' and has no conditions the rule is always successful

# Impacts
Authenication rule matching.

# Issue
fixes #3768

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries

## Bug Fixes
* When a rule match is 'any' and has no conditions the rule is always successful (#3768)


nqb (approved):
Could you rebase ? Then I will merge it.

PR #7162, opened 2022-08-31
fix/7158

### LINKED ISSUE #7158: v12: DNS audit log entries are deleted before their expiry
**Describe the bug**
The DNS audit log entries are deleted before their planned expiration (1 hour).

**To Reproduce**
1. Enable the DNS audit log
2. Get some DNS going through pfdns
3. Check the DNS audit log (entries should be there)
4. Wait a minute so pfcron does a pass
5. Check the DNS audit log
6. Entries are gone

**Expected behavior**
Entries should be cleaned only after an hour

**Additional context**
The pfcron container is on the same date/time as the host OS and the entries do have the correct date in the DB

jrouzierinverse:
# Description
Set the database location to the system Local timezone

# Issue
fixes #7158

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Set the database location to the system Local timezone (#7158)

### DIFF
--- a/go/db/db.go
+++ b/go/db/db.go
@@ -83,6 +83,7 @@ func ReturnURI(ctx context.Context, user, pass, host, port, dbName string) strin
 	host = strings.TrimSpace(host)
 	port = strings.TrimSpace(port)
 	dbName = strings.TrimSpace(dbName)
+	location, _ := time.LoadLocation("Local")
 
 	proto := "tcp"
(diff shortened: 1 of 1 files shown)


julsemaan:
Tested and it works!

PR #7169, opened 2022-09-01
exit post installation if manage-images.sh failed

### LINKED ISSUE #7148: Install/update should fail if manage-images.sh fails at any step
**Describe the bug**
I had an upgrade fail in a weird way where manage-images.sh wouldn't be able to pull some of the images (issue with ghcr.io I think) but it did prune images at the end.

We need to make it so that if **any** step of the manage-images.sh fails, it makes the packaging install/upgrade fail immediately. Otherwise you get weird unrelated issues like pfconfig connection failures because that service fails to start.

nqb:
# Description
Exit post installation if manage-images.sh failed

# Impacts
RPM installation

# Issue
fixes #7148 

# Delete branch after merge
YES

### DIFF
--- a/rpm/packetfence.spec
+++ b/rpm/packetfence.spec
@@ -720,41 +720,44 @@ echo "Restarting rsyslogd"
 #removing old cache
 /bin/systemctl enable docker
 /bin/systemctl restart docker
-# get containers image and tag them locally
-/usr/local/pf/containers/manage-images.sh
 
-rm -rf /usr/local/pf/var/cache/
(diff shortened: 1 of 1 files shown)


julsemaan:
Anything you need to do before we merge? I reviewed the change and it looks good

nqb:
I'm testing package and I will let you know my results.

nqb:
Works as expected 👌 
After a failure, you got message displayed and no containers are running.

I will just add a comment in `rpm/packetfence.spec` to explain why we added ìf` like this and we can merge.

nqb:
Ready

Issue #6396, opened 2021-06-15
Officially support external MariaDB database and related architecture


nqb:
**Is your feature request related to a problem? Please describe.**
PacketFence can work with an external MariaDB database but it requires to tweak configuration at several places.

We should officially support this kind of installation and architecture because it seems to be used more and more.

**Describe the solution you'd like**
Full support of external MariaDB database:
- [ ] installation (standalone, cluster of PacketFence servers attached to a remote MariaDB instance)
- [ ] upgrade
- [X] backups

nqb:
As discussed today during our meeting, we should certainly handle this through ProxySQL to keep same pf.conf and pfconfig.conf configuration as for a "local" cluster.

nqb:
Also, what I noticed recently is that our automatic upgrade process will fail to handle this properly because it expects a local database running.

PR #4090, opened 2019-03-19
Update backup script to support remote databases

nqb:
# Description
- Backup remote MariaDB databases (standalone or Galera clusters)
- Refactor code

# Impacts
Backup script

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests
- [ ] Add acceptance tests (TestLink)

# NEWS file entries
## Enhancements
* Backup remote MariaDB databases  (standalone or Galera clusters) from pf servers

PR #7170, opened 2022-09-01
Disable database backups if DB is remote

nqb:
# Description
We assume that a remote DB is backup by a tool directly on remote DB

This commit also deprecate the ability to backup a remote DB from a PacketFence server added by #4090

I will an upgrade note to warn about the deprecation even I don't think a lot of users are using this feature.

# Impacts
Backups of remote databases

# Issue
Related to #6396

# Delete branch after merge
YES

# UPGRADE file entries

```asciidoc
=== Remote database backups

Ability to backup a remote database configured in PacketFence has been deprecated. A dedicated tool on the database server itself must be used to backup PacketFence database.
```

### DIFF
--- a/addons/backup-and-maintenance.sh
+++ b/addons/backup-and-maintenance.sh
@@ -89,7 +89,6 @@ should_backup(){
 
     if [ $MARIADB_REMOTE_CLUSTER -eq 1 ]; then
         echo "Remote database detected: backup should be done on database server itself."
-        SHOULD_BACKUP=0
         exit $BACKUPRC
     fi
 
@@ -98,7 +97,6 @@ should_backup(){
         MARIADB_LOCAL_CLUSTER=1
         FIRST_SERVER=`mysql -u$REP_USER -p$REP_PWD -e 'show status like "wsrep_incoming_addresses";' | tail -1 | awk '{ print $2 }' | awk -F "," '{ print $1 }' | awk -F ":" '{ print $1 }'`
         if ! ip a | grep $FIRST_SERVER > /dev/null; then
-            SHOULD_BACKUP=0
             echo "Not the first server of the cluster: database backup canceled."
             exit $BACKUPRC
         else
--- a/addons/backup-and-maintenance.sh
+++ b/addons/backup-and-maintenance.sh
@@ -37,10 +37,12 @@ NODE2_HOSTNAME=''
 NODE1_IP=''
 NODE2_IP=''
 
-# Only for MariaDB remote clusters
-MARIADB_REMOTE_CLUSTER=0
-# should be equal to return of hostname command
-MARIADB_BACKUP_SERVER_HOSTNAME=''
+# to detect MariaDB remote DB
+if [ "$DB_HOST" != "localhost" ] && [ "$DB_HOST" != "100.64.0.1" ]; then
+    MARIADB_REMOTE_CLUSTER=1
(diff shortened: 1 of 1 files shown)


addons/backup-and-maintenance.sh
```diff
@@ -84,6 +84,13 @@ should_backup(){
     # Default choices
     SHOULD_BACKUP=1
     MARIADB_LOCAL_CLUSTER=0
+
+    if [ $MARIADB_REMOTE_CLUSTER -eq 1 ]; then
+        echo "Server not configured to do DB backups: database backup canceled."
+        SHOULD_BACKUP=0
```
julsemaan: No need to set this variable if we're going to exit right below

julsemaan (commented):
I think @fdurand should also take a look at this

addons/backup-and-maintenance.sh
```diff
@@ -84,6 +84,13 @@ should_backup(){
     # Default choices
     SHOULD_BACKUP=1
     MARIADB_LOCAL_CLUSTER=0
+
+    if [ $MARIADB_REMOTE_CLUSTER -eq 1 ]; then
+        echo "Server not configured to do DB backups: database backup canceled."
```
extrafu: canceled -> cancelled

extrafu: I would probably just write to be consistent:

"Server is not configured to do database backups: the operation has been cancelled."

nqb: I agree with you @extrafu. I prefer: "Remote database detected: backup should be done on database server itself."

extrafu: Fine to me!

nqb:
I made adjustments requested.

addons/backup-and-maintenance.sh
```diff
@@ -37,10 +37,12 @@ NODE2_HOSTNAME=''
 NODE1_IP=''
 NODE2_IP=''
 
-# Only for MariaDB remote clusters
-MARIADB_REMOTE_CLUSTER=0
-# should be equal to return of hostname command
-MARIADB_BACKUP_SERVER_HOSTNAME=''
+# to detect MariaDB remote DB
+if [ "$DB_HOST" != "localhost" ] && [ "$DB_HOST" != "100.64.0.1" ]; then
```
fdurand, line 41: if 127.0.0.1 ?

JeGoi (approved):
Looks good to me with others revue.

addons/backup-and-maintenance.sh
nqb: Is `127.0.0.1` a value still valid in 12.0.0 ?

I suppose it's when you are using `haproxy-db` in place of `proxysql`

fdurand: it´s just in case we forgot to change the value in the conf. Just to prevent no backup just because of that.

julsemaan: 127.0.0.1 is not a valid value anymore in pf.conf whether its haproxy-db or proxysql (there is a FATAL checkup in a cluster for that)

The setup will be hosed if that value is left there so I think a non-working backup will be the least of your trouble if 127.0.0.1 is left there

extrafu:
@julsemaan if all is fine can you please merge?

PR #7180, opened 2022-09-02
Feature/rework cluster upgrade

julsemaan:
# Description
Refreshes the cluster upgrade guide by 
* Reusing the `do-upgrade.sh` script we have for standalone.
* Detaching node C at the start of the upgrade process (mimics what we actually do in real life)
* Removing notes from pre-v11 since this guide only applies to 11+

Other:
* Build HTML and PDF docs in GitLab

# Impacts
Documentation, upgrade script

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [X] Document the feature
- [N/A] Add unit tests

### DIFF
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -218,7 +218,7 @@ If you are happy about the state of your upgrade on node C, you can move on to u
 [source,bash]
 ----
 export UPGRADE_CLUSTER_SECONDARY=yes
-systemctl start packetfence-mariadb
+systemctl restart packetfence-mariadb
 ----
 
 Then, <<PacketFence_Installation_Guide.asciidoc#_automation_of_upgrades,apply the upgrade process described here>> **on nodes A and B**.
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -204,12 +204,6 @@ Once you are feeling confident to try your failover to node C again, you can do 
 
 If you are happy about the state of your upgrade on node C, you can move on to upgrading the other nodes.
 
-.On nodes A and B
-[source,bash]
-----
-export UPGRADE_CLUSTER_SECONDARY=yes
-----
-
 .On node A
 ----
 /usr/local/pf/bin/cluster/node node-B-hostname disable
@@ -218,6 +212,13 @@ export UPGRADE_CLUSTER_SECONDARY=yes
 .On node B
 ----
 /usr/local/pf/bin/cluster/node node-A-hostname disable
+----
+
+.On nodes A and B
+[source,bash]
+----
+export UPGRADE_CLUSTER_SECONDARY=yes
+systemctl start packetfence-mariadb
 ----
 
 Then, <<PacketFence_Installation_Guide.asciidoc#_automation_of_upgrades,apply the upgrade process described here>> **on nodes A and B**.
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -124,7 +124,7 @@ PacketFence application services on it:
 
 IMPORTANT: `packetfence-config` should stay started in order to run `/usr/local/pf/bin/cluster/node` commands.
   
-Next, you can upgrade your operating system and/or PacketFence on node C.
+In the next following steps, you will be upgrading PacketFence on node C.
 
 ===== Detach node C from the cluster
 
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -58,7 +58,7 @@ In this procedure, the 3 nodes will be named A, B and C and they are in this ord
 
 ==== Backups
 
-First, ensure you have taken backups of your data. We highly encourage you to perform snapshots of all the virtual machines prior to the upgrade. You should also take a backup of the database and the `/usr/local/pf` directory using <<PacketFence_Upgrade_Guide.asciidoc#__database_and_configurations_backup,Database and configurations backup instructions>>.
+Re-importable backups will be taken during the upgrade process. We highly encourage you to perform snapshots of all the virtual machines prior to the upgrade if possible.
 
 ==== Disabling the auto-correction of configuration
 
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -267,7 +267,7 @@ You should now sync the configuration by running the following **on nodes A and 
 
 [source,bash]
 ----
-/usr/local/pf/bin/cluster/sync --from=192.168.1.5 --api-user=packet --api-password=fence
+/usr/local/pf/bin/cluster/sync --from=192.168.1.5 --api-user=packet --api-password=anotherMoreSecurePassword
 /usr/local/pf/bin/pfcmd configreload hard
 ----
 
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -204,11 +204,13 @@ Once you are feeling confident to try your failover to node C again, you can do 
 If you are happy about the state of your upgrade on node C, you can move on to upgrading the other nodes.
 
 .On node A
+[source,bash]
 ----
 /usr/local/pf/bin/cluster/node node-B-hostname disable
 ----
 
 .On node B
+[source,bash]
 ----
 /usr/local/pf/bin/cluster/node node-A-hostname disable
 ----
--- a/docs/cluster/appendix.asciidoc
+++ b/docs/cluster/appendix.asciidoc
@@ -69,7 +69,6 @@ In order to do so, go in _Configuration->System Configuration->Maintenance_ and 
 
 Once this is done, restart `pfcron` on all nodes using:
 
-.For PacketFence versions 10.2 and later
 [source,bash]
 ----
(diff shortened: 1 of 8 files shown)


julsemaan:
This has to be tested. Just did the steps logically but haven't tested the whole recipe yet.

nqb (changes requested):
* By default, a cluster upgrade through automation of upgrades is disabled:
```
Upgrading a cluster is not supported by this tool at the moment.
You can use it **at your own risk** by setting the following environment variable:
  export ALLOW_CLUSTER_UPGRADE=yes
```
I used:

`root@nqb-deb3:~# ALLOW_CLUSTER_UPGRADE=yes /usr/local/pf/addons/upgrade/do-upgrade.sh`

EDIT: we can use same recipe as for `UPGRADE_CLUSTER_SECONDARY=yes`

julsemaan:
> * By default, a cluster upgrade through automation of upgrades is disabled:
> 
> ```
> Upgrading a cluster is not supported by this tool at the moment.
> You can use it **at your own risk** by setting the following environment variable:
>   export ALLOW_CLUSTER_UPGRADE=yes
> ```
> 
> I used:
> 
> `root@nqb-deb3:~# ALLOW_CLUSTER_UPGRADE=yes /usr/local/pf/addons/upgrade/do-upgrade.sh`

I'll remove this now that the official recipe will allow to do this

nqb:
We need to adjust the backup and export usage inside `run-upgrade.sh` when running inside a cluster for following reasons: 
   * In a cluster, we take a DB backup only on first member of the cluster. It should have been done at the beginning of the cluster upgrade. So it means we will not have any packetfence-db-dump* under /root/backup for at least 2 servers
   * When running `run-upgrade.sh` a new backup is taken but at this time, we are not really in a cluster because servers operate in a standalone mode. EDIT: So there is no new backups taken on each server
   * Database is stopped on node A and B at step `Stop services on nodes A and B`. EDIT: that is not the case on node C.

Important: our backup script doesn't fail on an error at the moment. It means that upgrade process is not stopped even if there is a failure during backup

julsemaan:
> * In a cluster, we take a DB backup only on first member of the cluster. It should have been done at the beginning of the cluster upgrade. So it means we will not have any packetfence-db-dump* under /root/backup for at least 2 servers

MariaDB should be operating in standalone on node C when run-upgrade.sh is happening which means the backup should be taken (at least in theory)

> * When running `run-upgrade.sh` a new backup is taken but at this time, we are not really in a cluster because servers operate in a standalone mode. EDIT: So there is no new backups taken on each server

Can you confirm my understanding: The backup is not being taken when a node is being upgraded and that MariaDB is in standalone mode

> * Database is stopped on node A and B at step `Stop services on nodes A and B`. EDIT: that is not the case on node C.

I've adjusted this so that the DB is started in standalone mode prior to upgrading the secondary nodes

> Important: our backup script doesn't fail on an error at the moment. It means that upgrade process is not stopped even if there is a failure during backup

We need to check the effort required to make this happen

julsemaan:
TODO: 
- [X] Cleanup the upgrade guide to remove outdated cluster instructions

nqb:
TODO:
- [x] backport fixes in `run-upgrade.sh` to maintenance/11.X branches (currently only backported to 11.2)

julsemaan:
Fixes and adjustments have been backported in all maintenance/11.x

nqb:
I removed useless informations in Upgrade guide:
- instructions specific for cluster upgrade
- DB schema upgrade

Removal of DB schema upgrade should simplify release and post-release procedure I guess.

nqb:
I tested upgrade procedure on Debian. It works well. Nicely done @julsemaan !

Except #7188 which prevent restart of services to occur correctly.

Some comments (they can be added in a separate PR or commit)
- [X] We should add somewhere how to switch from `proxysql` to `haproxy-db` with warnings regarding latency
- [ ] In the upgrade procedure, at some places, there is no PacketFence servers which provide services for a few seconds:
   - [12.3.4](https://inverse-inc.gitlab.io/-/packetfence/-/jobs/2995728548/artifacts/docs/PacketFence_Clustering_Guide.html#_stop_services_on_nodes_a_and_b): we stop all services on A and B **then** we start services on C. If there was an upgrade problem on node C and you can't start services on it, there is no PF services available.
 
 => IMHO, we should:
 * do the opposite: start services on C and if everything is ok, stop everything on A and B.
 * mention in the documentation that this can generate a small service failure

   * [12.3.6 - Elect node C as database master](https://inverse-inc.gitlab.io/-/packetfence/-/jobs/2995728548/artifacts/docs/PacketFence_Clustering_Guide.html#_elect_node_c_as_database_master): same remark. We are restarting database on the only node which offer service at the moment.
   
* [/] Amend "Restart all PacketFence services" in PacketFence Upgrade Guide:
   * https://inverse-inc.gitlab.io/-/packetfence/-/jobs/2995728548/artifacts/docs/PacketFence_Upgrade_Guide.html#_restart_packetfence_services
   * https://inverse-inc.gitlab.io/-/packetfence/-/jobs/2995728548/artifacts/docs/PacketFence_Upgrade_Guide.html#_restart_all_packetfence_services

because I don't think it' still necessary to restart `packetfence-config` at end of an upgrade (patch or major/minor upgrades) for following reasons:
  * `packetfence-config` is restarted by packaging (all upgrade cases)
  * `run-upgrade.sh` also restarted `packetfence-config` (cover only major/minor upgrades)
  * `configreload hard` doesn't seem to like when `packetfence-config` has been restarted just before:
  ```
  root@nqb-deb3:~# systemctl restart packetfence-config
/usr/local/pf/bin/pfcmd configreload hard

cannot connect to the server Connection timed outn at /usr/local/pf/lib/pfconfig/util.pm line 80.
```
=> Certainly because `pfconfig` is not yet operating correctly

**Comment by julsemaan** [comment]
All the comments above look good to me, we can discuss how to address these.

The only exception would be 'do the opposite: start services on C and if everything is ok, stop everything on A and B.'
If you do that, the servers will fight for the VIPs and when in unicast, may end up mounting the VIP in both places and you'll need manual arping after shutting down A,B to get C to invalidate the bad arp that was caused by doing that. Still we could put a note about the fact there will be some downtime caused by this

**Comment by nqb** [comment]
> The only exception would be 'do the opposite: start services on C and if everything is ok, stop everything on A and B.'
If you do that, the servers will fight for the VIPs and when in unicast, may end up mounting the VIP in both places and you'll need manual arping after shutting down A,B to get C to invalidate the bad arp that was caused by doing that. Still we could put a note about the fact there will be some downtime caused by this

100% agree. At least we can recommend to run a `checkup` on node C before stopping services on node A and B.

**Comment by julsemaan** [comment]
> because I don't think it' still necessary to restart `packetfence-config` at end of an upgrade (patch or major/minor upgrades) for following reasons:

Good idea, done!

File: docs/cluster/appendix.asciidoc [hunk 1]
```diff
@@ -97,30 +111,23 @@ Once this is done, stop `galera-autofix` service on *all* nodes using:
...
+----
+/usr/local/pf/bin/pfcmd service pf stop
+----
 
-Node C should currently be running the latest version of PacketFence while A and B should still be running the older version and still be offering service. The database is also still being synced to node C via the `packetfence-mariadb` service.
-
-NOTE: The steps below will cause a temporary loss of service.
+IMPORTANT: `packetfence-config` should stay started in order to run `/usr/local/pf/bin/cluster/node` commands.
+  
+Next, you can upgrade your operating system and/or PacketFence on node C.
```
JeGoi:
```suggestion
In the next following steps, you will be upgrading PacketFence on node C.
```

JeGoi:
Change `
/usr/local/pf/bin/cluster/sync --from=192.168.1.5 --api-user=packet --api-password=fence` to `
/usr/local/pf/bin/cluster/sync --from=192.168.1.5 --api-user=packet --api-password=anotherMoreSecurePassword` to be consistant with the other part of the cluster settings

JeGoi (changes requested):
Just two points. And all good for me.
Upgrade tested from 11.2 to 12.0 on debian cluster
Starting test it on el8

nqb:
I adjusted @JeGoi's comment related to password consistency.

I adjusted restart of PacketFence services in:
* a808cddc407dfa65b1428f0a7be8e115a6ebcdea
* d96f7823ad

I think we can also remove these commands after an upgrade:
```
/usr/local/pf/bin/pfcmd pfconfig clear_backend
/usr/local/pf/bin/pfcmd configreload hard
```

`configreload hard` is already called by `import.sh` and `run-upgrade.sh` so it means these instructions will only cover maintenance packages upgrade.

EDIT: discuss with @julsemaan, we will keep thing as-is.

Remaining remark:

> [12.3.6 - Elect node C as database master](https://inverse-inc.gitlab.io/-/packetfence/-/jobs/2995728548/artifacts/docs/PacketFence_Clustering_Guide.html#_elect_node_c_as_database_master): same remark. We are restarting database on the only node which offer service at the moment.

EDIT: @julsemaan will add a small note.

nqb:
TODO:
- [ ] add https://github.com/inverse-inc/packetfence/commit/1ffaf8d750b17ba66319f9ed95a024675a2d5469 in `import.sh`

EDIT: will go directly into `devel`

PR #7184, opened 2022-09-06
fix/x-crypto

jrouzierinverse:
# Description
Updating the go dependency for golang.org/x/crypto

# Delete branch after merge
YES


julsemaan:
Pipeline launched, will try a fresh install after it succeeds and we're good to merge: https://gitlab.com/inverse-inc/packetfence/-/pipelines/633437620

PR #7191, opened 2022-09-08
Feature/containers tooling

julsemaan:
# Description
Some tooling and doc for how we should now be working with containers when patching packaged deployments

# Delete branch after merge
YES

### DIFF
--- a/addons/dev-helpers/build-local-container.sh
+++ b/addons/dev-helpers/build-local-container.sh
@@ -1,12 +1,25 @@
 #!/bin/bash
 
+set -o nounset -o pipefail -o errexit
+
 source /usr/local/pf/addons/functions/helpers.functions
 
 cd /usr/local/pf/
 
 main_splitter
 
-name="$1"
+name="${1:-}"
+if [ -z "$name" ]; then
+  echo "Unspecified container name"
+  
+  sub_splitter
+
+  echo "The following images can be used with this tool:"
+  output_all_container_images
+
+  exit 1
+fi
+
 dockerfile=containers/$name/Dockerfile
 
 if ! [ -f $dockerfile ]; then
@@ -19,8 +32,6 @@ if ! [ -f $dockerfile ]; then
 
   exit 1
 fi
-
-set -o nounset -o pipefail -o errexit
 
 source /usr/local/pf/conf/build_id
 
--- a/addons/dev-helpers/build-local-container.sh
+++ b/addons/dev-helpers/build-local-container.sh
@@ -14,24 +14,8 @@ if ! [ -f $dockerfile ]; then
 
   sub_splitter
 
-  # find all directories with Dockerfile
-  # excluding non necessary images
-  DOCKERFILE_DIRS=$(find containers/ -type f -name "Dockerfile" \
-                         -not -path "*/pfdebian/*" \
-                         -not -path "*/radiusd/*" \
-                         -not -path "*/pfconnector-*/*" \
-                         -printf "%P\n")
+  output_all_container_images
 
-  for file in ${DOCKERFILE_DIRS}; do
-    # remove /Dockerfile suffix
-    CONTAINERS_IMAGES+=" ${file%/Dockerfile}"
-  done
-  
-  echo "The following images can be built using this tool:"
-  for img in ${CONTAINERS_IMAGES}; do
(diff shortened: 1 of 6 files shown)


julsemaan:
I've assigned the expected main users of this (including the CPE)

Please make sure you're all happy with these so that we use these tools only when working on customers deployments and not each "fork" our own tools

NOTE: This doesn't address the need of making a patch survive to the maintenance. This involves more work and tooling which are IMO not necessary for 12.0 since we're currently already working with this behavior in our current release.

addons/dev-helpers/build-local-container.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
```
nqb, line 2:
```suggestion
set -o nounset -o pipefail -o errexit
```

addons/dev-helpers/build-local-container.sh
```diff
@@ -0,0 +1,41 @@
...
+  sub_splitter
+
+  echo "The following images can be built using this tool:"
+  output_all_container_images
+
+  exit 1
+fi
+
+set -o nounset -o pipefail -o errexit
```
nqb:
```suggestion

```

nqb: Not sure if it's expected.

docs/developer/containers_recipes.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+
+ * A server running the packages: A PacketFence server installed using the RPM or DEB packages and not converted to a dev machine (i.e. what PacketFence users deploy)
+
+=== Applying a custom patch on a server running the packages
+
+When applying a custom patch or adding debug in the code on a PacketFence 12.0+ server, additional steps must be taking for your change to be effective in the service(s) running in container mode.
+
+To know if a service runs in container mode, run `systemctl status` on the service (ex: `systemctl status packetfence-pfpki`) and check if one of the child processes contains `docker run`.
+
+After you've applied you've altered the code on the server and you're satisfied with the changes, run the following command:
```
nqb:
```suggestion
After you've altered the code on the server and you're satisfied with the changes, run the following command:
```

nqb (changes requested):
I didn't test recipes yet (except development environment using our previous instructions) but it looks good.

addons/dev-helpers/build-local-container.sh
julsemaan: Given there are no commands that run without checking the result, I put this line after validating the input. Otherwise, if you provide no args to the script, it fails with '$1' not being initialized instead of outputing the usage of the script. We could make it so it works with that at the top, just didn't find it useful to do it before the script args where handled

nqb: I think you will not have the issue with "$1" if you use same recipe as for `commit-local-container.sh`:

`name="${1:-}"`

docs/developer/containers_recipes.asciidoc
```diff
@@ -0,0 +1,54 @@
+// to display images directly on GitHub
+ifdef::env-github[]
+:encoding: UTF-8
+:lang: en
+:doctype: book
+:toc: left
+:imagesdir: ../images
```
JeGoi:
```suggestion
:imagesdir: ../../images
```
This rep is under doc/

docs/developer/dev_env_setup.asciidoc
```diff
@@ -0,0 +1,51 @@
+// to display images directly on GitHub
+ifdef::env-github[]
+:encoding: UTF-8
+:lang: en
+:doctype: book
+:toc: left
+:imagesdir: ../images
```
JeGoi:
```suggestion
:imagesdir: ../../images
```

JeGoi (approved):
Looks good for me, two minor modifications

addons/dev-helpers/build-local-container.sh
julsemaan: Done!

nqb (approved):

fdurand (approved):

JeGoi (approved):

julsemaan:
All approved except the CPE, let's merge this!

PR #7199, opened 2022-09-12
Permit to disable kaniko cache on demand

nqb:
# Description
Permit to disable kaniko cache on demand using `KNK_CACHE=false`
 
`--cache-repo` option has been removed because we are using the default location for cache image.

# Impacts
Build of containers

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests


nqb:
Tested on a pipeline and it works as expected. I flagged it at v12.1 but if you think it's better to have it for v12.0, it's fine. Anyway, we can push it through maintenance later.

julsemaan:
I think just to be on the really safe side, I'd wait until we have released

nqb:
@julsemaan, let's merge this if you agree.

julsemaan:
@julsemaan agrize

PR #7219, opened 2022-09-22
Fix 7218 (pfconnector dyn reverse cache issue)

### LINKED ISSUE #7218: Pfconnector server active dyn reverse cache checks can fail
**Describe the bug**
The pfconnector server uses a cache for the currently active dynamic reverses. When a reverse is closed due to inactivity (or any other reason), the cache entry stays. There is a check that tries to bind the port in the cache and if it fails to bind that port, it believes the reverse is still alive. On occasion, that inactively closed port can be reused by another process and the pfconnector will think the tunnel is still active while it isn't (the existing bind is for another process)

The solution is to cycle through all the active dynamic reverses when one closes to purge it from the cache

julsemaan:
# Description
rework active dyn reverses cache in pfconnector to stop using network binds to test

# Impacts
pfconnector

# Issue
fixes #7218

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed issue with pfconnector where it would reuse a dynamic reverse that isn't active anymore (#7218)

### DIFF
--- a/go/chisel/server/server_handler.go
+++ b/go/chisel/server/server_handler.go
@@ -13,19 +13,18 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/inverse-inc/go-utils/sharedutils"
-	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
-	"github.com/inverse-inc/packetfence/go/unifiedapiclient"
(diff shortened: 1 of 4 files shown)


julsemaan:
Tested and it works as expected. Ready for review

fdurand:
The logic is ok and the code too.

PR #7241, opened 2022-10-03
Fix source IP for pfconnector-remote deauth

### LINKED ISSUE #7239: RADIUS CoA/Disconnect reply is not seen when using a remote pfconnector (cluster)
**Describe the bug**
RADIUS CoA/Disconnect reply is not seen when using a remote pfconnector when running in a cluster because the packet is sent using the VIP to the local IP of the pfconnector-server. The issue is that the pfconnector-server sends the reply from the VIP to the VIP which prevents the packet from being seen by the socket since the IPs don't match the originating packet.

There is a workaround possible which is to disable `active_active.use_vip_for_deauth` but it will stop using the VIP to deauth both when using the pfconnector and when you're not using it.

**Expected behavior**
This needs to be handled transparently where it will not use the VIP when talking to a local pfconnector-server that is running in a cluster

julsemaan:
# Description
Fix source IP for pfconnector-remote deauth. It prevented the CoA-ACK or Disconnect-ACK from being seen by pfqueue so all attempts would log as failed although they succeeded

# Impacts
RADIUS deauth

# Issue
fixes #7239 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Fixed RADIUS deauth through pfconnector-remote in a cluster where it was logging as failed although it succeeded

### DIFF
--- a/conf/iptables.conf.example
+++ b/conf/iptables.conf.example
@@ -59,7 +59,6 @@
 # Netdata
 %%netdata%%
 
-
 # RADIUS
 -A input-management-if --protocol tcp --match tcp --dport 1812 --jump ACCEPT
 -A input-management-if --protocol udp --match udp --dport 1812 --jump ACCEPT
--- a/conf/iptables.conf.example
+++ b/conf/iptables.conf.example
@@ -58,8 +58,7 @@
 #-A input-management-if --protocol tcp --match tcp --dport 1026 --jump ACCEPT
 # Netdata
 %%netdata%%
-# pfconnector
-%%pfconnector%%
+
 
 # RADIUS
 -A input-management-if --protocol tcp --match tcp --dport 1812 --jump ACCEPT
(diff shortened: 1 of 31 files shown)


fdurand (approved):
The code looks ok and the logic too.

PureOxygen-Dev:
@julsemaan I see that this did not make it yet into the patch update.  We are not too clear how long it takes to come from the Git commit to us being able to pull the patch

Out git_commit_id  is currently 2377d23bccffc9774f9581671584a6b452b68aca that is the commit one behind this one but it was on the same day.

Can you shed some light on this process so we understand better?

julsemaan:
Unfortunately our Gitlab build system was out of sync with our repo when that build was triggered and we're experiencing issues with our nightly build process of the maintenance patches.

I launched a forced build for maintenance/12.0 with pipeline ID 665922342 (this number or a higher number will be in the deb package name)

That should be ready in ~1-2 hours

PR #7255, opened 2022-10-06
Feature/venom pf security event Suricata

JeGoi:
# Description
Add suricata security event based on inline l2 test suite.

# Impacts
- [X] Suricata test can be removed from manual testing

# Delete branch after merge
YES

# CI Pipeline:
https://gitlab.com/inverse-inc/packetfence/-/jobs/3137554969


extrafu:
Very nice! 👍

t/venom/test_suites/security_event_suricata/25_prepare_security_event_suricata.yml
```diff
@@ -0,0 +1,142 @@
...
+#- name: create_security_event
+#  steps:
+#  - type: pf_api_security_events_create 
+#    id: "{{.security_event_suricata.event.id}}"
+#    access_duration: "{{.security_event_suricata.event.access_duration}}"
+#    actions: ["autoreg", "role"]
+#    desc: "{{.security_event_suricata.event.desc}}"
+#    enabled: "Y"
+#    target_category: "{{.mac_auth.roles.headless_device.id}}"
+#    triggers: '[{"dhcp_fingerprint": "{{.security_event_suricata.node.dhcp_fingerprint_id}}"}]'
```
nqb: Can be removed I guess

t/venom/test_suites/security_event_suricata/25_prepare_security_event_suricata.yml
```diff
@@ -0,0 +1,142 @@
...
+            "mac": null,
+            "mac_vendor": null,
+            "nessus": null,
+            "nessus6": null,
+            "nexpose_event_contains": null,
+            "nexpose_event_starts_with": null,
+            "openvas": null,
+            "provisioner": null,
+            "role": null,
+            "suricata_event": "ET POLICY ",
```
nqb, line 77: Not sure if space after `POLICY` is expected

t/venom/test_suites/security_event_suricata/27_create_suricata_event.yml
```diff
@@ -0,0 +1,18 @@
+name: Create the event
+testcases:
+- name: get_node01_dhcp_on_reg
```
nqb: The name of the test case is a bit confusing because there is no "registration" network because you are using an inline L2 network.

t/venom/test_suites/security_event_suricata/teardown/40_delete_security_event.yml
```diff
@@ -0,0 +1,11 @@
+name: SecurityEventAutoreg
```
nqb: Wrong title

t/venom/test_suites/security_event_suricata/teardown/50_delete_file.yml
```diff
@@ -0,0 +1,6 @@
+name: SecurityEventAutoreg
```
nqb: Wrong title

t/venom/vars/all.yml
```diff
@@ -596,6 +596,19 @@ security_event_autoreg.node.mac_url_encoded: '{{.node01_ens7_mac_address_url_enc
...
+## Security event suricata test suite specific variables
+#################################################################################
+security_event_suricata.event.id: '88888889'
+security_event_suricata.syslog_parser.id: 'Suricata'
+security_event_suricata.event.access_duration: '12h'
+security_event_suricata.event.desc: 'ET P2P Kaaza Media'
+security_event_suricata.node.mac: '{{.node01_ens7_mac_address}}'
+security_event_suricata.node.mac_url_encoded: '{{.node01_ens7_mac_address_url_encoded}}'
+security_event_suricata.node.dhcp_fingerprint: 1,28,2,3,15,6,119,12,44,47,26,121,42
+security_event_suricata.node.dhcp_fingerprint_id: 276
```
nqb: I don't think these variables are used

t/venom/lib/pf_api_syslog_parser_action.yml
```diff
@@ -0,0 +1,16 @@
...
+  time_to_sleep: 0
+steps:
+- type: pf_api_action
+  body: '{{.input.body}}'
+  method: "{{.input.method}}"
+  status_code: "{{.input.status_code}}"
+  url: 'config/syslog_parser/{{.input.id}}'
+  time_to_sleep: "{{.input.time_to_sleep}}"
+output:
+  pf_api_realm_action_: "{{.result.pf_api_action_}}"
```
nqb: I think there is a mistake here (`realm`)

t/venom/test_suites/security_event_suricata/TESTSUITE.md
```diff
@@ -0,0 +1,37 @@
+# Inline L2
```
nqb: Wrong title

nqb (changes requested):
General remarks:
- security_events_suricata test suite should be integrated into inline scenarios. Otherwise, we are adding another scenario which take around 25 minutes to run
- I don't see a reason to rename existing scenario "security_events" to "security_events_autoreg". If we want to add other "security events" test suites inside this scenario, it will be a problem.
- There are several changes in Venom executors which are not used by security_events_suricata test suite but by other test suites. Could you move these changes into a dedicated PR ? In that way, we will be able to check if they have impacts on other test suites.
- In TESTSUITE.md, some steps are mentioned but they are missing in fact because the logic is different.
- The device is registered in (20_local..) but in fact, we don't rely on that later. So it's not really necessary to do it.

t/venom/test_suites/security_event_suricata/25_prepare_security_event_suricata.yml
JeGoi: Still the issue with venom 1.0.1 version... I should have remove it.

JeGoi:
@nqb :
1) Q. security_events_suricata test suite should be integrated into inline scenarios. Otherwise, we are adding another scenario which take around 25 minutes to run
A. Why not... It is just that it will be hidden inside this scenario which is not talking about security events...

2) Q. I don't see a reason to rename existing scenario "security_events" to "security_events_autoreg". If we want to add other "security events" test suites inside this scenario, it will be a problem.
A. Ok but how to manage it with point 1. ~~Another solution is to name scenarios with vms deployed inside...~~(bad idea)

3) Q. There are several changes in Venom executors which are not used by security_events_suricata test suite but by other test suites. Could you move these changes into a dedicated PR ? In that way, we will be able to check if they have impacts on other test suites.
A. Yes and no, at the beginning, we used security_events_autoreg as a reference them, libs have been created/modified. Because we tried to add more tests (web access), it has been rebased on inline/l2 which is easier to manage nodes, add a dhcp server, etc.
As you can see there, https://gitlab.com/inverse-inc/packetfence/-/jobs/3137913477 it is running even with these modifications. So, before revert them, why not just run the others? On current pipeline test, issues are mostly not related to venom.

4) Q. In TESTSUITE.md, some steps are mentioned but they are missing in fact because the logic is different.
A. I will double check. All steps are mostly the same except the suricata settings and how the flag is risen.

5) Q. The device is registered in (20_local..) but in fact, we don't rely on that later. So it's not really necessary to do it.
A. I think here you are wrong, it needs to be register to check the access to internet. It is the equivalent of: ./t/venom/test_suites/inline/l2/20_local_user_authentication.yml

t/venom/test_suites/security_event_suricata/25_prepare_security_event_suricata.yml
JeGoi: Fixed

t/venom/test_suites/security_event_suricata/25_prepare_security_event_suricata.yml
JeGoi: It is coming from the request. No clue if it is, but it is a copy paste. So...

t/venom/test_suites/security_event_suricata/27_create_suricata_event.yml
JeGoi: Correct. Renamed.

t/venom/test_suites/security_event_suricata/teardown/50_delete_file.yml
JeGoi: Fixed

t/venom/vars/all.yml
JeGoi: Your are right connected to the first iteration of this test_suite.
Fixed

t/venom/lib/pf_api_syslog_parser_action.yml
JeGoi: Fixed

t/venom/test_suites/security_event_suricata/TESTSUITE.md
JeGoi: Fixed with other improvements.

t/venom/test_suites/security_event_suricata/25_prepare_security_event_suricata.yml
```diff
@@ -0,0 +1,130 @@
+name: Create autoreg security event
```
nqb: Wrong title

nqb:
### 1 and 2

`security_events_suricata`  is an inline test, `security_event_autoreg` can be seen as a RADIUS test. So perhaps you can rename things accordingly.

### 3

I think we can keep changes in this PR if we are sure there is no regression on other test suites.

If I'm not wrong, changes in these PR have impact on other test suites (not only on `security_events_autoreg`).
But again, it's really more easy to review small PR with limited changes on specific things.

### 4

These two steps are still mentioned but they are not executed:
```
1. Change the role of the device to guest
1. Test if the device is in the role ID 2 ipset set
```

### 5

Yes you are right.

JeGoi:
### 1&2
Ok I have already revert it and add security_event_suricata in inline scenario.

### 3
Looks fine to me https://gitlab.com/inverse-inc/packetfence/-/pipelines/661194453
Errors are in the inline l2 and inline/suricata. My guess is that inline/suricata is using the same data as the inline/l2 and i may cause trouble.

### 4
Oh, you are right! Will be  fixed.

nqb:
Running a pipeline with all tests: https://gitlab.com/inverse-inc/packetfence/-/pipelines/663779856

t/venom/test_suites/security_event_suricata/30_tests.yml
```diff
@@ -0,0 +1,53 @@
...
+  - type: pf_api_node_security_events
+    id: "{{.security_event_suricata.node.macaddress}}"
+    vars:
+      all_events:
+        from: result.pf_api_node_security_event_
+
+- name: get_event_id
+  steps: 
+  - type:
+    script: echo '{{.get_security_event_id.all_events}}' | jq '.[] | if (.security_event_id == 88888889 and .status == "open") then .id else false end ' | grep -v "false"
```
JeGoi:
```suggestion
    script: echo '{{.get_security_event_id.all_events}}' | jq '.[] | if (.security_event_id == {{.security_event_suricata.event.id}} and .status == "open") then .id else false end ' | grep -v "false"
```

mellalahmed:

PR #7232, opened 2022-09-29
Debian package for PacketFence Connector

nqb:
# Description
Build a Debian package for PacketFence Connector.
 
Build is identical to other packetfence packages.

# Impacts
packetfence-pfconnector installation

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [X] Document the feature

# NEWS file entries
## Enhancements
* Debian package for PacketFence Connector

PR #7256, opened 2022-10-06
Feature/fingerbank via connector

julsemaan:
# Description
Adds support for the Fingerbank Collector on a pfconnector-remote instance

# Impacts
Fingerbank in PacketFence

# Code / PR Dependencies
Needs the pfconnector-remote debian packaging to work #7232 

# NEW Package(s) required
fingerbank-collector-1.4.1 and fingerbank-collector-remote-1.4.1 (both in devel repos)

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [X] Document the feature
- [N/A] Add unit tests

# NEWS file entries
## New Features
* Added support for the Fingerbank Collector on a PacketFence connector

### DIFF
--- a/addons/pfconnector/configure.sh
+++ b/addons/pfconnector/configure.sh
@@ -41,3 +41,4 @@ if ! prompt "Should the pfconnector server TLS certificate be validated?"; then
 fi
 
 echo "FETCH_REMOTES_VIA_API=true" >> /usr/local/pfconnector-remote/conf/pfconnector-client.env
+echo "PFCONNECTOR_REMOTE=true" >> /usr/local/pfconnector-remote/conf/pfconnector-client.env
--- a/go/chisel/server/server_handler.go
+++ b/go/chisel/server/server_handler.go
@@ -11,6 +11,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/gin-gonic/gin"
 	"github.com/inverse-inc/go-utils/sharedutils"
 	chshare "github.com/inverse-inc/packetfence/go/chisel/share"
@@ -19,7 +20,6 @@ import (
 	"github.com/inverse-inc/packetfence/go/chisel/share/tunnel"
 	"github.com/inverse-inc/packetfence/go/pfconfigdriver"
 	"github.com/inverse-inc/packetfence/go/unifiedapiclient"
-	"github.com/inverse-inc/packetfence/go/util"
 	"github.com/phayes/freeport"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/sync/errgroup"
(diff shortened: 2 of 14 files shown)


julsemaan:
# TODO

- [ ] High level schema for how it works
- [x] Add fingerbank-collector-remote as a dependency to packetfence-pfconnector-remote

jrouzierinverse:
Looks good so far. What is the best way to test?

addons/pfconnector/configure.sh
```diff
@@ -41,3 +41,4 @@ if ! prompt "Should the pfconnector server TLS certificate be validated?"; then
 fi
 
 echo "FETCH_REMOTES_VIA_API=true" >> /etc/pfconnector-client.env
+echo "PFCONNECTOR_REMOTE=true" >> /etc/pfconnector-client.env
```
nqb: We will need to rebase this PR on `devel` once #7232 is merged because this path will not be valid.

julsemaan:
> Looks good so far. What is the best way to test?

It's pretty manual now since the packaging for the pfconnector-remote is not completed yet but here are the general guidelines

- Install the pfconnector-remote from this branch
- Install fingerbank-collector-remote
- Connect your pfconnector-remote to PF
- Test some stuff related to Fingerbank for a node behind the pfconnector-remote

nqb (approved):

julsemaan:
@fdurand and @jrouzierinverse, can you provide a review of the code?

Thanks!

lib/pf/iptables.pm
```diff
@@ -154,6 +154,18 @@ sub iptables_generate {
     }
     $tags{'netdata'} .= "-A input-management-if --protocol tcp --match tcp --dport 19999 --jump DROP\n";
 
+    # The dynamic range used to access the fingerbank collector that are connected via a remote connector
+    $tags{'pfconnector'} = "";
```
fdurand, line 158: i don´t see pfconnector tag in iptables.conf.example

go/chisel/server/server_handler.go
```diff
@@ -253,22 +278,165 @@ func (s *Server) handleDynReverse(w http.ResponseWriter, req *http.Request) {
...
+	connectorId := req.URL.Query().Get("connector-id")
+	if connectorId == "" {
+		w.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(w).Encode(unifiedapiclient.ErrorReply{Status: http.StatusNotFound, Message: "Missing connector-id query parameter"})
+		return
+	}
+
+	if o, ok := activeTunnels.Load(connectorId); ok {
+		tun := o.(*tunnel.Tunnel)
+		index := s.computeConnectorIndex(connectorId)
```
jrouzierinverse, line 294: It looks like this index only increases.

jrouzierinverse (changes requested):

go/chisel/server/server_handler.go
julsemaan: It does, I'm assuming that we're connecting less than 256 connectors to a single server which seems like a fair limit for the moment

jrouzierinverse: Can connectors be removed?

julsemaan: They can but it's pretty unlikely a lot of them get removed. Worst case, the service can be restarted to clear the counter

lib/pf/iptables.pm
julsemaan: Good catch, it's fixed

fdurand (approved):

PR #7286, opened 2022-10-17
Fixes VPN attribute not returned each time

fdurand:
# Description
In some cases the vpn attribute is not returned by PacketFence and it should.

# Impacts
RADIUS VPN workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Return radius attribute if define for each VPN request.

# UPGRADE file entries

Run to-12.1-move-rolebyname-to-vpnbyname-fortigate.pl to copy the Role attributes in the VPN one for FortiGate

### DIFF
--- a/lib/pf/Switch/Fortinet/FortiGate.pm
+++ b/lib/pf/Switch/Fortinet/FortiGate.pm
@@ -269,7 +269,7 @@ sub returnAuthorizeVPN {
     if ( isenabled($self->{_VpnMap}) && $self->supportsVPNRoleBasedEnforcement()) {
         $logger->debug("Network device (".$self->{'_id'}.") supports roles. Evaluating role to be returned");
         if ( defined($args->{'user_role'}) && $args->{'user_role'} ne "" ) {
-            $role = $self->getRoleByName($args->{'user_role'});
+            $role = $self->getVpnByName($args->{'user_role'});
         }
         if ( defined($role) && $role ne "" ) {
             $radius_reply_ref = {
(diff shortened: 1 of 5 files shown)


julsemaan (approved):

julsemaan:
I don't think this should be a +1 though, this should be in 12.1 as it is more a feature than a bug fix

julsemaan:
I'm good with the PR

@jrouzierinverse, your review was requested too, can you review this and merge

lib/pf/Switch/Fortinet/FortiGate.pm
```diff
@@ -269,7 +269,7 @@ sub returnAuthorizeVPN {
     if ( isenabled($self->{_VpnMap}) && $self->supportsVPNRoleBasedEnforcement()) {
         $logger->debug("Network device (".$self->{'_id'}.") supports roles. Evaluating role to be returned");
         if ( defined($args->{'user_role'}) && $args->{'user_role'} ne "" ) {
-            $role = $self->getRoleByName($args->{'user_role'});
+            $role = $self->getVpnByName($args->{'user_role'});
```
jrouzierinverse, line 272: Should we have a migration script to copy the role map to the VPN map?

jrouzierinverse (changes requested):

lib/pf/Switch/Fortinet/FortiGate.pm
julsemaan: Looks like it does make sense to have one specific for the Fortigate since it is indeed a change of behavior

fdurand: Upgrade script uploaded

jrouzierinverse:
It looks tested it and it works

PR #7301, opened 2022-10-25
Remove set +e to stop Debian installation/upgrade at first error

nqb:
# Description
If you use `do-upgrade.sh` script to perform an upgrade and you got an issue in the steps **after** `set +e`, full upgrade will not stop and you got inconsistent results.

# Impacts
- Debian installation
- Debian upgrades
- Debian ISO

# Issue
I didn't open any issue but I have seen this behavior during a customer upgrade

# Delete branch after merge
YES


nqb:
Before merging this, we need to be sure `docker-minimal-rules.sh` doesn't exit if conditions are not met. Otherwise, we will break installations.

nqb:
I can confirm `docker-minimal-rules.sh` is not failing `iptables -L DOCKER` returns 1.

nqb:
Running a pipeline with a configurator scenario to ensure I didn't miss anything.

I prefer to not backport this fix to maintenance because it has the potential to break some things.

nqb:
Pipeline passed, merging. ISO will be tested during Q&A to be sure it doesn't cause issues.

PR #7306, opened 2022-10-26
Remove ip*tables.conf from config files

### LINKED ISSUE #7298: do-upgrade.sh failed with a custom iptables.conf during v12.0.0 upgrade
**Describe the bug**
If you modified your `iptables.conf` file manually, when you will upgrade to v12.0.0 using `do-upgrade.sh`, your `iptables.conf` will be kept as is and will prevent some services to start correctly during package upgrade.

This is because we force `apt` to keep existing files and we handle (later in the process) `.dpkg-dist` files. However, this logic doesn't work for v12.0.0 because we need to have iptables started before other services to permit communication with containers.

**To Reproduce**
1. Edit `iptables.conf` on a setup before v12.0.0
2. Upgrade to v12.0.0 using `do-upgrade.sh`

=> Your upgrade process will failed at some point because iptables is not started correctly.

**Expected behavior**
We should always install `iptables.conf` from packaging. Custom rules can go in `.conf.inc` files.

nqb:
# Description
Remove iptables.conf and ip6tables.conf from config files.

All custom rules should go in .conf.inc files.

# Impacts
- Custom rules in iptables.conf or ip6tables.conf
- Upgrade to v12.0.0

# Issue
fixes #7298

# Delete branch after merge
YES

### DIFF
--- a/conf/ip6tables.conf.example
+++ b/conf/ip6tables.conf.example
@@ -1,3 +1,6 @@
+# Copyright (C) Inverse inc.
+# iptables template
+# This file is manipulated on PacketFence's startup before being given to iptables
 *raw
 :PREROUTING ACCEPT [0:0]
 :OUTPUT ACCEPT [0:0]
--- a/conf/iptables.conf.example
+++ b/conf/iptables.conf.example
@@ -81,9 +81,6 @@
 -A input-management-if --protocol tcp --match tcp --dport 9392 --jump ACCEPT
 # Nessus Administration Interface
 -A input-management-if --protocol tcp --match tcp --dport 8834 --jump ACCEPT
-# PacketFence-PKI
-# -A input-management-if --protocol tcp --match tcp --dport 9393 --jump ACCEPT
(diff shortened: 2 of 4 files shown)


nqb:
There is an entry in Upgrade guide related to .conf.inc files: https://www.packetfence.org/doc/PacketFence_Upgrade_Guide.html#_support_of_custom_rules_in_iptables_conf

nqb:
I tested to upgrade from 11.X and from 12.X to packages built on this branch on Debian 11 and EL8, in all cases, iptables.conf and ip6tables.conf are replaced.

On EL8, existing ip*tables.conf files are saved into .rpmsave. On Debian, files are replaced.

Ready for a review and a backport to maintenance.

julsemaan:
I put this in the maintenance/12.0 branch, should this go back to 11.x too?

nqb:
> I put this in the maintenance/12.0 branch, should this go back to 11.x too?

AFAIK, it doesn't cause any issues on 11.X so I suggest to not backport.
On top of that, you will certainly get some trouble with ip6tables.conf added later.

Issue #7182, opened 2022-09-06
v12(upgrade): bad handling of .dpkg-dist file for pfconfig.conf


nqb:
**Describe the bug**
While testing #7180 on node C (first step of the upgrade), I used `run-upgrade.sh` to upgrade packages, configuration and database on node C.

`packetfence-config` was not able to start at the end because `pfconfig.conf` was empty. Logs:

```
Handling .dpkg-dist conf/pfconfig.conf.dpkg-dist
Downloading https://raw.githubusercontent.com/inverse-inc/packetfence/e122a7b3d3236be44336875dc9f4007961fb495f/conf/pfconfig.conf.example
Moving conf/pfconfig.conf.dpkg-dist -> conf/pfconfig.conf and creating backup file conf/pfconfig.conf.upgrade-backup
Attempting a dry-run of the patch on conf/pfconfig.conf
checking file conf/pfconfig.conf
Hunk #1 FAILED at 3.
1 out of 1 hunk FAILED
Patching conf/pfconfig.conf failed. Will put the .dpkg-dist file in place. This should be addressed manually after the upgrade is completed. Press enter to continue
```

=> At the end, we got a `pfconfig.conf` which is empty (identical to `pfconfig.conf.example` installed by `packetfence-config` package) where we have kept `pfconfig.conf.upgrade-backup` 

I propose following solution: `pfconfig.conf` should not generate a `.dpkg-dist` file anymore because it's empty and everything is in `pfconfig.conf.defaults`.

julsemaan:
I created a "dirty" patch for this in maintenance/11.2 through 5fef35a3066725095dbd87e337f68a34b247a920 and bf6281264f25e3719d9d266d5cbab65ebb12c2eb 

I noticed that the dpkg-dist behavior is there for other config files too (like authentication.conf). We should remove this behavior from the configuration files that are handled by our upgrade scripts

nqb:
There is also this fix: https://github.com/inverse-inc/packetfence/commit/a481781d8260c955eb1a50e13cb8e966614c027c

nqb:
`authentication.conf` currently generate a .rpmnew file.

[...]

PR #7308, opened 2022-10-26
Create authentication.conf.defaults to avoid .rpmnew and .dpkg-dist at each upgrade

nqb:
# Description
Move defaults auth sources into authentication.conf.defaults to keep authentication.conf.example empty. In this way, no .rpmnew or .dpkg-dist will be created during upgrade.

Changes in authentication.conf will be manage using upgrade scripts

# Impacts
- Install
- Upgrades

# Issue
Related to #7182 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* FIXME


nqb:
Did some tests on a fresh EL8:
- Not able to validate if sources still work on captive portal due to #7309 
- New auth sources appear in authentication.conf
- Defaults auth sources appear in web admin with a locker but you can still modify these sources.

I'm not sure about the approach we should have with defaults auth sources because there are used in default portal module.

nqb:
Needs to make wide changes on backend. I don't think we have the time to do it.

PR #7328, opened 2022-10-31
Feature/7173

JeGoi:
# Description
Should be able to fix 7173 by adding the connector option available for firewall sso config

# Impacts
Add connector option in firewall sso

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* connector option in all firewall sso

### DIFF
--- a/addons/upgrade/to-12.1-firewallsso.pl
+++ b/addons/upgrade/to-12.1-firewallsso.pl
@@ -1,0 +1,65 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+firewallsso_to_update -
+
+=head1 DESCRIPTION
+
+firewallsso_to_update
+
+=cut
+
+use strict;
+use warnings;
+use lib qw(/usr/local/pf/lib /usr/local/pf/lib_perl/lib/perl5);
+use pf::IniFiles;
+use pf::constants::config;
+use pf::file_paths qw(
+    $firewall_sso_config_file
+);
+
+
+my $fsso = pf::IniFiles->new( -file => $firewall_sso_config_file, -allowempty => 1);
+my @groups;
+my @switches;
+
+if (length ($fsso->Sections())) > 0
+    for my $section ($fsso->Sections()) {
+        if (!($fsso->exists($section, "use_connector"))) {
+            $fsso->newval($section, 'use_connector', 'enabled');
+        } else {
+            print "The section $section has already the option use_connector defined"
+        }
+    }
+    $fsso->RewriteConfig();
+} else {
+    print "Nothing to do\n";
+}
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2021 Inverse inc.
+
+=head1 LICENSE
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+as published by the Free Software Foundation; either version 2
+of the License, or (at your option) any later version.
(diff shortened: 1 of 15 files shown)


go/firewallsso/base.go
```diff
@@ -242,7 +243,7 @@ func (fw *FirewallSSO) logger(ctx context.Context) log15.Logger {
 }
 
 func (fw *FirewallSSO) getDst(ctx context.Context, proto string, toIP string, toPort string) string {
-	if cc := connector.ConnectorsContainerFromContext(ctx); cc != nil {
+	if fw.UseConnector == "enabled", cc := connector.ConnectorsContainerFromContext(ctx); cc != nil {
```
julsemaan: Small detail, you can use `sharedutils.IsEnabled(fw.UseConnector)` for a cleaner handling in case the value is set to `yes` or `1`

JeGoi: Done!
Thanks!

julsemaan:
You seem to be missing:
- Form elements in html/pfappserver/lib/pfappserver/Form/Config/Firewall_SSO/ so that your field is recognized by the backend
- An upgrade script that adds use_connector=enabled to all firewall SSO entries so that the default behavior continues to be using the connectors. Otherwise, with this change, the empty/non-existant use_connector value will make it so existing entries will not use the connector on 12.1 while they were using it on 12.0

html/pfappserver/root/src/views/Configuration/firewalls/_components/FormTypeBarracudaNg.vue
```diff
@@ -52,6 +52,12 @@
       :column-label="$i18n.t('Default realm')"
       :text="$i18n.t('The default realm to be used while formatting the username when no realm can be extracted from the username.')"
     />
+
+    <form-group-use-connector namespace="use_connector"
+      :column-label="$i18n.t('Use Connector')"
+      :text="$i18n.t('Use the available PacketFence connectors to connect to this authentication source. By default, a local connector is hosted on this server. Using remote connectors is only supported on a standalone instance at the moment.')"
```
julsemaan:
```suggestion
      :text="$i18n.t('Use the available PacketFence connectors to connect to this firewall.')"
```

It has to be changed in the other form files too

addons/upgrade/to-12.1-firewallsso.pl
```diff
@@ -0,0 +1,63 @@
...
+    $firewall_sso_config_file
+);
+
+
+my $fsso = pf::IniFiles->new( -file => $firewall_sso_config_file, -allowempty => 1);
+
+if (length ($fsso->Sections()) > 0) {
+    for my $section ($fsso->Sections()) {
+        if (!($fsso->exists($section, "use_connector"))) {
+            $fsso->newval($section, 'use_connector', 'enabled');
```
julsemaan: Your 'checkbox_value' is `1` so the value for `enabled` here should be `1` too

julsemaan (changes requested):

addons/upgrade/to-12.1-firewallsso.pl
JeGoi: Will do.
I just remember a code that says: enable=1=true, disable=0=false, but it may not be applied here.

JeGoi: Done

JeGoi:
I also rebased on devel

PR #7333, opened 2022-11-01
Removed monitor on eduroam authentication source

fdurand:
# Description
Since the eduroam radius authentication sources are defined as a radius source, it doesn´t make sense anymore to monitor the eduroam source but just monitor the radius one.

# Impacts
pfstats

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Deprecated eduroam test from pfstats

### DIFF
--- a/go/cmd/pfstats/main.go
+++ b/go/cmd/pfstats/main.go
@@ -177,65 +177,6 @@ func (s ldaptype) Test(source interface{}, ctx context.Context) {
 	}
 }
 
-var EDUROAM eduroamtype
-
-type eduroamtype struct{}
-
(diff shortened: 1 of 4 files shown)


julsemaan:
Not sure this is really a +1 patch since the eduroam rework is unreleased so I'm not backporting

Let me know if it should be backported for some other reason

fdurand:
I set +1 in case just in case it was too late for 12.1

julsemaan:
It's never too late for fixes ❤️

PR #7338, opened 2022-11-02
Fix / Add sameorigin in haproxy admin for netdata

JeGoi:
# Description
Add X-Frame-Options SAMEORIGIN in haproxy-admin
(source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#configuring_haproxy)

# Impacts
x-frame-options: SAMEORIGIN is now in header

# Delete branch after merge
YES

## Bug Fixes
Possible Clickjacking for netdata reverse proxy


JeGoi:
@fdurand should it be applied in maintenance branches?

JeGoi:
Pushed in maintenance/11.0 to maintenance/12.0

PR #7350, opened 2022-11-07
UI e2e testing framework

satkunas:
# Description
Integrate cypress end-to-end testing framework

# Impacts
Venom workflows
CI/CD results

# NEW Package(s) required
## RHEL
```
xorg-x11-server-Xvfb gtk2-devel gtk3-devel libnotify-devel GConf2 nss libXScrnSaver alsa-lib
```
## Debian
```
libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb
```

# Delete branch after merge
YES 

# Checklist
(REQUIRED) - [yes, no or n/a]
- [x] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Adding end-to-end testing framework to UI for CI/CD pipelines


.gitlab-ci.yml
```diff
@@ -1310,6 +1310,24 @@ configurator_deb11_branches:
     - .test_script_job
     - .test_branches_only_rules
 
+pfappserver_el8_branches:
```
nqb, line 1329: You need to create two other jobs `pfappserver_el8` and `pfappserver_deb11` which will be used on `devel` and maintenance branches.

t/html/pfappserver/Makefile
```diff
@@ -0,0 +1,68 @@
...
+	@echo "make [target] TAG=production"
+	@echo ""
+	@echo "DISPLAY=\$$(cat /etc/resolv.conf | grep nameserver | awk '{print \$$2; exit;}'):0.0 make [target]"
+
+install-rhel:
+	yum install -y xorg-x11-server-Xvfb gtk2-devel gtk3-devel libnotify-devel GConf2 nss libXScrnSaver alsa-lib
+	yum install -y firefox
+
+install-debian:
+	apt-get update
```
nqb:
```suggestion
	apt-get update -y
```

t/html/pfappserver/Makefile
```diff
@@ -0,0 +1,68 @@
...
+	npm install cypress -g --allow-root --unsafe-perm=true
+	npm ci --legacy-peer-deps
+
+test:
+		DEBUG=$(DEBUG) \
+		DISPLAY= \
+		NO_COLOR=1 \
+		BROWSERSLIST_IGNORE_OLD_DATA=true \
+		CYPRESS_baseUrl=$(BASE_URL) \
+		cypress run --browser $(BROWSER) --config-file $(CONFIG_FILE) --config $(CONFIG) --e2e --headless --env tags=[$(TAGS)] ; \
```
nqb: Here we use `TAGS` but there is no default value for `TAGS` in Makefile (at the beginning of file).
Also, in `make all`, usage is `TAG=production` (TAG without `s`)

t/html/pfappserver/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+See [`Makefile`](Makefile) for usage.
+
+* __DEBUG__: Cypress debug (example: DEBUG=cypress:*, default: none), (see https://docs.cypress.io/guides/references/troubleshooting#Log-sources).
+* __BASE_URL__: Base URL used within tests (default: BASE_URL=https://localhost:1443).
+* __CONFIG_FILE__: Cypress configuration file (default: cypress/config/cypress.config.js).
+* __CONFIG__: Comma-separated list of configuration file overloads. (example CONFIG=downloadsFolder=/tmp/downloads,screenshotsFolder=/tmp/screenshots,videosFolder=/tmp/videos).
+* __PROJECT_ID__: Cypress Cloud Project ID. Only used in `make test-e2e`.
+* __RECORD_KEY__: Cypress Cloud Recording Key. Only used in `make test-e2e`.
+* __TAG__: Optional Tag shown in Cypress Cloud test runs.
```
nqb, line 50: Same note regarding `TAG/TAGS`

t/venom/scenarios/pfappserver/playbooks/run_tests.yml
```diff
@@ -0,0 +1,35 @@
...
+    - pfappserver_configurator
+
+  tasks:
+    - name: Install Cypress
+      community.general.make:
+        chdir: /usr/local/pf/t/html/pfappserver
+        target: install
+        file: /usr/local/pf/t/html/pfappserver/Makefile
+      become: true
+      tags: cypress
```
nqb: I suggest to move this task directly in `configure.yml` playbook.

t/venom/vars/all.yml
```diff
@@ -671,3 +671,9 @@ firewall_sso.radius.radius_user_name: foo
 firewall_sso.radius.radius_user_password: bar
 firewall_sso.radius.radius_mock_dir: '{{.pfserver_root_dir}}/t/mock_servers/node_radius'
 firewall_sso.radius.type: FortiGate
+
+#################################################################################
+## html e2e tests
+#################################################################################
+html.pfappserver_baseurl: 'https://{{.pfserver_mgmt_ip}}:1443'
```
nqb, line 678: You can use directly: `{{.pfserver_webadmin_url}}`

This is the value used by default in `t/venom/lib/html_e2e.yml`

nqb (changes requested):

nqb:
### Difference between  `pfappserver_configurator` and `configurator` test suites

It looks like there are differences between `pfappserver_configurator` and `configurator` (existing) test suites.
* Step 1: `configurator` will configure dhcp-listener, registration and isolation interfaces, DNS servers
* Step 2: `configurator` will configure alerting and send an email
* Post-configurator: `configurator` will use new admin password defined at step 2 to login against API 

### [X] Packaging

We need to package `t/html/` directory in `packetfence-test` package. Not sure if this is the case at the moment.

EDIT: packaging checked and `t/html` is part of `packetfence-test` on RPM and Deb packages.

t/html/pfappserver/Makefile
satkunas: `TAG` and `TAGS` are different env variables.

 * `TAG` is used for only display when running e2e through the Cypress cloud. It allows easy sorting and filtering past tests based on keyword(s).
 * `TAGS` are used to filter `describe()` and `it()` using pre-defined keywords, thus allowing to conditionally run specific test units. This support was added for future use, but is currently not used since the configurator requires a single-unit (single session SPA). I'll add some documentation to clarify this.

t/venom/scenarios/pfappserver/playbooks/run_tests.yml
satkunas: For some reason the cypress binary gets removed and needs to be reinstalled before each test. It's been moved from configure to here intentionally to work-around this issue.

t/venom/vars/all.yml
satkunas: I didn't want to hardcode this in the vars since the tests may eventually need to work on the captive-portal, or some other host, port or path.

.gitlab-ci.yml
```diff
@@ -1161,6 +1161,22 @@ configurator_deb11:
...
+    - .test_script_job
+    - .test_branches_only_rules
+
+pfappserver_deb11:
+  variables:
+    VAGRANT_PF_DOTFILE_PATH: /var/local/gitlab-runner/vagrant/vagrant-${CI_JOB_NAME}-${CI_JOB_ID}
+  extends:
+    - .test_job
+    - .test_script_job
+    - .test_branches_only_rules
```
JeGoi:
```suggestion
    - .test_devel_and_maintenance_rules
```

.gitlab-ci.yml
```diff
@@ -1161,6 +1161,22 @@ configurator_deb11:
     - .test_script_job
     - .test_devel_and_maintenance_rules
 
+pfappserver_el8:
+  variables:
+    VAGRANT_PF_DOTFILE_PATH: /var/local/gitlab-runner/vagrant/vagrant-${CI_JOB_NAME}-${CI_JOB_ID}
+  extends:
+    - .test_job
+    - .test_script_job
+    - .test_branches_only_rules
```
JeGoi:
```suggestion
    - .test_devel_and_maintenance_rules
```

t/venom/Makefile
```diff
@@ -254,6 +254,18 @@ cluster_el8:
...
+	PF_VM_NAMES=pfel8$(DEV_ENV) \
+	SCENARIOS_TO_RUN=pfappserver \
+	$(MAKE_TARGET)
+
+pfappserver_deb11:
+	make \
+	PF_VM_NAMES=pfdeb11$(DEV_ENV) \
+	SCENARIOS_TO_RUN=pfappserver \
+	$(MAKE_TARGET)
+
```
JeGoi: This part of that file is more for testing.
I think it should be moved around line 218.

t/html/pfappserver/Makefile
```diff
@@ -0,0 +1,69 @@
...
+all:
+	@echo "Usage:"
+	@echo "make install-debian|install-rhel: install libraries"
+	@echo "make install: install cypress"
+	@echo "make test: run tests headless"
+	@echo "make test-e2e: run tests headless, requires PROJECT_ID and RECORD_KEY"
+	@echo "make test-x11: run tests interactively (w/ X11 \$$DISPLAY)"
+	@echo ""
+	@echo "make [target] DEBUG=cypress:* (see https://docs.cypress.io/guides/references/troubleshooting#Log-sources)"
+	@echo "make [target] BROWSER=firefox:nightly"
```
JeGoi, line 26:
```suggestion
	@echo "make [target] BROWSER= chrome" # or for ex: firefox:nightly
```
To be consistant with default values

t/html/pfappserver/cypress/config/cypress.config-configurator.js
```diff
@@ -0,0 +1,13 @@
...
+  ...config,
+
+  e2e: {
+    ...config.e2e,
+
+    specPattern: [
+      'cypress/specs/e2e/*-configurator/**/*.cy.{js,jsx,ts,tsx}',
+    ],
+  }
+};
```
JeGoi, line 13: Same as the other one => t/html/pfappserver/cypress/config/cypress.config-configuration.js
Duplicate?

t/html/pfappserver/cypress/config/cypress.config.js
```diff
@@ -0,0 +1,51 @@
...
+  webpackOptions: require('../../webpack.config'),
+  watchOptions: {}
+}
+
+// see https://docs.cypress.io/guides/references/configuration#Global
+
+module.exports = {
+//  defaultCommandTimeout: 10000, // 10s
+  e2e: {
+    baseUrl: 'https://localhost:1443',
```
JeGoi, line 12: Isn't it supposed to be variables? or there is an override system

t/venom/scenarios/pfappserver/playbooks/run_tests.yml
JeGoi: I think @nqb is right, even if it is deleted, this is part of the configuration not the test themself.
Both configure and run_tests are run before tests.
You should be able to add it after `Install Cypress dependencies` so at the end.
Perhaps need more details.

t/venom/scenarios/pfappserver/playbooks/localdev.yml
```diff
@@ -0,0 +1,63 @@
+---
+- hosts: pfservers
+  name: localdev pre run_tests setup
```
JeGoi, line 3: So, are we agreed that is only played when not in the CI, correct?

t/html/pfappserver/cypress/support/commands.js
```diff
@@ -0,0 +1,73 @@
...
+      cy.get('form button[type="submit"]').first().click()
+    })
+  })
+  */
+  return cy.pfUnifiedSystemPassword().then(password => {
+    return cy.request('POST', '/api/v1/login', { username: 'system', password }).then(response => {
+      return window.localStorage.setItem('user-token', response.body.token)
+    })
+  })
+})
```
JeGoi, line 16: Are we using API or the browser?

JeGoi (changes requested):
Some modifications/answers on my sides

t/html/pfappserver/Makefile
satkunas: The default was changed, firefox is preferred due to lower memory usage. It also gives us better coverage that will prevent firefox-only issues in the future.

t/html/pfappserver/cypress/config/cypress.config-configurator.js
satkunas: That's a stub for differentiation so it's more explicit for the next person to setup a new scenario using the same framework.

t/html/pfappserver/cypress/config/cypress.config.js
satkunas: Cypress uses NodeJS, so this is what's expected

t/html/pfappserver/cypress/support/commands.js
satkunas: The configurator is using the Browser, which doesn't require /login

This was created during the PoC to provide a support helper for writing tests. I left it there to help with future development.

t/html/pfappserver/cypress/config/cypress.config.js
satkunas: I misunderstood which line you were referring to. I understand now.

This is the default value. Our vars are injected in the scenario through make using the `BASE_URL` environmental value.

nqb:
I resolved some conversation.
Two items remaining for me:
- Difference between pfappserver_configurator and configurator test suites
- Tag and Tags

JeGoi (changes requested):
Move t/venom/scenarios/pfappserver/README.md to t/venom/test_suites/pfappserver_configurator/TESTSUITE.md to be consistant with the other tests.

t/venom/scenarios is more for ansible to prepare the environment and the test and it can be shared with multiple test_suites.

satkunas:
@nqb repo rebased w/ devel.

JeGoi (approved):
Ok, looks good to me.

nqb:
Running a pipeline to see if artifacts are saved correctly: https://gitlab.com/inverse-inc/packetfence/-/pipelines/891933439

t/html/pfappserver/Makefile
nqb: Doc adjusted.

nqb:
Regarding Difference between `pfappserver_configurator` and `configurator` test suites, @satkunas is aware that test suites are different. We will keep things as is. Once we want to deprecate `configurator` test suite, we will extend what `pfappserver_configurator` is doing.

nqb (changes requested):
@satkunas, I identified few issues:

* We need to configure 2nd interface (eth1) as management during configurator and not 1st interface (eth0). Otherwise, it can cause issues because iptables will reject SSH requests on eth1
* Artifacts were not correctly pulled using Ansible because hostname is changed during configurator. This issue should be fixed now
* I'm not able to see if screenshots are available in GitLab because it looks like current Cypress test doesn't generate screenshots, could you tell me how I can enable this ?
* There is a firefox error when we run tests on Debian (I provided logs in DM)

Also I opened https://github.com/inverse-inc/packetfence/issues/7704 because I got it while testing.

satkunas:
@nqb 

> We need to configure 2nd interface (eth1) as management during configurator and not 1st interface (eth0). Otherwise, it can cause issues because iptables will reject SSH requests on eth1

Adjustments have been made

> I'm not able to see if screenshots are available in GitLab because it looks like current Cypress test doesn't generate screenshots, could you tell me how I can enable this ?

This is automatic. By default at least 1 screenshot should be created w/ a failed test. This is where I mentioned I had some trouble extracting the them from the VM.

> There is a firefox error when we run tests on Debian (I provided logs in DM)

I added a `-headless` Cypress command-line flag for firefox. My reasoning is based on `libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)` in venom.log

If the test still fails, we may need to revert back to using the default [browser](https://docs.cypress.io/guides/guides/launching-browsers).

```
diff --git a/t/venom/test_suites/pfappserver_configurator/run_e2e.yml b/t/venom/test_suites/pfappserver_configurator/run_e2e.yml
index 7d0c2ece7a..f8c498b75c 100644
--- a/t/venom/test_suites/pfappserver_configurator/run_e2e.yml
+++ b/t/venom/test_suites/pfappserver_configurator/run_e2e.yml
@@ -5,7 +5,6 @@ testcases:
   steps:
   - type: html_e2e
     base_url: "{{.html.pfappserver_baseurl}}"
-    browser: firefox
     config_file: "cypress/config/cypress.config-configurator.js"
     tag: "pfappserver_configurator"
     info:
```

nqb:
Running a new pipeline to see if it's better: https://gitlab.com/inverse-inc/packetfence/-/pipelines/893321590

PR #7353, opened 2022-11-08
Sysprep on ZEN installation

nqb:
# Description
Delete/Recreate files which should be specific for each installation.

# Impacts
ZEN build

# Delete branch after merge
YES

# NEWS file entries
## Bug fixes
* Regenerate secrets at first boot of ZEN

### DIFF
--- a/ci/packer/zen/provisioners/inventory/group_vars/pfservers/rc_local.yml
+++ b/ci/packer/zen/provisioners/inventory/group_vars/pfservers/rc_local.yml
@@ -2,6 +2,18 @@
 rc_local_service_name: rc-local
 rc_local_commands:
   - |
+    if [ ! -f "/etc/ssh/ssh_host_rsa_key.pub" ]; then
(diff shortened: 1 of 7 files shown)


ci/packer/zen/provisioners/inventory/group_vars/pfservers/rc_local.yml
```diff
@@ -2,6 +2,10 @@
 rc_local_service_name: rc-local
 rc_local_commands:
   - |
+    make -C /usr/local/pf conf/ssl/server.pem
```
julsemaan, line 12: I was all proud of my `cd` commands and my 12 lines script and then @nqb comes in and does it in 3 lines :P

nqb:
One important thing to have in mind: first boot of ZEN can be long because `rc.local` will need to generate DH. On my personal laptop, it takes around 3 minutes to generate it. It should be less on dedicated servers.

nqb:
Ready for review and/or merge.

PR #7406, opened 2022-12-05
Fix/dyn reverse reconnect

julsemaan:
# Description
Clears the active dynamic reverses that exist when a pfconnector reconnects (these reverses are tied to a tunnel that is dead)

# Impacts
pfconnector

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Clear the active dynamic reverses that exist when a pfconnector reconnects


julsemaan:
I suggest we get this into maintenance/12.1 along with https://github.com/inverse-inc/packetfence/compare/fix/7323

Both fixes should cover all the known issues with the pfconnector

fdurand (approved):

PR #7421, opened 2022-12-08
Fix/venom stability

JeGoi:
# Description
Improve Venom tests stability

# Backport
Can be used on all v12.x branches

# Impacts
Venom is more stable.

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] wait for end of pipeline


t/venom/lib/pf_api_service_pf_start_async.yml
```diff
@@ -0,0 +1,19 @@
...
+  method: POST
+  time_to_sleep: "0"
+  status_code: 202
+  body: >-
+    {
+      "async": "true"
+    }
+
+- type: pf_api_service_status
+  service: "radiusd-auth"
```
nqb, line 17: not sure to understand why `radiusd-auth` is here.

t/venom/lib/pf_api_service_restart_async.yml
```diff
@@ -19,6 +19,8 @@ steps:
   vars:
     task_id:
       from: result.pf_api_service_action_json.task_id
+    task_token:
+      from: result.pf_api_service__token
 
 - type: pf_api_poll
```
nqb: We don't use `pf_api_poll_with_token` here while it's the case for `t/venom/lib/pf_api_service_start_async.yml` which seems pretty similar

t/venom/scenarios/configurator/playbooks/rsync.yml
```diff
@@ -16,3 +16,21 @@
...
+      ansible.builtin.lineinfile:
+        path: /root/.profile
+        regexp: '^export TERM='
+        line: export TERM=linux
+      when: ansible_facts['os_family'] == "Debian"
+
+    - name: Add vim
+      ansible.builtin.package:
+        name: vim
+        state: present
```
nqb: Could you move this in a dedicated playbook ? It's not related to rsync stuff.

t/venom/lib/pf_api_service_pf_start_async.yml
nqb: I think I understand, perhaps add a comment here to clarify this behavior.

nqb (changes requested):
- Could we remove `pf_api_poll.yml` ? If all `pf_api_poll` calls have been replaced by `pf_api_poll_with_token`, I suggest we remove it.
- I don't think we fixed issue related to expired tokens

t/venom/lib/pf_api_service_pf_start_async.yml
JeGoi: One of the latest services to restart.
Comment in progress

t/venom/scenarios/configurator/playbooks/rsync.yml
JeGoi: Yes! I used it there because it is only started when not the CI.
Could I rename rsync to development?

nqb: > Yes! I used it there because it is only started when not the CI. Could I rename rsync to development?

Good idea !

t/venom/lib/pf_api_service_restart_async.yml
JeGoi: Hum, I think I will remove all reference to token. As @julsemaan said, the token is not important. It just needs to be valid.

PR #7441, opened 2023-01-05
Allow Intune Compliance Check to handle InGracePeriod

PureOxygen-Dev:
# Description
(REQUIRED)
The current intune compliance checker does not handle the InGracePeriod returned by the Graph API.

This request added a check on the condition and if there returns as compliant.  

In addition, it gives the user the option in the provisioner setup form to select a targeted security event

# Impacts
(REQUIRED)
Describe what to reviewer should look at. Will also help for testing purposes.
Provisioner setup form and Intune Compliance Check

### DIFF
--- a/lib/pf/provisioner/intune.pm
+++ b/lib/pf/provisioner/intune.pm
@@ -285,7 +285,6 @@ sub verify_compliance {
     my $azuremac = uc($mac);
     $azuremac =~ s/://g;
 
-
     if($info != $pf::provisioner::COMMUNICATION_FAILED){
         my $not_compliant = $FALSE;
         foreach my $entry (@{$info->{value}}) {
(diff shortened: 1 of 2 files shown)


lib/pf/provisioner/intune.pm
```diff
@@ -294,7 +293,12 @@ sub verify_compliance {
                 if ($entry->{complianceState} eq 'compliant') {
                     $logger->info("Device $mac is compliant.");
                     return $TRUE;
-                } else {
+                } 
+                elsif ($entry->{complianceState} eq 'inGracePeriod') {
+					$logger->info("Device $mac is InGracePeriod.");
```
fdurand: incorrect indentation

PureOxygen-Dev: Hi Fabrice, sorry I missed in the developer guide about tabs.  Converted to spaces.

PureOxygen-Dev:
@fdurand Hang on this one a little more please.  I think the whole intune compliance part is not working as expected and I am dropping a message to @julsemaan on the subject.

PureOxygen-Dev:
Based on what this feature does I would only accept the compliant value back to trigger the captive portal form.  InGracePeriod could be misleading and result in the incorrect end result.

mellalahmed:

PR #7455, opened 2023-01-13
Fix/venom clear cache

JeGoi:
# Description
Clear the mem cache after a test in ci 

# Impacts
it clears the cache at the end

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Test pipeline

## Bug Fixes
Let's hope that it will help runners

### DIFF
--- a/t/venom/Makefile
+++ b/t/venom/Makefile
@@ -48,7 +48,9 @@ run_tests:
 	SCENARIOS_TO_RUN="$(SCENARIOS_TO_RUN)" \
 	./test-wrapper.sh run_tests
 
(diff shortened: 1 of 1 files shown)


nqb (changes requested):
The logic need to be adjusted a bit because we only call `clean` when a test passed (see `ci/lib/test/clean-test-environment.sh`) 

If a test failed, I think we want to run `clean_cache` target too.

So `halt` target need to call `clean_cache` target too.

PR #7466, opened 2023-01-19
feature/cisco_dacl

jrouzierinverse:
# Description
Add support for Dynamic ACLs and improved API warnings and errors when updating ACLs for roles and switches.
New api call /api/v1/config/role/{role_id}/bulk_reevaluate_access

# NEW Package(s) required
perl(Cisco::AccessList::Parser)

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [x] Add unit tests

# NEWS file entries
## New Features
* New api call /api/v1/config/role/{role_id}/bulk_reevaluate_access

## Enhancements
* Add warnings/errors when updating ACLs for roles and switches.


docs/installation/enforcement/supported_enforcement_modes.asciidoc
```diff
@@ -103,6 +103,18 @@ Web authentication is a method on the switch that forwards HTTP traffic of the d
...
+===== Downloadable ACLs
+
+Downloadable ACLs is a method that can be used when the ACL list is greater than the size of a RADIUS access-accept paquet.
+Some vendor support it, like Cisco Switches (IOS 15.2) and Dell (n1500 fw 6.8)
+
+The RADIUS flow is something close to the normal one but in the Access-Accept reply there is an extra RADIUS attribute that tell the equipment to trigger another RADIUS request to retreive the ACL.
+
+A second RADIUS request is made with the ACL name as a value of the username and multiples Access-Challenge are made in order to retreive the complete ACL.
+
+To enable it you need first to enable the RADIUS filter in the PacketFence authorize section. To do that go in Configuration -> System Configuration -> RADIUS -> General and enable "Use RADIUS filters in packetfence authorize" then restart RADIUS Auth service.
```
julsemaan:
```suggestion
To enable it you need first to enable the RADIUS filter in the PacketFence authorize section. To do that go in Configuration -> System Configuration -> RADIUS -> General and enable "Use RADIUS filters in packetfence authorize" then restart the `radiusd-auth` service.
```

docs/network/networkdevice/cisco.asciidoc
```diff
@@ -572,7 +572,18 @@ Or also you could restrict your users to use only your DNS server where 192.168.
 image::cisco-downloadable-acl-force-dns.png[scaledwidth="100%",alt="Force DNS ACL"]
 
 
-===== Web auth and Downloadable ACLs
+===== Downloadable ACLs
+
+Coming in IOS 15.2, Cisco switches supports Downloadable ACLs.
```
julsemaan:
```suggestion
Starting from IOS 15.2, Cisco switches supports Downloadable ACLs.
```

docs/network/networkdevice/cisco.asciidoc
```diff
@@ -572,7 +572,18 @@ Or also you could restrict your users to use only your DNS server where 192.168.
...
+The size of the radius packet limit the number of ACLs a switch can receive from a single Access-Accept answer, so Cisco Switches supports Downloadable ACLs which mean that the RADIUS server will do multiples Access-Challenge to send the complete ACL.
+
+Use the Cisco::Catalyst_2960 switch module to use the DACLs method and use the same Global settings from the section above.
```
julsemaan:
```suggestion
Use the Cisco::Catalyst_2960 switch module to use the DACLs method and use the same Global settings as the 'Dynamic ACLs' section above.
```

docs/network/networkdevice/dell.asciidoc
```diff
@@ -223,6 +225,32 @@ Web-Auth:
   1010 permit tcp any any eq http
   1020 permit tcp any any eq 443
 
+==== N1500 Series (FW >= 6.8.0)
+
+Downloadable ACLs:
+
+This configuration has been tested on FW 6.8.1.
+Important, even if "authentication allow-srcipanyacl enable" has been enable on the switch, it doesn't support ACL with source ip and the ACL direction are only in.
+So per example you have this configured in PacketFence:
```
julsemaan:
```suggestion
So, for example, if you have this configured in PacketFence:
```

html/pfappserver/lib/pfappserver/Form/Config/Roles.pm
```diff
@@ -104,6 +108,19 @@ has_field 'inherit_web_auth_url' => (
     default => 'disabled',
 );
 
+sub _validate_acl {
```
julsemaan: This function should be in a util module somewhere in `lib/` so it can be reused elsewhere

html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
```diff
@@ -453,15 +455,41 @@ has_field coaPort =>
...
+
+has_field DownloadableACLsLimit => (
+    type => 'PosInteger',
+);
+
+has_field ACLsLimit => (
+    type => 'PosInteger',
+);
+
+sub _validate_acl {
```
julsemaan: Same comment about reusing the code

html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
```diff
@@ -453,15 +455,41 @@ has_field coaPort =>
     },
   );
 
+has_field UseDownloadableACLs => (
+    type => 'Toggle',
```
julsemaan, line 460: Don't we need to set the enabled/disabled value here?

html/pfappserver/root/src/views/Configuration/switchGroups/_components/TheForm.vue
```diff
@@ -154,6 +154,22 @@
...
+              <b-card v-show="supports(['DownloadableListBasedEnforcement'])"
+                class="mb-3 pb-0" no-body
+              >
+                <form-group-use-downloadable-acls namespace="UseDownloadableACLs"
+                  :column-label="$i18n.t('Use downloadable ACLs instead of Dynamic ACLs')"
+                  :text="$i18n.t('This option parameter will allow you to do enable the Downloadable ACLs radius feature instead of using the Dynamic ACLs.')"
+                />
+                <form-group-downloadable-acls-limit namespace="DownloadableACLsLimit"
+                  :column-label="$i18n.t('Maximum number of ACLs PacketFence can return')"
+                  :text="$i18n.t('This option parameter will allow you to do define the maximum number of ACLs PacketFence can send to the switch.')"
```
julsemaan, line 166:
```suggestion
                  :text="$i18n.t('This option parameter will allow you to do define the maximum number of downloadable ACLs PacketFence can send to the switch.')"
```

html/pfappserver/root/src/views/Configuration/switches/_components/TheForm.vue
```diff
@@ -154,19 +154,36 @@
...
+                <form-group-use-downloadable-acls v-show="supports(['DownloadableListBasedEnforcement'])"
+                  namespace="UseDownloadableACLs"
+                  :column-label="$i18n.t('Downloadable ACLs')"
+                  :text="$i18n.t('Enable the Downloadable ACLs radius feature instead of using the Dynamic ACLs.')"
+                />
+
+                <form-group-downloadable-acls-limit v-show="supports(['DownloadableListBasedEnforcement'])"
+                  namespace="DownloadableACLsLimit"
+                  :column-label="$i18n.t('Maximum ACLs per switch')"
+                  :text="$i18n.t('The maximum number of ACLs PacketFence can send to the switch.')"
```
julsemaan, line 171:
```suggestion
                  :text="$i18n.t('This option parameter will allow you to do define the maximum number of downloadable ACLs PacketFence can send to the switch.')"
```

lib/pf/dal/_node.pm
```diff
@@ -329,6 +332,12 @@ BEGIN {
             is_primary_key => 0,
             is_nullable => 0,
         },
+        bypass_acls => {
```
julsemaan, line 335: It doesn't look like there is any validation of the syntax of the ACL that is being put in this field like there is for roles/switches ACLs

lib/pf/Switch/Cisco/Catalyst_2960.pm
```diff
@@ -498,21 +499,35 @@ sub returnRadiusAccessAccept {
...
+                    $access_list = $self->acl_chewer($access_list);
+                    my @acl = split("\n", $access_list);
+                    $args->{'acl'} = \@acl;
+                    $args->{'acl_num'} = '101';
+                    push(@av_pairs, "subscriber:service-name=$mac-".$self->setRadiusSession($args));
+                } else {
+                    my $acl_num = 101;
+                    while($access_list =~ /([^\n]+)\n?/g){
+                       my $acl = $1;
+                       if ($acl !~ /^permit/i && $acl !~ /^deny/i) {
```
julsemaan, line 514: I think we should have an else debug statement to say we ignored the line

lib/pf/Switch/Cisco/Catalyst_2960.pm
```diff
@@ -542,6 +557,104 @@ sub returnRadiusAccessAccept {
     return [$status, %$radius_reply_ref];
 }
 
+=head2
+
+Return Radius reply in the access-challenge
+
+=cut
+
+sub returnRadiusAdvanced {
```
julsemaan, line 566: If this is the same code than in the ASA module, I think it would make sense to reuse it instead of copy/pasting it

lib/pf/Switch/Dell/N1500.pm
```diff
@@ -468,6 +483,64 @@ sub parseExternalPortalRequest {
     return \%params;
 }
 
+=head2
+
+Return Radius reply in the access-challenge
+
+=cut
+
+sub returnRadiusAdvanced {
```
julsemaan, line 492: Same comment about reusing the code if possible

lib/pf/UnifiedApi/Controller/Config/Roles.pm
```diff
@@ -318,7 +323,64 @@ sub check_reassign_args {
...
+        push @items, \%item;
+    }
+
+    return \@items;
+}
+
+sub get_nodes_for_role {
+    my ($name) = @_;
+    my ($status, $sth) = pf::dal::node->db_execute($NODES_IN_CATGEORY, $name);
+    if (is_error($status)) {
```
julsemaan, line 390: Looks like it would make sense for the request to error out if this is true. From what I can see atm it will succeed but won't do anything

t/unittest/UnifiedApi/Controller/Config/Roles.t
```diff
@@ -37,6 +38,23 @@ my $collection_base_url = '/api/v1/config/roles';
...
+{
+    my $id = "test_role_${$}_1";
+    my $acl = <<ACL;
+permit ip 172.16.1.0 0.0.0.255 host 192.168.3.154
+ACL
+    $t->post_ok($collection_base_url => json => { id => $id, acls => $acl })
+      ->status_is(201);
+}
+
+{
```
julsemaan, line 50: I think it will be important to test the various cases where there should be warnings (ACL too large, equipment doesn't support it, etc)

html/pfappserver/lib/pfappserver/Form/Config/Roles.pm
```diff
@@ -136,6 +153,23 @@ sub validate {
 
     }
 
+    my $acls = $self->field('acls')->value;
```
julsemaan, line 144: From what I can tell this will not validate that an ACL has been added to a role but some configured switches do not support dACLs (or downloadable ACL if necessary)

Looking at fullfuling this requirement from our documented list:
> Make sure the size of the ACL is compatible with all the configured equipment
> ---- If some equipment is incompatible with the ACL, return it in the warnings

julsemaan (changes requested):

html/pfappserver/lib/pfappserver/Form/Config/Roles.pm
jrouzierinverse: The check is done on line 167

t/unittest/UnifiedApi/Controller/Config/Roles.t
jrouzierinverse: Ok, I would need to add the equipment does not support it.
I already have the ACL to large check.

html/pfappserver/lib/pfappserver/Form/Config/Roles.pm
julsemaan: According to what I see in Switch.pm, if the switch doesn't support dynamic or downloadable ACLs, no warnings are returned.

That means that if you configure an ACL in a role and some equipment doesn't support dACLs at all, then there will not be any warning

nqb:
@fdurand, do we need to add new Perl libs to packetfence-perl ?

jrouzierinverse:
The package is not on CPAN.
It is only on github

fdurand:
> @fdurand, do we need to add new Perl libs to packetfence-perl ?

no

lib/pf/Switch/Cisco/Catalyst_2960.pm
fdurand: It's not the same code as the ASA

lib/pf/Switch/Dell/N1500.pm
fdurand: It's not the same code as the ASA

fdurand:
BUMP

PR #7476, opened 2023-01-26
Template to track problems related to tests

nqb:
# Description
Create a template to manage problems related to tests in CI.

### DIFF
--- a/.github/ISSUE_TEMPLATE/test_problem.md
+++ b/.github/ISSUE_TEMPLATE/test_problem.md
@@ -1,7 +1,7 @@
 ---
 name: Test problem
 about: Report a problem with tests performed by CI (for internal use)
-title: ''
+title: 'tests: '
 labels: 'tests'
 assignees: ''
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/test_problem.md
```diff
@@ -0,0 +1,21 @@
...
+
+**Describe the problem**
+
+
+**Mandatory informations**
+
+ * Branch:
+ * Scenario:
+ * Pipeline ID: 
+
```
JeGoi, line 19:
```suggestion
 * Running Date:

```
We may not be able to do it the same day.

JeGoi (commented):
Just added a comment
We may make it evolve after a couple of days

.github/ISSUE_TEMPLATE/test_problem.md
nqb: I think it's not necessary because you can have it with pipeline ID.

PR #7482, opened 2023-01-27
Azure SAML documentation

fdurand:
# Description
Documentation on how to integrate Azure SAML

# Impacts
Documentation

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Azure SAML integration documentation


julsemaan:
One limit of this is that we must use an authorization source to be able to assign admin roles based on various rules.

One use-case this needs to tackle is to be able to allow admin login based on group membership. Can we use Azure AD to authorize the user?

The initial idea was to use OpenID + the attributes it pushes to be able to perform group membership rules. I guess as long as we can use Azure SAML+AD to accomplish this, that can also work fine for our customers

fdurand:
It not mandatory to use a authorization source, i was just lazzy. Btw SAML return the userPrincipalName so i think i can use Azure AD as an authentication source. Let me try it.

fdurand:
So it's possible to use the UserPrincipalName from the SAML response and use it to test the groupmembership from the Azure AD source. Do you want me to adapt the documentation ?

julsemaan:
> So it's possible to use the UserPrincipalName from the SAML response and use it to test the groupmembership from the Azure AD source. Do you want me to adapt the documentation ?

I think it would be worth it to give a more complete example since it seems logic to pair SAML + Azure AD as the autz source when using Azure SAML

extrafu:
I'll also have a look.

PR #7491, opened 2023-02-01
Use the ENV variable for the logging level

### LINKED ISSUE #7487: Change log levels of Perl services using environment variable
If you want to change log level of Perl services, you need to edit files under `conf/log.conf*` and replace hard-coded value `INFO` by another value on `log4perl.rootLogger` line.

In Cloud NAC environment, we can't do that easily.

Change log level of Perl services using an environment variable. Fallback to `INFO` is variable is not defined.

As discussed with @jrouzierinverse, perhaps we can use same recipe as in `pf.conf`  (template and `ENV.env_or_default`).

jrouzierinverse:
# Description
Change log levels of Perl services using environment variable

# Issue
fixes #7487

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Change log levels of Perl services using environment variable (#7487)

### DIFF
--- a/conf/log.conf.d/httpd.aaa.conf.example
+++ b/conf/log.conf.d/httpd.aaa.conf.example
@@ -1,6 +1,6 @@
 # Copyright (C) Inverse inc.
 ### httpd.aaa logger ###
-log4perl.rootLogger = INFO, HTTPD_AAA
+log4perl.rootLogger = sub { ($ENV{PF_LOG_LEVEL} || 'INFO' ) . ", HTTPD_AAA" }
 
 ### Categories ###
 # Below, you can specify different categories (based on package names) for different logging levels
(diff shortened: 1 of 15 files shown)


nqb (commented):
1. Is it normal that `conf/log.conf.d/pfconfig.conf.example` is not handle ?
2. I wonder if we should not try to use another variable name like `PF_PERL_LOG_LEVEL` in order to have something different than Golang services which already used `LOG_LEVEL`

Otherwise, code looks good.

nqb:
Regarding 2, `PF_SERVICES_PERL_LOG_LEVEL` is more consistent regarding current environment variables we already used.

jrouzierinverse:
Let us use a single variable like 'LOG_LEVEL' instead to reduce the likelihood of a mistake.

nqb:
> Let us use a single variable like 'LOG_LEVEL' instead to reduce the likelihood of a mistake.

Let's try like this.
Could you update the files to reflect that change ?

PR #7506, opened 2023-02-07
Use Nodejs 14 to build web admin

nqb:
# Description
Use Nodejs 14 to build web admin

# Impacts
web admin

# Delete branch after merge
YES


satkunas:
I've removed scrolling the iframe to leave fragment hrefs untouched instead of easing the iframe vertical scroll. 

Using the native behaviour reveals a chromium bug where clicking in an iframe causes the parent document to scoll

https://bugs.chromium.org/p/chromium/issues/detail?id=611982

satkunas:
PR split and nodejs changes moved to a new branch in order to avoid confusion

PR #7507, opened 2023-02-07
Improve pfconnector dynreverse error handling

### LINKED ISSUE #7498: pfconnector: avoid race-condition on ports used
**Describe the bug**
In pfconnector, a race-condition can occur when a port has been selected but it not used yet. The port could have been pick up by another session. We need to prevent this in the code.

Symptoms:

```
Feb 06 02:19:21 pf02 pfconnector-server-docker-wrapper[2980768]: t=2023-02-06T02:19:21+0430 lvl=eror msg="Error binding remote R:38723=>172.30.152.22:161/udp: server: session#1: tun: proxy#R:38723=>172.30.152.22:161/udp: listen: listen udp 0.0.0.0:38723: bind: address already in use" pid=1
```

**To Reproduce**
1. Generate a lot of traffic which involve simultaneous usage of pfconnector

**Expected behavior**
Each session must used a port available and not selected by another session.

julsemaan:
# Description
Properly catch dynreverse bind errors and retry on error

# Impacts
pfconnector dynreverse binds

# Issue
fixes #7498 

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Improved error handling of dynamic port reverses for pfconnector

### DIFF
--- a/go/chisel/server/server_handler.go
+++ b/go/chisel/server/server_handler.go
@@ -31,6 +31,11 @@ import (
 var activeTunnels = sync.Map{}
 var apiPrefix = "/api/v1/pfconnector"
 
+const (
+	DYNREVERSE_BIND_ATTEMPTS = 10
+	DYNREVERSE_ERR_WAIT      = 50 * time.Millisecond
+)
+
 // handleClientHandler is the main http websocket handler for the chisel server
 func (s *Server) handleClientHandler(w http.ResponseWriter, r *http.Request) {
 	log.LoggerWContext(r.Context()).Info(fmt.Sprintf("Handling %s %s", r.Method, r.URL.Path))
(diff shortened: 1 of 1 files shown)


go/chisel/server/server_handler.go
```diff
@@ -254,34 +259,72 @@ func (s *Server) handleDynReverse(w http.ResponseWriter, req *http.Request) {
...
-			w.WriteHeader(http.StatusBadRequest)
-			json.NewEncoder(w).Encode(unifiedapiclient.ErrorReply{Status: http.StatusBadRequest, Message: fmt.Sprintf("The format for the remote (%s) is invalid: %s", to, err)})
-			return
-		}
-
-		remote.Dynamic = true
-		remote.LastTouched = time.Now()
-		settings.ActiveDynReverse.Store(cacheKey, remote)
-		go func() {
-			ctx := context.Background()
```
jrouzierinverse, line 277: We should simplify this logic by using context.WithTimeout

```
ctx, cancel := context.WithTimeout(context.Background(), DYNREVERSE_ERR_WAIT)
defer cancel();
```

jrouzierinverse (changes requested):

go/chisel/server/server_handler.go
julsemaan: It looks like this would close the tunnel after the DYNREVERSE_ERR_WAIT

TBH, I used what I felt the most comfortable with to make sure it's as rock solid as I can given I likely won't be there when this goes onto customer deployments.

I'm happy for you to rework this recipe to something that you feel will be easier to maintain by the team but I'm not sure I'll have the bandwidth to rework this and retest it as thoroughly as I did before submitting the PR.

jrouzierinverse: Another approach would have been to set the LocalPort to 0 and let the OS pick the first free port available.
This would simplify the code by having zero need to do a retry.
However, this would require some additional changes as this would require us to return back the port picked from the OS.
I will mark that as future todo.

PR #7551, opened 2023-02-23
Fix/iptables in venom

### LINKED ISSUE #7489: tests: SSH access is lost on VM after configurator test suite
**Describe the problem**
In a random way, Ansible is not able to reach PF VM after configurator test suite has been passed.

It happened on different scenario and OS.

**Mandatory informations**

 * Branch: `maintenance/12.0`
 * Scenario: `inline_deb11`
 * Pipelines URL:
   * https://gitlab.com/inverse-inc/packetfence/-/jobs/3693333159#L607
 
**Symptoms**
- Configurator test suite passed
- Error message:
```yaml
Wednesday 01 February 2023  07:05:31 -0500 (0:00:00.850)       0:01:04.363 **** 
changed: [pfdeb11dev] => (item=configurator_inline)
fatal: [pfdeb11dev]: FAILED! => 
  msg: 'Failed to connect to the host via ssh: '
```

**Additional context**
@JeGoi has performed some investigation and it looks issue is caused by restart of `packetfence-iptables` (through API) by `global_config` test suite.

JeGoi:
# Description
(REQUIRED)
This will use perl to reniew iptables configuration.
It looks more stable than systemctl restart or the api (async or not)

# Issues
fixes #7489 

# Impacts
Should improve the pipelines
https://gitlab.com/inverse-inc/packetfence/-/pipelines/786217518

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [x] Pipeline tested

## Bug Fixes
* https://github.com/inverse-inc/packetfence/issues/7489

### DIFF
--- a/t/venom/lib/perl_iptables_restore.yml
+++ b/t/venom/lib/perl_iptables_restore.yml
@@ -1,0 +1,6 @@
+executor: perl_iptables_restore
+input:
+  backup_file: "/root/iptables.bak"
+steps:
+- type: exec
+  script: iptables-restore < "{{ input.backup_file }}"
--- a/t/venom/lib/perl_iptables_restart.yml
+++ b/t/venom/lib/perl_iptables_restart.yml
@@ -1,0 +1,19 @@
+executor: perl_iptables_restart 
+input:
+  backup_file: "/root/iptables.bak"
+steps:
+- type: pf_api_service_status
+  service: "iptables"
+  retry: "{{.input.i_retry}}"
+  delay: "{{.input.i_delay}}"
+  assertions:
+    - result.pf_api_service_status_json.alive ShouldEqual 1
+
+- type: exec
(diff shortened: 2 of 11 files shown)


t/venom/lib/perl_iptables_restore.yml
```diff
@@ -0,0 +1,6 @@
+executor: perl_iptables_restore
```
nqb, line 1: I don't think this executor is used.

nqb (approved):

t/venom/lib/perl_iptables_restore.yml
JeGoi: Nope, you are right. I create it for teardown but not used it yet.

nqb: Ok, let's keep it for now.

PR #7564, opened 2023-03-02
Disable DNS lookup made by sudo

### LINKED ISSUE #7335: Restarting iptables causes it to drop all INPUT+FORWARD
**Describe the bug**
If the hostname of the server is not resolvable via /etc/hosts, then restarting iptables causes it to make it drop all of INPUT and FORWARD. Adding the hostname on the 127.0.0.1 line fixes that.

**To Reproduce**
1. Install PacketFence
2. Change the hostname during the configurator
3. Restart iptables

**Expected behavior**
iptables shouldn't misbehave because the host entry is missing

**Additional context**
This only applies when you change the hostname after installing the OS

### LINKED ISSUE #7403: Zen 12.1 - ESXi - No interfaces found in Step 1
**Describe the bug**
After importing the OVA into vSphere and Launching the Machine, when trying to run the initial setup no interfaces can be found.
This stops the installation, as it cannot be progressed without a management interface.
After a short while, a notification is given stating a 504 error.

When directly going API, all other call's function, providing the initial hostname, gateway and DNS.
However the same error is seen when calling:
`https://xxx.xxx.xxx.xxx:1443/api/v1/configurator/config/interfaces?sort=id&fields=is_running,id,ipaddress,netmask,ipv6_address,ipv6_prefix,network,type,additional_listening_daemons,high_availability,buttons&limit=1000`

Restarting or Changing Network adapter types does not make a difference.
When trying Zen 12.0 - same issue is seen

**To Reproduce**
Import Packetfence Zen 12.1 on ESXi
Start the appliance and go to https://ipaddress:1443

**Screenshots**
image
image
image

**Expected behaviour**
Interfaces are populated in the configurator and can be selected to progress installation of the appliance

### LINKED ISSUE #7552: tests: domain join is failing randomly
**Describe the problem**
Domain join is failing randomly

**Mandatory informations**

 * Branch: `devel`
 * Scenario: `dot1x_eap_peap` el8/deb11
 * Pipelines URL:
   * https://gitlab.com/inverse-inc/packetfence/-/jobs/3827112670
   * https://gitlab.com/inverse-inc/packetfence/-/jobs/3827112662
 
**Symptoms**

- Nothing under /chroots
- `domain Timed out"` in /usr/local/pf/logs/api-frontend.log
- There is nothing related to domain join in pfperl-api logs

This problem seems to be directly related to #7403. When I checked pfperl-api logs, I see ` sudo: unable to resolve host pfel8dev: Name or service not known`

=> It makes sense because we change hostname of VM during configurator test suite but VM is not rebooted after.

nqb:
# Description
Disable DNS lookup made by sudo.

On Debian, if there is no entry in `/etc/hosts` which match current hostname of server, sudo will display error like:

`
sudo: unable to resolve host pfdeb11dev: Name or service not known
`

and can generate API timeouts because it is not able to resolve hostname of server.

Because `/etc/hosts` is not shared between containers and host, containers need to have this entry too in their own `/etc/hosts` file.

Currently, `/etc/sudoers` is shared with all containers so I take the decision to disable that feature in `sudo` according to this [bug report](https://bugzilla.sudo.ws/show_bug.cgi?id=916). According to my tests, it was not necessary on EL systems but I prefer to enforce that parameter on both distributions.

I create a dedicated file in order to simplify backport and upgrades.

# Impacts
- sudo configuration on host and containers
- upgrade

# Issue
fixes #7403 
fixes #7335
fixes #7552

# Delete branch after merge
YES

# NEWS file entries
## Bug Fixes
* Disable DNS lookup in sudo to prevent API timeouts

### DIFF
--- a/debian/packetfence.conffiles
+++ b/debian/packetfence.conffiles
@@ -1,4 +1,5 @@
 /etc/sudoers.d/packetfence
+/etc/sudoers.d/disable-dns-lookup
 /usr/local/pf/conf/adminroles.conf
 /usr/local/pf/conf/authentication.conf
 /usr/local/pf/conf/billing_tiers.conf
--- a/debian/rules
+++ b/debian/rules
@@ -147,6 +147,7 @@ install: build
 	#Sudoer
 	install -oroot -groot -d -m0750 $(CURDIR)/debian/packetfence/etc/sudoers.d
 	install -oroot -groot -m0440 $(CURDIR)/debian/packetfence.sudoers $(CURDIR)/debian/packetfence/etc/sudoers.d/packetfence
+	install -oroot -groot -m0440 $(CURDIR)/disable-dns-lookup.sudoers $(CURDIR)/debian/packetfence/etc/sudoers.d/disable-dns-lookup
 	#PacketFence Mariadb plugin
 	make src/mariadb_udf/pf_udf.so
 	install -D -m0755 src/mariadb_udf/pf_udf.so $(CURDIR)/debian/packetfence/$$(pkg-config mariadb --variable=plugindir)/pf_udf.so
--- a/disable-dns-lookup.sudoers
+++ b/disable-dns-lookup.sudoers
@@ -1,0 +1,1 @@
+Defaults        !fqdn
--- a/disable-dns-lookup.sudoers
+++ b/disable-dns-lookup.sudoers
@@ -1,1 +1,2 @@
 Defaults        !fqdn
+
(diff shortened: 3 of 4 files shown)


nqb:
Pipeline triggered with tests: https://gitlab.com/inverse-inc/packetfence/-/pipelines/794197527

nqb:
New pipeline after latest commit: https://gitlab.com/inverse-inc/packetfence/-/pipelines/795457113

Configurator passed.

JeGoi (approved):
Fine with me.
Found it in https://linux.die.net/man/5/sudoers

satkunas:
backport to patch releases

PR #7573, opened 2023-03-06
fix #7559

### LINKED ISSUE #7559: tests:  "systemctl_service_status" unable to unmarshal: error converting YAML to JSON
**Describe the problem**
When we test the systemctl service status, we check in the output if the service is started, but in some cases there is a ' or " in the output and it mess the json creation => failed.

**Mandatory informations**

 * Branch: Devel
 * Scenario: inline
 * Pipelines URL:
   * https://gitlab.com/inverse-inc/packetfence/-/jobs/3851176338
 
**Symptoms**

The main idea is to use the api to test the service status, or even the restart.
If systemctl needs to be used then two options reduce the number of lines from the output or not tests the status in the restart.

JeGoi:
# Description
This is a way to fix issue like what is happening in 7559.
We are using the api instead of the service, so no more json stuff.

# Impacts
Should fix inline scenario

# Issue
fixes #7559 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Pass the pipeline

### DIFF
--- a/t/venom/test_suites/security_event_suricata/teardown/05_deconfigure_packetfence.yml
+++ b/t/venom/test_suites/security_event_suricata/teardown/05_deconfigure_packetfence.yml
@@ -75,36 +75,30 @@ testcases:
 
 - name: restart_pfdns_service
   steps:
-  - type: systemctl_service_restart
-    service: packetfence-pfdns
-    time_to_sleep: 5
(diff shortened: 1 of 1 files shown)


nqb:
Pipeline had failures but not related to that fix.
Merging.

PR #7630, opened 2023-04-20
Samba4ad: New Fixes. 

abprasa:
# Description
Modified the old samba4 configuration to make it more dynamic and also fixed the frequent appearing resolution issues that we have come across during earlier testing.

# Impacts
Possible impact during backporting while replacing MS-AD with Samba4.

# Issue
1. Fixed the host's file to remove DNS resolution errors seen post-reload.
2. To have the config persistent after reloading updated the dhclient.conf file with the nameserver configured for the domain_name listed in the correct order.
3. Introduced new variables to make the code more dynamic and avoid duplicate reinstallation.

YES 

# Checklist
- [ ] Document the feature

# New Features
* Doing venom test samba4 AD.

### DIFF
--- a/addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
+++ b/addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
@@ -1,5 +1,5 @@
 ---
-samba4ad__mgmt_ip: "{{ packetfence_server_mgmt_ip }}"
+samba4ad__mgmt_ip: "{{ hostvars[inventory_hostname]['mgmt_ip'] }}"
 
 samba4ad__base_packages:
   - 'samba'
--- a/addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
+++ b/addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
@@ -1,4 +1,6 @@
 ---
+samba4ad__mgmt_ip: "{{ hostvars[inventory_hostname]['mgmt_ip'] }}"
+
 samba4ad__base_packages:
   - 'samba'
   - 'winbind'
--- a/addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
+++ b/addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
@@ -1,5 +1,5 @@
 ---
-samba4ad__mgmt_ip: "{{ hostvars[inventory_hostname]['mgmt_ip'] }}"
+samba4ad__mgmt_ip: "{{ packetfence_server_mgmt_ip }}"
 
 samba4ad__base_packages:
   - 'samba'
--- a/addons/vagrant/playbooks/linux_servers/samba4ad.yml
+++ b/addons/vagrant/playbooks/linux_servers/samba4ad.yml
@@ -2,12 +2,6 @@
   name: Manage Samba service
   become: True
 
-  collections:
-    - debops.debops
-    - debops.roles01
-    - debops.roles02
-    - debops.roles03
-
   tasks:
   - name: Check if samba4ad configuration is there
     ansible.builtin.lineinfile:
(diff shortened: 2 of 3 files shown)


addons/vagrant/inventory/group_vars/linux_servers/samba4ad.yml
```diff
@@ -1,4 +1,6 @@
 ---
+samba4ad__mgmt_ip: "{{ hostvars[inventory_hostname]['mgmt_ip'] }}"
```
nqb, line 2:
```suggestion
samba4ad__mgmt_ip: "{{ packetfence_server_mgmt_ip }}"
```

addons/vagrant/playbooks/linux_servers/samba4ad.yml
```diff
@@ -9,190 +9,217 @@
...
-      src: "/etc/krb5.conf"
-      dest: "/var/lib/samba/private/krb5.conf"
-      state: link
-
-  - name: Unmask and start Services
-    ansible.builtin.systemd:
-      enabled: yes
+    check_mode: yes
+    register: check_if_samba
+    ignore_errors: true
```
nqb, line 12: I don't think you need "ignore_errors"

addons/vagrant/playbooks/linux_servers/samba4ad.yml
```diff
@@ -9,190 +9,217 @@
...
+        cmd: "{{ item }}"
+      loop:
+      - "samba-tool user setexpiry {{ samba4ad__username }} --noexpiry"
+      - "samba-tool user setexpiry {{ packetfence_domain_account }} --noexpiry"
+      - "samba-tool group add sponsor"
+      - "samba-tool group add IT"
+      - "samba-tool group addmembers sponsor {{ packetfence_domain_account }}"
+      - "samba-tool group addmembers IT {{ packetfence_domain_account }}"
+
+    when: check_if_samba.failed is true
```
nqb: If `ansible.builtin.lineinfile` module returns "CHANGED" when it found a line in a file (even in check mode), you can certainly simplify this line by:

`when: check_if_samba is changed`

Which is, IMHO, more readable.

nqb (changes requested):
Using the `check_if_samba` variable, you can certainly skip more tasks on an existing installation.

At the moment, you only skip "Configure Samba AD Stuff" task but there is no need to re-run previous tasks (especially reboot).

What you can do is to move "installation" tasks into a dedicated playbook called "installation.yml" and use a condition based on `check_if_samba` variable to run that playbook ([example](https://github.com/inverse-inc/packetfence/blob/1aaf6ec7e5ee3e5f688e371262af8502111c64fb/t/venom/scenarios/template/site.yml#L6-L7))

addons/vagrant/playbooks/linux_servers/samba4ad.yml
JeGoi: Unfortunately, It does not work without it on a fresh install. I think it is due to the ansible version that we use.
It is joining the fact that our tasks order may need to be reviewed.
There are currently following the samba4ad wiki.

On a fresh installation with the current tasks order:
```
TASK [Check if samba4ad configuration is there] ********************************
Friday 21 April 2023  08:05:00 -0400 (0:00:00.827)       0:00:04.151 **********
fatal: [ad]: FAILED! => changed=false
  msg: Destination /etc/samba/smb.conf does not exist !
  rc: 257
...ignoring
```
Without the ignore, the playbook is failing.
The best should be to install samba first and them test if the file is here and modified, but I am not sure how to handle the multiple painful reboot

addons/vagrant/playbooks/linux_servers/samba4ad.yml
JeGoi: This is not working when there is no installation.
To use that we need to change tasks order.

JeGoi:
Nope, we skip almost everything due to the block line 21.
It is like, the file is here and contains the right config then just restart the service.
But you are right about the installation. It should be better but take more time.

PR #7633, opened 2023-04-24
Remove switch validation

### LINKED ISSUE #7508: Access list sending to Aruba 2930M done wrong
**Describe the bug**
Packetfence seems to send them all according to log, every line as its own NAS-Filter-Rule, but it seems like it actually sends them as one line, and character limit is reached:
W 02/07/23 14:21:08 00712 radius: ACL error - entry too long, client
3448EDxxxx port 41

I can send 2 lines, without above error presenting it self, but it anyways only takes one line in to use.

**To Reproduce**
Create accesslist and try to send multiple lines to Aruba 2930M. 
Check from Aruba that only the first line is in use, or not at all.

Fix according to Aruba
"If you are using multiple attributes, you need add "+" denoting multiple attributes as below -

Nas-filter-Rule = "deny in tcp from any to any 23 ",
Nas-filter-Rule += "deny in tcp from any to any 80 ",

------------------------------
xx
Aruba
------------------------------"

jrouzierinverse:
# Description
Don't Validate switch ACLs

# Issue
fixes #7508

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Bug Fixes
* Passthrough ACLs defined in it's config (#7508)

### DIFF
--- a/html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
+++ b/html/pfappserver/lib/pfappserver/Form/Config/Switch.pm
@@ -490,7 +490,7 @@ sub addRoleMapping {
 addRoleMapping("VlanMapping", "vlan");
 addRoleMapping("UrlMapping", "url");
 addRoleMapping("ControllerRoleMapping", "controller_role");
-addRoleMapping("AccessListMapping", "accesslist", [validate_method => \&_validate_acl ]);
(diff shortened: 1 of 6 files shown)


fdurand:
Code looks good, i will test it first.

PR #7644, opened 2023-05-02
Push acls with ansible

fdurand:
# Description
This allow PacketFence to provision the compatible switches with pre configured ACLs.
When the feature is enable in the switch config, PacketFence will retrieve the ACLs configured in the Role configuration, transform them to be compatible with the equipment and via ansible it will push the ACLs in the configuration.

If there is a mix between Dynamic/Downloadable and PushACLs then the priority will be to use PushACLs.
Once the ACLs are pushed in the switch configuration then PacketFence will be able to return the RADIUS attribute to apply this ACL.
 
# Impacts
- RADIUS authorize workflow
- Dynamic/Downloadable ACLs.

# Delete branch after merge
YES

# NEWS file entries
## New Features
* Pre-create ACLs on equipment (Cisco WLC, CIsco IOS, Aruba CX) and apply it via a RADIUS attribute.


satkunas:
Error being generated from Roles (`GET /api/v1/config/roles`) when viewing the switch page:

```
Jun 29 08:36:13 172 pfperl-api-docker-wrapper[226135]: pfperl-api(17) ERROR: [mac:[undef]] Can't call method "new" on an undefined value at /usr/local/pf/lib/pf/UnifiedApi/Controller/Config.pm line 238.
```

lib/pf/UnifiedApi/Controller/Config/Roles.pm
```diff
@@ -41,6 +41,23 @@ use pf::ConfigStore::Source;
...
+        my $switch = pf::SwitchFactory->instantiate($switch_id);
+        $switch->generateAnsibleConfiguration();
+        # Need to wait between each switch to avoid error on semaphore
+        sleep(1);
+    }
+}
+
+sub post_create {
+    my ($self, $id) = @_;
+    $self->port_update($id);
```
jrouzierinverse: This is a typo

jrouzierinverse (changes requested):

conf/pfsetacls/collections/requirements.yml
```diff
@@ -0,0 +1,5 @@
+collections:
```
JeGoi, line 1: I think you should add a version there for all collections.

containers/pfsetacls/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM golang:bullseye
```
JeGoi: Why didn't you use the pfbuild-debian-bullseye?
If I remember correctly pfbuild-debian-bullseye has the go installed on it.

containers/pfsetacls/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+ENV SEMAPHORE_VERSION="development" SEMAPHORE_ARCH="linux_amd64" \
+    SEMAPHORE_CONFIG_PATH="${SEMAPHORE_CONFIG_PATH:-/etc/semaphore}" \
+    APP_ROOT="/go/src/github.com/ansible-semaphore/semaphore/"
+
+# hadolint ignore=DL3013
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+
+RUN apt update
+RUN apt install -y gcc g++ make git mariadb-client-10.5 python3 pip python3-openssl openssl ca-certificates curl libcurl4-openssl-dev openssh-client tini nodejs bash rsync && \
```
JeGoi: Two lines could be merged
This dockerfile has a lot of layers. Perhaps we can reduce that.

JeGoi (commented):
Please, could you address my questions points?
Thanks

JeGoi:
More as a reminder change that too when it is ready:
https://github.com/inverse-inc/packetfence/blob/devel/addons/dev-helpers/bin/switch_options_table.pl
Due to `PushACLs` in Switch libs

lib/pf/Switch.pm
```diff
@@ -747,6 +747,8 @@ sub getAccessListByName {
         $fb_acl = $self->fingerbank_dynamic_acl($mac);
     }
 
+    print STDERR Dumper($acls);
```
satkunas, line 750: Debug that needs to be removed?

PR #7646, opened 2023-05-02
[Venom] Fix #7587

### LINKED ISSUE #7587: tests: "Gathering facts" task timed out
**Describe the problem**
For some VM (Debian 11 or EL8), Ansible is not able to complete "Gathering Facts" task in "Configure environment variables used for tests" play. The task didn't time out so the job is running for 170 minutes without doing anything.

When GitLab call after_script, we can see that the VM is still reachable because Ansible will gather facts (again).

At this stage, the VM is in pristine state because we only:
- Gather facts (one time)
- Start and enable serial-getty service

Consequently, jobs 

**Mandatory informations**
 * Branch: all maintenance branches
 * Scenario: all scenarios
 * Pipelines URL:
   * https://gitlab.com/inverse-inc/packetfence/-/jobs/3926221984
   *  https://gitlab.com/inverse-inc/packetfence/-/jobs/3926216596
 
**Symptoms**

image

JeGoi:
# Description
Set Gathering facts for each playbook. What we have done on all playbooks:
* move the gathering facts as a pre_tasks
* set the default gathering facts as the [min](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/setup.py#L202) with the !all
* add the missing facts, mostly network, that are used by debops collections when needed
* remove the gathering fact for apt update on linux_servers since all linux servers are the same.

# Impacts
Should have fix the gathering facts

# Issue
fixes #7587 

# Delete branch after merge
YES

# Checklist
- [ X ] pipeline tested on maintenance/12.0
https://gitlab.com/inverse-inc/packetfence/-/pipelines/854598479

### DIFF
--- a/addons/vagrant/playbooks/config_dev_repos.yml
+++ b/addons/vagrant/playbooks/config_dev_repos.yml
@@ -1,6 +1,8 @@
 - hosts: dev
   name: Install packetfence development packages
   become: True
+  gather_facts: False
+
   collections:
     - debops.debops
     - debops.roles01
@@ -8,6 +10,11 @@
     - debops.roles03
 
   pre_tasks:
+    - name: Gather min only
+      ansible.builtin.setup:
+        gather_subset:
+          - "!all"
+
     # PPA repositories are removed each day and cause issues when refreshing cache
     - name: Clean previous ppa source file if present (Debian)
       file:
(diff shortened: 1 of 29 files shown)


nqb (commented):
Looks good. If it has not been done yet, I think we need to run all tests to be sure we have all facts necessary when doing:
- initial provisioning (VM doesn't exist)
- refresh provisioning (VM already exist)

JeGoi:
Ok tested on initial provisioning and refresh
Ready to merge

Issue #7538, opened 2023-02-16
pfqueue stuck even if there is job in the queue


fdurand:
**Describe the bug**
I noticed on multiples setup that pfqueue sometime is just doing nothing even if there is job in redis.

A restart the service pfqueue fix the issue.
- `systemctl restart packetfence-pfqueue`
- `/usr/local/pf/bin/pfcmd service pfqueue restart`

You can check the pf queue status on specific a list with the following command line:
`/usr/local/pf/bin/pfcmd pfqueue list`
than check the list status
`/usr/local/pf/bin/pfcmd pfqueue count pfdhcplistener`

If the count never decrease, it means that you have the issue and you need to restart the service.

extrafu:
@jrouzierinverse and @fdurand is it fixed? If so please close.

PR #42, opened 2012-08-08
Feature/report sponsor

fdurand:
For review

PR #7647, opened 2023-05-02
fix/7538

jrouzierinverse:
# Description
Remove the use of pthread_atfork.

# Issue
Related to #7538

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Bug Fixes
* Remove the use of pthread_atfork (#42)

### DIFF
--- a/lib/pf/AtFork.pm
+++ b/lib/pf/AtFork.pm
@@ -1,44 +1,38 @@
-package pf::Syslog;
+package pf::AtFork;
 
 =head1 NAME
 
-pf::Syslog -
+pf::AtFork -
 
 =head1 DESCRIPTION
 
-pf::Syslog
+pf::AtFork
 
 =cut
 
 use strict;
 use warnings;
-use CHI;
-use Log::Any::Adapter;
-Log::Any::Adapter->set('Log4perl');
-use pf::AtFork;
-use Net::Syslog;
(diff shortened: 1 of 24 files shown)


nqb:
@jrouzierinverse, could you:

1. Explain in details what issue you are trying to solve (with POSIX::AtFork)
1. Give us more details about your choice here (move from POSIX::AtFork to our own implementation pf::AtFork which seems pretty simple)

The POSIX::AtFork seems to use [XSLoader](https://metacpan.org/module/XSLoader/source), I wonder what is the difference between two implementations. 

Last thing, I think this change is too big to be backported to maintenance.

nqb:
@jrouzierinverse, could you look at my comments ?

jrouzierinverse:
The reason is that [pthread_atfork](https://pubs.opengroup.org/onlinepubs/009604499/functions/pthread_atfork.html) seems to have an issue with deadlocking processes in some scenarios.
So to reduce this chance, I emulated this functionality with this fix.

nqb (approved):

nqb:
Changes look good. I'm just not sure which impact it has if it goes in maintenance.
I will let @jrouzierinverse take a decision on this.

PR #7677, opened 2023-05-24
Feature/switch module update: DRAFT

abprasa:
# Description
The existing code has a list of switch modules based on the Product model number.
This new code updates the existing list of Cisco switch Modules and maps them to the relevant versions and the features they support.

# Impacts

# NEW Package(s) required

# Feature
The idea of updating the existing list of Cisco switch Modules defined based on the model number to respective versions is to make it more convenient for users to choose modules based on the version of the switch.

# Delete branch after merge
YES 

# Checklist
- [ ] Document the feature

# NEWS file entries
## New Features
* Updating existing switch model for v13


extrafu:
@abprasa Great work! Correct me if I'm wrong but it looks like you added files and then renamed them. Perhaps you should rename the existing files instead - otherwise we'll lose history. Also, the files should be edited to reflect the name changes - there are references of switch models in the files themselves.

Have you also tested your changes in the GUI of PacketFence?

Also, a migration script should be provided to updates conf/switches.conf (and other parts) to reflect the name change.

The network configuration guide should also be adapted since we talk about switch models there too.

The PR can be left open, and you can push your changes gradually.

Thanks!

abprasa:
@extrafu Thanks Ludo for reviewing this and providing suggestions. Please find the answers in line:

`Correct me if I'm wrong but it looks like you added files and then renamed them. Perhaps you should rename the existing files instead - otherwise we'll lose history. Also, the files should be edited to reflect the name changes - there are references to switch models in the files themselves.`
Yes, I added new files and renamed them. 

`Have you also tested your changes in the GUI of PacketFence?`
Not yet because there is still some code cleanup to be performed on newer modules as suggested by Fabrice the newer modules have similar codes defined for features that are already defined in previous versions and can be fetched from there, so there is no need to redefine them. 

`Also, a migration script should be provided to update conf/switches.conf (and other parts) to reflect the name change.
`
Sure I will check with Fabrice and see how we can update that to reflect the name change.

`The network configuration guide should also be adapted since we talk about switch models there too.`
Sure we will update that too.

extrafu:
If you added files that were identical to old ones and after you renamed them, I think it's wrong because we lose history. I think you should just rename existing files and then do changes in them if required.

abprasa:
@extrafu Sure, Ludo I will check with Fabrice to find the best way to do that and we will get it done.

abprasa:
Will Create a new Draft PR, there is too much history and so starting a new one.

Yannik:
This can be closed since your other PR was merged @abprasa, right?

extrafu:
@abprasa @fdurand  I think we don't need this anymore, right? If so, close it.

PR #7689, opened 2023-06-02
Feature/switch module per version

abprasa:
# Description
[This is a New PR for "switch module update" and we will trash the old draft PR as the commits made in the old one made the module history inaccessible]
[OLD draft PR:https://github.com/inverse-inc/packetfence/pull/7677]

The existing code has a list of switch modules based on the Product model number.
This new code updates the existing list of Cisco switch Modules and maps them to the relevant versions and the features they support.

# Impacts

# NEW Package(s) required

# Feature
The idea of updating the existing list of Cisco switch Modules defined based on the model number to respective versions is to make it more convenient for users to choose modules based on the version of the switch.

YES 

# Checklist
- [x] Document the feature
- [x] Test the new modules on UI

# NEWS file entries
## New Features
* Updating existing switch model for the next release


lib/pf/Switch/Cisco/Cisco_IOS_15_0.pm
```diff
@@ -477,89 +477,6 @@ sub wiredeauthTechniques {
...
 
-=head2 returnRadiusAccessAccept
-
-Prepares the RADIUS Access-Accept reponse for the network device.
-
-Overrides the default implementation to add the dynamic acls
-
-=cut
-
-sub returnRadiusAccessAccept {
```
fdurand, line 488: This mean that our Cisco 15 module doesn't support the ACL and the Webauth

fdurand (changes requested):
I think this PR needs to be reworked a bit. We should add a kind of bypass acl from the upper switch per example then we will be able to keep the logic in 15 and add new logic in 15.5.
Kind of logic added here (removed since) with compute_action https://github.com/inverse-inc/packetfence/pull/7466/commits/455c7cf10fb5ed04d26e6759d446709efd9684b9

lib/pf/Switch/Cisco/Cisco_IOS_15_0.pm
abprasa: Thanks @fdurand I think I might need some help to fix that. Maybe you or @jrouzierinverse could help me with it?

PR #7690, opened 2023-06-02
Feature/upgrade to ansible 7 0

JeGoi:
# Description
Upgrade to ansible core 2.14, upgrade vagrant to 2.3.6, upgrade vagrant-libvirt to 0.12.1, upgrade cumulus to 4.3.0
The main idea is to have latest fixes from ansible and vagrant.
It will only runner on new runner with the right versions.

# Impacts
Should improve issues.

# Code / PR Dependencies
Need to be related with debops projects

# Delete branch after merge
YES

# Checklist
- [ ] Have new runner to make it work
- [ ] Have the right version on runners

### DIFF
--- a/addons/vagrant/Vagrantfile
+++ b/addons/vagrant/Vagrantfile
@@ -12,13 +12,17 @@ Vagrant.configure("2") do |config|
   config.vm.synced_folder ".", "/vagrant", disabled: true
 
   # The maximum amount of time to wait for a response
-  config.winrm.timeout = 600
+  config.winrm.timeout = 180
   
   # use same private key on all machines
   config.ssh.insert_key = false
 
   # Number of seconds to wait for establishing an SSH connection to the guest
-  config.ssh.connect_timeout = 180
+  config.ssh.connect_timeout = 90
+
+  # The time in seconds that Vagrant will wait for the machine to gracefully halt when vagrant halt is called.
+  # https://developer.hashicorp.com/vagrant/docs/vagrantfile/machine_settings#config-vm-graceful_halt_timeout
+  config.vm.graceful_halt_timeout = 120
 
   config.vm.provider "libvirt" do |vm|
     vm.default_prefix = DOMAIN_PREFIX
--- a/addons/vagrant/Vagrantfile
+++ b/addons/vagrant/Vagrantfile
@@ -18,7 +18,7 @@ Vagrant.configure("2") do |config|
   config.ssh.insert_key = false
 
   # Number of seconds to wait for establishing an SSH connection to the guest
-  config.ssh.connect_timeout = 60
+  config.ssh.connect_timeout = 120
 
   config.vm.provider "libvirt" do |vm|
     vm.default_prefix = DOMAIN_PREFIX
--- a/addons/vagrant/Vagrantfile
+++ b/addons/vagrant/Vagrantfile
@@ -11,9 +11,6 @@ Vagrant.configure("2") do |config|
   # disable synced folders
   config.vm.synced_folder ".", "/vagrant", disabled: true
 
-  # The maximum amount of time to wait for a response
-  config.winrm.timeout = 180
-  
   # use same private key on all machines
   config.ssh.insert_key = false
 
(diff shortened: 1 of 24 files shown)


JeGoi:
Need to add tags for runners per branch

addons/vagrant/Vagrantfile
```diff
@@ -12,13 +12,17 @@ Vagrant.configure("2") do |config|
   config.vm.synced_folder ".", "/vagrant", disabled: true
 
   # The maximum amount of time to wait for a response
-  config.winrm.timeout = 600
+  config.winrm.timeout = 180
```
nqb: I don't think we need to keep it since we have removed Windows VM.

addons/vagrant/cumulus/Vagrantfile
```diff
@@ -109,12 +119,20 @@ node03_netmask = user_vars['vms']['node03']['netmask']
...
+
   wbid = 1
   offset = wbid * 100
 
   ##### DEFINE VM for switch01 #####
   config.vm.define "switch01" do |device|
 
+
+    # Number of seconds to wait for establishing an SSH connection to the guest
+    config.ssh.connect_timeout = 120
```
nqb, line 134: This value is different from `addons/vagrant/Vagrantfile`, is it expected ?

addons/vagrant/inventory/group_vars/all/common.yml
```diff
@@ -26,3 +26,6 @@ pipeline_id: '{{ lookup("env","CI_PIPELINE_ID") | default("123456789", true) }}'
 # get current PacketFence minor release (X.Y)
 pf_minor_release: '{{ lookup("env", "PF_MINOR_RELEASE") | default("99.9", true) }}'
 
+# Cumulus Invormation
+switch01_api_user: cumulus
+switch01_api_password: CumulusLinux!
```
nqb: These variables are specific to Cumulus devices so they must move to a dedicated inventory group.

addons/vagrant/inventory/group_vars/clusters/venom_local_vars.yml
```diff
@@ -33,10 +33,10 @@ venom_local_vars:
     value: '{{ mailhog__api_port }}'
 
   - name: 'internet_int_ip'
-    value: "{{ ansible_default_ipv4['address'] }}"
+    value: "{{ ansible_default_ipv4.address }}"
```
nqb, line 36: Is it mandatory ?
I'm asking because we are mixing dot syntax (`ansible_default_ipv4.address`) and bracket syntax (`ansible_date_time['tz']`)

nqb (changes requested):
- Notable change: RHEL8 image has been updated too.
- One thing to have in mind also: we used Ansible not only to provision VM used for integration tests and to run Venom tests but also to:
  * build containers used to build images and packages
  * build our own Vagrant images
  * build ZEN images
  
We need to be sure there are no regressions on each of this component otherwise it's a recipe for disaster for the next release.

addons/vagrant/Vagrantfile
JeGoi: I though I removed it... My bad will do

addons/vagrant/inventory/group_vars/clusters/venom_local_vars.yml
JeGoi: I had a error with `ansible_default_ipv4['address']` but no error with `ansible_default_ipv4.address` and `ansible_date_time['tz']`. I did not search deep why new ansible version, new way to grab that gathering facts

addons/vagrant/cumulus/Vagrantfile
JeGoi: Yes, I found that when switch01 is short in memory it is taking more time than other VM.
It is more true with cumulus 5.5.0 which needs more 2GB of RAM than the one we had with v 3.X (~0.7GB).

JeGoi:
Merged for tests: https://github.com/inverse-inc/packetfence/pull/7855
Still open for other steps: https://github.com/inverse-inc/packetfence/issues/7858

PR #7705, opened 2023-06-07
Toggle ux fix

VakarisZ:
# Description
Fixes the toggle UI element:

## Before
before

## After
after

# Delete branch after merge
YES


html/pfappserver/root/src/components/new/BaseInputRange.vue
```diff
@@ -39,6 +39,7 @@
         :step="step"
         :disabled="isLocked"
         @change="onInput"
+        @mouseup="onClick"
```
VakarisZ: @click doesn't work

html/pfappserver/root/src/components/new/BaseFormGroupToggle.vue
```diff
@@ -54,11 +51,11 @@ import { useInputValidator, useInputValidatorProps } from '@/composables/useInpu
 import { useInputValue, useInputValueProps } from '@/composables/useInputValue'
 import { useInputValueToggle, useInputValueToggleProps } from '@/composables/useInputValueToggle'
 import BaseFormGroup from './BaseFormGroup'
-import BaseInputRange from './BaseInputRange'
+import BaseInputToggle from '@/components/new/BaseInputToggle.vue';
```
satkunas: Remove trailing semi-colon

html/pfappserver/root/src/components/new/BaseFormGroupToggle.vue
```diff
@@ -18,11 +18,8 @@
     :invalid-feedback="inputInvalidFeedback"
     :valid-feedback="inputValidFeedback"
   >
-    <base-input-range
+    <base-input-toggle
```
satkunas: 3-state toggles have regressed to 2-states only.

Before:

Screenshot 2023-06-08 at 9 43 19 AM


After:

Screenshot 2023-06-08 at 9 42 20 AM

html/pfappserver/root/src/components/new/BaseInputToggle.vue
```diff
@@ -98,6 +100,14 @@ export const setup = (props, context) => {
     validFeedback
   } = useInputValidator(metaProps, value)
 
+  const toggle = () => {
```
satkunas: This only handles 2-states. It should be able to handle N states (no limit).

satkunas (changes requested):

html/pfappserver/root/src/components/new/BaseFormGroupToggle.vue
VakarisZ: Hmmm... Have we concidered adding an eslint rule for this? https://eslint.org/docs/latest/rules/semi

html/pfappserver/root/src/components/new/BaseInputToggle.vue
VakarisZ: As we discussed on slack, switch components are just a fancy checkboxes, they are not meant to handle n states. For components requiring more than 2 states we should either use toggle buttons, range sliders, dropdowns, or radio buttons

html/pfappserver/root/src/components/new/BaseInputSwitch.vue
```diff
@@ -0,0 +1,287 @@
...
+    onBlur
+  } = useInput(metaProps, context)
+
+  const {
+    value,
+  } = useInputValue(metaProps, context)
+
+  const onChange = (value) => {
+    setFormNamespace(props.namespace.split('.'), unref(form), value)
+  }
```
VakarisZ: Ideally, low-level components like this should only be tracking their state. The switch should only track whether it's ON or OFF. Instead of namespace prop, the parent component should pass on change event. The boundary with the back end (view component) can then decide what true/false means (map true to "yes", "enabled" or "1")

VakarisZ: I haven't made that change because it would be inconsistent. All other low-level components track and modify the store. We can make the change, but we then have to make an effort to refactor other components as well, which seems to be a colossal effort

html/pfappserver/root/src/views/Configuration/domains/_components/index.js
```diff
@@ -23,10 +20,10 @@ export {
   BaseFormGroupInput                  as FormGroupAdServer,
   BaseFormGroupInput                  as FormGroupDnsServers,
   BaseFormGroupInput                  as FormGroupOu,
-  BaseFormGroupToggle                 as FormGroupNtlmv2Only,
-  BaseFormGroupToggle                 as FormGroupRegistration,
+  BaseFormGroupSwitch                 as FormGroupNtlmv2Only,
+  BaseFormGroupSwitch                 as FormGroupRegistration,
 
-  BaseFormGroupToggleDisabledEnabled  as FormGroupNtlmCache,
+  BaseFormGroupSwitch                 as FormGroupNtlmCache,
```
VakarisZ, line 26: These named exports is not a good idea. It makes it harder to refactor and read, since you don't know what component it actually is until you look at this reference

satkunas (changes requested):
The code and formatting changes look good.

Get the BaseInputSwitch and BaseFormGroupSwitch components working with both v-model (:value @change), and namespace; then deprecate the OnChangeFormGroupSwitch component.

Also ensure npm run lint is clean.

PR #7712, opened 2023-06-12
Change port for deployment

JeGoi:
# Description
Changes are needed to bypass port 22/TCP restrictions on Lab Access.

# Impacts
Using another port to bypass port 22/TCP restrictions.
Should not impact the current CI or workflow.
NEED TO BE BACKPORTED TO OTHER BRANCHES THAT ARE USING THE CI.

# Delete branch after merge
YES

# Checklist
- [x] Do a pipeline

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -422,6 +422,7 @@ variables:
   variables:
     DEPLOY_USER: reposync
     DEPLOY_HOST: web.inverse.ca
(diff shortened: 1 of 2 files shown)


nqb (approved):
Changes looks good.
We need to be sure remote server is listening on port 444.

The backport will be a bit tricky because you need to backport until 9.2.

PR #7721, opened 2023-06-22
Notify admins when unexpected devices pop up on network

### LINKED ISSUE #7529: Receive a notification each time a device is on registration VLAN
I would like to be notify each time a device is on registration VLAN because it's something uncommon: all my devices should be on production VLAN. So I want to detect it.

Using following VLAN filter:
```
[test]
condition=node_info.status == "unreg"
status=enabled
description=test
run_actions=enabled
scopes=RegistrationRole
top_op=and
action.0=trigger_security_event: mac, $mac, tid, registration, type, CUSTOM
```
and a security event, a node will match this VLAN filter in two cases;
1. if device is not registered when a RADIUS request is received (expected behavior)
1. if device reach unregistration date and is unregistered by nodes_maintenance task (side effect)

I would like my VLAN filter to match only in 1 because 2 will apply to almost all devices.

Have a way to match VLAN filter only in 1.

I look at conditions on VLAN filter but I don't find a suitable condition to match 1 and not 2.

I think a solution could be to use a dedicated scope when we deauthenticate a device. Currently, we call  `fetchRoleForNode` from `_should_we_reassign_vlan` which means that we are in the same code path in 1 and 2 at some point.

**Describe alternatives you've considered**
Use a different approach like a report

nqb:
# Description
Be informed (through security event) when a device pops up into a VLAN or a subnet that shouldn’t be there.

In order to trigger a security event, PacketFence must receive: DHCP traffic or RADIUS **accounting**. We don't trigger on RADIUS authentication requests.

If we receive several DHCP or RADIUS requests for the same device but IP isn't updated, security event will not be triggered again due to:

```
pfqueue(3805535) DEBUG: [mac:unknown] No recent data found for 33:33:33:33:33:33, will not trigger device profiling. Was last updated at 2023-06-21T10:17:12.067524888Z and last process timestamp is 2023-06-21T10:17:12 (pf::fingerbank::process)
```

# Issue
fixes #7529

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Be informed (through security event) when a device pops up into a VLAN or a subnet that shouldn’t be there


nqb:
@satkunas, I need your help to make "Device is not" field perform a search exactly like "Device" field is doing (using `api/v1/fingerbank/all/devices/search`). I was not able to figure out how to do it.

nqb:
I just need help to finalize frontend but nothing more to do on my side.

nqb:
@jrouzierinverse I had a discussion with @satkunas and in fact, the code needs to be adjusted on backend. `pfperl-api` currently returns:

```console
# pfperl-api get -M OPTIONS /api/v1/config/security_events
[..]
"device": {                                                                                                                                                                                                                       │
            "allow_custom": false,                                                                                                                                                                                                          │
            "allowed_lookup": {                                                                                                                                                                                                             │
              "field_name": "name",                                                                                                                                                                                                         │
              "search_path": "/api/v1/fingerbank/all/devices/search",                                                                                                                                                                       │
              "value_name": "id"                                                                                                                                                                                                            │
            },               
[..]
"device_is_not": {                                                                                                                                                                                                                │
            "default": null,                                                                                                                                                                                                                │
            "implied": null,                                                                                                                                                                                                                │
            "placeholder": null,                                                                                                                                                                                                            │
            "required": false,                                                                                                                                                                                                              │
            "type": "string"                                                                                                                                                                                                                │
          },
```

I would like to have similar results for `device_is_not` but I'm not able to figure out which portion of the code should I changed.
I think the logic is under `/usr/local/pf/lib/pf/UnifiedApi/Controller/Config.pm`.
Could you provide some help @jrouzierinverse ?

t/unittest/security_events.t
```diff
@@ -103,6 +103,21 @@ is(@security_events, 0);
...
+# Test a security_event using network
+@security_events = $pf::security_event::SECURITY_EVENT_FILTER_ENGINE->match_all({last_ip => "172.30.0.1"});
+is(@security_events, 1);
+is($security_events[0], "1100020");
+
+# Test a security_event using network and device_is_not
+@security_events = $pf::security_event::SECURITY_EVENT_FILTER_ENGINE->match_all({last_ip => "172.31.0.1", device_id => 1});
+is(@security_events, 1);
+is($security_events[0], "1100021");
+
```
jrouzierinverse, line 120: Add the following tests
```perl
# Test a security_event using network and device_is_not
@security_events = $pf::security_event::SECURITY_EVENT_FILTER_ENGINE->match_all({last_ip => "172.31.0.1", device_id => 33453});
is(@security_events, 0);
```

jrouzierinverse (changes requested):

t/unittest/security_events.t
nqb: Done

nqb:
@jrouzierinverse, I made changes you requested in `Trigger.pm` + unit tests. Frontend behaves as it should now.
If you are happy with the state of this PR, please merge.

mellalahmed:

PR #7729, opened 2023-06-27
Create a tag for sourceforge

### LINKED ISSUE #7728: ci: new runners are unable to upload to sourceforge
**Describe the bug**
When a build (ISO or ZEN) is running on new runners (lab), they can't upload things to Sourceforge because SSH to outside is denied on these runners.

We need to make sure these jobs will never run on these runners. Certainly using tags.

JeGoi:
# Description
Add tag shell-sourceforge to be applied on shell runners out of runner access 

# Impacts
Fixes #7728 
Needs to be backported if we create new iso or zen on maintenance branch until 11.0
Needs to be done in relation with https://git.inverse.ca/inverse/debops/issues/2

# Delete branch after merge
YES

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -531,7 +531,7 @@ variables:
     - make -e -C ${ZENDIR} clean_all
   dependencies: []
   tags:
-    - shell
+    - shell-sourceforge
 
 .build_pf_img_iso_job:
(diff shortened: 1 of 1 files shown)


nqb:
No need to backport.

You just need to unregister/register runners.

PR #7784, opened 2023-08-01
Ldap explorer manual input

### LINKED ISSUE #7768: PF 13: LDAP explorer not working
PF version:

````
root@debian-pf1:/usr/local/pf# dpkg -l packetfence
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version                                              Architecture Description
+++-==============-====================================================-============-=========================================================
ii  packetfence    13.0.0+20230717231108+934755451+0011+devel+bullseye1 all          PacketFence network registration / worm mitigation system
````

image

image

### PR DESCRIPTION
**Pull request by VakarisZ**
# Description
In some cases we can't be sure that all ldap options appear in the select. For those cases we want to allow the user to manually input what he wants to match. 

Example case: 
User wants to specify memeberOf. To get all the `memberOf`'s we pull all entries that have a non-empty `memeberOf`. If there's a limit (default is 1000) to how many entries are returned, it's possible that 1000 entries returned have the same `memberOf` value and entries with different `memberOf` values simply were not returned, so they were not included into the select. In that case we want to signal to the user that some attributes might be missing and to allow manual input.

# Demo

ldap_manual_input

Fixes #7768

### CODE REVIEW

File: conf/I18N/api/fr.po [hunk 1]
```diff
@@ -14546,6 +14546,15 @@ msgstr "Chemin du script"
 msgid "Search"
 msgstr "Rechercher"
 
+#: html/pfappserver/root/src/views/Configuration/sources/_components/ldapCondition/LdapSearchInput.vue:53
+msgid "Results might be missing"
+msgstr "Les résultats peuvent manquer"
+
+#: html/pfappserver/root/src/views/Configuration/sources/_components/ldapCondition/LdapSearchInput.vue:64
+msgid "Input manually"
+msgstr "Saisir manuellement"
```
VakarisZ, line 14555: Check these translations

html/pfappserver/root/src/main.js
```diff
@@ -87,6 +88,7 @@ import 'vue-awesome/icons/play'
 import 'vue-awesome/icons/play-circle'
 import 'vue-awesome/icons/plug'
 import 'vue-awesome/icons/plus-circle'
+import 'vue-awesome/icons/plus'
```
VakarisZ, line 91: This is not a good pattern, each component should import what it needs

satkunas (approved):

PR #7808, opened 2023-08-22
support ip address column for bulk node import

### LINKED ISSUE #7791: Node import - Consider IP address
When importing nodes from a CSV file, we should offer the option to import an IP address.

This is particularly useful if one imports assets that aren't doing DHCP - like PLC, some medical devices, etc. Those typically have a static IP assigned.

We could create an open entry in the iplog for those, so at least we'll also have IP information in PacketFence.

stgmsa:
# Description
allow users to specific a ip4 for each node while doing the bulk import

# Impacts
accepts a parameter ip4 in the front end and it will be used for import node

# Issue
fixes #7791 

# Delete branch after merge
YES

# Checklist
- [no] Document the feature
- [no] Add unit tests

### DIFF
--- a/html/pfappserver/root/src/globals/mysql.js
+++ b/html/pfappserver/root/src/globals/mysql.js
@@ -219,11 +219,6 @@ export const MysqlDatabase = {
       type: MysqlDatetime,
       format: 'YYYY-MM-DD HH:mm:ss',
       default: '0000-00-00 00:00:00'
-    },
-    ip: {
-      type: MysqlString,
-      maxLength: 16,
-      default:null
     }
   },
   password: {
--- a/html/pfappserver/root/src/globals/mysql.js
+++ b/html/pfappserver/root/src/globals/mysql.js
@@ -219,6 +219,11 @@ export const MysqlDatabase = {
       type: MysqlDatetime,
       format: 'YYYY-MM-DD HH:mm:ss',
       default: '0000-00-00 00:00:00'
+    },
+    ip: {
+      type: MysqlString,
+      maxLength: 16,
+      default:null
     }
   },
   password: {
--- a/html/pfappserver/root/src/views/Nodes/_config/index.js
+++ b/html/pfappserver/root/src/views/Nodes/_config/index.js
@@ -148,7 +148,7 @@ export const importFields = [
     text: i18n.t('IP'),
     types: [fieldType.SUBSTRING],
     required: false,
-    validator: validatorFromColumnSchemas(MysqlDatabase.node.ip)
+    validator: yup.string().nullable().isIpv4(i18n.t('Invalid IPv4 Address.'))
   },
 ]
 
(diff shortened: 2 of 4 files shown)


html/pfappserver/root/src/globals/mysql.js
```diff
@@ -219,6 +219,11 @@ export const MysqlDatabase = {
       type: MysqlDatetime,
       format: 'YYYY-MM-DD HH:mm:ss',
       default: '0000-00-00 00:00:00'
+    },
+    ip: {
+      type: MysqlString,
+      maxLength: 16,
```
satkunas: 15 chars

html/pfappserver/root/src/views/Nodes/_config/index.js
```diff
@@ -142,7 +142,14 @@ export const importFields = [
...
       labelRight: true
     }
-  }
+  },
+  {
+    value: 'ip4',
+    text: i18n.t('IP'),
+    types: [fieldType.SUBSTRING],
+    required: false,
+    validator: validatorFromColumnSchemas(MysqlDatabase.node.ip)
```
satkunas: This is a weak validator, use isIpv4 instead https://github.com/inverse-inc/packetfence/blob/devel/html/pfappserver/root/src/utils/yup.js#L359

lib/pf/UnifiedApi/Controller/Nodes.pm
```diff
@@ -1051,6 +1051,11 @@ sub import_item {
         node_update_last_seen($mac);
     }
 
+    if (exists($node{-ip4})) {
```
satkunas: @jrouzierinverse @zma should we namespace this key? `ip4` => `ip4log.ip` in consideration for our future golang DAL?

satkunas (changes requested):

lib/pf/UnifiedApi/Controller/Nodes.pm
jrouzierinverse: That makes sense.
That should be changed to ip4log.ip
```if (exists($node{'ip4log.ip'})) {```

File: lib/pf/UnifiedApi/Controller/Nodes.pm [hunk 4]
```diff
@@ -1051,6 +1051,11 @@ sub import_item {
...
+        my $ip4= $node{ip4};
```
jrouzierinverse: should be changed to
```my $ip4= $node{'ip4log.ip'};```

**Review by jrouzierinverse — CHANGES_REQUESTED** [review]

File: html/pfappserver/root/src/views/Nodes/_config/index.js [hunk 5]
```diff
@@ -142,7 +142,14 @@ export const importFields = [
       ],
       labelRight: true
     }
-  }
+  },
+  {
+    value: 'ip4',
```
jrouzierinverse: value: 'ip4log.ip',

jrouzierinverse (changes requested):

PR #7908, opened 2023-11-17
Fix update to the Debian sudoers file

### LINKED ISSUE #7738: Change to debian sudoers file prevents passing of $TERM to sudo environment
**Describe the bug**
Changes made to the debian sudoers file in this commit stop the `$TERM` value from being passed to the sudo environment:

https://github.com/inverse-inc/packetfence/commit/c37b32b012a934282ecd91f10f1d09dd7e402a60

When sudoing, it sets `$TERM=unknown`, which creates issues with running various commands (including `nano`, `less`, `vi`, `journalctl`) as sudo, e.g.:

```
$ sudo nano
Error opening terminal: unknown.
```

When we change the line to `Defaults env_check="PF_UID PF_GID TERM", !requiretty` the error disappears and we can run commands with sudo again, but it feels like maybe these defaults (or at least the `env_check`) should only apply to the `pf` user?

**To Reproduce**
1. Install PF 12.1+ on Debian 11
2. Try to run commands listed above as sudo

**Expected behavior**
Sudo behaviour for non-`pf` users to behave as usual, passing `$TERM` into sudo environment as expected

Edit:  https://github.com/inverse-inc/packetfence/issues/7648 - replied there initially but with the issue being closed it might not get seen

andrew-grasso:
# Description
Fix for issue  ["Change to Debian sudoers file prevents passing of $TERM to sudo environment"](https://github.com/inverse-inc/packetfence/issues/7738)

Commit https://github.com/inverse-inc/packetfence/commit/c37b32b012a934282ecd91f10f1d09dd7e402a60 added two environment variables to the `env_check` sudoers config option which allows them to be passed through when using sudo, however the use of the `=` assignment operator in that commit overwrites the default environment variables normally included in `env_check` including TERM. Instead, it is better to use the `+=` assignment operator to append the new environment variables to the already existing defaults.

# Impacts
Before this fix, installing PacketFence on Debian caused issues with sudo when running commands like `nano`, `less`, `vi`, `systemctl`, and `journalctl` because the TERM environment variable was being stripped. After the fix, sudo once again works as expected and the two custom `PF_` environment variables are still included in the check list which can be verified by running `sudo -V` as the root user with and without the fix applied and checking the "Environment variables to check for safety" section of the output.

# Issue
fixes #7738

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

## Bug Fixes
* Fixes issue  ["Change to Debian sudoers file prevents passing of $TERM to sudo environment"](https://github.com/inverse-inc/packetfence/issues/7738)


andrew-grasso:
Any feedback from maintainers on this? Any concerns or issues preventing it from being merged? @fdurand

PR #7911, opened 2023-11-21
feature/kakfa-service

jrouzierinverse:
# Description
Add kafka as a service.

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Add kafka as a service.


extrafu:
@jrouzierinverse some comments, @satkunas please follow-up:

Is there anything missing around:

1. iptables handling - between PacketFence servers, but also for external applications consuming stuff from our Kafka service
2. GUI configuration option to have the Fingerbank Collector log flow information to Kafka for: nothing, all devices, only registered devices
3. GUI configuration option to have Fingerbank Collector log flow information to localhost or an other IP/port

I think we need the following in our official documentation:

1. installation instructions of Kafka (if required)
2. basic Kafka maintenance and troubleshooting (like corrupted data)

Not entirely related to this but we need also documentation on ERSPAN -> ipt-netflow -> Fingerbank Collector in our official documentation.

extrafu:
@jrouzierinverse Also, related to having the Fingerbank Collector log for certain devices or not, we should probably handle this either:

1. after we're done doing a RADIUS authentication (so maybe on an accounting packet). We should also support doing this when we se a DHCP packet.
2. the option above won't work if we aren't doing RADIUS, nor DHCP. Think about an OT environment where devices are manually registered in PacketFence but we only want to report flow information. We don't want to do RADIUS. In this case, maybe a better approach is to have the Collector query PacketFence when it needs to report flows to see if it must report it or not for a certain "device" and cache the result. We'll have to deal with caching... (like a device being unregistered and moving to a registered one...."

Talk to @fdurand on this, for the best approach to take.

conf/documentation.conf
```diff
@@ -598,6 +598,13 @@ description=<<EOT
 Location of the proxysql binary.
 EOT
 
+[services.kafka]
+type=toggle
+options=enabled|disabled
+description=<<EOT
+Should kafka be started? Keep disabled unless you know what you're doing.
```
extrafu: kafka -> Kafka

conf/pf.conf.defaults
```diff
@@ -732,6 +732,11 @@ proxysql_binary=/usr/bin/proxysql
 #
 # Should pfsetacls be started?
 pfsetacls=enabled
+#
+# services.kafka
+#
+# Should kafka be started?
```
extrafu: kafka -> Kafka

html/pfappserver/root/src/views/Configuration/services/_components/TheForm.vue
```diff
@@ -287,6 +287,13 @@
                             enabled-value="enabled"
                             disabled-value="disabled"
     />
+
+    <form-group-mysql-probe namespace="kafka"
+                            column-label="kafka"
+                            :text="$i18n.t(`Should kafka be started? Keep disabled unless you know what you're doing.`)"
```
extrafu: kafka -> Kafka

jrouzierinverse:
> @jrouzierinverse some comments, @satkunas please follow-up:
> 
> Is there anything missing around:
> 
> 1. iptables handling - between PacketFence servers, but also for external applications consuming stuff from our Kafka service
> 2. GUI configuration option to have the Fingerbank Collector log flow information to Kafka for: nothing, all devices, only registered devices
> 3. GUI configuration option to have Fingerbank Collector log flow information to localhost or an other IP/port
> 
> I think we need the following in our official documentation:
> 
> 1. installation instructions of Kafka (if required)
> 2. basic Kafka maintenance and troubleshooting (like corrupted data)
> 
> Not entirely related to this but we need also documentation on ERSPAN -> ipt-netflow -> Fingerbank Collector in our official documentation.

At the moment, we have a way to configure the collector via the config; however, there is no GUI for it and/or instructions for it.

We would need to expose the collector section in `/usr/local/fingerbank/conf/fingerbank.conf`
Also, we would need to add the help for the environmental variables in the go in additional_env.
I will discuss with @satkunas the best way forward with this.

```
[collector]
host=100.64.0.1
port=4723
use_https=enabled
inactive_endpoints_expiration=168
arp_lookup=disabled
query_cache_time=1440
db_persistence_interval=60
cluster_resync_interval=120
network_behavior_analysis=disabled
additional_env=
```

satkunas (commented):
Fix errors w/ empty config
```
 /usr/local/pf/bin/pfcmd configreload hard

Use of uninitialized value in split at /usr/local/pf/lib/pfconfig/namespaces/config/Kafka.pm line 37.
Use of uninitialized value in split at /usr/local/pf/lib/pfconfig/namespaces/config/Kafka.pm line 37.
Use of uninitialized value in string eq at /usr/local/pf/lib/pfconfig/util.pm line 98.
```

PR #7923, opened 2023-12-07
Force mechanism LOGIN PLAIN

### LINKED ISSUE #7813: Unable to select SASL mechanism and timeout not honored
When configure SMTP settings in System Configuration > Alterting, it's not possible to select the SASL mechanism to use if SMTP authentication is required.

For example, if the server advertises "GSSAPI NTLM LOGIN", we'll incorrectly pick the first, GSSAPI which will lead to SMTP send failures.

We should force the user to pick between PLAIN and LOGIN, the only two we should support. PLAIN should be the default.

Also, the timeout isn't honored correctly when sending mails, especially in-between the TLS negotiation and the SMTP authentication. If it takes a couple of seconds, it'll fail independently of the timeout set in System Configuration > Alterting.

@fdurand please put the 2 lines of code you added at a customer to make it work, so anyone working on this will have some inspiration.

jrouzierinverse:
# Description
Force mechanism LOGIN PLAIN for SMTP.

# Issue
fixes #7813

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Enhancements
* Force mechanism LOGIN PLAIN for SMTP. (#7813)

### DIFF
--- a/lib/pf/config/util.pm
+++ b/lib/pf/config/util.pm
@@ -629,7 +629,7 @@ sub send_using_smtp_callback {
     {
         if ( $smtp->supports( 'AUTH', 500, ["Command unknown: 'AUTH'"] ) ) {
             my $sasl = Authen::SASL->new(
-              mechanism => 'LOGIN PLAIN',
+              mechanism => 'PLAIN LOGIN',
               callback => {
                 pass => $args{AuthPass},
                 user => $args{AuthUser},
--- a/lib/pf/config/util.pm
+++ b/lib/pf/config/util.pm
@@ -44,6 +44,7 @@ use Encode qw(encode encode_utf8);
 use File::Basename;
 use Net::MAC::Vendor;
 use Net::SMTP;
+use Authen::SASL;
 use MIME::Lite;
 use MIME::Lite::TT;
 use POSIX();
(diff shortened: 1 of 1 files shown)


lib/pf/config/util.pm
```diff
@@ -627,7 +628,15 @@ sub send_using_smtp_callback {
         and !$args{NoAuth} )
     {
         if ( $smtp->supports( 'AUTH', 500, ["Command unknown: 'AUTH'"] ) ) {
-            $smtp->auth( $args{AuthUser}, $args{AuthPass} )
+            my $sasl = Authen::SASL->new(
+              mechanism => 'LOGIN PLAIN',
```
extrafu: I don't understand this.

Is the Perl library smart enough to prioritize LOGIN and then PLAIN? "LOGIN PLAIN" is not a single SASL mechanism, those are two different mechanisms and PLAIN should be prioritized, it's much more popular than LOGIN.

extrafu: @jrouzierinverse Any updates on my comments?

jrouzierinverse: By doing some source diving it seems to be the case.
It will attempt them in order removing which one failed and retry.
You can see how it used here.
https://metacpan.org/pod/Authen::SASL

extrafu: Ok good, thanks for the explanation. Just flip LOGIN and PLAIN then, since PLAIN is a lot more popular.

extrafu: What about the timeout? In the bug report, it's mentioned that is is not used and should be.

satkunas:
@jrouzierinverse update timeout

@fdurand review after

jrouzierinverse:
@fdurand please review

PR #7954, opened 2024-01-04
Sort OAS Schemas

satkunas:
# Description
Sort OAS schemas to prevent random re-ordering of spec everytime it's regenerated.

# Delete branch after merge
YES


satkunas:
@jrouzierinverse please add sorting in `subTypesMetaSchema`

PR #8010, opened 2024-02-13
Feature/Extreme switch module per version

abprasa:
# Description
The existing code has a generic Extreme module, this new code updates the existing Extreme Module, adds more modules based on the currently available releases, and maps them to the relevant versions and the features they support.

# Impacts

# NEW Package(s) required

# Feature
The idea of updating the existing list of Extreme Modules defined based on the model number to respective versions is to make it more convenient for users to choose modules based on the version of the switch.

YES 

# Checklist
- [ ] Document the feature In the Network Device Config Guide
- [ ] Migration Script
- [ ] Test the new modules on UI

# NEWS file entries
## New Features
* Updating existing switch model for the next release


lib/pf/Switch/Extreme/EXOS_v30_x.pm
```diff
@@ -0,0 +1,201 @@
...
+}
+
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2023 Inverse inc.
```
fdurand: 2024

lib/pf/Switch/Extreme/EXOS_v30_x.pm
```diff
@@ -0,0 +1,201 @@
...
+=head1 SNMP
+
+This switch can parse SNMP traps and change a VLAN on a switch port using SNMP.
+
+=cut
+
+use strict;
+use warnings;
+use Net::SNMP;
+use base ('pf::Switch::Extreme::Summit::EXOS');
```
fdurand: It probably miss the base module in the PR

PR #8044, opened 2024-03-18
NT key caching support in ntlm_auth_api

### LINKED ISSUE #1246: Avoid accounts being locked due to password changes in AD
Issue:  Alice has her password being changed by Windows administrator Zaymmitte. She forgets to change the password her password on her iPhone which uses WiFi/EAP-PEAP. After X failed logins from her iPhone, her account gets locked out and she sadly has to call Zaymmitte - who is likely busy bitching about manouches.

Solution:
1. you can enable N-x password history in AD - so AD will keep x previously used password and if one of them is being used, it will NOT lock the account (the WiFi connection will still be rejected)
2. if a wrong and unknown password is being sent (ie.,  brute-force attack or anything else), the badPwdCount attribue for the user will be incremented each time a wrong password is tried. It will be reset to 0 upon a successful "login"
3. one thing can be done, is to use a LDAP filter in FreeRADIUS **prior** doing the ntlm_auth call. This filter can check if the badPwdCount attribute is greater than a preferred value and if so, we reject the connection immediately. This will prevent from doing ntlm_auth calls with a bogus password and thus, potentially lock the account.
4. if we detect such cases in 3, we could always unconditionally accept the EAP-PEAP connection but raise a violation showing the captive portal page with password informations. This will not work for supplicants doing NT key/LM session key validations but for those who don't, they will at least see some instructions.

### LINKED ISSUE #5945: Improve the mechanism to reject devices using an old password
When the password of a user gets changed, he needs to change it on all his devices using EAP-PEAP MSCHAPv2. If the user forgets to update a device, then it may lead to its account being locked.

Currently we have a cache that if a device uses the wrong password for X times in a timeframe, then we automatically reject the device for a configurable time T until we try to send it to the AD again. This means that if the user finally updates the device with the right password, he will have to wait until T has elapsed for his connection to work again.

The base of this mechanism is good but could use the following improvements:
- Use the security events to manage this mechanism where we'll stop sending it to the AD if a specific security event is opened
 - That security event will auto-close and have grace time configuration so that should allow for easy configuration of this feature through the admin
- When the security event is opened, don't simply reject the request but try to hit the NTLM cache only (i.e. if the NTLM cache misses then we don't send it to the AD). That means the device will succeed if at least one of the user's device has previously authenticated with the right password while the NTLM cache is enabled.

**Additional context**

### LINKED ISSUE #8013: Error using domain admin password containing spaces
**Describe the bug**
Adding an ActiveDirectory through `Configuration/Policies and Access Control/Domains/Active Directory Domains` using a `Domain administrator password` containing **spaces** leads to an error in Packetfence `13.1.0+20240207213823+1168611722+0011+maintenance~13~1+bullseye1`.

**To Reproduce**
1. Try adding an Active Directory Server and use a password containing **spaces** like `this is my password` or `please execute $(this)`
2. Click on Save
3. Error messages in notification area
4. Error message in `packetfence.log`:

`packetfence pfperl-api-docker-wrapper[66715]: pfperl-api(10) WARN: [mac:[undef]] Problem trying to run command: LANG=C /usr/local/pf/bin/impacket-addcomputer -computer-name packetfence$ -computer-pass 1234567890ABCDEFGHIjkL -dc-ip 192.168.1.1 -dc-host 'dc.intern.doma.in' -baseDN 'DC=intern,DC=doma,DC=de' -computer-group all-computers intern/domainadmin:password containing spaces   called from (eval). Child exited with non-zero value 2 (pf::util::pf_run)`

**Expected behavior**
If there are Active Directory Server out there that accept passwords containing spaces packetfence should allow them to be used.

**Desktop (please complete the following information):**
 - Debian Bookworm
 - Firefox 122.0.1

stgmsa:
# Description
Introduced NT key caching support in ntlm_auth_api 

# Impacts
changed ntlm_auth_api authenticate flow - added caching machenisim 
changed RADIUS authorize workflow - added mac address during NTLM auth (reporting by ntlm_auth_wrapper)

# NEW Package(s) required
python dependencies:
pymysql
flaskext.mysql 

# Issue
fixes #1246 
fixes #5945
fixes #8013 

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [na] Add unit tests
- [na] Add acceptance tests (TestLink)

# NEWS file entries

## New Features
bin/pyntlm_auth/config_generator.py
bin/pyntlm_auth/ncache.py
bin/pyntlm_auth/utils.py

## Enhancements
* introduced nt_key caching machemisn

# UPGRADE file entries
bin/pyntlm_auth/app.py
conf/radiusd/mschap.conf.example
lib/pf/services/manager/ntlm_auth_api.pm
src/ntlm_auth_wrap.c


conf/adminroles.conf.example
```diff
@@ -11,3 +11,7 @@ allowed_access_levels=User Manager
 [Security Event Manager]
 description=Security Events managements
 actions=SECURITY_EVENTS_READ,SECURITY_EVENTS_CREATE,SECURITY_EVENTS_UPDATE,SECURITY_EVENTS_DELETE,USERS_READ,NODES_READ
+
+[External API Receiver]
+description=Receives and execute external api calls
+actions=API_CALL_READ, API_CALL_CREATE, API_CALL_UPDATE, API_CALL_DELETE
```
satkunas: These nouns should be more descriptive. All of these roles are used by the "API", so this is confusing.

Include "NTLM" in the naming

Suggestion:
```
[NTLM Webhooks]
description=NTLM methods received from Active Directory
actions=NTLM_READ, NTLM_CREATE, NTLM_UPDATE, NTLM_DELETE
```

lib/pf/constants/admin_roles.pm
```diff
@@ -216,6 +216,11 @@ our @ADMIN_ACTIONS = qw(
     MFA_DELETE
     MFA_READ
     MFA_UPDATE
+
+    API_CALL_CREATE
+    API_CALL_READ
+    API_CALL_UPDATE
+    API_CALL_DELETE
```
satkunas: Rename (same as above)

bin/pyntlm_auth/rpc.py
```diff
@@ -0,0 +1,156 @@
...
+    workstation = global_vars.c_workstation
+    password = global_vars.c_password
+    domain = global_vars.c_domain
+    username = global_vars.c_username
+    server_name = global_vars.c_server_name
+
+    lp = param.LoadParm()
+    try:
+        config_generator.generate_empty_conf()
+        lp.load("/root/default.conf")
```
extrafu: Why /root ?

bin/pyntlm_auth/utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+# simplified IPv4 validator.
+def is_ipv4(address):
+    ipv4_pattern = re.compile(r'^(\d{1,3}\.){3}\d{1,3}$')
+    return bool(ipv4_pattern.match(address))
+
+
+def nt_time_to_datetime(nt_time):
+    if nt_time == 9223372036854775807:
```
extrafu: put that into a contant value with a descriptive name

conf/adminroles.conf.example
stgmsa: changed role to "Windows Event Receiver NTLM",
permission to "Windows Events READ" but accepts a POST requests - from PacketFence side, it's reading the windows events, not creating.

bin/pyntlm_auth/rpc.py
stgmsa: moved from original app.py,
for persistence purpose. put it to somewhere like /usr/local/pf/var/conf ?

bin/pyntlm_auth/config_loader.py
```diff
@@ -0,0 +1,242 @@
...
+        server_name_or_hostname = server_name_raw
+        if server_name_raw.strip() == "%h":
+            server_name_or_hostname = socket.gethostname().split(".")[0]
+
+        ad_fqdn = config.get(identifier, 'ad_fqdn')
+        ad_server = config.get(identifier, 'ad_server')
+        realm = config.get(identifier, 'dns_name')
+        workgroup = config.get(identifier, 'workgroup')
+        dns_servers = config.get(identifier, 'dns_servers')
+        username = server_name_or_hostname.upper() + "$"
```
fdurand, line 81: username is the correct term since it's a machine name ?

bin/pyntlm_auth/handlers.py
```diff
@@ -0,0 +1,154 @@
...
+        return "Test machine account failed. Access Denied", HTTPStatus.UNAUTHORIZED
+    if error_code != 0:
+        return "Error while establishing secure channel connection: " + error_message, HTTPStatus.INTERNAL_SERVER_ERROR
+
+    return "OK", HTTPStatus.OK
+
+
+def ntlm_auth_handler():
+    try:
+        required_keys = {'username', 'mac', 'request-nt-key', 'challenge', 'nt-response'}
```
fdurand, line 87: MAC is required ?

conf/radiusd/mschap.conf.example
```diff
@@ -243,7 +243,7 @@ mschap chrooted_mschap {
     # Call ntlm_auth through the logging wrapper. Make sure to preserve the -- separator to distinguish between
     # the args to the wrapper and those to the ntlm_auth executable itself
     ntlm_auth = "/usr/local/pf/bin/ntlm_auth_wrapper -p [% statsd_port %] -a %{PacketFence-NTLM-Auth-Host} -t %{PacketFence-NTLM-Auth-Port} -- \
-         --request-nt-key --username=%{%{control:AD-Samaccountname}:-%{%{Stripped-User-Name}:-%{mschap:User-Name:-None}}} --challenge=%{mschap:Challenge:-00} --nt-response=%{mschap:NT-Response:-00} %{PacketFence-NTLMv2-Only}"
+         --request-nt-key --mac=%{Calling-Station-Id} --username=%{%{control:AD-Samaccountname}:-%{%{Stripped-User-Name}:-%{mschap:User-Name:-None}}} --challenge=%{mschap:Challenge:-00} --nt-response=%{mschap:NT-Response:-00} %{PacketFence-NTLMv2-Only}"
```
fdurand: In the case of mschap for a VPN access there is no Calling-Station-Id attribute. We should have an option to have the mac and not having the mac.

bin/pyntlm_auth/handlers.py
stgmsa: Yes. since mac address will be used for identifying a specific device, it's required. 
not sure if we can get a mac address in this case: "In the case of mschap for a VPN access there is no Calling-Station-Id attribute. We should have an option to have the mac and not having the mac."

bin/pyntlm_auth/config_loader.py
stgmsa: username will be the "machine account name" in windows domain controller.

html/pfappserver/root/src/views/Configuration/domains/schema.js
```diff
@@ -72,7 +79,43 @@ export default (props) => {
       }),
     machine_account_password: yup.string().nullable().label(i18n.t('Machine Account Password'))
       .required(i18n.t('Password required.'))
-      .min(8),
+      .minAsInt(8),
```
fdurand: This create an issue and prevent the edition of the domain configuration.

PR #8131, opened 2024-05-02
Fixing missing OU in DN when creating machine account

### LINKED ISSUE #8112: PF 13.2: ntlm domain does not use the OU path to create object
image
image

Drop the account in Computers by default

image

stgmsa:
# Description
Fix missing OU settings when creating machine account,
when OU presents, add computers using LDAPS instead of SAMR. 

# Impacts
changed the logic of adding / updating machine account, will not do "-no-add", will do "delete" and "add" instead
changed protocols used in adding machine account. LDAPS will be chosen if OU presents. 
LDAPS service is required on Windows Domain Controller (Port 636 with MS AD Cert Service)
changed NT Hash authentication back to password auth because LDAPS failed to bind when using NT Hash.
escapes multiple possible strings to allow special characters such as ", ', or \, etc...

# Issue
fixes #8112 

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature
- [n/a] Add unit tests

## Bug Fixes
* Issue with OU parameter not taken when creating machine account. (#8112)

### DIFF
--- a/lib/pf/domain.pm
+++ b/lib/pf/domain.pm
@@ -26,7 +26,7 @@ use Encode qw(encode);
 use File::Slurp;
 
 # This is to create the templates for the domain info
-our $TT_OPTIONS = {ABSOLUTE => 1};
+our $TT_OPTIONS = { ABSOLUTE => 1 };
 our $template = Template->new($TT_OPTIONS);
 
 our $ADD_COMPUTERS_BIN = '/usr/local/pf/bin/impacket-addcomputer';
@@ -43,7 +43,7 @@ sub run {
     my $result = `$cmd`;
     my $code = $? >> 8;
 
-    return ($code , $result);
+    return ($code, $result);
 }
 
 =head2 test_join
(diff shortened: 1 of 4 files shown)


lib/pf/domain.pm
```diff
@@ -101,7 +113,50 @@ sub escape_bind_user_string {
...
+    my $ret = "";
+    my ($dns_name, $ou) = @_;
 
+    my $base_dn = generate_base_dn($dns_name);
+
+    # for OU=Computer or OU="", we put the machine account to CN=Computers.
+    if (!defined($ou)) {
+        $ou = ""
+    }
+    $ou =~ s/^\s+|\s+$//g;
```
fdurand: duplicate code

lib/pf/domain.pm
```diff
@@ -101,7 +113,50 @@ sub escape_bind_user_string {
...
+    $ou =~ s/^['"]|['"]$//g;
```
fdurand: duplicated code

PR #8156, opened 2024-05-16
fix an exception when nt key cache is enabled but no max login attempt was defined

stgmsa:
# Description
fix an exception when NT key cache is enabled and no max password attempts is defined, parsing old_password_allowed_peroid is ignored issue.

# Impacts
fix an exception when NT key cache is enabled and no max password attempts is defined, parsing old_password_allowed_peroid is ignored issue.
without this fix, an specific authentication will fail.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

### DIFF
--- a/bin/pyntlm_auth/config_loader.py
+++ b/bin/pyntlm_auth/config_loader.py
@@ -168,6 +168,16 @@ def config_load():
                 print(f"  NT key cache: expire value '{nt_key_cache_expire}' too large, set to maximum value: 864000")
                 nt_key_cache_expire = 864000
 
+            ad_old_password_allowed_period, error = get_int_value(ad_old_password_allowed_period)
+            if error is not None:
+                print(f"  NT Key cache: unable to parse 'ad_old_password_allowed_period', cache disabled.")
+                nt_key_cache_enabled = False
+                break
+            if ad_old_password_allowed_period < 0 or ad_old_password_allowed_period > 99999:
+                print(f"  NT Key cache: 'ad_old_password_allowed_period' ranges from 0..99999, cache disabled.")
+                nt_key_cache_enabled = False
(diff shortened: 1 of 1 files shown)


bin/pyntlm_auth/config_loader.py
```diff
@@ -168,6 +168,16 @@ def config_load():
                 print(f"  NT key cache: expire value '{nt_key_cache_expire}' too large, set to maximum value: 864000")
                 nt_key_cache_expire = 864000
 
+            ad_old_password_allowed_period, error = get_int_value(ad_old_password_allowed_period)
+            if error is not None:
+                print(f"  NT Key cache: unable to parse 'ad_old_password_allowed_period', cache disabled.")
+                nt_key_cache_enabled = False
+                break
+            if ad_old_password_allowed_period < 0 or ad_old_password_allowed_period > 99999:
```
satkunas, line 176: Are these values inclusive? `<= 0 or >= 99999` ?

bin/pyntlm_auth/config_loader.py
```diff
@@ -168,6 +168,16 @@ def config_load():
...
+                print(f"  NT Key cache: 'ad_old_password_allowed_period' ranges from 0..99999, cache disabled.")
+                nt_key_cache_enabled = False
+                break
```
stgmsa, line 179: 0 and 99999 are valid ad_old_password_allowed_peroid. 
if it's set to 0, no old password will be accepted, changing password will immeidately revoke access of old password. 
while 99999 means old password will still valid for 99999 minutes. but no admin will gonna do that in production environment.

satkunas:
@stgmsa please backport

PR #8179, opened 2024-06-12
Feature/osquery

stgmsa:
# Description
adds osquery support to PacketFence

# Impacts
added new authentication method - http basic auth when handling fleetdm api calls.
added fleetdm roles by default
added fleetdm api handlers
added fleetdm event task processors
added new event types related to fleetdm policy / cve
added default isolation template for fleetdm policy / cve

# Delete branch after merge
YES

# Checklist
- [yes] Document the feature
- [x] Add unit tests


docs/PacketFence_Upgrade_Guide.asciidoc
```diff
@@ -1926,6 +1926,19 @@ Since 13.2 PacketFence has reworked the Cisco, Juniper and Meraki switch modules
 /usr/local/pf/addons/upgrade/to-13.2-convert-merakiswitch-types.pl
 ----
 
+== Upgrading from a version prior to 14.0.0
+
+
+=== Admin Role
+
+Since 13.2 PacketFence is able to receive events from the FleetDM servers, which allows PacketFence to detect policy violations or CVEs of devices managed by FleetDM. To add a new admin role to receive these events through the PacketFence API perform the following steps:
```
fdurand: 14.0 instead

docs/images/fleetdm-webhook-policy.jpg
fdurand, line 1: /api/v14/ doesn't look to be correct

docs/installation/provisioner/fleetdm.asciidoc
```diff
@@ -0,0 +1,84 @@
...
+    See PacketFence_Installation_Guide.asciidoc
+    for authors, copyright and license information.
+
+////
+
+
+==== Background
+
+In order to provide flexible osquery support, regular vulnerability check and policy compliance, we introduced basic support for FleetDM.
+From PacketFence v14.0, you can config your FleetDM server to report Policy violations and CVE vulnerabilities of managed nodes to PacketFence.
```
fdurand: you can configure ...

go/caddy/fleetdm/fleetdm.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+var cfg = &fleetDMCfgS{}
+var FDMToken = &CachedTokenS{
+	Token:     "",
+	ExpiresAt: time.Now(),
+}
+
+func init() {
+	//c := GetFleetDMConfig(context.Background())
```
fdurand: remove please

lib/pf/task/fleetdm.pm
```diff
@@ -0,0 +1,393 @@
...
+pf::task::fleetdm
+
+=cut
+
+use strict;
+use warnings;
+use pf::config qw(%Config);
+use base 'pf::task';
+use JSON;
+use HTTP::Tiny;
```
fdurand, line 20: Looks to be a new lib, most of the time curl is used in the code.

conf/chi.conf.defaults
```diff
@@ -217,3 +217,7 @@ storage=distributed
 #
 # Amount of time from the current time to expire an entry
 expires_in=5s
+
+[namespace fleetdm]
+storage=distributed
```
fdurand, line 222: Is it necessary to be distributed ?

fdurand (changes requested):

satkunas (commented):
Remove the need to prefix fleetdm payloads in the email templates with [% notes %], so that [% notes.policy.id %] becomes [% policy.id %], along with all the other variables.

conf/chi.conf.defaults
stgmsa: it would be better.

we only caches the fleetdm hostid <-> packetfence device mac mapping. 
and 
fleetdm token (if they use a username + password auth when calling fleetdm APIs) - but I don't think this will be a common solution for fleetdm users. I think most of them will use an API token instead.

a cross-node cache will avoid unnecessary fleetdm API calls no matter on which node "security_event_trigger" falls on

lib/pf/task/fleetdm.pm
stgmsa: I've notice we are using LWP::UserAgent in many times, 
also several occurance of "WWW::Curl::Easy" 

since HTTP::Tiny is also built-in ones, shall we switch to `LWP` or `WWW::Curl::Easy` ?

PR #8205, opened 2024-07-03
Fix/dedicated machine account for cluster

stgmsa:
# Cluster support for domain.conf 
* allow PacketFence nodes in a cluster have their own dedicated settings in domain.conf
* adds HA support when establishing secure connections.

# Impacts
* changed the domain.conf structure - For now the section name will always be `hostname domain_id`
* changed the steps when init secure connection - perform a dns query and obtain all available domain controllers, choose one to connect.

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

## Enhancements
* allow different settings in domain.conf for each nodes in a cluster. e.g., each node has its own machine account name and password.
* supports domain controller high availibity now with a SRV query of DNS

### DIFF
--- a/addons/upgrade/to-14.0-update-domain-config-section.pl
+++ b/addons/upgrade/to-14.0-update-domain-config-section.pl
@@ -56,7 +56,7 @@ Inverse inc. <info@inverse.ca>
 
 =head1 COPYRIGHT
 
-Copyright (C) 2005-2023 Inverse inc.
+Copyright (C) 2005-2024 Inverse inc.
 
 =head1 LICENSE
 
--- a/addons/upgrade/to-14.0-update-domain-config-section.pl
+++ b/addons/upgrade/to-14.0-update-domain-config-section.pl
@@ -1,0 +1,79 @@
+#!/usr/bin/perl
+
+=head1 NAME
+
+addons/upgrade/to-14.0-update-domain-config-section.pl
+
+=cut
+
+=head1 DESCRIPTION
+
+changing section name from `domain_identifier` to `hostname domain_identifier`
+so PF cluster node can have individual config values.
+
+=cut
+
+use strict;
+use warnings;
+use lib qw(/usr/local/pf/lib /usr/local/pf/lib_perl/lib/perl5);
+use pf::IniFiles;
+use pf::file_paths qw($domain_config_file);
+use pf::util;
+use Sys::Hostname;
+
+my $ini = pf::IniFiles->new(-file => $domain_config_file, -allowempty => 1);
+
+unless (defined $ini) {
+    print("Error loading domain config file. Terminated. Try re-run this script or edit domain settings in admin UI manually. /\n");
+    exit;
+}
+
+my $updated = 0;
+my $hostname = hostname();
+
+for my $section (grep {/^\S+/} $ini->Sections()) {
+    print("Processing section '$section' in domain.conf: ");
+    if ($section =~ /^[a-zA-Z0-9\-\._]+ [a-zA-Z0-9]+$/) {
+        print("already up to date\n");
+    }
+    else {
+        my $new_section_name = "$hostname $section";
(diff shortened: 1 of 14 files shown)


addons/upgrade/to-14.0-update-domain-config-section.pl
```diff
@@ -0,0 +1,79 @@
...
+
+
+
+=head1 AUTHOR
+
+Inverse inc. <info@inverse.ca>
+
+=head1 COPYRIGHT
+
+Copyright (C) 2005-2023 Inverse inc.
```
fdurand: 2024

lib/pf/UnifiedApi/Controller/Config/Domains.pm
```diff
@@ -33,14 +33,75 @@ use Socket;
...
 
+sub get {
+    my ($self) = @_;
+    my $item = $self->item;
+    if ($item) {
+        $item->{id} =~ s/$host_id //i;
+        $item = $self->cleanupItemForGet($item);
+        return $self->render(json => { item => $item }, status => 200);
+    }
+    return $self->render_error(500, "Unknown error getting item");;
```
fdurand: double ;;

bin/pyntlm_auth/utils.py
```diff
@@ -53,6 +53,44 @@ def dns_lookup(hostname, dns_server):
         return "", str(e)
 
 
+def find_ldap_servers(domain, dns_server):
+    query_name = f'_ldap._tcp.{domain}'
```
fdurand: We should use _ldap._tcp.dc._msdcs. instead

lib/pf/UnifiedApi/Controller/Config/Domains.pm
```diff
@@ -33,14 +33,75 @@ use Socket;
...
+}
+
+sub item_shown {
+    my ($self, $item) = @_;
+    if ($item->{id} =~ s/$host_id //i) {
+        return 1;
```
fdurand: Use constant $TRUE

lib/pf/UnifiedApi/Controller/Config/Domains.pm
```diff
@@ -33,14 +33,75 @@ use Socket;
...
+    }
+    return 0;
```
fdurand: Use constant $FALSE

fdurand (changes requested):

leonardosecci:
Hello,

I have seen the changes made, but in my opinion, they could create problems in the event that a domain controller becomes unavailable.
In fact, if such a scenario occurs, it would be queried cyclically, generating an error.
This scenario is not uncommon in a production environment, as domain controllers are often taken down for maintenance, but this is not evident just by reading the SRV record from DNS.
For example, Samba, using the net.finddc method, first retrieves the list of domain controllers reading SRV record but returns the information of the first one that passes the operational check.

Another issue that concerns me is the performance in a production environment.
I presume that because Samba is not declared compatible with multi-threading, you have implemented a lock that, in addition to atomicizing access to global variables, effectively serializes all requests.
Given my limited knowledge of Samba, it seems that there could be compatibility with a multi-process architecture. Wouldn't it be simpler to integrate a pre-fork system, similar to Gunicorn?

Best regards

satkunas:
@leonardosecci thank you for your feedback

We currently handle fail-over, and LB - which isn't useful due to SMB single-thread. 

We have multi-threading on our road-map that will monitor alive/dead of all hosts from SRV DNS and manage a pool used for multi-threaded LB.

leonardosecci:
Thank you @satkunas, do you have an idea when the update that will monitor the health of domain controllers and for multi-threading will be released?

Best regards

satkunas:
No ETA, but we are likely to continue to prioritizing work on Domains in the future

PR #8324, opened 2024-09-25
Backport trusted domain support to PF v13.2

stgmsa:
# Description
Backport Trusted Domain support to PF v13.2

# Impacts
Changed NTLM Auth work flow. 
When there's a "domain" key in HTTP payload, use the given domain value instead of default one.
which is the default behavior of Samba

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

### DIFF
--- a/bin/pyntlm_auth/handlers.py
+++ b/bin/pyntlm_auth/handlers.py
@@ -102,6 +102,11 @@ def ntlm_auth_handler():
         challenge = data['challenge']
         nt_response = data['nt-response']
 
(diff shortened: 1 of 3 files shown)


stgmsa:
@fdurand and @lzammit 
Please create patch and test with a PF v13.2 installation. 

ntlm_auth_wrapper needs be recompiled 
ntlm_auth_api docker image needs to be rebuilt

PR #8356, opened 2024-10-23
Feature/improve dpsk

fdurand:
# Description
Improved the logic to find the associated user with the PSK 

# Impacts
- RADIUS authorize workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* DPSK improvement on heavy number of RADIUS requests

### DIFF
--- a/db/upgrade-X.X-X.Y.sql
+++ b/db/upgrade-X.X-X.Y.sql
@@ -68,7 +68,7 @@ ALTER TABLE `pki_certs`
 ALTER TABLE radius_audit_log MODIFY created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
 
 \! echo "Make psk unique";
-ALTER TABLE person ADD CONSTRAINT person_psk (`psk`);
+ALTER TABLE person ADD CONSTRAINT UNIQUE person_psk (`psk`);
 
 \! echo "Incrementing PacketFence schema version...";
 INSERT IGNORE INTO pf_version (id, version, created_at) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION), NOW());
--- a/db/upgrade-X.X-X.Y.sql
+++ b/db/upgrade-X.X-X.Y.sql
@@ -67,6 +67,9 @@ ALTER TABLE `pki_certs`
 \! echo "Adding default timestamp to RADIUS audit logs";
 ALTER TABLE radius_audit_log MODIFY created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
 
+\! echo "Make psk unique";
+ALTER TABLE person ADD CONSTRAINT person_psk (`psk`);
+
 \! echo "Incrementing PacketFence schema version...";
 INSERT IGNORE INTO pf_version (id, version, created_at) VALUES (@VERSION_INT, CONCAT_WS('.', @MAJOR_VERSION, @MINOR_VERSION), NOW());
 
--- a/db/pf-schema-X.Y.sql
+++ b/db/pf-schema-X.Y.sql
@@ -79,7 +79,8 @@ CREATE TABLE person (
   `potd` enum('no','yes') NOT NULL DEFAULT 'no',
   `otp` MEDIUMTEXT NULL DEFAULT NULL,
   `sponsored_date` DATETIME DEFAULT NULL,
-  PRIMARY KEY (`pid`)
+  PRIMARY KEY (`pid`),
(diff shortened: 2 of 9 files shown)


E-Thang:
Would it be possible to add RADIUS/VLAN filter option for owner.psk? There is owner.password, but it doesn't appear to match on the PSK.

fdurand:
> Would it be possible to add RADIUS/VLAN filter option for owner.psk? There is owner.password, but it doesn't appear to match on the PSK.

You mean it doesn't appear in the admin UI rule form ?

E-Thang:
> > Would it be possible to add RADIUS/VLAN filter option for owner.psk? There is owner.password, but it doesn't appear to match on the PSK.
> 
> You mean it doesn't appear in the admin UI rule form ?

The admin UI Filter Engines VLAN and RADIUS filters have **owner.password** as a filter criterion. However, a VLAN filter that has the sole condition **`owner.password =~ "^.+$"`** doesn't obtain any hits for a node that is owned by a local user created with both a password and a PSK. I also tried with **`defined(owner.password, "")`** and tried in a RADIUS filter as well, still no hits. I changed it to only filter on the SSID and it obtains hits, confirming that the filter is otherwise functional.

I wasn't sure if **owner.password** refers to the user's local password or the PSK, but in either case, the filter criteria **owner.password** doesn't seem to work. My inclination is that it refers to the node owner's actual password, not the PSK. I can create a separate bug report for the **owner.password** issue if you like. I mention it here since you are working on improving the DPSK feature and this is tangentially related to that topic.

It would seem that there should either be a new filter criterion **owner.psk** to go alongside **owner.password**, or **owner.password** should be renamed to **owner.psk** if it is supposed to refer to the node owner's PSK and not the password.

db/upgrade-X.X-X.Y.sql
```diff
@@ -67,6 +67,9 @@ ALTER TABLE `pki_certs`
 \! echo "Adding default timestamp to RADIUS audit logs";
 ALTER TABLE radius_audit_log MODIFY created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
 
+\! echo "Make psk unique";
+ALTER TABLE person ADD CONSTRAINT UNIQUE person_psk (`psk`);
```
satkunas, line 71: This statement will fail if there's more than one row with the same psk. Before this statement we should first empty it to avoid duplicates: `UPDATE users SET psk=NULL WHERE 1;`

fdurand: Yes but i can't empty it first since the psk are actively used

fdurand:
> > > Would it be possible to add RADIUS/VLAN filter option for owner.psk? There is owner.password, but it doesn't appear to match on the PSK.
> > 
> > 
> > You mean it doesn't appear in the admin UI rule form ?
> 
> The admin UI Filter Engines VLAN and RADIUS filters have **owner.password** as a filter criterion. However, a VLAN filter that has the sole condition **`owner.password =~ "^.+$"`** doesn't obtain any hits for a node that is owned by a local user created with both a password and a PSK. I also tried with **`defined(owner.password, "")`** and tried in a RADIUS filter as well, still no hits. I changed it to only filter on the SSID and it obtains hits, confirming that the filter is otherwise functional.
> 
> I wasn't sure if **owner.password** refers to the user's local password or the PSK, but in either case, the filter criteria **owner.password** doesn't seem to work. My inclination is that it refers to the node owner's actual password, not the PSK. I can create a separate bug report for the **owner.password** issue if you like. I mention it here since you are working on improving the DPSK feature and this is tangentially related to that topic.
> 
> It would seem that there should either be a new filter criterion **owner.psk** to go alongside **owner.password**, or **owner.password** should be renamed to **owner.psk** if it is supposed to refer to the node owner's PSK and not the password.

In fact owner.password doesn't exists in the owner hash so we should just remove it.
But owner.psk is part of the hash but visible in the admin UI.
I did a commit in devel to add the missing attributes https://github.com/inverse-inc/packetfence/commit/d5e5952c9c00a669007edca0f2564c08f1fbae28 , i just have to remove owner.password then.

E-Thang:
> 
> In fact owner.password doesn't exists in the owner hash so we should just remove it. But owner.psk is part of the hash but visible in the admin UI. I did a commit in devel to add the missing attributes [d5e5952](https://github.com/inverse-inc/packetfence/commit/d5e5952c9c00a669007edca0f2564c08f1fbae28) , i just have to remove owner.password then.

Perfect, thank you!

satkunas:
Waiting on tests with client with @fdurand

PR #8366, opened 2024-10-26
Change useOpenLdap.js to case-insensitive attribute name indices

### LINKED ISSUE #8088: LDAP Conditions not working for LDAP authentication Source
**Describe the bug**
I have two authentication sources. One is LDAP and the other is Active Directory.
They are both configured the same:
- Same LDAP account
- Same AD/LDAP server over port 636 with SSL

I can successfully authenticate only using the LDAP auth source. The AD one does not seem to work for authentication for some reason.
Whilst the AD auth source is able to setup LDAP conditions, the LDAP source does not.
When searching for a LDAP attribute using LDAP auth source, nothing shows up.

**Steps to reproduce the behavior:**
1. Create a new authentication source using the type LDAP.
2. Setup LDAP and make sure to test the authentication.
3. Add a new authentication rule and try to provide a LDAP condition.
4. It does not provide a list of attributes:
ldap

5. However, in the Active Directory auth source, the LDAP condition attributes are shown like they should:
AD

**Expected behavior**
Being able to use LDAP conditions using the LDAP authentication source.

**Tested on**
 - Two different PF ZEN installations (13.1.0)
 - PF on Debian 11 installation (13.1.0)
 - PF ZEN (13.2.0)

E-Thang:
Changed useOpenLdap.js to case-insensitive attribute name indices, set base_dn for getSubSchemaDN to null, and added explicit limits to sendLdapSearchRequest function calls.

# Description
Fixes issue with case mismatch of subSchemaSubEntry, subSchema, and attributeTypes names on various LDAP servers. This resulted in AD LDAP and RedHat Directory389 LDAP (And probably others) not able to use LDAP Authentication Source conditions.

# Impacts
AD LDAP and RedHat Directory389 schema download

# Delete branch after merge
YES

# Checklist
- [ N/A] Document the feature
- [ N/A] Add OpenAPI specification
- [ N/A] Add unit tests
- [N/A ] Add acceptance tests (TestLink)

## Bug Fixes
Partially fixes #8088 "LDAP Conditions not working for LDAP authentication Source"
There's another unresolved bug that remains for that issue though. see: https://github.com/inverse-inc/packetfence/issues/8088#issuecomment-2434450924

### DIFF
--- a/html/pfappserver/root/src/views/Configuration/sources/_components/ldapCondition/useOpenLdap.js
+++ b/html/pfappserver/root/src/views/Configuration/sources/_components/ldapCondition/useOpenLdap.js
@@ -10,7 +10,7 @@ import {
 function useOpenLdap(form) {
 
   const performSearch = (filter, scope, attributes, base_dn) => {
-    return sendLdapSearchRequest({...form.value}, filter, scope, attributes, base_dn)
+    return sendLdapSearchRequest({...form.value}, filter, scope, attributes, base_dn, 1000)
       .then((result) => {
           return {results: parseLdapResponseToAttributeArray(result, extractAttributeFromFilter(filter)), success: true}
         }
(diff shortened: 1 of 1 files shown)


satkunas:
@E-ThanG thank you for your contribution

I will need to apply these changes to type=AD as well.

E-Thang:
> @E-ThanG thank you for your contribution
> 
> I will need to apply these changes to type=AD as well.

I considered that, but in my testing AD wasn't case sensitive. In fact, the returned attribute case seemed to change depending on the case of my query. It certainly wouldn't hurt to also have it there though.

PR #8448, opened 2024-12-19
Fix NTLM auth API missing logs

### LINKED ISSUE #8414: Not enough logs available with NTLM Auth API
**Describe the bug**
Improve logging for both success and failures to help with debugging.

stgmsa:
# Description
Fix missing NTLM auth API logs
added several common error cases for easier troubleshooting.

# Impacts
* changed logging handler, from directly using print to built-in logging component.
* added several hints when specific NTError occurred.

# Issue
fixes #8414 

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [no] Document the feature
- [no] Add OpenAPI specification
- [no] Add unit tests

## Enhancements
added more detailed logs of ntlm-auth-api
added several common NTErrors

## Bug Fixes
* issue with not enough logs in ntlm-auth-api for troubleshooting (#8414)

### DIFF
--- a/bin/pyntlm_auth/config_loader.py
+++ b/bin/pyntlm_auth/config_loader.py
@@ -8,6 +8,7 @@ from configparser import ConfigParser
 import config_generator
 import global_vars
 import utils
+import log
 
 
 def get_boolean_value(v):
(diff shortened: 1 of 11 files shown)


bin/pyntlm_auth/log.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # file_handler = logging.FileHandler("app.log")
+    # file_handler.setLevel(default_logging_level)
+
+    formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s', datefmt="%Y-%m-%d %H:%M:%S")
+    console_handler.setFormatter(formatter)
+    # file_handler.setFormatter(formatter)
+
+    logger.addHandler(console_handler)
+    # logger.addHandler(file_handler)
```
satkunas: Cleanup

bin/pyntlm_auth/log.py
```diff
@@ -0,0 +1,46 @@
+import logging
+
+
+def init_logging():
+    default_logging_level = logging.DEBUG
```
JeGoi, line 15: How will you handle something different than the level DEBUG?
Why not using the env, or a config file?

bin/pyntlm_auth/config_loader.py
```diff
@@ -38,45 +39,51 @@ def get_int_value(v):
...
+    log.debug(f"Load database config from {conf_db}")
     try:
         with open(conf_db, 'r') as file:
             cp_db.read_file(file)
         if 'DB' not in cp_db:
-            print(f"  Section [DB] not found, ntlm-auth-api starts without NT Key caching capability.")
+            log.warning(f"Section [DB] not found, ntlm-auth-api starts without NT Key caching capability.")
     except FileNotFoundError:
-        print(f"  {conf_db} not found, ntlm-auth-api@{identifier} starts without NT Key caching capability.")
+        log.warning(f"{conf_db} not found, ntlm-auth-api@{identifier} starts without NT Key caching capability.")
```
JeGoi, line 84: To follow the previous way, it should be critical.

bin/pyntlm_auth/config_loader.py
```diff
@@ -193,59 +199,59 @@ def config_load():
...
-                print(f"  NT Key cache: '{s_device}' ranges from 0..999, set to '{s_threshold}' by default.")
+                log.warning(f"  NT Key cache: '{s_device}' ranges from 0..999, set to '{s_threshold}' by default.")
                 break
             if max_allowed_attempts_per_device > ad_account_lockout_threshold:
-                print(f"  NT Key cache: '{s_device}' larger than '{s_threshold}', set to '{s_threshold}' by default.")
+                log.warning(f"  NT Key cache: '{s_device}' larger than '{s_threshold}', set to '{s_threshold}' by default.")
 
     if None in (c_db_host, c_db_port, c_db_user, c_db_pass, c_db, c_db_unix_socket):
-        print(f"  DB config: Missing settings, NT Key cache will be disabled")
+        log.warning(f"  DB config: Missing settings, NT Key cache will be disabled")
```
JeGoi, line 246: This one should be a log.error no?
It will not allow to enable the cache.

bin/pyntlm_auth/rpc.py
```diff
@@ -94,22 +95,31 @@ def init_secure_connection():
...
-        print(f"  workstation: {workstation}")
-        print(f"  username: {username}")
-        print(f"  password: {utils.mask_password(password)}")
-        print(f"  set_NT_hash_flag: True")
-        print(f"  domain: {domain}")
-        print(f"  server_name(ad_fqdn): {server_name}")
+        log.warning(f"Error in init secure connection: NTError: {hex(error_code)}, {error_message}.")
+
+        if error_code == 0xc0000001:
+            log.warning("Did you give the wrong 'workstation' parameter in domain configuration ?")
```
JeGoi: Why not using the other logger option levels? like error?

bin/pyntlm_auth/log.py
stgmsa: the default log level is set to debug so everything will be logged.
we have other logs levels like critical / warning / error... , they are used something unexpected happens.

putting default log level to an env file makes sense. but administrators might need to manually modify the config file to chage log levels (No Admin UI options). Is there a best practise ?

bin/pyntlm_auth/config_loader.py
stgmsa: I would suggest no. 
This can be a warning level because NTLM auth API is still able to authenticate any type of accounts.
The only difference is caching capability is not available.

bin/pyntlm_auth/config_loader.py
stgmsa: similar to the previous one, cache disabled. but still works.
I don't think this is an error or critical scenario.
they will only see the message during ntlm auth api starts.

bin/pyntlm_auth/log.py
JeGoi: Here is an example in config file:
https://github.com/inverse-inc/packetfence/blob/devel/conf/log.conf.d/httpd.portal.conf.example#L3
I think anything is good as long as it is documented.
Most of the time debug is used from the backend for specific use case. I am not sure if there is anything related to that in the front end.
So, making it changeable at least from backend (then doc) is a good start, no?

JeGoi: Here is an example in config file:
https://github.com/inverse-inc/packetfence/blob/devel/conf/log.conf.d/httpd.portal.conf.example#L3
I think anything is good as long as it is documented.
Most of the time debug is used from the backend for specific use case. I am not sure if there is anything related to that in the front end.
So, yes, making it changeable even from backend (then doc) is a good start, no?

stgmsa: yes
mostly for backend. And a changable + default makes sense.

PR #8761, opened 2025-09-10
docs update

satkunas:
# Description
Docs update using Claude

# Delete branch after merge
YES


JeGoi:
Are you sure about the [appendix] and it's content when the == Appendix is defined just after with a file?

JeGoi:
Like here https://github.com/inverse-inc/packetfence/blob/feature/docs-update-2025/docs/installation/appendix.asciidoc?plain=1#L21 and now also here https://github.com/inverse-inc/packetfence/blob/5a2d1753022d1077bf27eecad663c6a9cdaf688e/docs/PacketFence_Installation_Guide.asciidoc?plain=1#L162

JeGoi (changes requested):
I think there is an error with [appendix] and the 
```
== Appendix
include::cluster/appendix.asciidoc[]
```
It needs to be fixed before merged

PR #8866, opened 2026-01-05
Send the job in the queue

fdurand:
# Description
Queue generateAnsibleConfiguration job in order for the api to reply quickly.

# Impacts
NO

# Issue
Fix the delay to create a role

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* Fix role delay creation

### DIFF
--- a/lib/pf/api.pm
+++ b/lib/pf/api.pm
@@ -1967,7 +1967,7 @@ This is called after role configuration changes to avoid blocking the API.
 
 =cut
 
-sub generate_ansible_configuration_all_switches : Public {
+sub generate_ansible_configuration_all_switches : Public : Fork {
     my ($class, %postdata) = @_;
     my $logger = get_logger();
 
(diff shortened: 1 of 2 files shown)


lib/pf/api.pm
```diff
@@ -1959,6 +1960,41 @@ sub push_acls : Public {
...
 }
 
+=head2 generate_ansible_configuration_all_switches
+
+Generate Ansible configuration for all switches as a background job.
+This is called after role configuration changes to avoid blocking the API.
+
+=cut
+
+sub generate_ansible_configuration_all_switches : Public {
```
jrouzierinverse: We should also have this fork since it could take a long time to finish since it is sleeping in the loop.
You can add the `: Fork` attribute to the function.

jrouzierinverse (changes requested):

PR #8908, opened 2026-02-04
Better detection of the unifi controller and rely on the cookie to reconnect.

fdurand:
# Description
Reworked the connection code in the Unifi module in order to detect correctly the old and new controller.
Removed the cache usage and just rely on the cookie.

# Impacts
Ubiquiti webauth reevaluation

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Better detection between old and new unifi controller and rely on the cookie to handle re-connection.

### DIFF
--- a/lib/pf/Switch/Ubiquiti/Unifi.pm
+++ b/lib/pf/Switch/Ubiquiti/Unifi.pm
@@ -268,6 +268,10 @@ sub _deauthenticateMacWithHTTP {
 
     my $response = $ua->get("$base_url/api/self/sites");
 
+    if ($response->code == 404) {
+        $response = $ua->get("$base_url/proxy/network/api/self/sites");
+    }
+
     unless($response->is_success) {
         $logger->error("Can't have the site list from the Unifi controller: ".$response->status_line);
         return;
@@ -384,12 +388,14 @@ sub populateAccessPointMACIP {
 
     my $response = $ua->get("$base_url/api/self/sites");
 
+    if ($response->code == 404) {
+        $response = $ua->get("$base_url/proxy/network/api/self/sites");
+    }
+
     unless($response->is_success) {
         $logger->error("Can't have the site list from the Unifi controller: ".$response->status_line);
         return;
     }
-
-    $response = $ua->get("$base_url/api/self/sites");
 
     my $json_data = decode_json($response->decoded_content());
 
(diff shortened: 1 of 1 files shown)


lib/pf/Switch/Ubiquiti/Unifi.pm
```diff
@@ -200,26 +200,29 @@ sub _connect {
     my $api_prefix = "";
 
     my $response = $ua->get($base_url."/proxy/network/status");
+    my $cookie_invalid = $FALSE;
 
-    if ($response->code == 401) {
+    if ($response->code == 401 || $response->is_success) {
```
codedude, line 205: Duplicated line 209 and 215, put one out of the if/else. You can then remove declaration at line 203.

lib/pf/Switch/Ubiquiti/Unifi.pm
```diff
@@ -200,26 +200,29 @@ sub _connect {
...
-    if (!defined($auth) || $auth == $FALSE) {
+    if ($cookie_invalid) {
+        # Clear old cookies before re-authenticating to avoid sending invalid session
+        $ua->cookie_jar->clear();
         $response = $ua->post($base_url.$login_path, Content => '{"username":"'.$username.'", "password":"'.$password.'", "remember": "true"}');
 
         unless($response->is_success) {
             $logger->error("Can't login on the Unifi controller: ".$response->status_line);
-            $cache->set("Ubiquiti-" . $controllerIp ."-auth" , $FALSE );
             die;
```
codedude, line 224: Cobra Kai Never Dies

codedude (changes requested):

lib/pf/Switch/Ubiquiti/Unifi.pm
codedude: You can even get rid of $cookie_invalid, and cache "$authorized = $response->code != 401"

fdurand: Hum line 215 i can't remove it since $response is coming from line 214

codedude: Y, so my second comment does not apply, but the first one y. $cookie_invalid can be defined once after the if/else.

PR #8928, opened 2026-02-20
INV-232 - generate material json for a new packetfence site CI flow

takrasiu-inv:
https://guardicore.atlassian.net/browse/INV-232

# Description
The new website [packetfence.com ](https://www.packetfence.com/self-host/compatible-switches) displays a table which relies on JSON. Right now, there is only 1 way of updating it, it's a running a JS script which will traverse a packetfence repo "/lib/pf/Switch" to grep all switches a build the corresponding JSON structure. Instead, this change will re-use existing logic of generating switches for old side, and adding a new script to generate JSON and then deliver it to new [site](https://github.com/akainverse/website-pfcom) to `/data/switches.json`. 
The site has own github action, which will be triggered and update site public files.

The targe branch for CI is `ci-release` so this branch should be created before. The commit message will look like:
```
Automatic update by pipeline ${CI_PIPELINE_ID} 2026-02-23 11:23:35
```

# Impacts
- `.gitlab.ci` -> build_artifacts_material_job
- `generate-material.sh`
- `publish-to-git.sh`
- `switch_options_json.pl` new file
- `switch_options_table.pl` deleted file, not used
- `switch_options.pl` deleted, not used

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [x] Create a separate script for generating switches JSON.
- [x] Delete unused code and files which will not be used anymore.
- [x] Update, don't include switch Templates, since they don't support ACL

# NEWS file entries
- `switch_options_json.pl` new file


JeGoi (changes requested):
As discussed.
Create a new branch in order to not pollute main in website repo.
Use branch name and timestamp.

Also as you proposed, make the cleaning as much as possible for extra files that are not used anymore (perl script, bash script etc).
Thanks a lot

JeGoi (approved):
Looks ok for me.
Should we wait for GC, or merge this one and create a new PR later?

PR #8939, opened 2026-03-04
fix(services): fix _unitFileExists() for Docker and fix log string interpolation

JeGoi:
# Description
Fix `_unitFileExists()` method in the services manager to work inside Docker containers, and fix broken string interpolation in log messages.

The `systemctl cat` command fails inside Docker containers because it attempts to read unit files from disk. This replaces it with `systemctl show -p LoadState` which works via the D-Bus socket and is compatible with Docker environments.

Additionally, Perl log messages using `"$self->name ..."` do not interpolate method calls correctly. This fixes them to use proper concatenation: `$self->name . " ..."`.

# Impacts
- `lib/pf/services/manager.pm`: `_unitFileExists()` now uses `systemctl show` instead of `systemctl cat`
- `lib/pf/services/manager.pm`: Fixed string interpolation in `postStartCleanup` and `postStopCleanup` log messages
- `lib/pf/services/manager/netdata.pm`: Fixed string interpolation in `postStartCleanup` log message

# Issue
fixes #INV-196

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests

# NEWS file entries
## Bug Fixes
* Fixed `_unitFileExists()` failing inside Docker containers by using `systemctl show` instead of `systemctl cat` (INV-196)
* Fixed broken Perl string interpolation in service manager log messages

### DIFF
--- a/lib/pf/services/manager.pm
+++ b/lib/pf/services/manager.pm
@@ -173,7 +173,7 @@ sub postStartCleanup {
     my ($self,$quick) = @_;
     my $logger = get_logger();
     unless ($self->pid) { 
-        $logger->error("$self->name died or has failed to start");
+        $logger->error($self->name . " died or has failed to start");
         return $FALSE;
     }
     return $TRUE;
@@ -346,7 +346,7 @@ sub postStopCleanup {
     my ($self,$quick) = @_;
     my $logger = get_logger();
     if ($self->pid) { 
-        $logger->error("$self->name failed to stop");
+        $logger->error($self->name . " failed to stop");
         return $FALSE;
     }
     return $TRUE;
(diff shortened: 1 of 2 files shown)


JeGoi:
For now pf15.0 and devel you are not able to use api and UI to stop a service

JeGoi:
Looks fine https://gitlab.com/inverse-inc/packetfence/-/pipelines/2363629159
Tests that are failing are the same as in devel and not related to the current pr.
When I wrote the comment, the fingerbank_invalid_db test failure is a classic.

JeGoi:
Backported to maintenance/15.0

PR #8962, opened 2026-03-25
fix/8951

### LINKED ISSUE #8951: SSO portaltoken using wrong address
`api-frontend` uses public FQDN for internal portal token validation instead of the configured service URL
Component: `go/api-frontend/aaa/portal_authentication_backend.go`

**Problem:**

When validating an SSO portal token, `api-frontend` constructs the callback URL using the system FQDN:

```
https://<hostname>.domain.tld/portaltoken?token=<uuid>
```

PacketFence already has a correct abstraction for internal service-to-service communication in `pf.conf.defaults:`

```
[services_url]
httpd_portal=[% ENV.env_or_default("PF_SERVICES_URL_HTTPD_PORTAL", "http://containers-gateway.internal:8080") %]
```

However, `portal_authentication_backend.go` ignores this configuration and uses the FQDN instead.

**Why this should fail in any containerized deployment — not just mine:**

Docker assigns each container its own IP and sets up `/etc/hosts` inside the container with `-h <hostname>`, mapping the system hostname to the container's own IP. This is standard Docker behavior, not a misconfiguration.

As a result:

`api-frontend` container has IP e.g. `100.64.0.12`
Inside `api-frontend`, `<hostname>` resolves to `100.64.0.12` (itself)
`api-frontend` calls `https://100.64.0.12:443/portaltoken` — its own IP
Nothing listens on 443 in `api-frontend` → connection refused
SSO login returns "Wasn't able to authenticate those credentials"
This will reproduce on any containerized PacketFence installation where each service runs in its own container with its own network namespace. It is not topology-specific.

**Expected behavior:**

`portal_authentication_backend.go` should use `[services_url].httpd_portal` from `pf.conf` (or the `PF_SERVICES_URL_HTTPD_PORTAL` env var) to reach `httpd.portal`, consistent with how other services communicate internally.

**Workaround:**

Add `--add-host=<hostname>:<gateway_ip>` to the `api-frontend` container start arguments so the FQDN resolves to the host/gateway instead of the container itself.

jrouzierinverse:
# Description
Use the proper url for the portal token

# Issue
fixes #8951

# Delete branch after merge
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## Bug Fixes
* SSO portaltoken using wrong address

### DIFF
--- a/go/api-frontend/aaa/portal_authentication_backend.go
+++ b/go/api-frontend/aaa/portal_authentication_backend.go
@@ -33,6 +33,7 @@ func (pab *PortalAuthenticationBackend) Authenticate(ctx context.Context, userna
 
 	req, err := http.NewRequestWithContext(ctx, "GET", pab.url.String()+"?token="+password, nil)
 	sharedutils.CheckError(err)
+	req.Header.Set("X-Forwarded-For-PacketFence", "127.0.0.1")
 	resp, err := pab.httpClient.Do(req)
 	if err != nil {
 		return false, nil, err
--- a/go/plugin/caddy2/api-aaa/api-aaa.go
+++ b/go/plugin/caddy2/api-aaa/api-aaa.go
@@ -182,7 +182,7 @@ func (h *ApiAAAHandler) buildApiAAAHandler(ctx context.Context) error {
 
 	// Backend for SSO
 	if sharedutils.IsEnabled(adminLogin.SSOStatus) {
-		url, err := url.Parse(fmt.Sprintf("%s%s", adminLogin.SSOBaseUrl, adminLogin.SSOAuthorizePath))
+		url, err := url.Parse(fmt.Sprintf("%s%s", servicesURL.HttpdPortal, adminLogin.SSOAuthorizePath))
 		sharedutils.CheckError(err)
 		h.authentication.AddAuthenticationBackend(aaa.NewPortalAuthenticationBackend(ctx, url, false))
 	}


JeGoi (approved):
Looks a perfect simple fix

BUT in fact it is not enough.

The problem here is that the resolution is done on the browser side not on the server.

JeGoi:
I tested that without the internal rewriting things and it works.

JeGoi:
For the admin sso.

ON PF CONFIG:
01-config

02-config

03-config

04-config

ON GUI SIDE
05-gui

06-gui

07-gui

08-gui

On that step and on the first time you came in, a qrcode is generated.

09-gui

And then you get: https://xxxx.inverse.ca/admin-sso?callback=https://xxxx.inverse.ca:1443/admin#/login

Then, now you are correctly redirected:

"https://xxxx.inverse.ca:1443/admin#/login?token=<generated_temporary_token>"

10-gui

et voilà!

JeGoi (approved):

JeGoi:
How to test it:
apply the patch https://patch-diff.githubusercontent.com/raw/inverse-inc/packetfence/pull/8962.diff and rebuild the image.
```
root@jg-deb12-standalone:/usr/local/pf# curl https://patch-diff.githubusercontent.com/raw/inverse-inc/packetfence/pull/8962.diff | patch -p1 --dry-run
root@jg-deb12-standalone:/usr/local/pf# curl https://patch-diff.githubusercontent.com/raw/inverse-inc/packetfence/pull/8962.diff | patch -p1
root@jg-deb12-standalone:/usr/local/pf# ./addons/dev-helpers/build-local-container.sh api-frontend ; systemctl restart packetfence-api-frontend.service
```
After that, api-frontend images should be different:
```
root@jg-deb12-standalone:/usr/local/pf# docker image ls | grep api-frontend
packetfence/api-frontend                                  devel     000a3b3cf8c4   14 seconds ago   2.87GB
ghcr.io/inverse-inc/packetfence/api-frontend              devel     6074c5b5ceb7   3 weeks ago      2.87GB
```
To revert just retag, restart the service check that images are the same for tag and prune docker for extra unused images:
```
root@jg-deb12-standalone:/usr/local/pf# docker tag ghcr.io/inverse-inc/packetfence/api-frontend:devel packetfence/api-frontend:devel ; systemctl restart packetfence-api-frontend.service ; docker image ls | grep api-frontend ; docker system prune -f
```

ArcasCZ:
I just tested out the fix and I can confirm that the patch fixes the problem.

JeGoi:
@ArcasCZ Thanks a lot!

satkunas (approved):

PR #9054, opened 2026-05-15
pfqueue: hash pfdhcplistener_external by client MAC (4 workers)

fdurand:
# Description
  
  Reduces the per-packet work done by the external DHCP listener path (pfdhcplistener_external), which previously enqueued several redundant
  general-queue tasks for every DHCP packet — including the half-lease renewal packets every client sends. Under load (relayed DHCP through
  pfdhcplistener_external) this was the dominant source of pfqueue churn and webservices round-trips.

  The branch makes the listener horizontally scalable and suppresses work whose input hasn't changed:

  * Parallelize the queue — pfdhcplistener_external goes from a single worker to workers=4 hashed=enabled, hashed by client MAC so all
  packets for a given device pin to one worker (a prerequisite for the process-local caches below to be correct).
  * firewallsso — instead of queuing a firewallsso Update on every packet whenever any firewall has sso_on_dhcp enabled, gate it behind a
  process-local CHI cache keyed by (mac, ip, lease_length) with TTL = lease_length / 2. The session stays alive at the half-lease renewal
  cadence without a task per packet. IP changes still do Stop + Start and re-arm the cache for the new IP.
  * fingerbank — skip fingerbank_process when the device's DHCP signature (IPv4 + DHCPv6 fingerprint/vendor fields + computername) is
  unchanged. Classification is a pure function of those fields; the signature is cached per MAC (24h TTL, mirroring [storage fingerbank]).
  Filter-engine rules still run on every packet so their side effects are preserved.
  * trigger_scan — short-circuit when no scan engines are configured, avoiding a queued task that does nothing.
  * apiClient — pin the DHCP processor's apiClient to pf::api::can_fork so its notifies run in-process. Under the modern pfqueue-backend
  worker (which, unlike legacy sbin/pfqueue, never calls setClient), pf::client::getClient was defaulting to jsonrpcclient, turning each
  notify into an HTTPS POST to localhost webservices — 5–7 round-trips per DHCP packet.

# Impacts

  Reviewers should look at:
  * DHCP processing flow (lib/pf/dhcp/processor.pm) — processIPTasks, processFingerbank, and the apiClient attribute default.
  * pfqueue worker model — the pfdhcplistener_external queue now runs 4 hashed workers; the firewallsso/fingerbank caches are process-local
  and only correct because of MAC-based hashing (each MAC pinned to one worker). Verify the hashing config.
  * Firewall SSO behavior — confirm sessions stay alive across renewals and that IP-change does Stop/Start without a redundant Update on the
  same packet.
  * Fingerbank classification — confirm a change in only the DHCPv6 fingerprint/enterprise still triggers re-classification (was previously
  suppressed).
  * can_fork client — confirm :Fork methods (e.g. trigger_scan) still fork as before, and that running notifies in-process under
  pfqueue-backend has no unintended side effects.

# Code / PR Dependencies

  N/A

# NEW Package(s) required

  None.

# Issue

  N/A

# Delete branch after merge

  YES

# Checklist

  - [ ] Document the feature — n/a (internal performance optimization, no user-facing feature)
  - [ ] Add unit tests — no

## NEWS file entries

  Enhancements

  * Parallelized the pfdhcplistener_external pfqueue task (4 workers, hashed by client MAC) to remove the single-worker bottleneck on
  relayed DHCP
  * Reduced redundant general-queue tasks per DHCP packet: firewallsso Update is now gated to once per half-lease, fingerbank classification
  is skipped when the DHCP signature is unchanged, and trigger_scan short-circuits when no scan engines are configured
  * DHCP processor api calls now run in-process under pfqueue-backend instead of issuing HTTPS round-trips to webservices

## Bug Fixes

  * firewallsso no longer queued a redundant Update on the same packet as an IP-change Stop/Start
  * fingerbank classification was incorrectly suppressed when only the DHCPv6 fingerprint/enterprise number changed

### DIFF
--- a/lib/pf/dhcp/processor.pm
+++ b/lib/pf/dhcp/processor.pm
@@ -27,6 +27,7 @@ use pf::config qw(
     %connection_type_to_str
     $INLINE
     %ConfigFirewallSSO
+    %ConfigScan
 );
 use pf::config::util;
 use pf::constants qw($TRUE);
@@ -190,7 +191,12 @@ sub processIPTasks {
     # Conformity scan
     # 2017.03.20 - dwuelfrath@inverse.ca - There is currently no ipv6 support for conformity scan. Remove the condition once "resolved"
     unless ( $iptasks_arguments{'ipversion'} eq $IPV6 ) {
-        $self->apiClient->notify('trigger_scan', %iptasks_arguments );
+        # Mirror the early-return in pf::api::trigger_scan: skip the
+        # enqueue when no scan engines exist so we don't ship a task the
+        # general-queue worker would immediately drop.
+        if (scalar keys %ConfigScan) {
+            $self->apiClient->notify('trigger_scan', %iptasks_arguments );
+        }
     }
 
     # Parking security_event
--- a/lib/pf/dhcp/processor.pm
+++ b/lib/pf/dhcp/processor.pm
@@ -96,7 +96,7 @@ my @local_dhcp_servers_ip;
 # cadence DHCP clients renew on — so the firewall session stays alive without
 # enqueueing an Update on every renewal packet.
 my $sso_refresh_hash = {};
-my $sso_refresh_cache = CHI->new( driver => 'Memory', datastore => $sso_refresh_hash );
+my $sso_refresh_cache = CHI->new( driver => 'RawMemory', datastore => $sso_refresh_hash );
 
 # Tracks the last (dhcp_fingerprint|dhcp_vendor|computername) signature
 # observed per MAC, so fingerbank_process is only enqueued when the device's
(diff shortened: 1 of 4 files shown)


lib/pf/dhcp/processor.pm
```diff
@@ -41,7 +43,21 @@ use Moose;
...
+# class lazily avoids a compile-time circular dep with pf::api (pf::api
+# loads this module, and pf::api::can_fork is already loaded by pf::task::api
+# before any DHCP task runs).
+has 'apiClient'    => (
+    is      => 'ro',
+    default => sub {
+        require pf::api::can_fork;
+        pf::api::can_fork->new;
+    },
+);
```
jrouzierinverse, line 60: Remove this as this is handled in pfqueue

jrouzierinverse (changes requested):

lib/pf/dhcp/processor.pm
fdurand: You're right that it's already done — but only in the legacy worker, which isn't the one running this queue.

PR #9063, opened 2026-05-25
Fix non-deterministic radius filter merge with reply: prefix

fdurand:
# Description
A radius filter answer such as 'reply:Cisco-AVPair = ...' was stored under a different hash key than the bare 'Cisco-AVPair' set by returnRadiusAccessAccept, so handleAnswerInRule never merged them. Both keys then crossed the rlm_rest boundary as duplicates targeting the same RADIUS attribute, producing a non-deterministic Access-Accept: the filter answer randomly disappeared (or displaced an existing Cisco-AVPair slot) depending on hash iteration order.

Normalize the leading 'reply:' prefix and combine values into a single arrayref. Other section prefixes (control:, request:, session-state:) are left untouched so they don't collide with reply attributes.

# Impacts
RADIUS authorize workflow

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Bug Fixes
* If a radius filter return a Cisco-AVPair and the switch module returns another Cisco-AVPair then the 2 hashes are not correctly merged.

### DIFF
--- a/lib/pf/access_filter/radius.pm
+++ b/lib/pf/access_filter/radius.pm
@@ -86,19 +86,33 @@ sub handleAnswerInRule {
             return ($radius_reply, $status);
         }
 
-        foreach my $key (keys %$radius_reply_ref) {
-            if (exists($radius_reply->{$key})) {
-                my $type = reftype($radius_reply->{$key});
-                if (defined($type) && $type eq 'ARRAY') {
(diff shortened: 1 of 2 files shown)


satkunas:
@fdurand backport where needed

PR #9078, opened 2026-06-02
Fixes a pfcmd configreload hang and the slow switch-config admin UI, both rooted in unbounded per-role switch mappings, plus a UX improvement for blocked role deletions.

fdurand:
# Description
  1. pfcmd configreload no longer hangs on large configs

  expire_all/expire recursively hard-expired every namespace's child_resources with no de-duplication. Because shared children (e.g. resource::RolesReverseLookup is a child of 13 namespaces) were rebuilt once per parent, a configreload hard on a large config issued 1000+ redundant
  rebuilds and effectively never returned.

  - lib/pfconfig/manager.pm: thread a per-run %seen set through expire_all/expire so each namespace is rebuilt at most once per run (idempotent — config is static during a run). The new $seen param is optional, so all other callers are unaffected.

  2. Stop orphaned switch role-mappings from accumulating

  Deleting a role never pruned its per-role keys (<role>Vlan, <role>Url, <role>AccessList, …) from switches. Over repeated create/delete cycles, switches.conf accumulated thousands of orphaned mappings, bloating the switch-config API (a single GET /api/v1/config/switch/:id reached
  ~1 MB / ~19 s).

  - lib/pf/UnifiedApi/Controller/Config/Roles.pm: add prune_role_from_switches() (factored out of the existing rename/reassign path) and a post_remove() hook so deleting a role removes its mappings from all switches.
  - tools/prune_orphaned_switch_role_mappings.pl: one-time maintenance script to clean pre-existing orphans (dry-run by default, --commit to apply; preserves roles.conf entries and built-in/reserved roles).

  3. Show which entities still use a role when delete is blocked

  When a role can't be deleted because it's still in use, the 422 reported only the category ("switch") — not which ones. The referencing ids already existed in RolesReverseLookup but were discarded.

  - lib/pf/UnifiedApi/Controller/Config/Roles.pm: config_delete_error now includes a structured ids array and appends the ids to the message (e.g. Role still in use for switch: 10.255.255.2).
  - html/pfappserver/root/src/views/Configuration/roles/_components/TheSearch.vue: the reassign dialog lists those ids per category instead of the raw reason code.

# Impacts
Roles deletion and pfconfig reload

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* pfcmd configreload no longer hangs on large configs
* Stop orphaned switch role-mappings from accumulating
* Show which entities still use a role when delete is blocked

# UPGRADE file entries

### DIFF
--- a/html/pfappserver/root/src/views/Configuration/roles/_components/TheSearch.vue
+++ b/html/pfappserver/root/src/views/Configuration/roles/_components/TheSearch.vue
@@ -118,10 +118,10 @@
         <div class="mx-2">{{ $t('The role could not be deleted. Either manually handle the following errors and try again, or re-reassign the resources to another existing role.') }}</div>
       </b-media>
       <h5>{{ $t('Role is still in use for:') }}</h5>
-      <b-row v-for="error in deleteErrors" :key="error.reason">
-        <b-col cols="auto" class="mr-auto">{{ reasons[error.reason] }}</b-col>
-        <b-col cols="auto">{{ error.reason }}</b-col>
-      </b-row>
+      <div v-for="error in deleteErrors" :key="error.reason" class="mb-2">
+        <strong>{{ reasons[error.reason] || error.reason }}</strong>
+        <span v-if="error.ids && error.ids.length" class="text-secondary">: {{ error.ids.join(', ') }}</span>
+      </div>
       <template v-slot:modal-footer>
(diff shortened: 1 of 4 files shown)


satkunas:
@fdurand backport to 15.x (maybe 14.x?)

PR #9079, opened 2026-06-03
Require AUP and AUP Template in choice portal module

fdurand:
# Description
With the default portal module when a choice module use Sources by Class (dynamic source selection) then there is no way to enable or disable the AUP since there is no toggle and by default it´s enabled in the code.
This PR added the "Require AUP" and "AUP Template" in the choice module.

# Impacts
Portal modules

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries

## Enhancements
* Require AUP and AUP Template option in choice portal module

### DIFF
--- a/html/pfappserver/root/src/views/Configuration/portalModules/_components/FormTypeAuthenticationChoice.vue
+++ b/html/pfappserver/root/src/views/Configuration/portalModules/_components/FormTypeAuthenticationChoice.vue
@@ -29,6 +29,11 @@
       :text="$i18n.t('Require the user to accept the AUP')"
     />
 
+    <form-group-aup-template namespace="aup_template"
+      :column-label="$i18n.t('AUP template')"
+      :text="$i18n.t('The template to use for the Acceptable Use Policy')"
+    />
+
     <form-group-template namespace="template"
       :column-label="$i18n.t('Template')"
     />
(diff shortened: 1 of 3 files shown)


html/pfappserver/root/src/views/Configuration/portalModules/_components/FormTypeAuthenticationChoice.vue
```diff
@@ -71,7 +81,9 @@ import {
   FormGroupMultiSourceAuthClasses,
   FormGroupMultiSourceObjectClasses,
   FormGroupMultiSourceTypes,
-  FormGroupTemplate
+  FormGroupTemplate,
+  FormGroupWithAup,
+  FormGroupAupTemplate
```
satkunas, line 86: @fdurand are these files defined in index.js, or missing?

satkunas (changes requested):

html/pfappserver/root/src/views/Configuration/portalModules/_components/FormTypeAuthenticationChoice.vue
fdurand: It's defined in views/Configuration/portalModules/_components/index.js

PR #9091, opened 2026-06-07
feat(admin): cluster-wide live logs, historical logs, pftest in GUI

Antony1711:
# Description

Adds three cluster-aware troubleshooting features to the admin UI:

1. **Cluster-wide Live Logs** — the existing LiveLogs view now opens one tail session per cluster peer (via `X-PacketFence-Server` + the existing `passthrough_admin.lua` routing) and merges events client-side with a coloured `hostname / filename` tag per line. Standalone installs keep the legacy single-session path.

2. **Historical Log View** — new admin page under the same navbar as LiveLogs that reads from `/usr/local/pf/logs/` (active files + rotated `.log.*.gz`) across every cluster node. Cursor-based pagination, datetime-local pickers with quick-range buttons (1h / 6h / 24h / 7d), filename allow-list sourced from `resource::syslog_files`.

3. **pftest in the GUI** — `Auditing → pftest` runs `pftest authentication` and `pftest profile_filter` across all cluster peers in parallel (`Mojo::IOLoop->subprocess` + `Mojo::Promise->all`, 25–30s per-peer timeout). Output rendered in a terminal-style panel with the same colour palette `pf::util::console::colors` uses on the CLI. Audit log redacts the password via the existing `fields_to_mask` mechanism.

# Impacts

Reviewers should focus on:

- `lib/pf/UnifiedApi/Controller/Logs/{History,ClusterHistory}.pm` and `lib/pf/UnifiedApi/Controller/Pftest{,/Cluster}.pm` — new controllers, cluster fan-out mirrors `Services::ClusterStatuses` and `Services.pm:130-173` (`Mojo::IOLoop->subprocess` + `Mojo::Promise->all`).
- `lib/pf/UnifiedApi.pm` — three new `setup_api_v1_*_routes` subs, all registered from `setup_api_v1_routes`.
- `sbin/pfperl-api-docker-wrapper` — adds `-v /usr/local/pf/logs:/usr/local/pf/logs:ro` next to the existing lib/html mounts; the new History controller reads from that path.
- `html/pfappserver/root/src/views/{LiveLogs,HistoricalLogs,Auditing/pftest}/` — Vue 2 / Vuex modules following the existing `_router.js`/`_api.js`/`_store/`/`_components/` layout. LiveLogs additions are conditional on `cluster/isCluster` so single-node installs are unaffected.

Verified end-to-end on a 3-node active/active cluster and on the standalone fallback path.

The merged `docs/api/spec/openapi.{yaml,json}` are intentionally not regenerated in this PR — only the `docs/api/spec/static/` sources are touched, so the diff stays reviewable. CI or a maintainer can regenerate via `docs/api/spec/generate-openapi-spec.pl`.

# Delete branch after merge
YES

# Checklist
- [x] Document the feature
- [x] Add OpenAPI specification
- [x] Add unit tests
- [ ] Add acceptance tests (TestLink) — not in scope of this PR

# NEWS file entries
## New Features
* Cluster-wide live log streaming in admin UI
* Historical log view across cluster nodes (.gz-aware)
* pftest authentication and profile_filter exposed under Auditing in the admin UI


lib/pf/UnifiedApi/Controller/Logs/History.pm
```diff
@@ -0,0 +1,331 @@
...
+    my $now_ms = int(time() * 1000);
+    if (!%$cursor && !defined $start_ms) {
+        $start_ms = $now_ms - 24 * 60 * 60 * 1000;
+    }
+    $end_ms //= $now_ms + 60_000;
+
+    my $matcher;
+    if (defined $filter && length $filter) {
+        if ($is_regex) {
+            my $re = eval { qr/$filter/ };
```
satkunas: **ReDoS (blocking).** `qr/$filter/` compiles an arbitrary client-supplied regex and the resulting matcher runs against every line of every selected file across the default 24h window (including all `.gz` rotations). Perl regex has catastrophic backtracking, so a pattern like `(a+)+$` against a long line hangs this `pfperl-api` worker — an authenticated remote DoS, amplified because the cluster controller fans the same body out to every peer.

Note the irony: the Go `log-tailer` this duplicates uses `regexp.MustCompile` (RE2), which is immune to this. Please bound matching with a timeout (e.g. an `alarm`/`Time::HiRes` watchdog) and/or reject patterns containing nested quantifiers.

lib/pf/UnifiedApi/Controller/Logs/History.pm
```diff
@@ -0,0 +1,331 @@
...
+        my $floor_ms = $cursor->{$name} // $start_ms;
+        my @sources  = _enumerate_sources($name, $floor_ms, $end_ms);
+        my $last_ts_ms;   # carry forward for multi-line / continuation events
+        my $last_meta;
+
+        for my $src (@sources) {
+            my $fh = _open_source($src);
+            next unless $fh;
+            while (my $line = <$fh>) {
+                last FILE if @events >= MAX_EVENTS;
```
satkunas: **Unbounded scan.** `MAX_EVENTS` caps the number of events *returned*, not the work *done*: a filter that matches nothing still walks and gzip-decompresses the entire 24h window of every selected file synchronously in the worker before returning. Combined with the cluster fan-out this is an easy resource-exhaustion lever for any `SYSTEM`-level admin. Consider a wall-clock / bytes-scanned ceiling in addition to the event cap.

lib/pf/UnifiedApi/Controller/Logs/History.pm
```diff
@@ -0,0 +1,331 @@
...
+                            hostname           => $meta->{hostname},
+                            process            => $meta->{process},
+                            syslog_name        => $meta->{syslog_name},
+                            log_level          => $meta->{log_level},
+                            filename           => $name,
+                            log_without_prefix => $meta->{log_without_prefix},
+                        },
+                    },
+                };
+                $new_cursor{$name} = $ts_ms;
```
satkunas: **Cursor boundary can drop events.** The cursor advances with `$new_cursor{$name} = $ts_ms` and the next page filters `next if $ts_ms <= $floor_ms` (line 160). When more than one line shares the same millisecond — common with sub-second logging, and guaranteed for lines with no fractional part — and a 500-event page boundary falls inside that millisecond, the remaining same-ms lines are silently skipped on the next poll. The Go tailer avoids this by tracking byte/file position rather than timestamp. For paginated historical reads this is silent data loss at page edges.

lib/pf/UnifiedApi/Controller/Logs/History.pm
```diff
@@ -0,0 +1,331 @@
...
+
+=head1 NAME
+
+pf::UnifiedApi::Controller::Logs::History -
+
+=cut
+
+=head1 DESCRIPTION
+
+Reads historical log content from /usr/local/pf/logs/ (active file plus
```
satkunas: **Architectural duplication (please justify).** This controller reimplements `go/plugin/caddy2/log-tailer/` in Perl: `$LINE_RE`/`_parse_line` ↔ `logmeta.go`, `%LEVEL_MAP` ↔ the Go level mapping, `_enumerate_sources` (`.log.N.gz` enumeration) ↔ the Go rotation logic, and the `filter`/`filter_is_regexp` handling ↔ `handlers.go:84-91`. Two parsers for the same log format in two languages will drift over time. If the goal is time-windowed/offset reads, extending the existing Go tailer would avoid the fork — or at minimum, please note in the PR description why a separate Perl implementation was chosen.

lib/pf/UnifiedApi/Controller/Pftest/Cluster.pm
```diff
@@ -0,0 +1,130 @@
...
+        }
+        return $self->render(json => {
+            items => [{
+                host => pf::cluster::get_host_id() // 'localhost',
+                %{$resp->{json}},
+            }],
+        });
+    }
+
+    my @servers = pf::cluster::enabled_servers();
```
satkunas: **Auth amplification.** Fanning the request out to every enabled peer means a single `pftest authentication` call produces N real authentication attempts against the backend directory (LDAP/AD/RADIUS) — one per node. That can trip account-lockout policies and looks like a distributed brute-force from the directory's perspective. Please confirm this is intended; consider gating behind a more specific admin role than the implicit `SYSTEM_CREATE` and adding rate limiting.

lib/pf/UnifiedApi/Controller/Pftest/Cluster.pm
```diff
@@ -0,0 +1,130 @@
...
+
+sub authentication { $_[0]->_dispatch('authentication') }
+sub profile_filter { $_[0]->_dispatch('profile_filter') }
+
+sub _dispatch {
+    my ($self, $action) = @_;
+    my $body = $self->req->json // {};
+
+    if (!$pf::cluster::cluster_enabled) {
+        my $local = pf::UnifiedApi::Controller::Pftest->new(%$self);
```
satkunas: **Fragile controller construction.** `pf::UnifiedApi::Controller::Pftest->new(%$self)` shallow-copies the controller hash (tx, stash, match, …) by reference purely to reach `_run_$action`. Since `_run_authentication`/`_run_profile_filter` are effectively pure functions of `$body`, making them package subs (or using `$self->app->build_controller`) would be cleaner and avoid sharing mutable request state. Same pattern in `Logs/ClusterHistory.pm:37`.

html/pfappserver/root/src/views/Auditing/pftest/_router.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+  next()
+}
+
+export default [
+  {
+    path: 'pftest',
+    name: 'pftest',
+    component: TheView,
+    meta: { can: 'read services' },
```
satkunas: **Permission mismatch.** This declares `can: 'read services'` (`SERVICES_READ`), but the backend has no `pathAdminRolesMap` entry for `/api/v1/pftest/*`, so it falls through to the `SYSTEM` default → POST requires `SYSTEM_CREATE` (`go/api-frontend/aaa/authorization.go:231-233`). A user with `SERVICES_READ` but not `SYSTEM_CREATE` will see this page and the form, then get a 403 on submit. Please reconcile the frontend `can` with the role the backend actually enforces (and add the explicit ACL entry).

html/pfappserver/root/src/views/HistoricalLogs/_store/session.js
```diff
@@ -0,0 +1,205 @@
+/**
+* "$_historical_log" store module — per-session state for one historical
+* query. The shape mirrors LiveLogs/_store/session.js (so TheView can reuse
```
satkunas, line 3: **Copy-paste store (please factor out).** As the comment notes, this mirrors `LiveLogs/_store/session.js` — ~40% of the lines are identical. The `loadClusterConfig()` helper in this view's `_router.js` is also byte-identical to the LiveLogs one. Extracting the shared session-store logic and the cluster-config preload into a common module would prevent the two from drifting on every future fix.

html/pfappserver/root/src/views/LiveLogs/_api.js
```diff
@@ -10,32 +10,49 @@ const prefix = () => {
...
   },
-  delete: id => {
+  delete: (id, server) => {
     if (isSaas()) {
       return Promise.resolve({})
     }
-    return apiCall.delete([prefix(), 'tail', id]).then(response => {
-      return response.data
-    })
+    // 404 = session already gone; treat as success and mark response quiet so the interceptor stays silent.
```
satkunas: **Behavior change worth flagging.** Treating 404 as success (and marking the response `quiet`) is reasonable for the "session already reaped" race, but it now also silently swallows a 404 caused by a wrong session id or a request landing on the wrong cluster node — the operator gets no signal. Consider distinguishing "already gone" from "never existed here" if that turns out to matter in practice.

satkunas (changes requested):
Thorough, well-commented feature with genuine attention to edge cases (cluster fan-out, standalone fallback, password masking, read-only log mount). I verified there is **no shell injection, path traversal, XSS, or plaintext-password persistence** — those are handled correctly (see "verified-good" below). The blocking concern is a remote-DoS regex; the rest are an architectural duplication, a permission inconsistency, and some correctness edges. Inline comments anchor each finding.

## Must fix before merge
- **ReDoS in `Logs/History.pm`** — a client-supplied regex is compiled and run against every line of every selected file over the default 24h window (incl. all `.gz`). A pathological pattern hangs the `pfperl-api` worker (authenticated remote DoS, amplified by cluster fan-out). The Go tailer this duplicates uses RE2 and is immune. See inline.

## Should fix
- **Server-side ACL gap:** `/api/v1/pftest/*` has **no entry** in `go/api-frontend/aaa/authorization.go` `pathAdminRolesMap` (that file isn't touched by this PR, so I can't inline it). It's protected only by the implicit `→ SYSTEM` fall-through default (`authorization.go:231-233`). Please add an explicit prefix mapping like every other feature, so the protection is intentional rather than incidental.
- **Frontend/backend permission mismatch:** the pftest GUI route declares `can: 'read services'` (`SERVICES_READ`) but the backend resolves to `SYSTEM_CREATE`. A `SERVICES_READ`-only admin sees the UI then gets 403 on submit. Align both sides. See inline.
- **pftest auth amplification:** the cluster variant fans one auth attempt out to every peer in parallel = N real LDAP/AD/RADIUS attempts per call, which can trip account lockouts and resembles distributed brute force. Please confirm intended; consider a dedicated role + rate limiting. See inline.
- **Unbounded scan:** `MAX_EVENTS` caps output, not work scanned — a non-matching filter decompresses 24h of `.gz` synchronously. Add a time/size ceiling. See inline.

## Architectural / tech debt (please justify or address)
- **`Logs/History.pm` reimplements `go/plugin/caddy2/log-tailer/` in Perl** — line parsing, level mapping, `.gz` rotation enumeration, and filter handling now exist twice in two languages and will drift. Extending the Go tailer with a time-window mode would avoid the fork. At minimum, please explain in the PR description why a parallel Perl implementation was chosen. See inline.
- **`HistoricalLogs/_store/session.js` is ~40% identical to `LiveLogs/_store/session.js`** (the header comment says "mirrors"), and `loadClusterConfig()` is byte-identical across both `_router.js` files. Factor into a shared module. See inline.

## Correctness / side effects (non-blocking)
- **Cursor boundary data loss** in `History.pm`: when >1 line shares a millisecond and a 500-event page ends mid-millisecond, the remaining same-ms lines are dropped on the next page. See inline.
- **LiveLogs rewrite has no frontend tests:** every standalone/SaaS session now flows through the new group/peer machinery. Guards look correct, but it's a behavior change for all users with zero automated coverage.
- **`_api.js` `delete` now swallows 404 as silent success** — reasonable for the "already gone" race, but also hides genuinely-wrong-id/wrong-node 404s. See inline.
- **`new(%$self)` controller construction** in both `Cluster` controllers is fragile (copies tx/stash by reference); the `_query`/`_run_*` helpers are pure functions of the body and would be cleaner as package subs. See inline.

## Verified-good (not issues) 👍
Password masking via `fields_to_mask` (request-only audit, `output_raw` never logged); syslog-name allow-list checked before any path is built (no traversal); `escapeHtml` applied before highlight (no XSS); read-only `/usr/local/pf/logs` mount; in-process allow-listed pftest invocation (no shell). Also +1 on shipping unit tests and the OpenAPI static sources.

Antony1711:
Hi satkunas,

Thank you for the thorough review and feedback, as well as for the time you've taken to provide it. I will do my best to address all the outstanding issues. If you have any further concerns or questions, please don’t hesitate to contact me.

King Regards

satkunas:
> Hi satkunas,
> 
> Thank you for the thorough review and feedback, as well as for the time you've taken to provide it. I will do my best to address all the outstanding issues. If you have any further concerns or questions, please don’t hesitate to contact me.
> 
> King Regards

Thank you for your contributions @Antony1711

lib/pf/UnifiedApi/Controller/Logs/History.pm
Antony1711: Fixed at the root rather than bounded: the historical-log backend moved into the Go log-tailer plugin as you suggested (see the architecture reply below), so the filter is now `regexp.Compile` (RE2) — catastrophic backtracking is impossible by construction. Invalid patterns return 422 and filters are capped at 256 chars. Verified on the cluster: `(a+)+$` over a full 24h window returns in <0.3s.

lib/pf/UnifiedApi/Controller/Logs/History.pm
Antony1711: The Go implementation bounds the work, not just the output: a 5s wall-clock budget and a 256MB decompressed-byte ceiling on top of the 500-event cap. When a bound is hit the response carries `truncated: true` and a cursor that advanced over the scanned (also non-matching) regions, so the client resumes losslessly instead of rescanning. Rotations whose mtime cannot intersect the requested window are skipped without being decompressed.

lib/pf/UnifiedApi/Controller/Logs/History.pm
Antony1711: Implemented exactly the fix you pointed at: cursors are now byte/file positions (rotation index + uncompressed offset + a content signature), not timestamps, so pages can never drop lines that share a millisecond. The signature pins a fixed hashed prefix length (append-only files keep it stable), finds renamed sources after logrotate so the offset survives rotation, and falls back to the timestamp floor only when a rotation was pruned by retention. Regression tests cover same-millisecond runs across page boundaries, rotation between polls, empty-file cycles and continuation lines at page edges (`go/plugin/caddy2/log-tailer/history_test.go`).

lib/pf/UnifiedApi/Controller/Logs/History.pm
Antony1711: You were right that extending the Go tailer was the better shape — we did that instead of justifying the fork. `go/plugin/caddy2/log-tailer/history.go` adds a stateless time-window mode (POST/OPTIONS `/api/v1/logs/history`) to the existing plugin; both `History.pm` and `ClusterHistory.pm` are deleted, along with their routes, tests and the pfperl-api logs bind-mount. Line parsing and level mapping exist exactly once: `logmeta.go` gained an ISO-8601 branch shared by live tail and history — which also fixes a latent gap, since the rsyslog templates write ISO timestamps that `GlobalExtractionRe` never matched (live-tail meta was silently empty). The `/logs/cluster_history` fan-out endpoint disappears too: the frontend queries each node directly via the `X-PacketFence-Server` routing the cluster live logs already use. This also removes the second `->new(%$self)` instance you flagged.

lib/pf/UnifiedApi/Controller/Pftest/Cluster.pm
Antony1711: Not intended as a default — fixed threefold: (1) fan-out is now opt-in via a `cluster: true` body flag, surfaced in the GUI as an explicit "Run on all cluster nodes" checkbox with an account-lockout warning; default is the local node only. (2) authentication runs are rate-limited per tested user (one run per node per 10s window, redis-backed, HTTP 429) — checked at the cluster entry point before any dispatch and again on each node's public endpoint as defense in depth. (3) the endpoints sit behind the new dedicated PFTEST admin role (see the `_router.js` permission reply). `profile_filter` performs no authentication and is not limited.

lib/pf/UnifiedApi/Controller/Pftest/Cluster.pm
Antony1711: Fixed: the `pftest` runners are now package subs (pure functions of the request body), so neither the direct endpoint nor the cluster wrapper constructs a controller via `->new(%$self)` — there is no controller hash to shallow-copy and no shared tx/stash state to leak. The cluster fan-out calls the same package subs directly (commit `refactor(api): pftest runners as package subs, drop fragile ->new(%$self)`).

html/pfappserver/root/src/views/Auditing/pftest/_router.js
Antony1711: Aligned both sides on `PFTEST_CREATE`.

Backend: added the explicit mapping with the more specific role you implied — `adminRoleMapping{prefix: apiPrefix + "/pftest/", role: "PFTEST"}` in `authorization.go`, plus new `PFTEST_READ`/`PFTEST_CREATE` actions in `@ADMIN_ACTIONS` (the admin-roles form and the frontend ACL derivation pick them up automatically). `authorization_test.go` asserts `PFTEST_CREATE` is authorized on all four endpoints and that `SERVICES_READ` and `SYSTEM_CREATE` are now denied — the protection is intentional, not fall-through.

Frontend: the route now declares `can: 'create pftest'` (= `PFTEST_CREATE`, what the backend enforces on POST), matched on the Auditing nav item and the Auditing parent guard (the create verb could not ride the existing `$some('read', …)` list, so the guard became `$some(read, [logs]) || $can('create', 'pftest')`). A `SERVICES_READ`-only admin no longer sees the page; a `PFTEST_CREATE`-only admin can use it (the source picker degrades to empty without `USERS_SOURCES_READ` — the call was already quiet/catch-guarded).

html/pfappserver/root/src/views/HistoricalLogs/_store/session.js
Antony1711: Extracted: `src/utils/logEvents.js` now carries the shared scope construction, addMeta/delMeta counting, the isFiltered/eventsFiltered getter bodies and the filter recomputation; `src/utils/cluster.js` carries `loadClusterConfig()` used by both log routers (and now the pftest router too). Both stores unified on the sorting `addMeta` variant. While unifying we also fixed a Vue 2 reactivity bug: LiveLogs' `LOG_FILTER_ENABLE` used plain assignment on a key that does not exist before the first toggle.

html/pfappserver/root/src/views/LiveLogs/_api.js
Antony1711: Kept the idempotent-stop behavior (the server reaps idle sessions after 5 minutes, so the race is routine), but the result now carries `gone: true` and a `console.debug` breadcrumb with the session id and peer, so a wrong-id/wrong-node 404 stays diagnosable. The backend returns a plain 404 either way, so a stronger distinction is not available client-side.

Antony1711:
Thanks for the thorough review. The branch has been updated and re-tested on a 3-node cluster; replies to each inline thread are inline. A few points that did not map to a single line, plus the follow-ups from cluster testing:

### Frontend tests for the LiveLogs rewrite

Fair point — pfappserver currently has no frontend test infrastructure at all (no jest/vitest, no spec files, no test script), and introducing a framework felt out of scope for this PR; we would rather see that land as its own change. What we did instead: the historical-log logic that used to be untestable Perl now has a Go test suite (window/meta extraction, lossless pagination across same-millisecond runs, rotation between polls, allow-list/traversal rejects, budget truncation), and the shared frontend helpers extracted in the dedup commit are pure functions, ready for unit tests the day the project adopts a runner.

### Verified-good items preserved across the migration

The filename allow-list is enforced in Go before any path is built (basenames only, membership in the `syslog_files` resource, resolution strictly under `/usr/local/pf/logs` — with traversal tests; note the pre-existing live-tail session create did NOT validate paths upstream, the new endpoint does not share that). escapeHtml-before-highlight stays as-is, pftest stays in-process and allow-listed, password masking via `fields_to_mask` is untouched (route-level, verified). On the mount: the pfperl-api `:ro` logs mount this PR added is reverted with the Perl controllers; the api-frontend already mounted `/usr/local/pf/logs` upstream (rw) — we tightened that existing mount to `:ro` since only rsyslogd writes there. Two drive-by hygiene fixes in the existing tail handler: `regexp.MustCompile` on user input (panic → now 422) and a missing `return` after the empty-files 422 (a session was started anyway).

Antony1711:
Hey Satkunas, 
I hope our revisions meet your expectations and now satisfy all requirements. We access CLI logs daily for troubleshooting purposes, so having this feature implemented would be a huge benefit for us—and surely for many other administrators as well. We look forward to your feedback.

satkunas:
## Security review (follow-up)

Focused security pass over the diff (XSS, path traversal, command injection, authz, password handling, cluster SSRF). **No exploitable vulnerabilities — no security blockers to merge.** The three concerns from my earlier review (ReDoS, pftest ACL gap, auth amplification) are all resolved.

Two minor, non-blocking notes on surfaces this PR adds:

1. **pftest output echo** — pftest stdout is now returned over HTTP. If an auth source ever echoed the submitted password into its status message it would land in `output`/`output_raw`. Theoretical (PF's sources don't), but scrubbing the password substring from `output` would close it for good.
2. **`PFTEST_READ` unused** — all routes require `PFTEST_CREATE`; either wire it to a read-only view or drop it.

Antony1711:
Hi Satkunas,

Thank you for the follow-up review.
I will address the few remaining open points as soon as possible.

Antony1711:
Both notes are now addressed:

**1. pftest output echo** — the response now scrubs the submitted password from both `output` and `output_raw`, guarded on `length` so an empty password is a no-op. The cluster fan-out inherits it, since each node redacts its own output.

**2. `PFTEST_READ` unused** — dropped it. pftest is execute-only (every route is a POST → `PFTEST_CREATE`), so a `_READ` action would stay dead. Removed from `admin_roles.pm` and the `adminroles.conf.example` sample.

Re-tested on the 3-node cluster. Thanks again for the thorough review!

With both follow-up points resolved and no remaining blockers, is the PR good to merge now? Happy to address anything else if needed.

satkunas:
@Antony1711 thank you for your contribution

PR #9094, opened 2026-06-08
fix(security): replace shell command execution with safe_pf_run

jrouzierinverse:
## Summary

Replaces shell-interpolated command execution (`pf_run`, backticks, `qx`, `system`, piped `open`) with `safe_pf_run`, which runs commands via `open3` with an explicit argument list — no shell, so no command-injection or glob/word-splitting surprises. Also rewrites a couple of shell pipelines in pure Perl where that was cleaner than spawning a process at all.

## Changes

- **Command execution hardening** — convert call sites across the codebase to `safe_pf_run(BIN, @ARGS, \%opts)`:
  - `addons/AD/migrate.pl`, `addons/upgrade/to-13.1-move-ntlm-auth-to-rest.pl`, `bin/cluster/sync`, `bin/pfcmd_vlan`
  - `lib/pf/services/manager.pm` and managers (`pf`, `tracking`, `fingerbank_collector`) — `systemctl` start/stop/enable/disable/show/is-active
  - `lib/pf/ipset_cache.pm`, `lib/pf/iptables.pm` — feed `ipset restore` from a temp file via `stdin` instead of a shell pipe
  - `lib/pfconfig/git_storage.pm` — git pull/push/add/commit/rm and conf copies (globs expanded in Perl, `working_directory` instead of `cd &&`)
  - `lib/pfconfig/namespaces/config/Pf.pm` — `timedatectl set-timezone`
  - `sbin/pf-mariadb` — `mysqld_safe` launched with an argument list
- **Pure-Perl rewrites (no subprocess):**
  - `lib/pf/Switch/Xirrus.pm` — `Digest::MD5::md5_hex` instead of `printf … | md5sum`
  - `lib/pf/domain.pm` — parse AD config in Perl instead of shelling out
  - `lib/pf/ssl.pm` — `verify_chain`/`modulus_md5` without `/bin/bash -c` process substitution; tests in `t/unittest/ssl.t`
- **Robustness fixes** (latest commit):
  - `git_storage` pull/push and `pf-mariadb` `mysqld_safe`: merge stderr into stdout so `safe_pf_run` drains the child's output rather than leaving stderr undrained, which could fill the pipe buffer and hang `waitpid`.
  - `services::manager::isAlive`: accept exit codes 1–4 from `systemctl is-active` so inactive/failed services on this hot path aren't logged as command failures (matches `print_status`).

## Test plan

- `t/unittest/ssl.t` covers the rewritten chain verification.
- Manual verification of service start/stop/status, ipset/iptables rule loading, and pfconfig git sync recommended before merge.

### DIFF
--- a/addons/AD/migrate.pl
+++ b/addons/AD/migrate.pl
@@ -29,17 +29,29 @@ BEGIN {
 }
 
 use pf::util;
+use pf::util::dns;
 use pf::domain;
 use pf::ConfigStore::Domain;
 
-my $REALM = pf_run('grep default_realm /etc/krb5.conf | awk \'{print $3}\'');
-chomp($REALM);
-my $WORKGROUP = pf_run('grep workgroup /etc/samba/smb.conf | awk \'{print $3}\'');
-chomp($WORKGROUP);
-my $SERVER = pf_run('grep admin_server /etc/krb5.conf | head -1 | awk \'{print $3}\'');
-chomp($SERVER);
-my $NAMESERVER = pf_run('grep nameserver /etc/resolv.conf | head -1 | awk \'{print $2}\'');
-chomp($NAMESERVER);
+# Return the value of the first "<key> = <value>" line in $file, or '' if none.
+# Replaces the old `grep <key> <file> | awk '{print $3}'` shell pipelines.
+sub _first_config_value {
+    my ($file, $key) = @_;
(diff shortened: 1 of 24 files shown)


codedude:
Are you able to provide some exploits about this vulnerability ? So we can understand the issue and prove it's fixed, thx! I don't know how to test it

PR #9136, opened 2026-06-30
fix(backup): reapply DB triggers on import and carry them in the exportable backup

JeGoi:
# Description
(REQUIRED)

Restoring a PacketFence DB silently dropped all triggers, so the restored
instance stopped cleaning up related rows — most visibly, deleting a user no
longer removed its `password` row (re-creating the user then failed with `409
Conflict`), and the `ip4log`/`ip6log`/`locationlog` history triggers were lost.

Root cause: `mysqldump` runs as the `pf` user, which lacks the `TRIGGER`
privilege, so the dump has no triggers. On import the DB is dropped, recreated
and reloaded from that trigger-less dump, and the triggers were never reapplied.

Fix:
- **Reapply triggers on import.** `import_mysqldump` reapplies triggers when the
  dump has none — from `triggers.sql` in the export, falling back to the matching
  `pf-schema-<version>.sql`. Idempotent, and runs before `upgrade_database`.
- **Self-contained export.** `export.sh` (already root) also dumps `triggers.sql`
  (triggers only). Routines already ride in the base dump; there are no DB events.
- **Replace `--db-restore` with `--restore-as-is`.** A composable non-interactive
  modifier (with `--db`/`--conf`/full) that skips upgrades and keeps the backup's
  values.

# Impacts
(REQUIRED)
- `export.sh` emits `triggers.sql` alongside `grants.sql`.
- `database.functions` (`import_mysqldump`) reapplies triggers when the dump lacks
  them.
- `import.sh` drops `--db-restore`, adds `--restore-as-is`.

# Backport
(REQUIRED)
Bug was masked by cross-version upgrades but real for same-version restore, a
standard workflow since `exportable-backup` (14.1). Affected: **14.1 → 15.1**.
Backport to at least **maintenance/15.1**.

# Delete branch after merge
(REQUIRED)
YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [x] Document the feature — n/a (documented in the script's `--help`)
- [x] Add unit tests — n/a (shell backup/import path)
- [x] Add acceptance tests — venom `backup_db_and_restore` checks the export
  carries triggers and that triggers/routines survive the restore, and covers the
  `--restore-as-is` db/conf/full restores, each restarting PacketFence
  (`pfcmd service pf start`) so the API is reachable after.

# NEWS file entries
## Bug Fixes
* Database import/restore no longer drops triggers, fixing orphaned `password`
  rows after a user is deleted on restored databases
## Enhancements
* The exportable backup now includes the database triggers so a restore is
  self-contained; `import.sh` gains `--restore-as-is` for non-interactive imports

# UPGRADE file entries
Databases restored with older code may be missing triggers. Re-import with the
fixed code, or reapply from `/usr/local/pf/db/pf-schema-<version>.sql` as root.

### DIFF
--- a/addons/full-import/export.sh
+++ b/addons/full-import/export.sh
@@ -55,8 +55,17 @@ if echo "$last_db_dump" | grep '\.sql.gz$' >/dev/null; then
     mariadb_args="$mariadb_args -p$mariadb_root_pass"
   fi
 
+  db_name=$(/usr/local/pf/bin/get_pf_conf database db)
+  db_name=${db_name:-pf}
+
   echo "Database dump uses mysqldump. Exporting the grants from the database. Enter the MariaDB root password if prompted to"
   mysql $mariadb_args --skip-column-names -A -e"SELECT CONCAT('SHOW GRANTS FOR ''',user,'''@''',host,''';') FROM mysql.user WHERE user<>'' AND user<>'PUBLIC'" | mysql $mariadb_args --skip-column-names -A | sed 's/$/;/g' > grants.sql
+
+  # pf-user dump omits triggers (no TRIGGER privilege); dump them as root.
+  # Routines are skipped: the base pf dump already carries them.
+  echo "Exporting the triggers from the database"
+  mysqldump $mariadb_args --single-transaction --no-create-info --no-data --skip-add-drop-table \
+    --triggers "$db_name" > triggers.sql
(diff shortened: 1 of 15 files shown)


JeGoi:
There is an issue with the ignored table and a restore-as-is where these tables are missing

PR #9179, opened 2026-07-20
Switch role pool (Filter-Id pooling) via username hash

Antony1711:
## Summary

Adds pooling for the *Switch Role* mapping, mirroring the existing VLAN pool but for the per-switch role (`${role}Role`) that is returned as a role attribute such as `Filter-Id`. A `${role}Role` may now hold a comma-separated list of roles (e.g. `role_a,role_b,role_c`); PacketFence returns a single member per client, so role-based enforcement can spread devices evenly across several networks.

This addresses an open customer requirement (ticket 0011256): a WLAN controller enforces network segmentation through `Filter-Id`, and a large number of BYOD devices must be distributed evenly across several networks without a client hopping between them while roaming.

## Behaviour

Selection is a **deterministic hash** of the node owner (username), falling back to the device MAC when there is no username (MAC auth / machine auth / unowned nodes). Because it is a pure function of that key:

- a given user (or device) is mapped to the **same role on every re-authentication and while roaming** between APs;
- different users/devices **spread across the pool**;
- CoA / deauth re-enforcement picks the **same member** as the initial Access-Accept.

Only the deterministic username-hash technique is offered for role pools (unlike VLAN pools, which also offer random / round-robin): those are not roaming-stable for role strings, whose assigned value is not persisted.

## Changes

- `pf::role::pool`: `getRoleFromPool()` + `_isRolePool()` / `_poolMembers()`. The username-hash index is factored out into `_usernameHashIndex()` and shared with the existing VLAN `getVlanByUsername()`, so **VLAN pooling behaviour is unchanged**.
- `pf::Switch::getRoleByName($role, $args)`: resolves a role pool to a single member when node context is available; without `$args` it returns the raw value (backward compatible). The base `returnRadiusAccessAccept` and every vendor module that resolves the role itself pass `$args`.
- `pf::roles::performRoleLookup`: passes node context too (CoA / deauth path).
- Documentation and a unit test (`t/unittest/role/pool.t`).

## Notes / limitations

- All members of a role pool must be configured as **equally privileged** — the selection is deterministic and therefore predictable/influenceable by the client (notably under MAC auth). A client can only ever land on a member of the pool of its own already-assigned role.
- Role pooling applies to the *Switch Role* (Filter-Id) mapping; it does not apply to downloadable-ACL (push ACL) enforcement (the global role name is sent there) nor to the VoIP voice role.

## Testing

- Unit test covering: single (non-pool) pass-through, membership, determinism (roaming stability), distribution, VLAN hash parity, `getRoleByName` integration (with/without args), CoA parity via `performRoleLookup`, MAC fallback for pid-less nodes, trimming / empty members, and inherited (parent) role pools.
- Verified end-to-end on a 3-node cluster: PEAP-MSCHAPv2 authentication returns a pool member as `Filter-Id`, stable across repeated authentications, and a CoA carries the same member.

### DIFF
--- a/docs/installation/introduction_to_role_based_access_control.asciidoc
+++ b/docs/installation/introduction_to_role_based_access_control.asciidoc
@@ -51,6 +51,35 @@ algorithms: one based on a hash of the username (default one), another one based
 on a round-robin (last registered device +1) and one that selects a VLAN
 randomly in the pool.
 
+The same pooling mechanism is available for the _Switch Role_ mapping (_Role
+mapping by Switch Role_), which is useful when a controller distributes clients
+through a role attribute such as `Filter-Id` instead of a VLAN. Instead of a
+single role, set a comma-separated list of roles, e.g. `role_a,role_b,role_c` -
+PacketFence then returns one of these roles. Because roles are arbitrary strings
+(not numeric ranges), the `..` range syntax does not apply; list every role
+explicitly. The selection always uses the username hash, so a given user is
+mapped to the same role on every re-authentication and keeps that role while
+roaming between access points. For this stability to hold, configure the same
+role pool (same list, same order) on every switch the user can roam across - a
+_Switch Group_ is the simplest way to guarantee that.
+
+The pool member is chosen from the node owner (username): all devices of a user
+share the same role. For machine authentication, MAC authentication or otherwise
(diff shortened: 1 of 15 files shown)


abprasa:
@jrouzierinverse @fdurand could you please review and share feedbacks

PR #9181, opened 2026-07-21
fix(pfcron): stop empty management-network panic and self-heal on boot

JeGoi:
# Description
(REQUIRED)
At boot, pfcron (and any Go daemon fetching `ManagementNetwork`) panicked when pfconfig served an empty/scalar element for `interfaces::management_network` — e.g. a cold/stale cluster cache before the host's management interface is resolved. The panic exited the process, and the default systemd restart rate-limit left the unit dead.

This PR:
- `pfconfigdriver`: decode the element without panicking; return an error on an empty/scalar payload so a daemon can recover instead of crashing.
- `pfcron` systemd unit: self-heal on boot (`RestartSec=30`, `StartLimitIntervalSec=0`) instead of tripping the start limit.
- cluster venom: `configreload hard` before the final service restart so daemons read fresh, fully-synced config.

# Impacts
(REQUIRED)
- `go/pfconfigdriver` element-decode path — used by every daemon via `FetchDecodeSocket` (pfcron, pfacct, pfstats, pfudpproxy, firewallsso, connectors, chisel, pfsso). Verify normal namespaces still decode.
- `packetfence-pfcron.service` restart behavior at boot (never permanently gives up).
- cluster bring-up (`cluster_wrapup`): extra hard reload before restart.

# Delete branch after merge
(REQUIRED)
NO

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [no] Add unit tests
- [yes] Add acceptance tests (TestLink)

# NEWS file entries

## Bug Fixes
* pfcron and other Go daemons no longer crash when pfconfig serves an empty management network (cluster boot)

### DIFF
--- a/conf/systemd/packetfence-pfcron.service
+++ b/conf/systemd/packetfence-pfcron.service
@@ -3,6 +3,8 @@
 Description=PacketFence pfcron Service
 Wants=packetfence-base.target packetfence-config.service
 After=packetfence-base.target packetfence-config.service
+# Retry forever: at boot pfcron may start before pfconfig is ready to serve.
+StartLimitIntervalSec=0
 
 [Service]
 Type=notify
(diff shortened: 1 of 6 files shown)


JeGoi:
https://gitlab.com/inverse-inc/packetfence/-/pipelines/2693390546

PR #9209, opened 2026-08-09
Add wired 802.1X and dynamic VLAN support to UniFi driver

Ryanlewis2018:
## Summary
- Declare wired 802.1X support for the UniFi controller driver.
- Declare RADIUS dynamic VLAN assignment support.
- Cover both advertised capabilities in `t/switch_supports.t`.

## Validation
- Validated in a wired EAP-TLS pilot using a UniFi switch.
- A domain-joined Windows endpoint authenticated successfully and received its RADIUS-assigned VLAN.
- Confirmed both capabilities in the PacketFence runtime.


fdurand:
If the wire support is added to the Unify then we need to have the wiredeauthTechniques function added too in order to be able to send a CoA/Disconnect.

PR #9210, opened 2026-08-10
feat(pfconnector): enable chisel pprof via CHISEL_PPROF env var

fdurand:
# Description
Replace the compile-time 'pprof' build tag with a runtime CHISEL_PPROF environment variable. Unset/0/false keeps profiling off; 1/true listens on localhost:6060; a host:port value binds that address (lets client and server profile on the same host without colliding on 6060).

Forward CHISEL_PPROF into the containers from the systemd-service wrappers (mirrors the LOG_LEVEL pattern) and add commented Environment= examples to the pfconnector server/client/remote-combined units, disabled by default.

# Impacts
-pfconnector pprof

# Delete branch after merge
YES

# Checklist
- [ ] Document the feature
- [ ] Add unit tests

# NEWS file entries
## New Features
* Feature 1
* Feature 2

## Enhancements
* Use env variable to enable pprof in pfconnector

### DIFF
--- a/addons/pfconnector/containers/systemd-service
+++ b/addons/pfconnector/containers/systemd-service
@@ -35,6 +35,11 @@ function base_args {
     args="$args --env-file /usr/local/pfconnector-remote/var/conf/$name.env"
   fi
 
+  # Pass CHISEL_PPROF to container if set (enables the pfconnector pprof server)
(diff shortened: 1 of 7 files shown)


codedude (approved):
👍

PR #9233, opened 2026-08-25
fix(proxysql): route processlist SELECTs to the writer hostgroup

fdurand:
## Problem

netdata's `go.d` mysql collector logs this on every collection cycle when PacketFence uses ProxySQL with the reader/writer split:

```
Error 9006 (Y0000): ProxySQL Error: connection is locked to hostgroup 10 but trying to reach hostgroup 30
```

The collector runs `SET SESSION sql_log_bin=OFF` at the start of each cycle, which makes ProxySQL lock its session to the writer hostgroup (10). All of its `SHOW` queries keep working (catch-all rule → HG 10), but the process-list query is a plain `SELECT ... FROM information_schema.processlist`, which matched the `^SELECT` rule → HG 30. ProxySQL refuses to switch a locked session and returns error 9006, so the process-list charts never populate.

## Fix

- Add `rule_id=3` matching `(information_schema|performance_schema).processlist` with `destination_hostgroup=10, apply=1`, so the collector's one SELECT stays on the writer with the rest of its locked session.
- The former `^SELECT` and catch-all rules are renumbered to 4 and 5, and the read-only scheduler's `rule_ids` list in `proxysql.conf.example` is updated from `1,2,4` to `1,2,5`.
- Rule 3 is deliberately excluded from the degraded-mode rewrite: a SELECT still works on a read-only writer, and the `replace_pattern='ERROR DB IS IN R/O'` rewrite would break it.

## Deployment note

ProxySQL only parses its config file when no `proxysql.db` exists in the datadir, so existing installs need `proxysql-initial` (or an admin-interface update) for the renumbered rules to take effect.

### DIFF
--- a/conf/proxysql.conf.example
+++ b/conf/proxysql.conf.example
@@ -104,7 +104,7 @@ scheduler =
      filename="/proxysql-read-only-handler.sh"
      arg1="{\"username\":\"administrator\", \"password\":\"admin\", \"host\":\"127.0.0.1\", \"port\":6032}"
      arg2="{\"username\":\"[% mysql_pf_user %]\", \"password\":\"[% mysql_pf_pass %]\"}"
-     arg3="{\"read_write_hostgroup\":10, \"read_hostgroup\":30, \"rule_ids\":\"1,2,4\"}"
+     arg3="{\"read_write_hostgroup\":10, \"read_hostgroup\":30, \"rule_ids\":\"1,2,5\"}"
      arg4="/proxysql-read-only-handler.log"
   [% ELSE %]
      id=11
--- a/lib/pf/services/manager/proxysql.pm
+++ b/lib/pf/services/manager/proxysql.pm
@@ -273,8 +273,15 @@ EOT
     #   rule 2 — SELECT            → HG 30 (readers preferred)
     #            writer is also in HG 30 (weight=50) only as fallback
     #            if all pure readers go down ✅
-    #   rule 4 — catch-all                 → HG 10 in normal mode
-    #            scheduler rewrites 1,2,4 during degraded mode
+    #   rule 3 — SELECT on processlist     → HG 10 (writer)
+    #            netdata's mysql collector runs SET SESSION sql_log_bin=OFF
+    #            which locks its session to HG 10; without this rule its
+    #            processlist SELECT matches rule 4 → HG 30 and ProxySQL
+    #            rejects it with error 9006 (locked to hostgroup)
+    #   rule 5 — catch-all                 → HG 10 in normal mode
+    #            scheduler rewrites 1,2,5 during degraded mode
+    #            (rule 3 is deliberately NOT rewritten: a SELECT still works
+    #            on a read-only writer, and replace_pattern would break it)
     # Only generated when we actually have a reader split
     #
     # IMPORTANT: the proxysql.conf template MUST reference [% mysql_query_rules %]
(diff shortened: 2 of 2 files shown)


JeGoi (approved):

codedude (approved):

jrouzierinverse (commented):
## Rule renumbering

The old rule 4 wasn't removed — it was renumbered. Inserting the processlist rule at id 3 pushed the two rules below it down by one:

| before | after | pattern → destination |
|---|---|---|
| 1 | 1 | `^SELECT.*FOR UPDATE` → HG 10 |
| 2 | 2 | `^SELECT @@global.read_only` → HG 10 |
| — | **3** | `(information_schema\|performance_schema).processlist` → HG 10 *(new)* |
| 3 | **4** | `^SELECT` → HG 30 |
| **4** | **5** | `.*` catch-all → HG 10 |

Hence `rule_ids` going from `1,2,4` to `1,2,5` in `conf/proxysql.conf.example`. That bookkeeping checks out:

- Nothing else in the tree references these ids (`grep -rn rule_id` only hits the template).
- `containers/proxysql/Dockerfile` starts proxysql with `--initial`, so the config file is re-parsed and the on-disk DB rebuilt on every restart — no stale-rule-id hazard on upgrade.
- Leaving rule 4 out of `rule_ids` is still correct: `bin/proxysql-read-only-handler.sh:53` forces every listed rule's `destination_hostgroup` back to the writer on recovery, so listing the plain-`SELECT` rule would silently kill the read split.

## Findings

**1. The header comment block is now internally contradictory** (`lib/pf/services/manager/proxysql.pm:272-283`). It lists "rule 2 — SELECT → HG 30", but rule 2 is `^SELECT @@global.read_only` → *writer*; the plain-SELECT-to-readers rule is rule 4, which the block never lists — even though the new rule-3 note four lines below correctly says "matches rule 4 → HG 30". The block was already off by one before this commit (it omitted the `global.read_only` rule); the renumbering makes it off by two. Suggested:

```
#   rule 1 — SELECT ... FOR UPDATE     → HG 10 (writer always, locks rows)
#   rule 2 — SELECT @@global.read_only → HG 10 (writer)
#   rule 3 — SELECT on processlist     → HG 10 (writer)
#   rule 4 — SELECT                    → HG 30 (readers preferred)
#   rule 5 — catch-all                 → HG 10 in normal mode
```

**2. Unescaped dot in the new pattern** (`lib/pf/services/manager/proxysql.pm:313`): `(information_schema|performance_schema).processlist` — the `.` matches any character. Low impact, since it only over-matches toward the writer, which is where the catch-all would have sent a non-`SELECT` anyway, but `\.processlist` costs nothing.

**3. The fix is per-query, not per-session — expect siblings.** The root cause is ProxySQL's `mysql-set_query_lock_on_hostgroup` (default 1): the collector's `SET SESSION sql_log_bin` pins the session to HG 10, and *any* subsequent `SELECT` in that session that falls through to rule 4 gets error 9006. Pinning the processlist query fixes the one query observed here; if go.d's mysql collector issues another bare `SELECT` after the `SET`, it fails the same way and needs another rule. (I did not verify the collector's full query list.) A durable alternative is `mysql-set_query_lock_on_hostgroup=0` in `mysql_variables`, at the cost of relaxing session pinning for all clients — the targeted rule is the lower-risk change, so this is a judgement call rather than a request.

**4. Context, not a regression.** `rule_ids` only bites when the read-only-handler scheduler *and* the generated query rules coexist, which happens only on the `getWriteDB()` (geoDB) path. In the plain `scheduler == 'default'` branch (`lib/pf/services/manager/proxysql.pm:187`), `$has_reader_split` stays 0, so no `mysql_query_rules` block is emitted at all and `1,2,5` refers to nothing. That was equally true of `1,2,4` — noting it only so the arg3 edit isn't mistaken for broader coverage than it has.

Nothing here blocks the fix; 1 and 2 are the ones worth folding in before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9234, opened 2026-08-25
pfqueue: stop losing DHCP tasks to expiry on the pfdhcplistener queue

fdurand:
## Problem

pfqueue logs errors like:

```
pfqueue[1778380]: t=2026-08-25T13:16:06+0000 lvl=eror msg="Error runSingleWorkerQueue: Data not found for task Task:6E8A09A2-A086-11F1-8B96-8345D4E48283:Queue:pfdhcplistener:api:process_dhcpv4"
```

The producer stores each task's payload in a Redis hash with a 300s TTL and pushes the task ID onto the queue list. Task IDs were sitting in the `pfdhcplistener` list longer than that TTL: by the time a worker popped them the payload hash had expired, and the DHCP packet's processing was silently dropped.

The queue was **not** capped at its one dedicated worker — it also carried `weight=4`, which puts it in the weighted rotation served by the 8 shared workers (`[pfqueue] workers=8`). The problem is contention rather than a hard cap: every `pfdhcplistener` task competed with `general`, `pfdetect` and `priority` for those shared workers, and a DHCP burst arriving while they are busy with long tasks on another queue has to wait behind them. Same shape as the bottleneck previously fixed on `pfdhcplistener_external` (2ecf9bab79), and on `pfsnmp`/`pfsnmp_parsing` before it.

## Changes

- **Hash the `pfdhcplistener` queue by client MAC with 4 dedicated workers**, mirroring `pfdhcplistener_external`. Config-only: `sbin/pfdhcplistener` is the only producer to this queue and already submits both `process_dhcpv4` and `process_dhcpv6` through `notify_hashed` keyed by client MAC, so per-client packet ordering is preserved while different clients run in parallel. pfconfig expands a hashed queue into one single-worker entry per bucket (`pfdhcplistener_000..003`, `weight=0`), so the consumer needs no changes.

  The trade is a guaranteed 4 workers instead of a contended up-to-9; `workers` is the knob if a deployment needs more.

- **Fix the Go consumer's expired-task accounting**: the `ExpiredCounters` HINCRBY was queued on an already-executed pipeline, so it never reached Redis at all, and it used `-1` where the Perl consumer (`pf::pfqueue::consumer::redis`) uses `+1` — the miss counters in the queue stats never reflected expired tasks. Now a direct `HIncrBy(..., 1)`.

- **Typed `ErrTaskExpired`** naming the task that was dropped, instead of the bare "Data not found". It stays at **error** level, as the same condition does in the Perl consumer: an expired task is a submission that was silently thrown away, and that is exactly what an operator needs to see.

## Testing

- `go/pfqueueclient` had no tests at all, which is why the dead-pipeline HINCRBY survived. Added `consumer_test.go` (miniredis): an expired task returns `ErrTaskExpired` naming the task, increments `ExpiredCounters` **in Redis**, and still decrements the pending `TaskCounters`; an empty queue returns `redis.Nil` and is not mistaken for an expired task. Verified the expired-task test fails against the pre-fix code (`expired counter: got "", want "1"` — the command never reached Redis).
- `go build` / `go vet` clean on `pfqueueclient` and `cmd/pfqueue-go`.

## Upgrading

`conf/pfqueue.conf.defaults` only applies where the operator has not overridden it. Existing deployments that pinned `[queue pfdhcplistener]` in `conf/pfqueue.conf` need `workers=4` + `hashed=enabled` there (dropping `weight`), then `pfcmd configreload hard` and a restart of `pfqueue` and `pfdhcplistener`.


JeGoi (approved):

JeGoi:
Question: Can we create a venom test for that part? Does it make sens?

fdurand:
Good question — partly yes, but not for the bug itself.

**Not reproducible in venom.** Triggering the actual failure means pushing the queue backlog past the 300s payload TTL, which needs sustained DHCP load, not a functional scenario. That's a load test, and it would be flaky in CI.

**A Go unit test is the right level, and I've added one.** `go/pfqueueclient` had no tests at all, which is exactly why the dead-pipeline `HINCRBY` survived in the tree. `consumer_test.go` (miniredis) now covers it: an expired task returns `ErrTaskExpired` naming the task, increments `ExpiredCounters` *in Redis*, and still decrements the pending `TaskCounters`; an empty queue returns `redis.Nil` and isn't mistaken for an expired task. I checked it fails against the pre-fix code — `expired counter: got "", want "1"`, the empty string being the command that never left the pipeline.

**Where venom would add something** is the config half, which no test covers: after a `configreload hard`, assert that `Queue:pfdhcplistener_000..003` exist and are being consumed. That guards against a future edit silently dropping `hashed=enabled` and quietly sending every task to a list nobody reads. The pfcron suite already starts `pfqueue-go`/`pfqueue-backend`, so there's a natural place to hang it. Happy to add it here or as a follow-up — let me know which you'd prefer.

Note the PR description has been updated: the original diagnosis ("single dedicated worker") was wrong — the queue also carried `weight=4`, so the 8 shared workers served it too. The real problem is contention with `general`/`pfdetect`/`priority`, and the change trades a contended up-to-9 workers for a guaranteed 4.

codedude (approved):

PR #9249, opened 2026-09-01
pfacct: restore the pfacct_rate_limit toggle and always forward accounting Stop

fdurand:
# Description
Two fixes to the pfacct accounting rate limiter (`radius_configuration.pfacct_rate_limit`), which gates the forwarding of accounting packets from pfacct to httpd.aaa's `radius_accounting` handler:

1. **Restore the toggle** — commit 9df644a704 fixed the inverted error check on the rate-limit cache TTL parse but dropped two neighboring lines with it: the assignment of `pfAcct.RateLimit` from `radius_configuration.pfacct_rate_limit`, and the TTL default of 5 minutes. Since then the toggle decoded from pfconfig was never applied: `RateLimit` stayed false and every accounting packet was forwarded to httpd.aaa regardless of the setting.
2. **Always forward Stop** — the limiter treated Stop like Interim-Update: forwarded only when the Framed-IP changed, and dropped entirely when the session's Start key was not cached (e.g. sessions established before pfacct started). The AAA layer needs every Stop: locationlog close, `unreg_on_acct_stop` and floating-device handling depend on it. Stop now passes unconditionally and clears the session's Start key so the next session re-registers.

Verified on a loaded lab server (~460 Interim-Updates/s): with the toggle restored, notifies to httpd.aaa dropped from ~370/s to ~0 and Apache shed ~1.3 cores, while Accounting-Responses to the NAS were unaffected (pfacct acknowledges before the gate).

# Impacts
- pfacct accounting workflow: with `pfacct_rate_limit=enabled`, httpd.aaa now only receives Start (once per session/IP/location per TTL), Interim-Updates whose Framed-IP changed, and every Stop.
- Deployments that had `pfacct_rate_limit=enabled` set will observe the setting actually taking effect after upgrading.
- Default behavior (`pfacct_rate_limit=disabled`) is unchanged.

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature (setting already documented in documentation.conf)
- [x] Add unit tests (`go/cmd/pfacct/rate_limit_test.go`, run against the real pfacct dictionary)

# NEWS file entries
## Bug Fixes
* radius_configuration.pfacct_rate_limit had no effect: the toggle was decoded but never applied
* The pfacct accounting rate limiter no longer suppresses Accounting-Stop packets (locationlog close, unreg_on_acct_stop and floating-device handling rely on them)


codedude (approved):

fdurand:
All three addressed in 94981235aa, and rebased on devel.

**Stale Start key after a move seen through an Interim-Update** (`radius.go`). Real, and the reproduction is exactly the one described: Start on A, Interim on B (which registers the new location), back to A within the TTL -- the Start on A matched the stale `Start-A` key and was suppressed, so the AAA layer never saw it and locationlog was never refreshed for the NAS the device returned to. The Start path already purged the old key, but only in its `!exists` branch, and `handleInterim` did not purge at all.

Both of `handleInterim`'s branches move the device, and so does the IP-change branch of the Start path, so the purge now lives in one `recordLocation` helper that all of them call. `handleInterim` also takes the same `mac.Mac` and raw `Calling-Station-Id` as the Start path rather than a `net.HardwareAddr`, so both build the cache keys the same way instead of relying on the two `String()` implementations agreeing.

Two regression tests, `TestRateLimitForwardsRoamBackAfterInterimOnly` and `...AfterInterimIpChange`; both were verified to fail with the purge removed from `handleInterim` and pass with it.

**Coverage of the restored wiring.** A fair hit, and the sharpest of the three given this branch's history: the toggle was dead code for three releases precisely because the assignment vanished in a refactor, and every test stayed green because they all drive `rateLimit` directly, which never reads the configuration. The mapping now lives in `applyRateLimitConfig` and is covered for the enabled, disabled, unset, zero, negative and unparseable-TTL cases. `TestRateLimitConfigIsWiredUp` covers the other half -- that `SetupConfig` still calls it -- through `PfacctRateLimitCacheTtl`: the block either sets the configured value or floors it to the default, so a zero there means it never ran, and that catches a deleted call site regardless of what the host's toggle is set to. It also compares `RateLimit` against the live configuration, and says so in the log when the host has the toggle disabled and that half of the check is therefore inconclusive.

**NEWS.** Both entries from the description are now in the unreleased Bug Fixes section.

`go build`, `go vet` and the `cmd/pfacct` tests were run against pfconfig and the database on a test box.

PR #9250, opened 2026-09-01
pfacct: handle node.last_seen and ip4log natively instead of via httpd.aaa

fdurand:
# Description
Moves two per-packet accounting primitives from httpd.aaa's `radius_accounting` handler into pfacct itself:

- **node.last_seen**: `pf::radius::accounting` runs `node_update_last_seen` on every forwarded packet — a DB write per packet through mod_perl. pfacct now refreshes it natively on every accounting packet, rate-limited to once per MAC per `pfacct_rate_limit_cache_ttl` window.
- **ip4log** (when `advanced.update_iplog_with_accounting=enabled`): `pf::api::update_ip4log` is ported with its exact semantics — close the previous entry when the MAC moved to a new IP, auto-create the missing node row, upsert the entry with the zero end_time (accounting carries no lease). Rate-limited per MAC/IP pair; an IP change bypasses the limiter and is handled immediately.

This has two benefits: it removes recurring per-packet mod_perl work on busy deployments (measured on a lab server sustaining ~460 accounting packets/s), and it fixes the staleness gap introduced when `pfacct_rate_limit` is enabled — suppressed packets never reach Perl, so last_seen and iplog previously froze between forwarded packets.

The Perl code itself is unchanged apart from one guard: forwarded packets carry an `X-PacketFence-Handled-Natively` header, and `pf::radius::accounting` and `pf::api::handle_accounting_metadata` skip exactly the primitives it names, so the same row is not written twice. Every other caller of those primitives is untouched.

The marker is built per packet, not from the configuration: it names only what pfacct actually performed for that packet (or deliberately skipped because it had written the same row within the rate-limit TTL). A write that failed, or a case pfacct declines -- `pfdhcp.mac2ip_lookup` enabled, a `0.0.0.0` Framed-IP-Address, `last_seen` for a MAC with no node row -- is left out of the header, and httpd.aaa then runs that primitive exactly as it did before this branch. So the fallback is never lost on the packets that need it.

# Impacts
- pfacct accounting workflow: two new prepared statements per worker path (node last_seen update; ip4log mac-lookup/close/open, active only when `update_iplog_with_accounting` is enabled).
- node.last_seen now stays fresh under `pfacct_rate_limit=enabled`.
- ip4log SQL mirrors `pf::ip4log::open/close` and the existing pfdhcp upsert (`go/cmd/pfdhcp/utils.go`).

# Code / PR Dependencies
Complements #9249 (pfacct_rate_limit restore); no build dependency — both apply cleanly in either order.

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature (no new configuration; existing settings drive the behavior)
- [x] Add unit tests (`go/cmd/pfacct/accounting_primitives_test.go`, DB-backed, run on a PF server: both PASS)

# NEWS file entries
## Enhancements
* pfacct now updates node.last_seen and (when update_iplog_with_accounting is enabled) ip4log natively on accounting packets, removing per-packet httpd.aaa work and keeping them fresh when pfacct_rate_limit suppresses forwarding

### DIFF
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1958,7 +1958,11 @@ EOT
 [radius_configuration.pfacct_rate_limit_cache_ttl]
 type=numeric
 description=<<EOT
-If rate limiting on pfacct is enabled,  the Time to Live value in minutes represent the time of the accounting cache
+Time to Live in minutes of the pfacct accounting caches. When
+radius_configuration.pfacct_rate_limit is enabled it is how long pfacct
+suppresses forwarding a repeated accounting packet to httpd.aaa. Independently
+of that option, it is also how often pfacct refreshes node.last_seen and the
+ip4log entry of a given device.
 EOT
 
 [radius_configuration.pfacct_socket_recv_buffer]
--- a/conf/documentation.conf
+++ b/conf/documentation.conf
@@ -1409,7 +1409,9 @@ EOT
 type=toggle
 options=enabled|disabled
 description=<<EOT
-Use the information included in the accounting to update the iplog
+Use the information included in the accounting to update the iplog.
+Note: pfacct performs this update itself and reads this option once at
+startup, so restart pfacct for a change to this option to take effect.
 EOT
 
 [advanced.update_iplog_with_authentication]
--- a/conf/pf.conf.defaults
+++ b/conf/pf.conf.defaults
@@ -1044,7 +1044,9 @@ pfperl_api_processes=8
 #
 # advanced.update_iplog_with_accounting
 #
-# Use the information included in the accounting to update the iplog
+# Use the information included in the accounting to update the iplog.
+# Note: pfacct performs this update itself and reads this option once at
+# startup, so restart pfacct for a change to this option to take effect.
 update_iplog_with_accounting=disabled
 #
 # advanced.update_iplog_with_authentication
@@ -1473,7 +1475,11 @@ pfacct_work_queue_size = 1000
 pfacct_rate_limit=disabled
 # radius_configuration.pfacct_rate_limit_cache_ttl
 #
-# If rate limiting on pfacct is enabled,  the Time to Live value in minutes represent the time of the accounting cache
+# Time to Live in minutes of the pfacct accounting caches. When
+# radius_configuration.pfacct_rate_limit is enabled it is how long pfacct
+# suppresses forwarding a repeated accounting packet to httpd.aaa. Independently
+# of that option, it is also how often pfacct refreshes node.last_seen and the
+# ip4log entry of a given device.
 pfacct_rate_limit_cache_ttl=5
 # radius_configuration.pfacct_socket_recv_buffer
 #
--- a/go/cmd/pfacct/accounting_primitives.go
+++ b/go/cmd/pfacct/accounting_primitives.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"context"
+	"database/sql"
+
+	cache "github.com/fdurand/go-cache"
+	"github.com/inverse-inc/go-utils/mac"
+)
+
+// Native ports of the per-packet primitives historically performed by
+// httpd.aaa's radius_accounting handler (pf::radius::accounting and
+// pf::api::update_ip4log). Running them here keeps node.last_seen and ip4log
+// fresh on every accounting packet even when pfacct_rate_limit suppresses the
+// notification to httpd.aaa, and removes recurring per-packet mod_perl work.
+
+// updateNodeLastSeen refreshes node.last_seen for the MAC, at most once per
+// rate-limit TTL per MAC. Like pf::node::node_update_last_seen it only
+// updates an existing node row; node creation stays with the registration
+// paths.
+func (h *PfAcct) updateNodeLastSeen(ctx context.Context, m mac.Mac) {
+	key := m.String()
+	if _, found := h.LastSeenCache.Get(key); found {
+		return
+	}
+	h.LastSeenCache.Set(key, 1, cache.DefaultExpiration)
+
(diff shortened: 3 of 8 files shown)


codedude (approved):

jrouzierinverse:
please rebase

fdurand:
Addressed in b904ddee0e, and rebased on devel.

**The marker is now per packet, not per configuration** (`radius.go`). This was the root of several of the findings: `handledNatively()` was assembled from the startup toggles, so a packet whose native write failed still told httpd.aaa to skip the primitive -- both sides skipped it and the update was lost, with no fallback left. `updateNodeLastSeen` and `updateIp4log` now return whether the primitive is pfacct's business for that packet, and `nativePrimitives.header()` renders only what is true. A primitive pfacct performed, or deliberately skipped because it wrote the same row within the rate-limit TTL, is claimed; anything else is left to the AAA layer, which behaves exactly as before this branch. Both Perl consumers split the header into a set, so an empty header simply means "nothing was handled here".

**Failed close no longer falls through to the open** (`accounting_primitives.go`). This was the sharpest one: the old code set `complete = false` and carried on to `ip4logOpen`, leaving two entries active for the MAC -- and nothing could repair it afterwards, because `ip4logMac2Ip` returns the most recently started entry, so the retry would find the one just opened and never close the old one. Any failed statement now aborts the move and declines the primitive.

**pfdhcp.mac2ip_lookup.** Confirmed: `pf::ip4log::mac2ip` (ip4log.pm:129-143) queries the pfdhcp API first and only then falls back to SQL. pfacct has only the SQL half, so rather than reproduce half of `update_ip4log` it now hands the whole ip4log primitive back to httpd.aaa while that setting is on. Read at startup through a new `PfConfPfdhcp`, with a log line when it takes effect. (The default is `disabled`, so this costs nothing on stock deployments.)

**0.0.0.0 Framed-IP-Address.** Also confirmed: `pf::api::handle_accounting_metadata` tests the address for truthiness only, and `"0.0.0.0"` is truthy in Perl, so Perl does call `update_ip4log` with it -- resolving and closing the previous entry before `pf::ip4log::open` rejects the new address. pfacct still returns early, but no longer claims `ip4log` for that packet, so the close is not silently dropped.

**last_seen for an unknown MAC.** An `UPDATE` that matches no row returns no error, and caching the key there suppressed `last_seen` for a whole TTL for a node registered moments later. The key is now cached only when a row was actually updated; a no-op update is left to the AAA layer. Cost of the conservative reading of `RowsAffected` (a second write inside the same second also reports 0): one redundant Perl write, the behaviour that existed before this branch.

**PR description.** Corrected -- it claimed the Perl side was untouched and that forwarded packets still ran the same code, which stopped being true when the marker landed. It now describes the guard and, explicitly, which cases keep the fallback.

**On the Ip4logCache hiding another writer:** left as is, deliberately. The cache key is `MAC|IP`, so if pfdhcp moves the MAC to a different IP, the next accounting packet carrying that new IP misses the cache and goes through the full close/open path. The only case that short-circuits is an accounting packet still reporting the *old* IP after pfdhcp moved the binding -- where doing nothing matches the DHCP view rather than fighting it, and is bounded by the rate-limit TTL anyway. Checking the current binding on a cache hit would mean a SELECT per accounting packet, which is the cost this branch exists to remove. Happy to revisit if you would rather have the correctness guarantee.

Tests: the ip4log and last_seen tests now assert what is claimed as well as what is written, and cover the unknown-node, 0.0.0.0 and mac2ip_lookup paths; `TestNativePrimitivesHeader` replaces `TestHandledNatively`. The failed-close path is not covered -- it needs fault injection the suite has no hook for. `go build`, `go vet` and the `cmd/pfacct` tests were run against a database on a test box.

conf/documentation.conf
```diff
@@ -1956,7 +1958,11 @@ EOT
 [radius_configuration.pfacct_rate_limit_cache_ttl]
 type=numeric
 description=<<EOT
-If rate limiting on pfacct is enabled,  the Time to Live value in minutes represent the time of the accounting cache
+Time to Live in minutes of the pfacct accounting caches. When
+radius_configuration.pfacct_rate_limit is enabled it is how long pfacct
+suppresses forwarding a repeated accounting packet to httpd.aaa. Independently
+of that option, it is also how often pfacct refreshes node.last_seen and the
+ip4log entry of a given device.
```
jrouzierinverse, line 1965: - `go/cmd/pfacct/pfacct.go:242` — The new documentation (conf/documentation.conf, pf.conf.defaults) says pfacct_rate_limit_cache_ttl paces the native last_seen/ip4log refreshes, but the code floors refreshTtl at 5 minutes, so a TTL < 5 refreshes every 5 minutes, diverging from what the docs imply.

jrouzierinverse (changes requested):

jrouzierinverse:
please rebase

jrouzierinverse:
rebase

conf/documentation.conf
fdurand: Fixed in 5025ad6a17. Two things on this one:

The floor only applied to a TTL of 0 or less, not to any TTL below 5: `refreshTtl <= 0` was the condition, so a TTL of 1 refreshed every minute as documented. Since the rebase it was also dead code, because `applyRateLimitConfig` (from #9249) already replaces a non-positive value with `DefaultRateLimitCacheTtl` and logs a warning before the native caches are built. The floor is gone and `LastSeenCache` / `Ip4logCache` now take `PfacctRateLimitCacheTtl` as is, like the two rate-limit caches, so the value is validated in one place.

The docs did leave the non-positive case unsaid; documentation.conf and pf.conf.defaults now say the value must be at least 1 and that zero or a negative value falls back to the 5 minute default with a warning.

PR #9254, opened 2026-09-01
fix(services): tolerate an unset management network during the configurator

JeGoi:
# Description

During the configurator wizard the management interface does not exist yet, so
`$management_network` (a tied scalar) is `undef`. Four service managers called
`$management_network->tag(...)` unconditionally, which is a hard FATAL.

`packetfence-netdata.service` runs `generateConfig()` in `ExecStartPre` with
`Restart=on-failure` and is pulled in by `packetfence.target`, so that FATAL
failed the unit and every subsequent target/iptables restart re-triggered it —
a ~15s-cadence restart loop lasting the whole wizard (44 FATALs on pf1, 120
across the 3 nodes of `cluster_configurator_deb12`).

This guards the four managers so they generate config with an empty management
IP while the interface is unset, and pick up the real value once it is set.
The `pf::iptables` "Management Interface is not set" warnings are by design and
are left untouched — only the crash is fixed.

Guard idiom is `ref($management_network)`, matching the existing
`snmptrapd.pm::getManagementIp`, with `''` fallbacks so templates do not emit
uninitialized-value warnings.

Perl-side sibling of #9181, which fixed the same class of bug on the Go side
(pfcron / pfconfigdriver).

# Impacts

- `lib/pf/services/manager/netdata.pm` — IP/VIP lookups hoisted to the top of
  `generateConfig`; cluster-members block gated on the management network.
- `lib/pf/services/manager/radiusd_child.pm` — new `_management_ip()` helper
  replaces 7 duplicated `defined(...->tag('vip')) ? ... : ...` call sites.
- `lib/pf/services/manager/pfacct.pm` — only the IP-dependent part is guarded;
  `$port_save` is now set outside the guard because it is used later
  independently of the management IP.
- `lib/pf/services/manager/haproxy_db.pm` — single `//` chain.

Reviewer should confirm no behaviour change once the management network **is**
set: in that case every guard takes the same branch as before, and
`vip // ip` is equivalent to the previous `defined(vip) ? vip : ip`.

Worth exercising: cluster + standalone configurator runs, RADIUS config
generation (`radiusd_child` touches sites/main/auth/acct/eduroam/cli), and
`haproxy-db` in both slave and normal cluster mode.

# Backport

Backport scope is **15.x and devel**. Cherry-pick of `4524dfda53` was tested in
a throwaway worktree per branch and applies with no manual work:

| Branch             | Cherry-pick | Notes                                           |
|--------------------|-------------|-------------------------------------------------|
| `maintenance/15.2` | clean       | target branch of this PR                        |
| `maintenance/15.1` | clean       | —                                               |
| `maintenance/15.0` | clean       | —                                               |
| `devel`            | clean       | four files byte-identical to this branch's base |

The same unguarded pattern also exists on 14.x and older, but the surrounding
code has drifted there (cherry-pick conflicts in `pfacct.pm` /
`radiusd_child.pm` / `netdata.pm`), so those are deliberately out of scope.

# Delete branch after merge

YES

# Checklist

- [ ] Add unit tests — no (see note below)
- [ ] Add acceptance tests (TestLink) — no; covered by the existing
      `cluster_configurator_deb12` venom suite, which this fix targets

Note: these `generateConfig` paths have no existing unit-test harness, and the
failure only reproduces with an unset management network during cluster
bring-up, which the venom configurator suite already exercises.

# NEWS file entries

## Bug Fixes

* Service config generation no longer crashes when the management network is
  unset, which caused netdata / pfacct / radiusd / haproxy-db to restart-loop
  throughout the configurator wizard

### DIFF
--- a/lib/pf/services/manager/haproxy_db.pm
+++ b/lib/pf/services/manager/haproxy_db.pm
@@ -67,10 +67,10 @@ sub generateConfig {
          $tags{'os_path'} = '/usr/share/haproxy/';
     }
     
-    $tags{'management_ip'}
-        = defined( $management_network->tag('vip') )
-        ? $management_network->tag('vip')
-        : $management_network->tag('ip');
+    # management_network stays undef until the mgmt interface is configured (configurator)
+    $tags{'management_ip'} = ref($management_network)
+        ? ($management_network->tag('vip') // $management_network->tag('ip') // '')
+        : '';
 
     my $i = 0;
     my @mysql_backend;
--- a/lib/pf/services/manager/netdata.pm
+++ b/lib/pf/services/manager/netdata.pm
@@ -66,10 +66,15 @@ sub generateConfig {
     my $logger = get_logger();
     my %tags;
 
+    # management_network stays undef until the mgmt interface is configured (configurator)
+    my $has_mgmt = ref($management_network) ? 1 : 0;
+    my $mgmt_ip  = $has_mgmt ? $management_network->tag('ip')  : '';
+    my $mgmt_vip = $has_mgmt ? $management_network->tag('vip') : undef;
+
     $tags{'hosts_cluster_members'} = '';
-    if ($cluster_enabled) {
+    if ($cluster_enabled && $has_mgmt) {
         my $int = $management_network->tag('int');
-        $tags{'hosts_cluster_members'} = join(",", grep( {$_ ne $management_network->tag('ip')} values %{pf::cluster::members_ips($int)}));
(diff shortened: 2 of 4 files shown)


JeGoi:
Cluster tests: https://gitlab.com/inverse-inc/packetfence/-/pipelines/2809806603
Entrire CI tests: https://gitlab.com/inverse-inc/packetfence/-/pipelines/2810246445

codedude (approved):

PR #9255, opened 2026-09-01
pfacct: skip per-packet balance accounting when no node carries a balance

fdurand:
# Description
`handleTimeBalance` and `handleBandwidthBalance` run on every accounting packet: several node-table SELECTs (`IsUnreg`, zero-balance checks) plus two unconditional UPDATEs (`SoftNode*BalanceUpdate`). Every one of those statements filters on the balance column being `NOT NULL`, so on deployments that don't use time/bandwidth balances they are guaranteed no-ops — but the DB still pays for them.

Measured on a lab server at ~300 accounting packets/s with 30,026 nodes (zero with balances): this machinery was the majority of MariaDB's ~1,000 UPDATEs/s and a large share of the node table's ~6,400 rows read/s, with mariadbd as the top CPU consumer on the box (167%).

pfacct now probes `EXISTS(SELECT 1 FROM node WHERE time_balance IS NOT NULL OR bandwidth_balance IS NOT NULL)` at startup and every 5 minutes, and skips the balance path entirely while no node carries one.

# Impacts
- pfacct accounting workflow: balance enforcement semantics unchanged when balances are in use; when none are, the whole path is skipped.
- A balance assigned to a node activates within one 5-minute refresh interval (the probe fails open on DB errors).
- **The gate is deployment-wide, not per node.** One node anywhere carrying a balance re-enables the per-packet path for every packet from every node, so the win is all-or-nothing: a single time-limited guest voucher restores the original cost across the whole deployment.
- **Within one refresh interval of the *first* balance being assigned, accounting for that window is lost.** The interim path is what detects an exhausted balance mid-session (`SoftNodeTimeBalanceUpdate` zeroes the balance and fires the trigger) and the `Stop` path is the only debit. While the gate is still shut, interims are skipped and a `Stop` arriving in that window skips the debit outright — `Stop` happens once, so that session is never charged, and a `TimeExpired`/`BandwidthExpired` that would have fired may not. Nothing in the `set_time_balance` / `set_bandwidth_balance` path pokes pfacct, so there is no invalidation hook; this bounds the exposure at one interval per deployment, not per node.
- Sessions already in flight when the gate reopens are charged at most the time since it reopened. `handleTimeBalance` uses the `NodeSessionCache` entry to record the offset at which a node stopped being unregistered, and that entry is never primed while the gate is shut, so charging the full `AcctSessionTime` would bill time pfacct never accounted for — enough to zero a freshly assigned balance and lock the user out. A deployment that has balances from the start records no reopen instant and keeps the previous behaviour exactly.

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature (no configuration change)
- [x] Add unit tests (`go/cmd/pfacct/balance_gate_test.go`, PASS on a PF server; the gate test is scoped to its own node and the cap/stop tests need no database)

# NEWS file entries
## Enhancements
* pfacct skips the per-packet time/bandwidth balance accounting (several SELECTs and two UPDATEs per accounting packet) while **no node in the deployment** carries a balance, re-probed every 5 minutes; assigning a balance re-enables it within that interval

### DIFF
--- a/go/cmd/pfacct/balance_gate.go
+++ b/go/cmd/pfacct/balance_gate.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"time"
+)
+
+const balancesInUseRefreshInterval = 5 * time.Minute
+
+// refreshBalancesInUse probes whether any node currently carries a time or
+// bandwidth balance. When none do (the common case), the per-packet balance
+// accounting in handleAccountingRequest is skipped entirely: every statement
+// it runs filters on the balance being NOT NULL and is a guaranteed no-op,
+// yet together they cost several node-table SELECTs and two UPDATEs per
+// accounting packet. Fails open: with the probe unavailable or erroring, the
+// balance path stays on.
+func (h *PfAcct) refreshBalancesInUse() {
+	if h.anyNodeBalances == nil {
+		h.balancesInUse.Store(true)
+		return
+	}
+
+	var inUse bool
+	if err := h.anyNodeBalances.QueryRow().Scan(&inUse); err != nil {
+		logError(h.LoggerCtx, "anyNodeBalances: "+err.Error())
+		inUse = true
+	}
+	h.balancesInUse.Store(inUse)
+}
+
+// startBalancesInUseRefresher primes the balances-in-use flag and keeps it
+// fresh; a balance assigned to a node is picked up within one refresh
+// interval.
+func (h *PfAcct) startBalancesInUseRefresher() {
+	h.refreshBalancesInUse()
+	go func() {
+		for {
+			time.Sleep(balancesInUseRefreshInterval)
+			h.refreshBalancesInUse()
+		}
+	}()
+}
--- a/go/cmd/pfacct/balance_gate.go
+++ b/go/cmd/pfacct/balance_gate.go
@@ -6,36 +6,103 @@ import (
 
 const balancesInUseRefreshInterval = 5 * time.Minute
 
+// anyNodeBalancesQuery probes whether any node carries a time or bandwidth
+// balance. One predicate rather than two EXISTS subqueries: neither column is
+// indexed, and the case this gate optimises for (no balances anywhere) is the
+// one that cannot short-circuit, so two subqueries scan the node table twice.
+const anyNodeBalancesQuery = `SELECT EXISTS(SELECT 1 FROM node WHERE time_balance IS NOT NULL OR bandwidth_balance IS NOT NULL)`
+
 // refreshBalancesInUse probes whether any node currently carries a time or
 // bandwidth balance. When none do (the common case), the per-packet balance
 // accounting in handleAccountingRequest is skipped entirely: every statement
 // it runs filters on the balance being NOT NULL and is a guaranteed no-op,
 // yet together they cost several node-table SELECTs and two UPDATEs per
-// accounting packet. Fails open: with the probe unavailable or erroring, the
-// balance path stays on.
+// accounting packet. Fails open: if the probe errors the balance path stays on.
+//
+// The query is issued directly rather than through RadiusStatements. It runs
+// once per refresh interval, so preparing it buys nothing, and a prepared
+// statement would have to be read while setupStmt's retry goroutine may still
+// be writing the pointer.
 func (h *PfAcct) refreshBalancesInUse() {
-	if h.anyNodeBalances == nil {
-		h.balancesInUse.Store(true)
+	if h.Db == nil {
+		h.setBalancesInUse(true)
 		return
 	}
 
 	var inUse bool
-	if err := h.anyNodeBalances.QueryRow().Scan(&inUse); err != nil {
+	if err := h.Db.QueryRow(anyNodeBalancesQuery).Scan(&inUse); err != nil {
 		logError(h.LoggerCtx, "anyNodeBalances: "+err.Error())
 		inUse = true
 	}
-	h.balancesInUse.Store(inUse)
+
+	h.setBalancesInUse(inUse)
+}
+
+// setBalancesInUse records the flag and, when the gate reopens after having
+// been shut, the instant it did. Sessions already in flight then have no
+// NodeSessionCache entry while pfacct has no accounting at all for the period
+// the gate was shut, so that instant bounds what may be charged for them; see
+// handleTimeBalance. balancesInUse starts out true, so a deployment that has
+// balances from the start never records an instant and keeps the pre-existing
+// behaviour exactly.
+func (h *PfAcct) setBalancesInUse(inUse bool) {
+	if !inUse {
+		h.balancesInUse.Store(false)
+		return
+	}
+
+	if !h.balancesInUse.Swap(true) {
+		h.balancesEnabledAt.Store(time.Now().Unix())
+	}
+}
+
+// untrackedChargeCap returns the most that may be charged for a session pfacct
+// holds no NodeSessionCache entry for, and whether such a cap applies at all.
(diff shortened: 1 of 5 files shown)


go/cmd/pfacct/balance_gate.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// startBalancesInUseRefresher primes the balances-in-use flag and keeps it
+// fresh; a balance assigned to a node is picked up within one refresh
+// interval.
+func (h *PfAcct) startBalancesInUseRefresher() {
+	h.refreshBalancesInUse()
+	go func() {
+		for {
+			time.Sleep(balancesInUseRefreshInterval)
```
codedude: Use time.Ticker instead (thread safe, better in general).
Also, this goroutine never ends... no bueno
```
	ticker := time.NewTicker(time.Second)
	defer ticker.Stop()
	done := make(chan bool)
	go func() {
		time.Sleep(10 * time.Second)
		done <- true
	}()
	for {
		select {
		case <-done:
			fmt.Println("Done!")
			return
		case t := <-ticker.C:
			fmt.Println("Current time: ", t)
		}
	}
```

codedude (dismissed):

go/cmd/pfacct/balance_gate.go
fdurand: Fixed in 38517bd7e9 — you were right on both counts, and the second one mattered more than the ticker.

The refresher is now a `time.Ticker` selected against a stop channel, with an idempotent `stopBalancesInUseRefresher` wired into main's SIGINT/SIGTERM handler (next to `rs.Shutdown`) and into the test's cleanup. `NewPfAcct` is called from main plus five test files, so the old sleep loop leaked one goroutine per construction, each still querying the database every five minutes.

While in there, the first probe no longer runs through a prepared statement either: it runs once per interval so preparing bought nothing, and reading that pointer raced with the retry goroutine `setupStmt` spawns when the initial `Prepare` fails.

fdurand:
Pushed 38517bd7e9 addressing the review, plus three things a second pass turned up. The PR description above is updated too — two of its claims were wrong.

**The reviewer's thread (replied inline).** The refresher is a `time.Ticker` selected against a stop channel now, with an idempotent stop wired into main's SIGINT/SIGTERM handler and the test cleanup. `NewPfAcct` is called from main plus five test files, so the old sleep loop leaked a goroutine per construction.

**Reopening the gate could lock a user out.** `handleTimeBalance` uses the `NodeSessionCache` entry to record the offset at which a node stopped being unregistered, so only post-registration time is charged. While the gate is shut that entry is never primed, so the first packet after it reopened charged the **whole** `AcctSessionTime`. Concretely: a node sits in the portal from 09:00, registers at 09:20 with a one-hour balance, the gate reopens by 09:25, and the next packet bills ~1500s instead of ~300s — enough for `SoftNodeTimeBalanceUpdate` to zero the fresh balance and fire `ACCOUNTING_POLICY_TIME` against someone who just paid for an hour. Sessions with no cache entry are now charged at most the time since the gate reopened, which is all pfacct has accounting for. `balancesInUse` starts out `true`, so a deployment with balances from the start records no reopen instant and behaves exactly as before.

**The test wiped every node's balances.** The setup ran `UPDATE node SET time_balance = NULL, bandwidth_balance = NULL WHERE time_balance IS NOT NULL OR bandwidth_balance IS NOT NULL` — unscoped, and never restored. This suite needs a live database, so on any real server it silently destroyed every guest's remaining time and bandwidth. It now asserts the transitions its own node causes against whatever baseline the database holds. Verified on a PF server: balance count identical before and after the run.

**Probe halved and de-raced.** It was two `EXISTS` subqueries over unindexed columns, and the no-balances case this PR targets is precisely the one that cannot short-circuit — so it scanned the node table twice every five minutes. Now one predicate, and issued via `Db.QueryRow` rather than a prepared statement: it runs once per interval so preparing bought nothing, and reading that pointer raced with the retry goroutine `setupStmt` spawns when the initial `Prepare` fails.

**Corrected in the description.** The claim that gating "cannot suppress a legitimate trigger" was not right: within one interval of the *first* balance being assigned, interims are skipped and a `Stop` in that window skips the debit permanently. Also now stated explicitly that the gate is deployment-wide, so one guest voucher restores the full cost everywhere.

Verified: `go build` / `go vet` clean; the gate, cap and stop tests pass on a PF server. The `TestPacketServer_*` failures on that box are timing-related under load and reproduce identically on unmodified devel.

go/cmd/pfacct/balance_gate.go
```diff
@@ -0,0 +1,108 @@
...
+	"time"
+)
+
+const balancesInUseRefreshInterval = 5 * time.Minute
+
+// anyNodeBalancesQuery probes whether any node carries a time or bandwidth
+// balance. One predicate rather than two EXISTS subqueries: neither column is
+// indexed, and the case this gate optimises for (no balances anywhere) is the
+// one that cannot short-circuit, so two subqueries scan the node table twice.
+const anyNodeBalancesQuery = `SELECT EXISTS(SELECT 1 FROM node WHERE time_balance IS NOT NULL OR bandwidth_balance IS NOT NULL)`
```
jrouzierinverse: Create an index on time_balance and bandwidth_balance.

jrouzierinverse:
rebase

fdurand:
Both findings were real; fixed in 7b6cf2db80. Rebased on devel too.

**Gate publication ordering (`balance_gate.go:56`).** Correct: `Swap(true)` made the gate visible before `balancesEnabledAt` was stored, so a worker entering the time-balance path in that window read `since == 0`, got no cap from `untrackedChargeCap`, and charged the whole cumulative `AcctSessionTime`. The instant is now stored first and the gate opened after. Only `refreshBalancesInUse` calls this, from the refresher goroutine or before it starts, so the load/store pair needs no CAS -- and a CAS would leave the same window anyway, since the problem is the order of two separate fields, not the atomicity of one.

**Cached sessions bypassing the cap (`radius.go:225`).** Also correct, and it is the more damaging of the two. `NodeSessionCache` entries expire on a 12h idle timeout that only `getNodeSessionFromCache` refreshes, and nothing touches them while the gate is shut -- so any shutdown shorter than that leaves the entry in place, carrying an offset recorded before the shutdown. `AcctSessionTime - ns.timeBalance` then still spans the whole gated-off period. On the interim path that is worse than an over-charge: `softNodeTimeBalanceUpdate` is `UPDATE node SET time_balance = 0 WHERE mac = ? AND time_balance <= ?`, a threshold rather than a subtraction, so an inflated figure zeroes a balance that was only just assigned and fires the security event immediately -- the lockout this branch was supposed to prevent, reached by a different route.

The cap now applies with or without a cache entry, through a `capCharge` helper shared by the three call sites. It still returns the charge untouched until the gate has actually reopened after being shut, so deployments that have carried balances all along are unaffected.

**On the test for the ordering.** The first version raced a goroutine against the window and passed against the buggy code -- the window is a couple of instructions wide. `TestBalancesEnabledAtIsStoredBeforeTheGateOpens` instead observes the order directly, through what is already published when the instant's clock is read (`time.Now` indirected as `balanceGateNow`). Verified both ways: it fails on the previous ordering and passes on this one. `TestCapCharge` covers the helper's semantics.

The `radius.go` call sites themselves stay uncovered -- they need a live database and RADIUS packets. `go build`, `go vet` and the balance gate tests under `-race` were run on a test box (`go test ./cmd/pfacct/` cannot run on a dev machine: the package `init` parses `/usr/local/pf/raddb/dictionary.pfacct`).

PR #9265, opened 2026-09-04
fix(radius): add Message-Authenticator to non-EAP (MAB) Access-Accept

fdurand:
# Description
FreeRADIUS only adds a `Message-Authenticator` to an Access-Accept automatically when the request carried an `EAP-Message`. MAB is non-EAP, so the Access-Accept went out without attribute 80, and newer NAS firmware that requires it (e.g. Juniper Junos for MAB) rejected the session.

This adds a `Message-Authenticator := 0x00` placeholder in the `packetfence` virtual server post-auth section for non-EAP replies (`if (!EAP-Message)`). The value is a placeholder: radiusd computes the real HMAC-MD5 at send time. EAP flows are unaffected, FreeRADIUS already handles them.

# Impacts
- RADIUS authorize/post-auth workflow of the `packetfence` virtual server (`conf/radiusd/packetfence.example`): every non-EAP Access-Accept (MAB, and any other non-EAP authentication) now carries `Message-Authenticator`.
- To test: a MAB authentication against a NAS that enforces attribute 80 in the accept (Junos) must now succeed; check with `radsniff` or a `tcpdump` that the Access-Accept for a MAB request contains attribute 80 and that an 802.1X/EAP accept still contains exactly one.
- `radiusd-auth` must be restarted (the `.example` is rendered into the raddb at configuration time).

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature
- [no] Add unit tests (FreeRADIUS configuration change, no unit test harness for it)

# NEWS file entries
## Bug Fixes
* MAB (non-EAP) Access-Accept is sent without a Message-Authenticator, rejected by NAS firmware that requires it (e.g. Juniper Junos)

### DIFF
--- a/conf/radiusd/packetfence.example
+++ b/conf/radiusd/packetfence.example
@@ -404,6 +404,18 @@ server packetfence {
 			}
 		}
 
+		# Force a Message-Authenticator into non-EAP (e.g. MAB) Access-Accept
+		# replies. FreeRADIUS only adds it automatically when the request
+		# carried an EAP-Message; MAB is non-EAP, so newer NAS firmware that
+		# requires attr 80 in the accept (e.g. Juniper Junos) would otherwise
+		# reject it. 0x00 is a placeholder; radiusd computes the real HMAC-MD5
+		# at send time.
+		if (!EAP-Message) {
+			update reply {
+				Message-Authenticator := 0x00
+			}
+		}
+
 		attr_filter.packetfence_post_auth
 		linelog
 		#


fdurand:
Closing: not needed on devel, the behaviour it adds is already provided by FreeRADIUS itself on every version PacketFence now supports.

**Why the issue was real when reported.** It was seen on PacketFence 13.0, which shipped FreeRADIUS 3.2.2. On that version only rlm_eap added `Message-Authenticator` to a reply, so a MAB (non-EAP) Access-Accept went out without attribute 80 and a NAS requiring it on responses (e.g. Junos) dropped the session. The hunk in this PR is a correct workaround for 13.x / 14.0 (FreeRADIUS 3.2.2).

**Why it is moot now.** FreeRADIUS 3.2.5 (July 2024, BlastRADIUS mitigations) changed the encoder in `src/lib/radius.c` to *always* prepend `Message-Authenticator` to replies to Access-Request and to every Access-Accept / Access-Reject / Access-Challenge on UDP, with no configuration involved; a duplicate placed in the reply list is skipped. PacketFence 14.1 moved to 3.2.6 and 15.x requires >= 3.2.7 (ships 3.2.11), so no supported install lacks it. Confirmed on a 15.2 box running 3.2.11 with the unmodified site file: every captured MAB Access-Accept already carries attribute 80 as its first attribute.

| PacketFence | FreeRADIUS shipped | attr 80 on MAB accepts |
|---|---|---|
| 13.0 – 14.0 | 3.2.2 | no |
| 14.1 | 3.2.6 | yes |
| 15.x | >= 3.2.7 | yes |

Note for anyone hitting this on 13.x/14.0: upgrade to 14.1+ or add the 12-line post-auth block from this PR to your site's `conf/radiusd/packetfence`. On devel the only transport where the block would change anything is RadSec (the encoder skips it over TLS), which is not what it was written for.

PR #9272, opened 2026-09-08
[pfcron] pfflow: remove consumer throughput ceiling and recover from stranded group offset

fdurand:
# Description
The pfcron `pfflow` task (Kafka `pfflows_events` → `network_events`) could not keep up with a large deployment: with ~30K devices on a virtualSwitch producing ~110K flows/min, only ~25-30K network events/min were pushed and the consumer lag grew without bound. Separately, when the `pfflows_events` topic was deleted/recreated behind the consumer group, the consumer stopped producing anything, silently.

Root causes and fixes (all in `go/cron`):

* **Synchronous offset commit per message.** `kafka.Reader` had no `CommitInterval`, so kafka-go committed after every `ReadMessage`: one broker round trip per flow, ~650 msg/s ceiling. Offsets are now committed asynchronously once per second.
* **Flush on the ingest goroutine with per-event queries.** Every minute the aggregator ran two node-role queries per network event (~50K queries for 25K events) while not reading Kafka, pausing consumption ~19 s per minute. The flush now runs in its own goroutine (`aggregator_flush.go`), and ip4log MAC and node-role lookups are batched in chunks of 1000 distinct keys (`resolveMissingMacs`, `UpdateNetworkEvents` / `PolicyLookup.LookupWithRoles`). Semantics of the per-flow/per-event queries are preserved.
* **Per-flow INFO log line** ("Received N flows", ~4 GB/day of pfcron.log) is now DEBUG, replaced by one INFO summary per window with counts and timings.
* **Stranded group offset.** When the committed offset is past the partition log end, kafka-go retries forever and, with no `ErrorLogger`, silently. kafka-go's loggers are now wired into pfcron logging and, after a 60 s read timeout, the group offset is compared with the log end; if it is past the end the reader is closed, the log-end offset is committed and the reader reconnects.
* `MarkSwitchAsSeen` is skipped when the flow header has no valid agent address (it was inserting an `invalid IP` row into `switch_observability`).

Verified on a 15.2 load-test box: after deploy the guard reset the stranded offset within a minute, then ~110K flows/min were consumed with a ~0.8 s flush (down from ~19 s) and consumer lag stayed in the hundreds of messages. pfcron CPU ~11%.

# Impacts
* pfcron `pfflow` task: Kafka consumer (`go/cron/pfflowjob.go`), aggregator/flush (`go/cron/aggregator.go`, `go/cron/aggregator_flush.go`), enforcement lookup (`go/cron/policy_lookup.go`, `go/cron/network_event.go`).
* Offsets are now committed asynchronously (1 s), so a pfcron crash can replay up to ~1 s of flows into `network_events`.
* Log volume of pfcron.log drops sharply on flow-heavy deployments; look for the new `pfflow aggregator:` summary line and for `consumer group ... is past the log end offset` when a topic reset is detected.
* Node role and ip4log lookups are now `IN (...)` queries of up to 1000 bindings each.

# Delete branch after merge
YES

# Checklist
- [ ] Add unit tests — no (existing `TestAggregator` covers the aggregation semantics; the go/cron test binary needs a live pfconfig socket)

# NEWS file entries
## Enhancements
* pfcron pfflow: commit Kafka offsets asynchronously and batch the DB lookups of the network event flush, removing the ~650 flows/s ceiling of the pfflows_events consumer
* pfcron pfflow: replace the per-flow INFO log line with a per-window summary

## Bug Fixes
* pfcron pfflow: recover automatically when the consumer group offset is past the end of a recreated pfflows_events topic instead of silently stalling forever
* pfcron pfflow: do not record an "invalid IP" switch in switch_observability for flows without an agent address

### DIFF
--- a/go/cron/aggregator.go
+++ b/go/cron/aggregator.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/inverse-inc/go-utils/log"
-	"github.com/inverse-inc/packetfence/go/db"
 )
 
 type EventKey struct {
@@ -390,14 +389,16 @@ func logPfFlow(ctx context.Context, header *PfFlowHeader, f *PfFlow) {
 }
 
 func (a *Aggregator) handleEvents() {
-	ctx := context.Background()
+	// A context carrying a logger: with a bare context every Log*f call
+	// rebuilds a logger (including a syslog dial) before filtering the line.
+	ctx := log.LoggerNewContext(context.Background())
 	ticker := time.NewTicker(a.timeout)
 	defer ticker.Stop()
 
 	go a.flusher(ctx)
 	defer close(a.flushChan)
 
-	stats := windowStats{}
+	stats := newWindowStats()
 loop:
 	for {
 		select {
@@ -407,17 +408,14 @@ loop:
 					continue
 				}
 
-				log.LogDebugf(ctx, "Received %d flows of FlowType %s", len(*pfflows.Flows), flowType(pfflows.Header.FlowType))
 				stats.messages++
 				stats.flows += len(*pfflows.Flows)
-				// The agent address is the exporter (switch) IP. Older collectors do not
-				// fill it in for NetFlow/IPFIX, in which case it prints as "invalid IP",
-				// and an sFlow exporter without an agent-ip sends 0.0.0.0; neither must
-				// be recorded as a switch.
-				if addr := pfflows.Header.AgentAddr; a.db != nil && addr.IsValid() && !addr.IsUnspecified() {
-					if err := db.MarkSwitchAsSeen(a.db, pfflows.Header.AgentAddr.String()); err != nil {
-						log.LogErrorf(ctx, "handleEvents: failed to mark switch %s as seen: %s", pfflows.Header.AgentAddr.String(), err.Error())
-					}
+				// The agent address is the exporter (switch) IP. Older collectors
+				// leave it unset for NetFlow/IPFIX ("invalid IP") and an sFlow
+				// exporter without an agent-ip sends 0.0.0.0; neither is a switch.
+				// Marked as seen once per window by the flusher, off this goroutine.
+				if addr := pfflows.Header.AgentAddr; addr.IsValid() && !addr.IsUnspecified() {
+					stats.noteAgent(addr.String())
 				}
 
 				for _, f := range *pfflows.Flows {
@@ -431,21 +429,25 @@ loop:
 			}
 		case <-ticker.C:
 			if len(a.events) == 0 {
-				stats = windowStats{}
+				if stats.messages > 0 {
+					log.LogDebugf(ctx, "pfflow aggregator: %d messages carried no flows this window", stats.messages)
+				}
+				stats = newWindowStats()
 				continue
 			}
 
 			batch := flushBatch{events: a.events, stats: stats, tickedAt: time.Now()}
 			a.events = make(map[EventKey][]PfFlow, len(batch.events))
-			stats = windowStats{}
+			stats = newWindowStats()
 
-			// Hand the window to the flusher. If the previous flush is still
-			// running we block here (bounded memory), which is visible as
+			// Hand the window to the flusher. The channel holds one window, so
+			// this only blocks when the flusher is already a full window behind;
+			// then ingestion pauses (bounded memory) and the delay shows up as
 			// consumer lag rather than silent loss.
 			select {
 			case a.flushChan <- batch:
 			default:
-				log.LogWarnf(ctx, "pfflow aggregator: previous flush still running, pausing ingestion until it completes")
+				log.LogWarnf(ctx, "pfflow aggregator: flusher is one window (%s) behind, pausing ingestion until the previous flush completes", a.timeout)
 				a.flushChan <- batch
 			}
 		case <-a.stop:
--- a/go/cron/aggregator.go
+++ b/go/cron/aggregator.go
@@ -1,11 +1,9 @@
 package maint
 
 import (
-	"cmp"
 	"context"
 	"database/sql"
 	"fmt"
-	"math"
 	"net/netip"
 	"time"
 
@@ -33,6 +31,7 @@ func NewAggregator(o *AggregatorOptions) *Aggregator {
 		PfFlowsChan:      ChanPfFlow,
 		Heuristics:       o.Heuristics,
 		db:               o.Db,
+		flushChan:        make(chan flushBatch, 1),
 	}
 }
 
@@ -57,37 +56,12 @@ type Aggregator struct {
 	timeout          time.Duration
 	Heuristics       bool
 	db               *sql.DB
+	flushChan        chan flushBatch
 }
 
 func emptyMac(mac string) bool {
 	return mac == "00:00:00:00:00:00" || mac == ""
 }
-
-func updateMacs(ctx context.Context, f *PfFlow, stmt *sql.Stmt) {
-	if !emptyMac(f.SrcMac) && !emptyMac(f.DstMac) {
-		return
-	}
-
-	var srcMac, dstMac string
-	err := stmt.QueryRowContext(ctx, f.SrcIp.String(), f.DstIp.String()).Scan(&srcMac, &dstMac)
-	if err != nil {
-		log.LogErrorf(ctx, "updateMacs Database Error: %s", err.Error())
-	}
-
-	if emptyMac(f.SrcMac) {
-		f.SrcMac = srcMac
-	}
-
-	if emptyMac(f.DstMac) {
-		f.DstMac = dstMac
-	}
-}
-
-const updateMacsSql = `
-SELECT
-	COALESCE((SELECT mac FROM ip4log WHERE ip = ?), "00:00:00:00:00:00") as src_mac,
-	COALESCE((SELECT mac FROM ip4log WHERE ip = ?), "00:00:00:00:00:00") as dst_mac;
-`
 
 func flowType(t uint16) string {
 	switch t {
@@ -418,100 +392,62 @@ func logPfFlow(ctx context.Context, header *PfFlowHeader, f *PfFlow) {
 func (a *Aggregator) handleEvents() {
 	ctx := context.Background()
 	ticker := time.NewTicker(a.timeout)
-	stmt, err := new(sql.Stmt), error(nil)
-	//	if a.db != nil {
-	stmt, err = a.db.PrepareContext(ctx, updateMacsSql)
-	if err != nil {
-		log.LogErrorf(ctx, "handleEvents Database Error: %s %s", updateMacsSql, err.Error())
-		stmt = nil
-	} else {
-		defer stmt.Close()
-	}
-	//	}
+	defer ticker.Stop()
 
+	go a.flusher(ctx)
+	defer close(a.flushChan)
+
+	stats := windowStats{}
 loop:
 	for {
 		select {
-		case pfflowsArray := <-ChanPfFlow:
+		case pfflowsArray := <-a.PfFlowsChan:
 			for _, pfflows := range pfflowsArray {
-				log.LogInfof(ctx, "Received %d flows of FlowType %s", len(*pfflows.Flows), flowType(pfflows.Header.FlowType))
+				if pfflows.Flows == nil {
+					continue
+				}
+
+				log.LogDebugf(ctx, "Received %d flows of FlowType %s", len(*pfflows.Flows), flowType(pfflows.Header.FlowType))
+				stats.messages++
+				stats.flows += len(*pfflows.Flows)
 				// The agent address is the exporter (switch) IP. Older collectors do not
 				// fill it in for NetFlow/IPFIX, in which case it prints as "invalid IP",
 				// and an sFlow exporter without an agent-ip sends 0.0.0.0; neither must
 				// be recorded as a switch.
-				if addr := pfflows.Header.AgentAddr; addr.IsValid() && !addr.IsUnspecified() {
+				if addr := pfflows.Header.AgentAddr; a.db != nil && addr.IsValid() && !addr.IsUnspecified() {
 					if err := db.MarkSwitchAsSeen(a.db, pfflows.Header.AgentAddr.String()); err != nil {
 						log.LogErrorf(ctx, "handleEvents: failed to mark switch %s as seen: %s", pfflows.Header.AgentAddr.String(), err.Error())
 					}
 				}
-				for _, f := range *pfflows.Flows {
-					if stmt != nil {
-						updateMacs(ctx, &f, stmt)
-					}
 
-					key := f.Key(&pfflows.Header)
-					val := a.events[key]
+				for _, f := range *pfflows.Flows {
 					if a.Heuristics {
 						f.Heuristics()
 					}
 
-					a.events[key] = append(val, f)
+					key := f.Key(&pfflows.Header)
+					a.events[key] = append(a.events[key], f)
 				}
 			}
 		case <-ticker.C:
-			networkEvents := []*NetworkEvent{}
-			for _, events := range a.events {
-				startTime := int64(math.MaxInt64)
-				endTime := int64(0)
-				connectionCount := uint64(0)
-				var networkEvent *NetworkEvent
-				for _, e := range events {
-					networkEvent = e.ToNetworkEvent()
-					if networkEvent != nil {
-						break
-					}
-				}
-
-				if networkEvent == nil {
-					continue
-				}
-
-				ports := map[AggregatorSession]struct{}{}
-				for _, e := range events {
-					startTime = min(startTime, e.StartTime)
-					endTime = max(endTime, e.EndTime)
-					sessionKey := e.SessionKey()
-					if _, ok := ports[sessionKey]; !ok {
-						ports[sessionKey] = struct{}{}
-						connectionCount += e.ConnectionCount
-					}
-				}
-
-				networkEvent.Count = cmp.Or(int(connectionCount), len(ports))
-				if startTime != 0 {
-					networkEvent.StartTime = uint64(startTime)
-				}
-
-				if endTime != 0 {
-					networkEvent.EndTime = uint64(endTime)
-				}
-
-				if networkEvent.EndTime == 0 {
-					networkEvent.EndTime = networkEvent.StartTime
-				}
-
-				networkEvents = append(networkEvents, networkEvent)
+			if len(a.events) == 0 {
+				stats = windowStats{}
+				continue
 			}
 
-			for _, e := range networkEvents {
-				e.UpdateEnforcementInfo(ctx, a.db)
-			}
+			batch := flushBatch{events: a.events, stats: stats, tickedAt: time.Now()}
+			a.events = make(map[EventKey][]PfFlow, len(batch.events))
+			stats = windowStats{}
 
-			if len(networkEvents) > 0 && a.networkEventChan != nil {
-				a.networkEventChan <- networkEvents
+			// Hand the window to the flusher. If the previous flush is still
+			// running we block here (bounded memory), which is visible as
+			// consumer lag rather than silent loss.
+			select {
+			case a.flushChan <- batch:
+			default:
+				log.LogWarnf(ctx, "pfflow aggregator: previous flush still running, pausing ingestion until it completes")
+				a.flushChan <- batch
 			}
-
-			clear(a.events)
 		case <-a.stop:
 			break loop
 		}
--- a/go/cron/aggregator_flush.go
+++ b/go/cron/aggregator_flush.go
@@ -1,0 +1,201 @@
+package maint
+
+import (
+	"cmp"
+	"context"
+	"database/sql"
+	"math"
+	"strings"
+	"time"
+
+	"github.com/inverse-inc/go-utils/log"
+)
+
+// windowStats counts what the aggregator ingested during one flush interval.
+type windowStats struct {
+	messages int
+	flows    int
+}
+
+// flushBatch is one interval's worth of aggregated flows handed to the flusher.
+type flushBatch struct {
+	events   map[EventKey][]PfFlow
+	stats    windowStats
+	tickedAt time.Time
+}
+
+// flusher turns each batch into network events off the ingest goroutine so
+// that Kafka consumption keeps going while the (DB heavy) flush runs.
+func (a *Aggregator) flusher(ctx context.Context) {
+	for batch := range a.flushChan {
+		a.flush(ctx, batch)
+	}
+}
+
+func (a *Aggregator) flush(ctx context.Context, batch flushBatch) {
+	start := time.Now()
+	ipsLookedUp := resolveMissingMacs(ctx, a.db, batch.events)
+	macDone := time.Now()
+
+	networkEvents := buildNetworkEvents(batch.events)
+	buildDone := time.Now()
+
+	UpdateNetworkEvents(ctx, a.db, networkEvents)
+	rolesDone := time.Now()
+
+	if len(networkEvents) > 0 && a.networkEventChan != nil {
+		a.networkEventChan <- networkEvents
+	}
+
+	log.LogInfof(
+		ctx,
+		"pfflow aggregator: %d messages, %d flows, %d keys -> %d network events (ip4log lookups %d in %s, aggregate %s, policy lookup %s, total %s, queued %s)",
+		batch.stats.messages, batch.stats.flows, len(batch.events), len(networkEvents),
+		ipsLookedUp, macDone.Sub(start).Round(time.Millisecond),
+		buildDone.Sub(macDone).Round(time.Millisecond),
+		rolesDone.Sub(buildDone).Round(time.Millisecond),
+		time.Since(start).Round(time.Millisecond),
+		start.Sub(batch.tickedAt).Round(time.Millisecond),
+	)
+}
+
+// buildNetworkEvents collapses each aggregation key into one network event.
+func buildNetworkEvents(events map[EventKey][]PfFlow) []*NetworkEvent {
+	networkEvents := make([]*NetworkEvent, 0, len(events))
+	for _, flows := range events {
+		startTime := int64(math.MaxInt64)
+		endTime := int64(0)
+		connectionCount := uint64(0)
+		var networkEvent *NetworkEvent
+		for i := range flows {
+			networkEvent = flows[i].ToNetworkEvent()
+			if networkEvent != nil {
+				break
+			}
+		}
+
+		if networkEvent == nil {
+			continue
+		}
+
+		ports := map[AggregatorSession]struct{}{}
+		for i := range flows {
+			e := &flows[i]
+			startTime = min(startTime, e.StartTime)
+			endTime = max(endTime, e.EndTime)
+			sessionKey := e.SessionKey()
+			if _, ok := ports[sessionKey]; !ok {
+				ports[sessionKey] = struct{}{}
+				connectionCount += e.ConnectionCount
+			}
+		}
+
+		networkEvent.Count = cmp.Or(int(connectionCount), len(ports))
+		if startTime != 0 {
+			networkEvent.StartTime = uint64(startTime)
+		}
+
+		if endTime != 0 {
+			networkEvent.EndTime = uint64(endTime)
+		}
+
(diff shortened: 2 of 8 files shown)


go/cron/aggregator_flush.go
```diff
@@ -0,0 +1,201 @@
...
+
+	for _, flows := range events {
+		for i := range flows {
+			f := &flows[i]
+			if !emptyMac(f.SrcMac) && !emptyMac(f.DstMac) {
+				continue
+			}
+
+			if emptyMac(f.SrcMac) {
+				f.SrcMac = cmp.Or(macs[f.SrcIp.String()], "00:00:00:00:00:00")
```
jrouzierinverse, line 159: **An ip4log query failure now silently drops the window's events** — this is a behaviour change from `updateMacs`.

The old code left the MAC as `""` when the query errored:

```go
var srcMac, dstMac string
err := stmt.QueryRowContext(...).Scan(&srcMac, &dstMac)
if err != nil { log.LogErrorf(...) }   // srcMac/dstMac stay ""
if emptyMac(f.SrcMac) { f.SrcMac = srcMac }  // -> ""
```

and `ToNetworkEvent()` only returns nil when both MACs are literally `"00:00:00:00:00:00"`, so those flows still became events (with nil inventory items).

Now, if `lookupIp4logMacs` fails, it logs and `continue`s, `macs` comes back empty, and `cmp.Or` stamps every unresolved MAC with the zero MAC — so `ToNetworkEvent()` returns nil and the events are discarded. A single DB hiccup silently costs a full minute of events, and the new INFO summary just reports `0 network events` without indicating why.

Suggest applying the zero-MAC default only when the lookup actually succeeded, or threading a failure flag through so the window summary says so.

(Note `sql.ErrNoRows` was impossible in the old path because of the `COALESCE`, so an error there always meant a real DB error.)

go/cron/aggregator_flush.go
```diff
@@ -0,0 +1,201 @@
...
+			args[i] = ip
+		}
+
+		rows, err := dbh.QueryContext(ctx, query, args...)
+		if err != nil {
+			log.LogErrorf(ctx, "lookupIp4logMacs Database Error: %s", err.Error())
+			continue
+		}
+
+		for rows.Next() {
```
jrouzierinverse, line 187: `rows.Err()` is never checked. If iteration aborts mid-stream (connection drop, `max_allowed_packet`, …) this returns a silently partial map.

Combined with the zero-MAC fallback above, a partial result drops exactly the events for the IPs that didn't make it, with no log line at all. `rows.Close()` placement is fine — this is just the missing error check.

go/cron/policy_lookup.go
```diff
@@ -354,6 +360,78 @@ func UpdateNetworkEvent(ctx context.Context, db *sql.DB, ne *NetworkEvent) {
...
+			args[i] = mac
+		}
+
+		rows, err := db.QueryContext(ctx, query, args...)
+		if err != nil {
+			log.LogErrorf(ctx, "nodeRoles Database Error: %s", err.Error())
+			continue
+		}
+
+		for rows.Next() {
```
jrouzierinverse, line 419: Same missing `rows.Err()` check as in `lookupIp4logMacs`. A partial result here means those MACs silently fall through to the empty-role policy path, i.e. potentially the wrong enforcement decision rather than no decision.

Nit while you're here: `var mac, role string` on the next line (and `for mac := range macSet` / `for i, mac := range chunk` above) shadow the imported `mac` package. It compiles, but it reads oddly a few hundred lines below `mac.NewFromString`.

go/cron/pfflowjob.go
```diff
@@ -157,59 +209,198 @@ func WaitForTopic(dialer *kafka.Dialer, ctx context.Context, brokerAddr string,
...
 		if err != nil {
 			// Check if it's just a timeout (no messages available) vs a real error.
 			// kafka-go's ReadMessage wraps the error ("fetching message: %w"), so a
 			// plain == comparison never matches — use errors.Is to unwrap.
 			if errors.Is(err, context.DeadlineExceeded) {
-				// Timeout waiting for messages is normal, just retry without reconnecting
+				// No message for a whole readTimeout: either the topic is idle
+				// or the group offset is stranded past the end of a recreated
+				// topic, which kafka-go never recovers from on its own.
+				stale, err := j.staleGroupOffsets(ctx, client)
```
jrouzierinverse, line 380: **The stranded-offset repair only runs when the entire topic is silent for 60s.**

Because the check lives in the `context.DeadlineExceeded` branch, a single stranded partition alongside partitions that are still delivering never triggers it: `ReadMessage` keeps returning messages, the read never times out, and that partition stays stuck indefinitely.

If `pfflows_events` is single-partition in practice this is latent rather than live, but running `staleGroupOffsets` on its own ticker instead of only on read starvation would close the gap and also cover the case where the topic is busy at the moment of the reset.

go/cron/pfflowjob.go
```diff
@@ -157,59 +209,198 @@ func WaitForTopic(dialer *kafka.Dialer, ctx context.Context, brokerAddr string,
...
+	var err error
+	for attempt := 0; attempt < 5; attempt++ {
+		if attempt > 0 {
+			time.Sleep(2 * time.Second)
+		}
+
+		var resp *kafka.OffsetCommitResponse
+		resp, err = client.OffsetCommit(ctx, &kafka.OffsetCommitRequest{
+			GroupID:      j.GroupID,
+			GenerationID: -1,
```
jrouzierinverse, line 311: `GenerationID: -1` is the admin/simple-consumer commit path, and the broker rejects it unless the group is `Empty`. Closing *this* reader is sufficient only when this process is the group's sole member.

With more than one consumer in `j.GroupID`, the 5×2s retry will always exhaust, and the whole cycle — close the reader, fail the commit, rebuild the reader — repeats every 60s. Worth spelling out in the doc comment, and the `failed to reset consumer group %s offsets` log at the call site could hint at that as a likely cause.

go/cron/pfflowjob.go
```diff
@@ -157,59 +209,198 @@ func WaitForTopic(dialer *kafka.Dialer, ctx context.Context, brokerAddr string,
...
+	}
+
+	stale := []kafka.OffsetCommit{}
+	for _, p := range fetched.Topics[j.ReadTopic] {
+		if p.Error != nil {
+			return nil, p.Error
+		}
+
+		last, ok := lastOffsets[p.Partition]
+		if !ok || p.CommittedOffset <= last {
```
jrouzierinverse, line 282: `last` is trusted without a sanity check. `ListOffsets` pre-seeds `LastOffset = 0` for every requested partition and only overwrites it from the response, so a partition that is present in the request but absent from the response (with no error code) yields `last == 0` — and then any healthy committed offset is judged stale and rewound to 0, reprocessing the whole topic.

A conformant broker shouldn't produce that (a partition-level problem comes back as `po.Error`, which you already return on), but given the blast radius a cheap `if last <= 0 { continue }` guard seems worth it.

go/cron/policy_lookup.go
```diff
@@ -249,6 +249,12 @@ type PolicyLookup struct {
 func (l PolicyLookup) Lookup(ctx context.Context, db *sql.DB, ne *NetworkEvent) *EnforcementInfo {
 	srcMac, srcRole := ne.GetSrcRole(ctx, db)
 	dstMac, dstRole := ne.GetDstRole(ctx, db)
+	return l.LookupWithRoles(ne, srcMac, srcRole, dstMac, dstRole)
```
jrouzierinverse, line 252: After this change `Lookup` itself has no remaining callers, and neither do `UpdateNetworkEvent`, `NetworkEvent.UpdateEnforcementInfo`, `GetSrcRole`, `GetDstRole`, or `getRoleFromInventory` — in production or in tests. I grepped the whole `go/` tree; they only reference each other.

That leaves a second, unbatched, non-nil-db-safe copy of the role lookup that will drift from `UpdateNetworkEvents` over time. Worth deleting, or adding a note if they're being kept deliberately.

go/cron/aggregator.go
```diff
@@ -418,94 +392,58 @@ func logPfFlow(ctx context.Context, header *PfFlowHeader, f *PfFlow) {
...
+			batch := flushBatch{events: a.events, stats: stats, tickedAt: time.Now()}
+			a.events = make(map[EventKey][]PfFlow, len(batch.events))
+			stats = windowStats{}
 
-			if len(networkEvents) > 0 && a.networkEventChan != nil {
-				a.networkEventChan <- networkEvents
+			// Hand the window to the flusher. If the previous flush is still
+			// running we block here (bounded memory), which is visible as
+			// consumer lag rather than silent loss.
+			select {
```
jrouzierinverse, line 441: Nit: this is just a blocking send with a warning — the `default` case falls straight through to the same send. That's fine, but the warning fires on a backlog of exactly one window, so it'll be noisy if flushes routinely run close to the tick interval. A slightly higher-water-mark buffer, or wording that makes clear this is "one window behind" rather than a stall, might read better in the logs.

go/cron/aggregator.go
```diff
@@ -418,94 +392,58 @@ func logPfFlow(ctx context.Context, header *PfFlowHeader, f *PfFlow) {
...
-				if endTime != 0 {
-					networkEvent.EndTime = uint64(endTime)
-				}
-
-				if networkEvent.EndTime == 0 {
-					networkEvent.EndTime = networkEvent.StartTime
-				}
-
-				networkEvents = append(networkEvents, networkEvent)
+			if len(a.events) == 0 {
```
jrouzierinverse, line 429: Nit: `stats` is discarded here, so messages that carried empty flow slices (counted in `stats.messages` but contributing no keys) never show up in any INFO summary. Purely cosmetic, but it makes the per-window counters slightly lossy as a diagnostic.

jrouzierinverse (commented):
Reviewed the full diff, checked the kafka-go v0.4.47 API semantics against the vendored source, and ran build/vet/tests. Solid change — the diagnosis is convincing and the three throughput levers (async `CommitInterval`, flush off the ingest goroutine, batched ip4log/node-role queries) look like the right ones.

**Verification:** `go build ./cron/...` and `go vet ./cron/...` clean; `go test -count=1 ./cron/` passes, and the aggregator/policy tests also pass under `-race`.

**Things I checked and confirmed correct:**

- **Role-query equivalence.** The old `SELECT name FROM node_category WHERE category_id IN (SELECT category_id FROM node WHERE mac = ?)` returned `(mac, "")` for an unknown node, an unknown category, and a NULL `category_id`. The new `LEFT JOIN … COALESCE(name,'')` plus "absent from map yields `""`" matches all three. `LookupWithRoles`'s `if srcMac != ""` gate behaves identically too, since `inventoryMac()` is exactly the old early-return logic.
- **ip4log batching is safe.** `ip4log` has `PRIMARY KEY (ip)`, so the batched `WHERE ip IN (…)` cannot select a different row than the old per-IP `QueryRow`.
- **The reordering in `handleEvents` is safe.** MAC resolution now happens after `f.Key()` and `f.Heuristics()`, but neither reads `SrcMac`/`DstMac`, and `ToNetworkEvent()` still runs after `resolveMissingMacs`.
- **kafka-go usage is correct.** `LastOffsetOf` → high watermark; `OffsetFetch` returns `-1` for an uncommitted group, so `-1 <= last` correctly skips; and `OffsetCommit`/`OffsetFetch` implement `protocol.GroupMessage`, so the transport resolves the group coordinator on its own — leaving `Addr` unset is fine.
- **Two real bug fixes bundled in:** the `pfflows.Flows == nil` guard prevents a nil deref on `len(*pfflows.Flows)`, and the `AgentAddr.IsValid()` guard stops the "invalid IP" `switch_observability` row.

Inline comments below: one behaviour change I think is worth fixing, two missing error checks, three notes on the offset-repair path, and some nits.

One broader note: none of the new code paths are covered by tests — `resolveMissingMacs`, `nodeRoles`/`UpdateNetworkEvents`, or the offset comparison. The last is the easiest win: pulling the "committed vs. last" comparison out of `staleGroupOffsets` into a pure function would make it unit-testable without a broker.

go/cron/aggregator_flush.go
fdurand: Fixed in a5cf9bd0ff: `lookupIp4logMacs` now reports whether every chunk succeeded, and `applyResolvedMacs` only coalesces to the zero MAC when it did; on a failed chunk the IP keeps an empty MAC and the flow still becomes an IP-only event, as before. Covered by `TestApplyResolvedMacsIncompleteKeepsUnknownEmpty`.

go/cron/aggregator_flush.go
fdurand: Fixed in a5cf9bd0ff: both lookups now go through one `queryStringPairs` helper that checks `rows.Err()` after the loop (and flags the result as incomplete).

go/cron/policy_lookup.go
fdurand: Fixed in a5cf9bd0ff via the shared `queryStringPairs` helper (`rows.Err()` checked, no more `mac` shadowing). While here: the map was keyed by the DB-stored `node.mac` but read with the raw inventory MAC; MariaDB compares case-insensitively, the Go map does not, so an upper-case MAC from ip4log silently lost its role. Keys are now lower-cased on both sides (`LOWER(node.mac)` / `roleKey`).

go/cron/pfflowjob.go
fdurand: Fixed in a5cf9bd0ff: the reader now sets `OffsetOutOfRangeError: true`, so the broker's own out-of-range answer triggers the repair immediately from the read-error branch, and a periodic check (every 5 min after a successful read) covers a stranded partition while others keep delivering.

go/cron/pfflowjob.go
fdurand: Addressed in a5cf9bd0ff: the doc comment now says it (Empty group only, never with a second pfcron in the group), and a failed reset backs off for 5 minutes with a log line naming that cause instead of closing and rebuilding the reader every readTimeout.

go/cron/pfflowjob.go
fdurand: Fixed in a5cf9bd0ff: the comparison is now `isStrandedOffset(committed, last)`, which requires `last > 0`, with a table test. Two related changes: committed offsets are fetched *before* the log ends (both only grow, so the old order could produce a false 'reset' when another member committed in between), and the reset target is the partition's first offset rather than its end, since everything on a recreated topic is still unread.

go/cron/policy_lookup.go
fdurand: Removed in a5cf9bd0ff: `PolicyLookup.Lookup`, `UpdateNetworkEvent`, `NetworkEvent.UpdateEnforcementInfo`, `GetSrcRole`, `GetDstRole` and `getRoleFromInventory` are gone; `LookupWithRoles` + `nodeRoles` is the only role path left.

go/cron/aggregator.go
fdurand: Reworded in a5cf9bd0ff to say the flusher is one window (interval) behind and that ingestion pauses until the previous flush completes.

go/cron/aggregator.go
fdurand: Addressed in a5cf9bd0ff: when a window ends with no keys but `stats.messages > 0`, a debug line reports how many messages carried no flows. Also moved `MarkSwitchAsSeen` off the per-message path: the window collects its distinct exporters and the flusher marks them once per window.

fdurand:
Pushed a5cf9bd0ff addressing all nine review threads (replies inline), plus three things the second pass turned up:

- **Fetch-before-list**: committed offsets are read before the log ends. Both only grow, so a commit by another group member between the two calls can no longer look like a topic recreated behind the consumer.
- **Case-insensitive role map**: `node.mac` is compared case-insensitively by MariaDB but the Go map was not, so a MAC written upper-case into ip4log lost its role. Keys are lower-cased on both sides.
- **Reset target**: a stranded group is reset to the partition's first offset rather than its end, since everything on a recreated topic is still unread.

Tests added: `TestIsStrandedOffset`, `TestApplyResolvedMacs*`, `TestWindowStatsNoteAgent`, `TestRoleKeyIsCaseInsensitive`, `TestUpdateNetworkEventsWithoutDatabase`. Note that the cron package's tests need a live pfconfig even for pure functions; they pass (with the existing aggregator/policy tests) against one.

go/cron/pfflowjob.go
```diff
@@ -157,59 +234,240 @@ func WaitForTopic(dialer *kafka.Dialer, ctx context.Context, brokerAddr string,
...
+				// or the group offset is stranded past the end of a recreated
+				// topic, which kafka-go never recovers from on its own.
+				if !repairStrandedOffsets("idle topic") {
+					log.LogDebugf(ctx, "no message on %s for %s", j.ReadTopic, readTimeout)
+				}
+
+				continue
+			}
+
+			if errors.Is(err, kafka.OffsetOutOfRange) {
```
jrouzierinverse, line 488: **High — no backoff on this path, so a no-op repair becomes an unbounded reconnect loop.**

This branch neither increments `consecutiveErrors` nor sleeps, unlike the generic error path right below it. `repairStrandedOffsets` returns without doing anything in three reachable cases:

1. `repairBlockedUntil` is in the future after a failed commit — a second pfcron in the same group makes this effectively permanent (5 min of spinning, one attempt, repeat);
2. `staleGroupOffsets` errors out (metadata/OffsetFetch/ListOffsets failure, or any single partition returning `p.Error`, which aborts the whole check);
3. `isStrandedOffset` reports nothing stale (see the comment on `isStrandedOffset`).

In each case the loop falls through to `r == nil` → `j.newReader(...)` → `ReadMessage` → `OffsetOutOfRange` again, as fast as the broker can answer: a continuous connect/join-group/fetch/leave-group storm against the cluster, one error logged per iteration. Suggest applying the same exponential backoff here, or at minimum `consecutiveErrors++` plus a sleep when `repairStrandedOffsets` returns false.

go/cron/pfflowjob.go
```diff
@@ -116,6 +117,82 @@ func (j *PfFlowJob) kafkaDialer() *kafka.Dialer {
...
+
+// isStrandedOffset reports whether a committed group offset lies past the end
+// of its partition, which happens when the topic was deleted and recreated
+// behind the consumer: kafka-go then fetches from an offset that does not
+// exist and never advances. lastOffset <= 0 means the log end is unknown or
+// the partition is empty and is never treated as stranded (an empty answer
+// from ListOffsets must not rewind a healthy group to 0); committed < 0 means
+// the group has no offset for the partition.
+func isStrandedOffset(committed, lastOffset int64) bool {
+	return lastOffset > 0 && committed > lastOffset
```
jrouzierinverse: **Medium — this can never repair a topic that was recreated empty, which is the primary scenario the change targets.**

The `lastOffset > 0` guard means that immediately after a delete-and-recreate, before any producer writes, `ListOffsets` reports `LastOffset == 0`, `isStrandedOffset(5_000_000, 0)` is false, and no reset is issued. Meanwhile kafka-go's `initialize` fails with `OffsetOutOfRange` on the very first fetch (offset > last), and with `OffsetOutOfRangeError: true` that error reaches `ReadMessage` right away. Combined with the missing backoff on the `OffsetOutOfRange` branch, the exact failure this PR exists to fix becomes a hot loop that persists until a producer happens to write to the topic.

The guard's stated intent (an empty/failed `ListOffsets` answer must not rewind a healthy group) can be kept by distinguishing an errored or absent partition response from a partition that genuinely reports `FirstOffset == 0 && LastOffset == 0` with no error — the latter is a valid empty partition and can safely be reset to 0.

go/cron/aggregator.go
```diff
@@ -416,102 +389,67 @@ func logPfFlow(ctx context.Context, header *PfFlowHeader, f *PfFlow) {
...
-				ports := map[AggregatorSession]struct{}{}
-				for _, e := range events {
-					startTime = min(startTime, e.StartTime)
-					endTime = max(endTime, e.EndTime)
-					sessionKey := e.SessionKey()
-					if _, ok := ports[sessionKey]; !ok {
-						ports[sessionKey] = struct{}{}
-						connectionCount += e.ConnectionCount
-					}
+			if len(a.events) == 0 {
```
jrouzierinverse: **Medium — windows that aggregate to zero keys drop the collected exporter set, so those switches are never marked as seen.**

`markSwitchesSeen` is only reached through `a.flush`, and `flush` is only reached when `len(a.events) > 0`. Before this change `db.MarkSwitchAsSeen` was called once per Kafka message, unconditionally on the `AgentAddr`.

Now an exporter whose messages carry an empty `flows` array for a whole window — template/options-only NetFlow packets, or a switch with no traffic to report that minute — contributes to `stats.agents` and then has that set thrown away, so its `switch_observability` row goes stale and the switch reads as not-seen while it is actively exporting. Suggest calling `markSwitchesSeen` (or handing the flusher an agents-only batch) before the `continue`.

go/cron/policy_lookup.go
```diff
@@ -346,14 +347,61 @@ func StorePolicyLookup(p *PolicyLookup) {
...
+// nodeRoles maps each known MAC (lower-cased, see roleKey) to its role name
+// ("" when the node has no role). MACs absent from the node table are absent
+// from the result, which also yields "" on lookup, the same as the per-event
+// query did. macs must already be roleKey-normalized.
+func nodeRoles(ctx context.Context, db *sql.DB, macs []string) map[string]string {
+	if db == nil {
+		return map[string]string{}
+	}
+
+	roles, _ := queryStringPairs(ctx, db, "nodeRoles",
```
jrouzierinverse: **Low — discarding the `complete` flag means a failed chunk silently blanks the roles of up to 1000 MACs.**

`resolveMissingMacs` deliberately threads `complete` through so a failed query is not confused with "no row"; the role path drops it. When one chunk query fails (connection reset, killed query), every MAC in that chunk resolves to role `""`, so `LookupWithRoles` skips all role-based policies and the events go out with missing or wrong `EnforcementInfo` — indistinguishable from "these nodes genuinely have no role".

The per-event query this replaced had the same blind spot but a blast radius of one event; here it is a whole 1000-MAC chunk of a window. The error is logged inside `queryStringPairs`, but the batching makes it worth surfacing to the caller so the role-policy stage can be skipped rather than applied wrongly.

go/cron/pfflowjob.go
```diff
@@ -157,59 +234,240 @@ func WaitForTopic(dialer *kafka.Dialer, ctx context.Context, brokerAddr string,
...
+		}
+
+		log.LogWarnf(ctx, "resetting consumer group %s offsets (attempt %d): %s", j.GroupID, attempt+1, err.Error())
+	}
+
+	return err
+}
+
 func (j *PfFlowJob) Run() {
+	ctx := context.Background()
```
jrouzierinverse: **Low — bare `context.Background()` here is the exact cost the aggregator change was made to avoid.**

`log.LoggerWContext` rebuilds the logger (syslog backend included) whenever the context carries no `LoggerKey`, which is why `handleEvents` was changed to `log.LoggerNewContext(context.Background())` with a comment saying so. This `ctx` backs every log line in `Run` *and* is captured by the `Logger`/`ErrorLogger` closures handed to `kafka.NewReader`, which kafka-go invokes from its internals — notably `commitOffsets`, now running every `CommitInterval` (1s). `log.LoggerNewContext(context.Background())` here would keep the two files consistent.

jrouzierinverse (commented):
Reviewed both commits on this branch (8 files, +809/-175). Build, `go vet` and the new unit tests pass.

The first two findings interact: `isStrandedOffset` skips a recreated-but-still-empty topic, and the `OffsetOutOfRange` branch that follows has no backoff — so the scenario this PR targets degenerates into a reconnect hot loop instead of a repair. The aggregator one is a separate regression on switch-seen tracking. Details inline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

go/cron/pfflowjob.go
fdurand: Fixed in 656e1db726. When `repairStrandedOffsets` resets nothing, the OffsetOutOfRange branch now goes through the same exponential backoff as the generic read error (shared `reconnectBackoff` closure, 1s doubling to 60s, reset on the next successful read). The helper also returns `false` when the commit itself failed, so the cooldown case backs off too instead of being reported as a repair. The idle path was left alone: it only runs after the 60s read timeout, so its cadence is already bounded.

go/cron/pfflowjob.go
fdurand: Fixed in 656e1db726. The distinction you suggest is not available from the struct: kafka-go pre-seeds `LastOffset = 0` for every partition it asks about and only overwrites it from the response, so a partition absent from the answer and a genuinely empty one are identical by value. Instead `isStrandedOffset` takes a `brokerConfirmed` flag: `last == 0` is trusted only when the broker itself just answered OffsetOutOfRange for the group offset (that error branch is the only caller passing `true`), which proves the offset does not exist on the partition. The idle and periodic checks keep the conservative rule. kafka-go skips ahead on its own when the offset is *before* the first offset, so the confirmed error always means "past the end". `TestIsStrandedOffset` now covers both dimensions, including the empty recreated topic.

go/cron/aggregator.go
fdurand: Fixed in 656e1db726. The ticker only skips a window when no message at all arrived; a window with messages but no flows is handed to the flusher, which marks the exporters and returns before the ip4log/role work (with a debug line giving message and exporter counts). Marking stays off the ingest goroutine. No unit test for this one: `markSwitchesSeen` is a no-op without a database and the aggregator test needs pfconfig.

go/cron/policy_lookup.go
fdurand: Fixed in 656e1db726. `nodeRoles` returns the completeness flag and `UpdateNetworkEvents` logs an error when it is false. The per-event work moved to `applyEnforcement`: with complete roles it is unchanged; with incomplete roles only the MAC policies are applied and the role-based and implicit policies are skipped, since a role missing from the map can no longer be told from "no role" and the implicit fallback is only correct once the role policies were checked. Events of that window go out without enforcement info, as they do when nothing matches. Covered by `TestApplyEnforcementIncompleteRolesSkipsRolePolicies`.

go/cron/pfflowjob.go
fdurand: Fixed in 656e1db726: `Run()` uses `log.LoggerNewContext(context.Background())` with the same comment as `handleEvents`, so both the loop and the two kafka-go logger closures reuse one logger.

go/cron/pfflowjob.go
```diff
@@ -157,59 +234,240 @@ func WaitForTopic(dialer *kafka.Dialer, ctx context.Context, brokerAddr string,
 	}
 }
 
+// staleGroupOffsets returns, for every partition of the read topic whose
+// committed group offset is past the partition's last offset, the commit that
+// moves it back to the log end. This happens when the topic is deleted and
+// recreated behind the consumer: the group keeps its old (now unreachable)
+// offset and kafka-go waits forever for the log to catch up with it.
```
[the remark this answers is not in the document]
fdurand: Fixed in 656e1db726: the `staleGroupOffsets` comment now says the reset goes to the partition's first offset and why (everything on a recreated topic is still unread), and points to `isStrandedOffset` for the recognition rule.

fdurand:
Pushed 656e1db726 addressing the second round (replies inline on each thread):

- **OffsetOutOfRange hot loop**: the branch now backs off like any other read error when the repair resets nothing, and a failed commit counts as "nothing reset".
- **Empty recreated topic**: `isStrandedOffset` accepts a log end of 0 only when the broker itself just reported the offset out of range (`brokerConfirmed`, set on that error path only); the idle and periodic checks keep refusing 0, since kafka-go pre-seeds `LastOffset = 0` and an absent partition is indistinguishable from an empty one by value.
- **Switch-seen tracking**: windows with messages but no flows are handed to the flusher, which marks the exporters and returns.
- **Role lookup completeness**: a failed chunk now applies MAC policies only for that window and logs it, instead of evaluating role and implicit policies with possibly missing roles.
- Logger-carrying context in `Run()`, and the `staleGroupOffsets` doc comment now matches the first-offset reset.

Tests: `TestIsStrandedOffset` extended with the `brokerConfirmed` dimension, new `TestApplyEnforcementIncompleteRolesSkipsRolePolicies`. `go build`, `go vet` and the pure cron tests pass; the rest of the package still needs a live pfconfig.

PR #9276, opened 2026-09-09
fix(ci): check runner disk per volume, and make the reclaim actually reclaim for maintenance/15.2

JeGoi:
Same as https://github.com/inverse-inc/packetfence/pull/9274 but for maintenance/15.2


JeGoi:
https://gitlab.com/inverse-inc/packetfence/-/jobs/16402598882

PR #9284, opened 2026-09-11
fix(switch_observability): MySQL-compatible upsert and skip flows without an agent address

fdurand:
# Description
pfcron (flow aggregator) and pfacct call `db.MarkSwitchAsSeen`, whose upsert was written as `INSERT ... WITH cte AS (...) SELECT ...`. MariaDB accepts that, but a real MySQL backend (as used in cloud NAC behind ProxySQL) rejects it with:

```
Error 1064 (42000): You have an error in your SQL syntax; ... near 'cte AS (SELECT ? as switch_id)
```

MySQL 5.7 has no CTE support at all and `INSERT ... WITH ... SELECT` is not portable, so the statement is rewritten as a plain `INSERT ... VALUES ... ON DUPLICATE KEY UPDATE` with an `IF()` guard that only refreshes `visibility_timestamp` when it is NULL or older than the TTL. `RowsAffected` semantics are unchanged (1 inserted, 2 updated, 0 untouched), so the existing cache logic still works. `VALUES()` is avoided because MySQL 8.0.20+ deprecates it.

Two related hardenings:
- A failing query is negatively cached for 10 seconds, so a broken database is not hit and logged on every single flow batch / accounting request.
- The flow aggregator only calls `MarkSwitchAsSeen` when `Header.AgentAddr` is valid. Collectors that do not set it for NetFlow/IPFIX made pfcron try to record a switch literally named `invalid IP` (the `netip.Addr` zero value). The collector-side fix is in akainverse/fingerbank-collector (fill the agent address from the exporter's UDP source).

# Impacts
- `go/db/switch_observability.go`: `MarkSwitchAsSeen` SQL and error path. Check on both MariaDB and MySQL that the row is inserted, refreshed after 1 minute, and left untouched (0 rows affected) within the minute.
- `go/cron/aggregator.go`: pfcron pfflow handling; no `handleEvents: failed to mark switch invalid IP as seen` errors on NetFlow/IPFIX batches.
- pfacct accounting path (same function, unchanged call site).

# Delete branch after merge
YES

# Checklist
- [n/a] Document the feature
- [no] Add unit tests

# NEWS file entries
## Bug Fixes
* switch_observability upsert failed with SQL error 1064 on MySQL backends (CTE in INSERT ... SELECT)
* pfcron tried to mark a switch named "invalid IP" as seen for NetFlow/IPFIX flows without an agent address

### DIFF
--- a/go/db/switch_observability.go
+++ b/go/db/switch_observability.go
@@ -14,6 +14,36 @@ var (
 	switchObservabilityNextAllowed = NewShardedCache[time.Time](16)
 )
 
+// markSwitchAsSeenSQL is a plain upsert, kept identical to $sql_mark_as_seen in
+// pf::Switch (lib/pf/Switch.pm): both processes write this table and must agree
+// on the statement (TestPerlMarkAsSeenStatementMatchesGo enforces it).
+//
+// The previous INSERT ... WITH cte ... SELECT ... ON DUPLICATE KEY UPDATE ...
+// VALUES() form is accepted by MariaDB 10.11 and MySQL 8.0/8.4 but rejected
+// with error 1064 by MySQL 5.7, which has no CTE support. This form runs on all
+// four (see switch_observability_integration_test.go) with the same
+// RowsAffected contract the callers rely on: 1 = inserted, 2 = refreshed,
+// 0 = row exists and is still fresh (left untouched). go/db never enables
+// CLIENT_FOUND_ROWS and pf::db sets mysql_client_found_rows=0, so 0 really
+// means untouched on both sides.
+const markSwitchAsSeenSQL = `INSERT INTO switch_observability (switch_id, visibility_timestamp)
+VALUES (?, NOW())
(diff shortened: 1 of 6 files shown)


go/db/switch_observability.go
```diff
@@ -27,16 +32,29 @@ func MarkSwitchAsSeen(db *sql.DB, switchID string) error {
...
 		switchID,
 		switchObservabilityCacheTTLInMinutes,
 	)
 
 	if err != nil {
+		// Remember the failure for a short while so a broken database does not
+		// get hit (and logged) again on every single flow batch / accounting request.
+		shard.Lock()
+		shard.Set(switchID, time.Now().Add(-1*(switchObservabilityCacheTTL-switchObservabilityErrorBackoff)))
+		shard.Unlock()
```
codedude: `defer shard.Unlock()` must be prefered over manual call

codedude (changes requested):

PR #9287, opened 2026-09-11
fix(radiusd): cache dynamic clients for 300s instead of per-packet lookups

fdurand:
## Problem

`raddb/sites-available/dynamic-clients` ships `lifetime = "0"`. Our FreeRADIUS build carries fork commit b04cb25fe ("Remove client if lifetime equals to 0"), which deletes a lifetime-0 dynamic client after every request instead of caching it forever as upstream does. Net effect: every RADIUS packet re-runs the `dynamic_clients` virtual server, i.e. three `rlm_sql` queries against `radius_nas`, a client add, and the BlastRADIUS warning block in radius.log.

Measured on a load-test box at ~140 authz/s: ~490,000 "Adding client" lines per hour, `radmin show client list` holding zero switches, ~480 extra SQL queries/s on the request path.

## Fix

`lifetime = 300` in `raddb/sites-available/dynamic-clients` and in the pfconnector variant. A switch is resolved once and cached; a changed secret/type in `radius_nas` is picked up within 300 s. The pfconnector variant reads flat files synced from the server; FreeRADIUS renews an unchanged file at expiry (mtime check) and only re-reads a changed one.

Companion: inverse-inc/freeradius-server `fix/dynamic-clients-lifetime-0-revert` reverts the fork patch so `lifetime = 0` means "cached until restart" again, as the file's comments say. This config change is effective on the current binary on its own, because the fork patch only fires when lifetime is exactly 0.

## Validation (2026-09-11, ~140 authz/s, 15-min windows)

| | before | after |
|---|---|---|
| "Adding client" lines | ~122,000 | 18 (6 switches × one refresh per 300 s) |
| MariaDB queries/s | 19,800 | 3,400 (with the buffer-pool change from the same session) |
| rsyslogd CPU | 0.8 cores | 0.27 cores |

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -29,7 +29,7 @@ For a list of compatibility related changes see the <<PacketFence_Upgrade_Guide.
 
 === Enhancements
 
-  - RADIUS dynamic clients (switches resolved from radius_nas) are now cached for 300 seconds instead of being re-resolved from the database on every packet
+  - RADIUS dynamic clients (switches resolved from radius_nas) are now cached for 300 seconds. With the FreeRADIUS build shipped by PacketFence, `lifetime = 0` deleted the client after every request and re-ran the radius_nas lookups for every packet (fork commit b04cb25fe, reverted separately in inverse-inc/freeradius-server#3). A changed switch secret or type now applies to an already-known switch within five minutes
 
 === Bug Fixes
 
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -29,6 +29,8 @@ For a list of compatibility related changes see the <<PacketFence_Upgrade_Guide.
 
 === Enhancements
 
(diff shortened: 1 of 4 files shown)


fdurand:
Re the review note about existing installations keeping `lifetime = "0"`:

`raddb/sites-available/dynamic-clients` is a dpkg conffile and an rpm `%config(noreplace)` file. Both package managers **replace it automatically when it is unmodified** (the on-disk checksum matches the previous package's default), which is the common case; only a locally modified copy is kept, with the new default shipped alongside (`.rpmnew` / dpkg conffile prompt). Preserving intentional local changes is exactly the behaviour we want, so no migration script rewrites the file.

Added a section to the Upgrade Guide ("Upgrading from a version prior to X.Y.Z") covering the customized-file case: set `lifetime = 300` in the `client dynamic` block and restart `packetfence-radiusd-auth`, plus the five-minute propagation delay for `radius_nas` changes on an already-known switch.

PR #9288, opened 2026-09-11
fix(radiusd): do not answer Access-Reject when the post-auth REST call fails

fdurand:
## Problem

In the `packetfence` post-auth section the `rest` call to httpd.aaa ran unguarded. When httpd.aaa was unreachable, timed out or answered 5xx, rlm_rest returned `fail`, the section failed and FreeRADIUS sent an **Access-Reject**. A NAS treats that as authoritative: on a periodic reauthentication it tears down a session it had already authorized. On a load-test site (5 switches × 10k ports, MAB reauth every 300 s) every httpd.aaa restart or stall disconnected thousands of devices; PF's own audit log showed 1050 rejects in one hour for a single NAS during two short restarts.

## Fix (after review)

- **Scope: MAB / PAP only.** For requests without EAP the REST failure now leaves the request unanswered (`do_not_respond`); the NAS retransmits and, if PacketFence stays silent, applies its own dead-server policy (e.g. Cisco critical-auth) while the existing session is untouched. **EAP keeps the Reject**: rlm_eap has already freed the conversation when post-auth runs, so a retransmit cannot be resumed and silence would only delay the Reject. The inner PEAP/TTLS tunnels are unchanged (a `handled` there would have produced a fail-open Access-Accept).
- **One shared policy**, `packetfence-post-auth-rest` in `raddb/policy.d/packetfence`: `rest { fail = 1 }`, User-Password scrubbed on both paths, `reject` for EAP, otherwise `packetfence-audit-log-no-response` + the existing `do_not_respond` policy.
- **Audit trail kept**: a `NoResponse` row (same shape as accept/reject, Module-Failure-Message in `reason`) is written for every silenced request, so an httpd.aaa outage stays visible in the RADIUS audit log instead of becoming a gap.
- **503 for infrastructure failures** in `lib/pf/radius/rest.pm`: `RLM_MODULE_FAIL` flagged `RADIUS_FAILURE` by pf::radius (the two "Database is unavailable" sites) and a missing result after a handler exception now answer 503 → rlm_rest `fail` → same policy. Policy denials ("Switch is not managed", "Mac is empty", USERLOCK, …) keep 401/deny. Unit test: `t/unittest/radius_rest_format_response.t` (15 checks).
- NEWS spells out the scope, the dead-server / critical-auth consequence and the 503 change. eduroam / pfconnector sites are EAP-only and untouched.

## Validation (172.235.26.61, ~130 authz/s)

| httpd.aaa restart | rest failures | Access-Rejects | max_request_time stalls | audit rows |
|---|---|---|---|---|
| before the branch (18:07) | many | yes → +1202 disconnected ports on one switch | yes | Reject rows |
| this branch (00:17) | 1869 | **0** | **0** | **1869 NoResponse** (reason: `rest: Request failed: 28 - Timeout was reached`) |

`freeradius -Cxm` OK; unit test green in the pfperl-api container.

Companion on the simulator side: akainverse/virtualSwitch#90 (a reject on reauth no longer unplugs the device).

### DIFF
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -31,7 +31,8 @@ For a list of compatibility related changes see the <<PacketFence_Upgrade_Guide.
 
 === Bug Fixes
 
-  - RADIUS: when the post-auth REST call to httpd.aaa fails (unreachable, timeout, 5xx), radiusd no longer answers with an Access-Reject but stays silent so the NAS retries or fails over; an Access-Reject tore down sessions the switch had already authorized on every httpd.aaa restart or stall
+  - RADIUS: for MAB and PAP requests, when the post-auth REST call to httpd.aaa fails (unreachable, timeout, 5xx) radiusd no longer answers with an Access-Reject but leaves the request unanswered, logging a "NoResponse" row in the RADIUS audit log. An Access-Reject tore down sessions the switch had already authorized on every httpd.aaa restart. EAP requests keep the Reject (the EAP conversation cannot be resumed on retransmit). Note: a NAS counts unanswered requests towards its dead-server criteria, so a stalled PacketFence now triggers the switch's fail-over behaviour (e.g. Cisco critical-auth) instead of de-authorizing ports
+  - RADIUS: /radius/rest/authorize now answers 503 instead of 401 when the request could not be evaluated (database unavailable, handler exception), so radiusd treats it as a failure rather than a deny; a database outage no longer turns every reauthentication into an Access-Reject
 
 === Security Fixes
 
--- a/NEWS.asciidoc
+++ b/NEWS.asciidoc
@@ -31,6 +31,8 @@ For a list of compatibility related changes see the <<PacketFence_Upgrade_Guide.
 
 === Bug Fixes
 
+  - RADIUS: when the post-auth REST call to httpd.aaa fails (unreachable, timeout, 5xx), radiusd no longer answers with an Access-Reject but stays silent so the NAS retries or fails over; an Access-Reject tore down sessions the switch had already authorized on every httpd.aaa restart or stall
+
 === Security Fixes
 
 == Version 15.2.0 released on 2026-08-24
--- a/conf/radiusd/packetfence-tunnel.example
+++ b/conf/radiusd/packetfence-tunnel.example
@@ -279,7 +279,20 @@ post-auth {
 		}
 	}
 	if !( ("%{client:shortname}" =~ /eduroam_tlrs/)  || (&request:PacketFence-ShortName && &request:PacketFence-ShortName =~ /eduroam_tlrs/)) {
-		rest
+		rest {
+			# httpd.aaa unreachable, timed out or answering 5xx: do not reply at all.
+			# Letting the section fail would send an Access-Reject, which tears down
+			# a session the NAS had already authorized (seen: every httpd.aaa restart
+			# disconnected thousands of devices on periodic reauth). Silence lets the
+			# NAS retry or fail over instead.
+			fail = 1
+		}
+		if (fail) {
+			update control {
+				&Response-Packet-Type := Do-Not-Respond
+			}
+			handled
+		}
 	}
 	update {
 		&request:User-Password := "******"
@@ -617,7 +630,20 @@ session {
 #  the ONLY safe way is to set "use_tunneled_reply = yes", and
 #  then update the inner-tunnel reply.
 post-auth {
-	rest
+	rest {
+		# httpd.aaa unreachable, timed out or answering 5xx: do not reply at all.
+		# Letting the section fail would send an Access-Reject, which tears down
+		# a session the NAS had already authorized (seen: every httpd.aaa restart
+		# disconnected thousands of devices on periodic reauth). Silence lets the
+		# NAS retry or fail over instead.
+		fail = 1
+	}
+	if (fail) {
+		update control {
+			&Response-Packet-Type := Do-Not-Respond
+		}
+		handled
+	}
 	update {
 		&request:User-Password := "******"
 	}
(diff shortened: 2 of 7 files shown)


codedude (approved):

fdurand:
On the two `RADIUS_FAILURE` findings (`lib/pf/radius.pm:217` and `:988`): the consumer they rely on is dead configuration. `raddb/mods-config/perl/packetfence.pm` is defined and loaded, but nothing invokes it.

It is declared in `raddb/mods-available/perl:85-93`:

```
perl packetfence {
	filename = ${modconfdir}/perl/packetfence.pm
}
```

and `mods-enabled/perl` symlinks that file, so radiusd instantiates the module at startup. But no virtual server or policy calls it — grepping every `conf/radiusd/*.example`, `raddb/sites-available/*`, `raddb/sites-enabled/*` and `raddb/policy.d/*` for a bare `packetfence` module invocation returns nothing:

```
$ grep -rnE "^[[:space:]]*(packetfence|packetfence\.[a-z]+)[[:space:]]*$" conf/radiusd/ raddb/
$ (no output)
```

What the main virtual server actually calls (`conf/radiusd/packetfence.example`):

```
:337   rest                            # accounting
:396   packetfence-post-auth-rest      # post-auth (the policy this PR adds)
:112   packetfence-multi-domain        # rlm_perl
:694   packetfence-multi-domain        # rlm_perl
:545   reply_in_db                     # rlm_perl
```

The two rlm_perl modules that *are* called do not consume a `radius_authorize` tuple. `packetfence-multi-domain.pm::authorize` only resolves the realm and sets `PacketFence-NTLM-Auth-Host` / `PacketFence-NTLM-Auth-Port` before returning `RLM_MODULE_UPDATED`; it never merges attributes into `%RAD_REPLY`. The `%RAD_REPLY = (%RAD_REPLY, %$attributes)` merge that only strips `RADIUS_AUDIT` exists solely in `packetfence.pm`, which is on the legacy path: its last three commits are copyright bumps and its header still documents the SOAP / `RPC_PORT => 9090` era.

So on a stock installation `RADIUS_FAILURE` can never reach `%RAD_REPLY`. The only way to hit it is a site running a customized raddb that still wires up the module — the rpm ships `mods-config` as `%config(noreplace)`, so such a local edit does survive upgrades. If we want insurance for that case, it is one line in `packetfence.pm` next to the existing `delete $attributes->{RADIUS_AUDIT}`:

```perl
delete $attributes->{RADIUS_FAILURE};
```

Happy to add it, but it is not a defect in any supported path.

The other two findings (CHAP/MS-CHAP falling into `do_not_respond`, and the NoResponse audit rows missing their `control:*` context because rlm_rest does not process the 503 body) are being looked at separately.

fdurand:
Addressed the audit-context finding (`raddb/policy.d/packetfence:251`) in fd54e37223.

The observation is correct: rlm_rest does not process the body of a 5xx answer (`5xx  server error  Process body no  fail`, per the status code table in `conf/radiusd/rest.conf`), so the `control:PacketFence-*` pairs `format_response` builds from `RADIUS_AUDIT` never reach radiusd on this path, and the NoResponse rows landed with every PacketFence-derived column empty.

`packetfence-audit-log-no-response` now fills in locally the two fields radiusd knows first-hand, before the Redis push:

```unlang
update control {
    &PacketFence-Request-Time := "%l"
    &PacketFence-Switch-Ip-Address := "%{request:NAS-IP-Address}"
}
request-timing
```

Neither is a guess:

* **Request time** — `lib/pf/radius.pm:444` sets `$args->{'time'} = time`, an absolute epoch, and the existing `request-timing` policy subtracts `control:Tmp-Integer-0` (the `%l` stamped in authorize, `conf/radiusd/packetfence.example:26`) to get the elapsed time. Assigning `%l` here feeds that same policy the same shape, so the column carries a real duration.
* **Switch IP** — `packetfence-nas-ip-address` runs first in authorize and guarantees `request:NAS-IP-Address` is set, defaulting it to `Packet-Src-IP-Address`. That is exactly the value PacketFence keys its switch lookup on (`_ip` → `PacketFence-Switch-Ip-Address` in `%SWITCH_ATTRIBUTES_TO_RADIUS_ATTRIBUTES`), so this reproduces it rather than approximating it.

The rest are deliberately left to the consumer's `N/A` default: role, status, profile, source, ifIndex, connection type, switch id, switch MAC and computer name are decisions PacketFence makes, and on this path it never evaluated the request. Deriving ifIndex from `NAS-Port` or the connection type from `EAP-Type` in unlang would duplicate `pf::Connection` and the per-vendor switch logic, and would be wrong often enough to make the audit log misleading. The request-side identity is not lost either way: `nas_ip_address`, `nas_port`, `called_station_id`, `calling_station_id` and the packet source IP all come from the request JSON the row already carries.

On the suggested alternative of preserving the context out of band from httpd.aaa: that only covers one of the three failure modes. When httpd.aaa is unreachable or times out, PacketFence never sees the request, so there is no context on that side to preserve and no row to write — radiusd is the only place the NoResponse row can be produced reliably.

Note this is unlang, so it is not covered by the unit tests in this PR; it wants a `radiusd -XC` plus a live httpd.aaa outage on a test box before merge.

fdurand:
On the CHAP/MS-CHAP scope finding (`raddb/policy.d/packetfence:287`): catching them is intended, so I took the second half of the suggestion and made the comment say so (ae470937cf) rather than narrowing the condition.

The premise of this PR is that an Access-Reject is destructive *because* the NAS treats it as authoritative on a periodic reauthentication and tears down a session it had already authorized. That harm is specific to MAB and 802.1X, which is what a switch reauthenticates live sessions with.

Direct `Auth-Type CHAP` / `MS-CHAP` in the main virtual server is a different shape. Per `conf/radiusd/packetfence.example`:

```unlang
Auth-Type MS-CHAP {
    if(NAS-Port-Type =~ /^Wireless/ || NAS-Port-Type =~ /^Ethernet/) { ok }
    else { packetfence-mschap-authenticate }
}
```

The branch that actually reaches PacketFence is the non-Wireless/Ethernet one: interactive CLI and VPN logins. Those are one-shot authentications with a person at a prompt, not a reauthentication of an established session, so nothing gets torn down whichever way we answer. A drop means the client retries and the user waits a little longer -- they are already waiting on a password prompt and typically willing to retry -- while a Reject during an httpd.aaa outage would tell them their credentials are wrong. Neither is harmful, and the drop is the marginally better of the two.

So the behaviour stays as is and the comment now describes the real scope instead of claiming MAB/PAP only.

jrouzierinverse:
Rebase

PR #9292, opened 2026-09-16
Update detection_mecanism_urls to include new Firefox detection URL

heapdavid:
# Description
Firefox is updating their captive portal detection url, ref https://support.mozilla.org/en-US/kb/captive-portal and https://bugzilla.mozilla.org/show_bug.cgi?id=2049252

# Impacts
pfdns/passthroughs

# Delete branch after merge

YES

# Checklist
(REQUIRED) - [yes, no or n/a]
- [n/a] Document the feature
- [n/a] Add unit tests
- [n/a ] Add acceptance tests (TestLink)


heapdavid:
> Add the trailing slash so Firefox detection requests are recognized correctly.

I was unsure about whether it needed that or not, so followed the pattern of the other pre-existing root only entries: `http://www.appleiphonecell.com,http://captive.apple.com`

PR #9300, opened 2026-09-16
Stop the constant pfconfig expiration: clock comparisons and cluster conflict threshold

jrouzierinverse:
# Description

A customer reported pfconfig constantly expiring the configuration. Three separate things in that
area were comparing timestamps that don't belong to the same clock, or resetting the state that
says whether a resource is still current.

**1. The cache validity checks compared a remote timestamp to the local clock**
(`c9a084b8`)

`pfconfig::manager::is_valid` compared the mtime of the control file, touched by whichever process
expired the namespace, with `time()` in the current process, and `pfconfig::cached::is_valid`
compared the last touch cache pfconfig reports with its own clock to the time the subcache was
filled. `pfconfigdriver.IsValid` did the same with `LoadedAt`.

Any clock difference between those processes made every namespace permanently invalid, which
rebuilds the whole configuration on every access, or hid expirations when the difference went the
other way. The processes sharing `var/control` are often separate containers, and cluster members
and k8s nodes have their own clocks.

They now store the timestamp observed at load time and consider what they have valid until that
timestamp *changes*, so only two observations of the same clock are ever compared. This also
catches a namespace expired twice within the same second, which the previous comparison accepted
as still valid.

**2. The cluster conflict resolution ignored its threshold** (`2b43149a`)

`cluster_check` compared `$last_conflict_at`, an epoch timestamp, to
`active_active.conflict_resolution_threshold`, a duration. An epoch timestamp is always greater
than it, so the five minutes of grace never applied and a conflict resolved on the first check
that saw a divergence — the elapsed time it was meant to use was computed and then unused.

Resolving syncs every configuration store to every member through `expire_cluster`, which hard
expires the store and all of its child resources. With the task running every minute, a divergence
that would have settled on its own (a member restarting, a configuration push still in flight)
turned into a cluster wide expiration storm for as long as the versions differed.

**3. A reply without a last touch cache reset it to zero** (`a35e39d6`)

Zero is what `IsValid` reads as "nothing was ever loaded", so a single such reply invalidated every
pfconfig resource of a Go process at once. The keys replies did it on every call (their nested
response field had no JSON tag), and the element branch did it whenever a reply couldn't be decoded,
since it stored the value before returning the error. A zero is now logged and ignored, and only a
reply that could be decoded updates the value.

# Impacts

- pfconfig cache invalidation, so anything reading the configuration: `pfconfig` itself, every Perl
  service through `pfconfig::cached`, and every Go service through `pfconfigdriver`.
- `pf::pfcron::task::cluster_check` and, through it, `handle_config_conflict` /
  `sync_config_as_master` on clusters.
- `FirewallSSOInt` now embeds `pfconfigdriver.PfconfigObject` instead of re-declaring three of its
  methods, so the new accessors don't have to be listed twice.

Worth looking at:
- the ordering in `cache_resource` / `get_cache`: the control file timestamp is read *after* the
  touch when we expired it ourselves, and *before* loading when we read from the backend, so a
  concurrent expiration costs a reload rather than going unnoticed;
- `set_in_subcache` keeps the first last touch cache seen since the subcache was emptied, so a touch
  that lands while the subcache is being filled invalidates all of it;
- the 5 second phone in in `pfconfig::cached` and `pfconfigdriver` is unchanged and is still what
  propagates a new last touch cache.

Tested: `t/unittest/pfconfig/*.t` and `t/unittest/pfcron/task/cluster_check.t` pass, the new tests
fail against the previous logic (verified by reverting each comparison), and `go build ./...` plus
the `pfconfigdriver` tests show no new failures — the `pfdemo.org` ones fail the same way before and
after on a box that isn't the reference test box.

# Delete branch after merge

YES

# Checklist
- [x] Add unit tests — yes: `t/unittest/pfconfig/validity.t`, `t/unittest/pfcron/task/cluster_check.t`,
  `TestIsValidWithPfconfigClockAhead`, `TestFetchKeysKeepsLastTouchCache`,
  `TestUpdateLastTouchCacheIgnoresMissingValue`

# NEWS file entries

## Bug Fixes
* pfconfig considered the configuration permanently expired when the processes sharing `var/control`
  or running pfconfig had different clocks, rebuilding it on every access
* A cluster configuration version conflict was resolved on the first check instead of after
  `active_active.conflict_resolution_threshold`, hard expiring every configuration store on every
  member every minute
* A pfconfig reply that carried no last touch cache invalidated every configuration resource of a Go
  service at once


satkunas:
1. TImestamps should be normalized from the systems TZ.
2. first check the mtime/ctime, if < or > than now, then compare the hash. If < or > then trigger, otherwise FP.

lib/pf/pfcron/task/cluster_check.pm
```diff
@@ -42,19 +42,35 @@ sub run {
...
             $cache->set('last_config_healthy_timestamp', $now);
         }
-        elsif($last_conflict_at > $conflict_resolution_threshold) {
-            get_logger->info("Configuration is unhealthy since $conflict_resolution_threshold seconds. Will attempt to resolve the conflict");
+        # How long the members have been running different versions is what is compared to the
+        # threshold, not the timestamp at which they last agreed. Comparing the timestamp made
+        # this true on the first divergent check, so a transient one (a member restarting, a
+        # configuration push still in flight) resolved right away, and resolving means a full
+        # expiration of every configuration store on every member, every time this task runs
+        elsif($unhealthy_for > $conflict_resolution_threshold) {
```
fdurand, line 67: **medium — the threshold postpones the expiration storm instead of rate-limiting it**

`handle_config_conflict()` is called here but `last_config_healthy_timestamp` is left untouched, and nothing else in this branch resets it. On the next run (1 minute later) `$unhealthy_for` is still above the threshold, so the resolution fires again — and again every minute for as long as the members run different versions.

Concrete case: a member that cannot be synced (webservices down on it, or `sync_config_as_master` failing repeatedly) keeps the divergence alive. Since resolving runs `expire_cluster`, which hard-expires every configuration store and all child resources on every member, the cluster takes a full expiration storm every minute indefinitely — exactly the failure mode the PR description sets out to stop, just delayed by 5 minutes rather than eliminated.

Suggest adding `$cache->set('last_config_healthy_timestamp', $now);` right after the `handle_config_conflict()` call so each resolution attempt gets its own grace period.

lib/pf/util.pm
```diff
@@ -1550,12 +1552,17 @@ sub touch_file {
 
     if (sysopen(my $fh,$filename,O_RDWR | O_CREAT)) {
         my ($seconds, $microseconds) = Time::HiRes::gettimeofday();
-        POSIX::2008::futimens(fileno $fh, $seconds, $microseconds * 1000,$seconds, $microseconds * 1000);
+        my $status = POSIX::2008::futimens(fileno $fh, $seconds, $microseconds * 1000,$seconds, $microseconds * 1000);
+        my $timestamp = defined($status) && $status == 0
```
fdurand, line 1556: **medium — the returned timestamp is compared with `==` against `stat`, so it depends on the filesystem storing nanosecond mtimes**

`touch_file` returns the timestamp it *asked* `futimens` for, and `pfconfig::manager::is_valid` (manager.pm:442) compares that stored value for exact equality with what `Time::HiRes::stat` reports for the same file.

That holds only when the filesystem under `var/control` actually stores sub-second mtimes. On a mount that truncates mtime (ext4 formatted with 128-byte inodes, some container volume drivers, older NFS), or if `Time::HiRes::d_hires_stat` is 0 (the module documents that case as a pass-through to `CORE::stat`), the recorded value can never equal the stat'ed one. `is_valid` then returns 0 forever and the pfconfig server rebuilds every namespace on every `get_cache` — the exact symptom this PR fixes, but permanent and silent.

I verified the float arithmetic here matches `Time::HiRes::stat` bit-for-bit on ext4 and tmpfs over 200 randomized nanosecond values, so this is purely about environments without ns mtime. A cheap guard would remove the whole failure class: after touching, compare against `control_file_timestamp` once and fall back to the stat'ed value if they differ.

(The `$status == 0` check itself is correct — `POSIX::2008` declares `futimens` as `SysRetTrue`, whose typemap returns the `"0 but true"` string on success and `undef` on failure.)

go/pfconfigdriver/fetch.go
```diff
@@ -349,13 +344,28 @@ func IsValid(ctx context.Context, o PfconfigObject) bool {
...
+// loaded", so storing it would invalidate every resource of the process at once and send all of
+// them back to pfconfig. The value we already have is kept instead, which makes the resources
+// reload when pfconfig reports a different one, like they do for any expiration.
+func updateLastTouchCache(ctx context.Context, lastTouchCache float64, identifier string) bool {
+	if lastTouchCache == 0 {
+		log.LoggerWContext(ctx).Warn(fmt.Sprintf("The reply for %s carried no last touch cache. Keeping the one we have.", identifier))
+		return false
+	}
+
+	globalMeta.setLastTouchCache(lastTouchCache)
```
fdurand, line 365: **medium — the global marker can be rolled backwards by a stale in-flight reply, and `IsValid` is now exact-equality**

`setLastTouchCache` is called unconditionally with whatever the reply carried, including a value *older* than the one already in `globalMeta` (`TestFetchDecodeSocketPublishesResponseMarker`'s "backward marker" case asserts this behaviour).

Race: goroutine A sends a request while pfconfig's marker is T1; pfconfig then expires a namespace and moves to T2; goroutine B fetches, gets T2, sets the global to T2 and stamps its object T2; A's slower reply finally lands carrying T1 and rolls the global back to T1. Every resource already stamped T2 now fails `o.GetLoadedTouchCache() == globalMeta.getLastTouchCache()` at line 347 and is refetched.

The old `float64(o.GetLoadedAt()...) >= globalMeta.getLastTouchCache()` comparison could not be tripped this way, because `LoadedAt` only ever moved forward. Making the setter monotonic (ignore a value lower than the current one, or only accept it alongside a decode we know is newer) keeps the clock-independence fix and removes the burst.

go/pfconfigdriver/fetch.go
```diff
@@ -391,40 +401,63 @@ func FetchDecodeSocket(ctx context.Context, o PfconfigObject) error {
...
 			return fmt.Errorf("element in response for %s was invalid. Response was: %s", query.GetIdentifier(), jsonResponse)
 		}
 	}
 
+	// Only a reply we could decode updates the last touch cache, so a failed fetch above leaves
+	// the resources that are loaded alone instead of sending all of them back to pfconfig
+	loadedTouchCache := lastTouchCache
+	if !updateLastTouchCache(ctx, lastTouchCache, query.GetIdentifier()) {
+		// The reply carried none, so the global is the only value we can stamp this one with
+		loadedTouchCache = globalMeta.getLastTouchCache()
```
fdurand, line 452: **low — a marker-less reply on a fresh process stamps the object 0, which `IsValid` reads as "never loaded" forever**

When the reply carries no `last_touch_cache` *and* `globalMeta` is still 0 (first fetch of a freshly started process, or every reply so far lacked the marker), `loadedTouchCache` ends up 0. `IsValid`'s first branch (`globalMeta.getLastTouchCache() == 0`) then returns false on every access, so the object is refetched on every single call and `updateLastTouchCache` logs a `Warn` each time — a hot fetch+warn loop rather than one warning.

This is reachable: `sbin/pfconfig`'s `encode_output` has a fallback (`{error => "Cannot transform object into JSON data: ..."}`) that discards the `last_touch_cache` key `send_output` had added. Worth either stamping from a sentinel or bailing out of the retry loop when both values are 0.

lib/pfconfig/cached.pm
```diff
@@ -236,7 +239,7 @@ sub _get_from_socket {
             print STDERR "$what $response";
             die $@;
         }
-        $LAST_TOUCH_CACHE = $result->{last_touch_cache} // 0;
+        $LAST_TOUCH_CACHE = $result->{last_touch_cache} // $LAST_TOUCH_CACHE;
```
fdurand, line 242: **low — the fail-safe direction flipped: a marker-less reply now keeps the client's subcache valid instead of forcing a reload**

Previously `// 0` meant a reply without `last_touch_cache` reset the global to 0, which `is_valid` reads as "never loaded" and forces a reload. With `// $LAST_TOUCH_CACHE` the old value is kept, while `$RELOADED_TOUCH_CACHE = time` on the next line is still set unconditionally — so the 5-second staleness branch never fires either.

Net effect: a client that starts receiving marker-less replies keeps serving its subcache indefinitely and never notices an expiration. Same fail-open direction as `updateLastTouchCache` on the Go side. Reachable via `sbin/pfconfig`'s `encode_output` JSON-error fallback, which drops the key that `send_output` had added (less likely for Perl clients, which default to sereal, but the branch is there).

fdurand (commented):
Careful single-pass correctness review of the diff (all 12 files). 8 findings below, all anchored to lines inside the diff hunks.

Things I checked that look right, for the record: `decodeJsonInterfaceErr` -> `json.Unmarshal` is safe because `query.encoding` is always `"json"` and passing `o` instead of `&o` behaves identically in `encoding/json`; the `destination`/`newo` swap genuinely makes a failed refresh atomic (devel zeroed the caller's struct before decoding); the `PfconfigKeys.Response` json tag + `SetKeysFromResponse` fix promotes correctly through `HostsIp`; `conflict_resolution_threshold` really is normalized to seconds by `config/Pf.pm` (`type=time`) and `$task->interval` is 60 from `schedule=@every 1m`, so the cluster_check arithmetic and its test hold; test counts in both new `.t` files (22 and 9) are correct; no `memorized_at` references remain. `go build ./...` and `go vet` on the changed packages pass.

lib/pfconfig/manager.pm
```diff
@@ -378,34 +387,48 @@ sub cache_resource {
...
+        # Keep our own expiration marker, including when bootstrap has no L2.
+        # Reading it back could adopt a concurrent writer's expiration.
+        $control_timestamp = $self->touch_cache($what);
     }
+    # External builds successfully written to L2 stay invalid until reloaded:
+    # we do not know the timestamp of their remote expiration.
     $self->{memory}->{$what}       = $result;
     delete $self->{memory}->{"$ordered_prefix$what"};
-    $self->{memorized_at}->{$what} = time;
+    $self->{control_timestamp}->{$what} = $control_timestamp;
```
fdurand, line 403: **low — a non-server manager stores `undef` here on the success path, so the namespace it just built is immediately invalid**

When `$self->{pfconfig_server}` is false and the L2 write succeeded, `$control_timestamp` is never assigned by the block above, so this line stores `undef`. `is_valid` (line 442) requires `defined($memory_timestamp)`, so the namespace that was *just* built is reported invalid and the next `get_cache` pays a full L2 (Redis/MySQL) round trip — the L1 `{memory}` entry is dead on arrival for that namespace.

I understand this is deliberate (the comment above says so) and it is harmless today only because every `pfconfig::manager->new` call site is short-lived (`pf::ConfigStore::commit`, `pf::config::configreload`, `bin/kafka-init.pl`, the `pf::cmd::pf::pfconfig` handlers). It becomes a per-access DB hit the moment anyone keeps a non-server manager around, and nothing in the code signals that. A comment at the `{memory}` assignment, or reading the control file timestamp back here, would make it safe by construction.

lib/pfconfig/manager.pm
```diff
@@ -285,13 +290,17 @@ sub get_cache {
     # we look in raw memory and make sure that it's not expired
     my $memory = $self->{memory}{$what};
     unless (defined($memory) && $self->is_valid($what)) {
+        # Read the control file timestamp before loading. If the namespace gets
+        # expired while we are loading it, we keep the timestamp from before the
+        # expiration and reload on the next access instead of missing the change
+        my $control_timestamp = $self->control_file_timestamp($what);
```
fdurand, line 296: **low — `$control_timestamp` can be `undef` here and is stored anyway at line 303**

`control_file_timestamp` returns `undef` when the control file does not exist yet (fresh install, `var/control` wiped). If the L2 lookup on the next line succeeds, line 303 stores that `undef`, so the value just loaded from the backend is marked invalid and the next access re-reads L2.

It converges only because `is_valid` creates the file as a side effect on the following call, but that is one wasted backend round trip per namespace after every `var/control` reset. Skipping the assignment when the timestamp is undefined (or touching the file first) avoids it.

t/unittest/pfconfig/validity.t
```diff
@@ -0,0 +1,195 @@
...
+# An external builder cannot identify the timestamp of its remote expiration.
+# It must reload from L2 before treating its memory entry as current.
+{
+    no warnings qw(redefine once);
+    local *pfconfig::manager::config_builder = sub { return { value => 'external' } };
+    local *pfconfig::git_storage::is_enabled = sub { return 0 };
+    local *pfconfig::util::socket_expire = sub { $manager->touch_cache($ns); return 1 };
+    local $manager->{pfconfig_server} = 0;
+    $manager->cache_resource($ns);
+    ok(!$manager->is_valid($ns), "external build does not adopt an unverified control timestamp");
```
fdurand, line 86: **low — this block silently requires a reachable L2 backend, or it fails**

`cache_resource` only leaves `control_timestamp` undefined (which is what this assertion checks) when the L2 write *succeeds*. On a test host where the pfconfig backend is not reachable, `$cache_w` is false, the new `$self->{pfconfig_server} || !$cache_w` fallback calls `touch_cache`, `control_timestamp` is defined, and this `ok(!...)` plus the `is_deeply ... "external build is reloaded from L2"` on line 87 both fail.

The sibling subtest below ("external builds remain usable before the database is configured") stubs a backend for exactly this reason; this block does not, so the file's result depends on whether the runner has a live backend. Stubbing `$manager->{cache}` here too (or asserting `$cache_w` first) would make it deterministic.

lib/pf/util.pm
jrouzierinverse: This is not a problem as we use modern filesystems

PR #9317, opened 2026-09-21
fix(cron): match portless ACLs on any port, strip ACL comments, and name the acl in parse errors

jrouzierinverse:
# Description

Two fixes to the ACL parser and matcher used by the network policy lookup in `go/cron/policy_lookup.go`.

**1. A portless ACL matched only port 0.**
An ACL with no port qualifier, such as `permit tcp any any`, parses to `Port: 0`. `Matcher.Matches` compared that literally against the event's destination port, so the rule matched only traffic destined to port 0 — effectively nothing — instead of any port. `Port == 0` now means "any port".

**2. ACL comments broke parsing.**
`ParseAcl` now trims surrounding whitespace and drops a trailing `#` comment, so an annotated line like `permit tcp any any #Supports comments` parses instead of failing on the comment words.

Additionally, every error `ParseAcl` returns now carries the offending acl string. The only caller, `UpdateMatchers`, logs the error and skips the acl, which until now gave no way to tell which of a policy's acls was rejected.

# Impacts

- Network policy lookup / ACL matching (`go/cron/policy_lookup.go`)
- Any role or implicit policy whose acls omit a port qualifier — these previously never matched and now match on any port. This is the intended behaviour, but it does widen what those rules match, so it's the main thing to look at.
- `UpdateMatchers` log output: parse failures now name the acl.

# Delete branch after merge

YES

# Checklist

- [ ] Document the feature — n/a (bug fix, no behaviour to document)
- [x] Add unit tests — yes, see below

Unit tests added in `go/cron/policy_lookup_test.go`:
- `TestMatcher`: parse cases for `permit tcp any any` and `permit tcp any any #Supports comments`
- `TestMatchNetworkEvent`: a portless acl matching an event on a non-zero destination port
- The error assertions in `TestMatcher` were also fixed. `errors.Is` was called as a bare statement, so the expected-error cases asserted nothing — and it could never have matched anyway, since these are plain `fmt.Errorf` values with no sentinel to compare against. The messages are now compared, and cases that unexpectedly parse without error are reported.

`go test ./cron/ -run 'TestMatcher|TestMatchNetworkEvent|TestPolicyLoad'` passes.

# NEWS file entries

## Bug Fixes
* ACLs without a port qualifier, such as `permit tcp any any`, only matched traffic to port 0 instead of any port
* ACLs with a trailing `#` comment failed to parse
* ACL parse errors did not identify which acl was rejected


go/cron/policy_lookup.go
```diff
@@ -151,7 +158,7 @@ type Matcher struct {
 }
 
 func (m *Matcher) Matches(ne *NetworkEvent) bool {
-	return m.Port == ne.DestPort && m.Proto == ne.IpProtocol && m.SrcNet.Contains(ne.SourceIp) && m.matchDest(ne)
+	return (m.Port == 0 || m.Port == ne.DestPort) && m.Proto == ne.IpProtocol && m.SrcNet.Contains(ne.SourceIp) && m.matchDest(ne)
```
fdurand: **Blocking: `deny` lines now match and return the policy's (allow) verdict.**

`ParseAcl` records `matcher.Action` (`permit`/`deny`), but neither `Matches` nor `matchEnforcementInfo` ever reads it, so a `deny` line has always behaved like a `permit` line. Until now that was mostly harmless: a portless `deny tcp any any` parsed to `Port: 0` and could only match traffic to port 0, so the usual catch-all deny was inert.

With `Port == 0` meaning "any port", that catch-all wakes up:

```
Policy 1  verdict=allow  rule-id=AAAA
  permit tcp any any eq 443
  deny   tcp any any

Policy 2  verdict=block  rule-id=BBBB
  permit tcp any any eq 22
```

Event: TCP to 10.1.2.3:22 from a device in that role.

- Before: policy 1 line 2 has `Port 0 != 22`, no match; policy 2 matches → `block` / `BBBB`.
- After: policy 1 line 2 matches (any port, tcp, any, any) → `matchEnforcementInfo` returns policy 1's info → `allow` / `AAAA`, and policy 2 is never evaluated.

The flow is attributed to the one rule that says the opposite, and because `Lookup` stops at the first hit, a portless deny in a source-role policy also shadows the destination-role and implicit policies. The same ACL text pushed to the switch really drops that traffic, so pfflow would report as allowed a flow the switch blocked.

Minimal fix, deny is a non-match:

```go
func (m *Matcher) Matches(ne *NetworkEvent) bool {
	if m.Action == "deny" {
		return false
	}
	return (m.Port == 0 || m.Port == ne.DestPort) && m.Proto == ne.IpProtocol && m.SrcNet.Contains(ne.SourceIp) && m.matchDest(ne)
}
```

(or skip `deny` lines in `UpdateMatchers`). Please also add a `TestMatchNetworkEvent` case with a `deny tcp any any` line ahead of a permit line so it stays locked in. Proper first-match ACL semantics (a matching deny means the policy does not apply) can be a follow-up.

